PR #1, opened 2015-01-31
Cassandra I/O layer for Metadata

velvia:
* Table definition and async methods for I/O operations
* `MetadataActor`, which takes `Command`s and returns `Response`s, and also throttles the number of outstanding async futures
* Tests for both of the above, which actually spin up or use a real Cassandra daemon


velvia:
@chitang updated with review feedback

PR #16, opened 2015-09-19
added implicit PatienceConfig

Shiti:
attempt to fix Travis test since there are error in test due to timeout


velvia:
@Shiti thanks, :+1:

PR #17, opened 2015-09-19
added script for filo-cli

Shiti:
updated Scala version to 2.10.5 since that was being picked by SBT and minor changes in the build script

### DIFF
--- a/filo-cli
+++ b/filo-cli
@@ -1,0 +1,42 @@
+#!/usr/bin/env bash
+
+SCALA_VERSION="2.10"
+FILO_VERSION="0.1-SNAPSHOT"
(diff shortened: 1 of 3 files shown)


filo-cli
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env bash
+
+SCALA_VERSION="2.10"
+FILO_VERSION="0.1-SNAPSHOT"
```
velvia: We should split out the version into a version.sbt, then it could be easily grep'ed.  For now this might be ok but we do need a more automatic way of getting the version.

velvia:
@Shiti thanks, going to merge this.

PR #19, opened 2015-09-22
updated README

Shiti:


README.md
```diff
@@ -32,130 +55,144 @@ FiloDB is a great fit for bulk analytical workloads, or streaming /  event data.
...
 
-Note that FiloDB uses an off-heap direct memory memtable, and you probably need to bump Spark's default direct memory size with a `-XX:MaxDirectMemorySize=1G` or some larger setting.
+Or, specifying the `partition_column`,
 
-- Write data without partition column
+```scala
+scala> csvDF.write.format("filodb.spark").
+    option("dataset", "gdelt").
+    option("sort_column", "GLOBALEVENTID").
+    option("partition_column","ActionGeo_FullName").
```
velvia: `ActionGeo_FullName` is not a good partitioning column, it generates too many partitions.   Instead use "Year".

velvia:
@Shiti Thanks.  Looks good overall!

* I think the Spark-SQL section should definitely be above the "Old Spark API" section.
* We should note that the CLI export command is not functional.
* Is the --init command documented?  It should be.
* I still think the main Spark CLIs and the query section (it wasn't taken out was it?  Hard to tell with so many changes) is slightly more important than the CLI....

Shiti:
I could move the spark-sql section above the old api section.
I am working on the cli export command but until its done I guess we should
mention it isnt working. Will do that.
No the init command is not documented. Will add that too.
The spark cli part is still there. I guess it'll be easier if you could
read from my repo.

PR #20, opened 2015-09-23
exportCSV for filo-cli

Shiti:


velvia:
@Shiti thanks!  Some relatively minor comments but make sure you test with a large number of rows, like exporting a million rows to a file.

velvia:
@Shiti good point.  Maybe we should encapsulate.  I think one way to do that is to have ColumnStore either take a Dataset, or more likely, a RichProjection.  Perhaps change KeyRange to embed a RichProjection.

velvia:
I'll try to do that tomorrow morning.

Shiti:
I tried export command with the limit as 1Mill and it got done in approx 4-5s

velvia:
@Shiti thanks.  The reason why we still need a type parameter for scanSegments and readSegments is because `Segment` is typed in `[K]`.   We could possibly have a method that reads rows out, that does not need to be typed.  However, not all queries will be by row - in fact going forward we'd want to explore doing columnar type queries.

velvia:
@Shiti you did a rebase and push?  i guess that's fine but then it's harder to see your revisions.  How about small fix for the whitespace and then we can merge, thanks!

velvia:
@Shiti LGTM for most part, going to merge.

PR #29, opened 2015-10-15
Refactor Scheduler, Reprojector, MemTable, and coordinator actors

velvia:
This is a fairly big change.  The first part refactors the scheduler and FlushPolicy into a new DatasetCoordinatorActor, which coordinates all ingestion, reprojection, and columnar flushing.  It also makes the scheduling event-based, rather than polling or time-based.   Unlike the previous scheduler, which was global, the DatasetCoordinatorActor coordinates writes for only one (dataset, version) at a time.  In turn, the NodeCoordinatorActor spins up DatasetCoordinatorActors, and acts as the one external API entry point for all ingestion and client interactions (other than Spark queries).  The implications:

* Both the CLI and Spark modules now use Actor client based ingestion, and talk to NodeCoordinatorActor through `RowSource`, instead of calling into a MemTable directly.  This makes it much easier in the future to move the core logic of FiloDB into a daemon server form factor, instead of being a library.
* Scheduler and FlushPolicy are gone from core, as is EndToEndTest
* `NodeCoordinatorActorSpec` becomes kind of the end to end ingestion test.

Currently, all tests are passing.

The second part, which is refactoring and simplifying MemTable and Reprojector, will be coming next.


helena:
Big change, starting review tomorrow afternoon.

velvia:
Change is pretty much done.  All tests pass - at least locally - and am able to ingest data through Spark, as well as query.   Curiously, ingestion seems to be much faster (about 2x), and query is a bit faster, but not sure why.  Offheap/direct memory use is better too.   However there is an issue - the ingestion seems to hang at the very end, the last Spark task seems not to finish.

trulite:
I know we are very close to what I am suggesting. I think the MemTable could still be simplified thusly. 


Memtable{
	addRows(rows: Seq[RowReader]):Unit

	size:Int

	getSorted(comparator:Comparator):Seq[RowReader]

	removeRow(partitionKey:PartitionKey,key:Key)

}



DataSetCoordinator{

	projectionOrders:Seq[Seq[ColumnId]]

	ingestRows(rows: Seq[RowReader]) ={

		if(memTable.size > N || memorySize > M){
			flushMemtable
		}
		memTable.addRows(rows)	
	}
	
	
	flushMemtable = {
		val segments:Seq[Segment] = projectionOrders.map{ projectionOrder =>
			reprojector.reproject(memTable,projectionOrder)
		}
		columnStore.writeSegments(segments)
	}
	
}

Reprojector{
	
	reproject(memTable:MemTable, projectionOrder:Seq[ColumnId]):Segment = {
		val comparator = getComparator(memTable.getSchema,projectionOrder)
		val rows:Seq[RowReader] = memTable.sort()
		newSegment(rows)
	}

}

cli/src/main/scala/filodb.cli/CsvImportExport.scala
```diff
@@ -20,8 +19,6 @@ import filodb.core._
 trait CsvImportExport {
   val system: ActorSystem
```
helena: shouldn't these vals be defs?

     trait Fu {
        def system: ActorSystem
     }

     class MyFu(val system: ActorSystem) extends Fu {

     }
     class MyOtherFu extends Fu {
         val system = ActorSystem("fu", ConfigFactory.load)
     }

cli/src/main/scala/filodb.cli/CsvImportExport.scala
```diff
@@ -57,51 +54,15 @@ trait CsvImportExport {
...
-        println(s"Dataset $dataset is not known, you need to --create it first!")
-        exitCode = 2
-        return
-      case CoordinatorActor.UndefinedColumns(undefCols) =>
-        println(s"Some columns $undefCols are not defined, please define them with --create first!")
-        exitCode = 2
-        return
-      case CoordinatorActor.BadSchema(msg) =>
-        println(s"BadSchema - $msg")
+    val csvActor = system.actorOf(CsvSourceActor.props(fileReader, dataset, version, coordinatorActor))
```
helena: Do we want all instances of csvActor to not have a supervisor? For failures etc.
I mean, by doing
 
     system.actorOf(...)

versus this in a supervisor actor

     context.actorOf(..) 

just a thought.

cli/src/main/scala/filodb.cli/CsvImportExport.scala
```diff
@@ -57,51 +54,15 @@ trait CsvImportExport {
   def ingestCSV(dataset: String, version: Int, csvPath: String, delimiter: Char) {
```
helena: IMO it's really nice for readers of code when functions have return types specified, especially when Unit

cli/src/main/scala/filodb.cli/CsvImportExport.scala
```diff
@@ -57,51 +54,15 @@ trait CsvImportExport {
...
-        exitCode = 2
-        return
-      case CoordinatorActor.UndefinedColumns(undefCols) =>
-        println(s"Some columns $undefCols are not defined, please define them with --create first!")
-        exitCode = 2
-        return
-      case CoordinatorActor.BadSchema(msg) =>
-        println(s"BadSchema - $msg")
+    val csvActor = system.actorOf(CsvSourceActor.props(fileReader, dataset, version, coordinatorActor))
+    actorAsk(csvActor, RowSource.Start, 99 minutes) {
```
helena: I feel like a timeout of 99 minutes should maybe not be burried down in the code but rather under the trait declaration

     trait Fu {
         private val askTimeout = 99.minutes
     }

cli/src/main/scala/filodb.cli/CsvImportExport.scala
```diff
@@ -57,51 +54,15 @@ trait CsvImportExport {
...
-      linesIngested += mappedLines.length
-      if (linesIngested % 10000 == 0) println(s"Ingested $linesIngested lines!")
-    }
-
-    coordinatorActor ! CoordinatorActor.Flush(dataset, version)
-    println("Waiting for scheduler/memTable to finish flushing everything")
-    Thread sleep 5000
-    while (memTable.flushingDatasets.nonEmpty) {
-      print(".")
-      Thread sleep 1000
```
helena: Great to not have this anymore ;)

coordinator/src/main/scala/filodb.coordinator/CoordinatorSetup.scala
```diff
@@ -9,7 +9,7 @@ import filodb.core.metadata.MetaStore
 import filodb.core.reprojector._
 
 /**
- * A trait to make setup of the CoordinatorActor stack a bit easier.
+ * A trait to make setup of the NodeCoordinatorActor stack a bit easier.
```
helena: Maybe for docs

     [[NodeCoordinatorActor]]

coordinator/src/main/scala/filodb.coordinator/CoordinatorSetup.scala
```diff
@@ -22,31 +22,16 @@ trait CoordinatorSetup {
...
   lazy val reprojector = new DefaultReprojector(columnStore)
-  lazy val scheduler = new Scheduler(memTable,
-                                     reprojector,
-                                     flushPolicy,
-                                     config.getInt("scheduler-max-tasks"))
 
   lazy val coordinatorActor =
-    system.actorOf(CoordinatorActor.props(memTable, metaStore, scheduler, columnStore,
-                                          config.getConfig("coordinator")),
+    system.actorOf(NodeCoordinatorActor.props(metaStore, reprojector, columnStore, config),
```
helena: Same idea - do we want a supervisor actor in cli?

coordinator/src/main/scala/filodb.coordinator/DatasetCoordinatorActor.scala
```diff
@@ -0,0 +1,237 @@
+package filodb.coordinator
+
+import akka.actor.{Actor, ActorRef, Props}
+import com.typesafe.config.Config
+import net.ceedubs.ficus.Ficus._
+import org.velvia.filo.RowReader
+import scala.concurrent.Future
```
helena: IMO imports should be

java.* (to see the hurt first)
space
scala.*
akka.*
third.party.*
filodb.*

coordinator/src/main/scala/filodb.coordinator/DatasetCoordinatorActor.scala
```diff
@@ -0,0 +1,237 @@
...
+  case object GetStats
+  case class Stats(flushesStarted: Int,
+                   flushesSucceeded: Int,
+                   flushesFailed: Int,
+                   numRowsActive: Int,
+                   numRowsFlushing: Int)
+
+  // Internal messages
+  case class FlushDone(result: Seq[String])
+  case class FlushFailed(t: Throwable)
```
helena: These case objects/classes need a sealed common trait to extend, that also explains the domain tree they fall under for FiloDB, perhaps that is Serializable

coordinator/src/main/scala/filodb.coordinator/DatasetCoordinatorActor.scala
```diff
@@ -0,0 +1,237 @@
...
+ * ==Configuration==
+ * {{{
+ *   memtable {
+ *     flush-trigger-rows = 50000    # No of rows above which memtable flush might be triggered
+ *     max-rows-per-table = 1000000
+ *     min-free-mb = 512
+ *   }
+ * }}}
+ */
+class DatasetCoordinatorActor[K](projection: RichProjection[K],
```
helena: Is this actor part of the public API? if not lets

     private[filodb] class DatasetCoordinatorActor...

it.

coordinator/src/main/scala/filodb.coordinator/DatasetCoordinatorActor.scala
```diff
@@ -0,0 +1,237 @@
...
+                                 reprojector: Reprojector,
+                                 config: Config) extends BaseActor {
+  import DatasetCoordinatorActor._
+  import context.dispatcher
+
+  val datasetName = projection.dataset.name
+  val nameVer = s"$datasetName/$version"
+  val flushTriggerRows = config.getLong("memtable.flush-trigger-rows")
+  val maxRowsPerTable = config.getInt("memtable.max-rows-per-table")
+  val minFreeMb = config.as[Option[Int]]("memtable.min-free-mb").getOrElse(DefaultMinFreeMb)
```
helena: Looks like later on we could consider adding a FiloDBCoreSettings class that would have all of these created once in load time vs n-times in each actor needing the settings.

coordinator/src/main/scala/filodb.coordinator/DatasetCoordinatorActor.scala
```diff
@@ -0,0 +1,237 @@
...
+      sys.runtime.gc()
+      return
+    }
+
+    // MemTable will take callback function and call back to ack rows
+    activeTable.ingestRows(rows) { ackTo ! NodeCoordinatorActor.Ack(seqNo) }
+
+    // Now, check how full the memtable is, and if it needs to be flipped, and a flush started
+    if (shouldFlush) self ! StartFlush()
+  }
```
helena: Hrm, 'return's used in scala... How about

```private def ingestRows(ackTo: ActorRef, rows: Seq[RowReader], seqNo: Long): Unit = activeRows match {
     case rows if rows >= maxRowsPerTable =>
         logger.debug(s"MemTable ($nameVer) is at $activeRows rows, cannot accept writes...")
      case rows if rows < maxRowsPerTable =>
          val freeMB = getRealFreeMb
          if (freeMB < minFreeMb) {
             // logging...
             sys.runtime.gc()
         }
     case _ =>
        activeTable.ingestRows(rows) { ackTo ! NodeCoordinatorActor.Ack(seqNo) }
        if (shouldFlush) self ! StartFlush()
    } 
}```

File: coordinator/src/main/scala/filodb.coordinator/DatasetCoordinatorActor.scala [hunk 13]
```diff
@@ -0,0 +1,237 @@
...
+      return
+    }
+    flushingTable = Some(activeTable)
+    activeTable = makeNewTable()
+    val newTaskFuture = reprojector.reproject(flushingTable.get, version)
+    curReprojection = Some(newTaskFuture)
+    flushesStarted += 1
+    newTaskFuture.map { results =>
+      self ! FlushDone(results)
+    }
```
helena: Or a for comprehension

for (results <- newTaskFuture) self ! FlushDone(results)

coordinator/src/main/scala/filodb.coordinator/DatasetCoordinatorActor.scala
```diff
@@ -0,0 +1,237 @@
...
+
+  private def shouldFlush: Boolean =
+    activeRows >= flushTriggerRows && curReprojection == None
+
+  private def handleFlushDone(): Unit = {
+    curReprojection = None
+    // Close the flushing table and mark it as None
+    flushingTable.foreach(_.close())
+    flushingTable = None
+    flushedCallbacks.foreach { ref => ref ! NodeCoordinatorActor.Flushed }
```
helena: Or maybe
     for (ref <- flushedCallbacks) ref ! NodeCoordinatorActor.Flushed

coordinator/src/main/scala/filodb.coordinator/DatasetCoordinatorActor.scala
```diff
@@ -0,0 +1,237 @@
...
+    if (shouldFlush) self ! StartFlush()
+  }
+
+  private def handleFlushErr(t: Throwable): Unit = {
+    // TODO: let everyone know task failed?  Or not until retries are all up?
+    flushedCallbacks.foreach { ref => ref ! FlushFailed(t) }
+    flushedCallbacks = Nil
+    flushesFailed += 1
+    // TODO: retry?
+    // Don't delete reprojection.  Just let everything suspend?
```
helena: let's add an error handling ticket?

coordinator/src/main/scala/filodb.coordinator/DatasetCoordinatorActor.scala
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  private def clearProjection(originator: ActorRef, projection: Projection): Unit = {
+    for { flushResult <- curReprojection.getOrElse(Future.successful(Nil))
+          resp <- columnStore.clearProjectionData(projection) }
+    { originator ! NodeCoordinatorActor.ProjectionTruncated }
+  }
```
helena: Nice!

coordinator/src/main/scala/filodb.coordinator/NodeCoordinatorActor.scala
```diff
@@ -0,0 +1,229 @@
...
+ * - Forwarding new changes (rows) to other NodeCoordinatorActors if they are not local
+ * - Forwarding rows to DatasetCoordinatorActors
+ *
+ * It is called by local (eg HTTP) as well as remote (eg Spark ETL) processes.
+ */
+object NodeCoordinatorActor {
+  /**
+   * Creates a new dataset with columns and a default projection.
+   */
+  case class CreateDataset(dataset: Dataset, columns: Seq[Column])
```
helena: Same idea here for a common sealed trait they extend

coordinator/src/main/scala/filodb.coordinator/NodeCoordinatorActor.scala
```diff
@@ -0,0 +1,229 @@
...
+  /**
+   * Ingests a new set of rows for a given dataset and version.
+   * The partitioning column and sort column are set up in the dataset.
+   *
+   * @param seqNo the sequence number to be returned for acknowledging the entire set of rows
+   * @return Ack(seqNo) returned when the set of rows has been committed to the MemTable.
+   */
+  case class IngestRows(dataset: String, version: Int, rows: Seq[RowReader], seqNo: Long)
+
+  case class Ack(seqNo: Long) extends Response
```
helena: Events like Ack need a different base sealed trait than state changes like DatasetCreated so we can have one common marker trait, one that extends that for state, another for events that these guys extend

coordinator/src/main/scala/filodb.coordinator/NodeCoordinatorActor.scala
```diff
@@ -0,0 +1,229 @@
...
+            columnStore: ColumnStore,
+            config: Config): Props =
+    Props(classOf[NodeCoordinatorActor], metaStore, reprojector, columnStore, config)
+}
+
+/**
+ * ==Configuration==
+ * {{{
+ *   {
+ *   }
```
helena: ?

coordinator/src/main/scala/filodb.coordinator/NodeCoordinatorActor.scala
```diff
@@ -0,0 +1,229 @@
...
+        logger.info(s"Undefined columns $undefinedCols for dataset $dataset with schema $schema")
+        originator ! UndefinedColumns(undefinedCols.toSeq)
+        None
+      } else {
+        Some(schema)
+      }
+    }
+  }
+
+  private def createDataset(originator: ActorRef, datasetObj: Dataset, columns: Seq[Column]) = {
```
helena: private def createDataset(originator: ActorRef, datasetObj: Dataset, columns: Seq[Column]): Unit =

coordinator/src/main/scala/filodb.coordinator/NodeCoordinatorActor.scala
```diff
@@ -0,0 +1,229 @@
...
+          val ref = context.actorOf(props, s"ds-coord-${datasetObj.name}-$version")
+          self ! AddDatasetCoord(dataset, version, ref)
+          notify(IngestionReady)
+        }
+      }).recover {
+        case NotFoundError(what) => notify(UnknownDataset)
+        case t: Throwable        => notify(MetadataException(t))
+      }
+    }
+  }
```
helena: IMO this is big and a big procedural. I'd break it up to smaller functions with sub private functions.

coordinator/src/main/scala/filodb.coordinator/RowSource.scala
```diff
@@ -79,27 +92,72 @@ trait RowSource extends Actor with StrictLogging {
...
       } else {
-        logger.debug(s"Marking isDoneReading as true: HiSeqNo = $currentHiSeqNo, lastAcked = $lastAckedSeqNo")
-        isDoneReading = true
+        logger.debug(s" ==> doneReading: HiSeqNo = $currentHiSeqNo, lastAcked = $lastAckedSeqNo")
+        if (currentHiSeqNo == lastAckedSeqNo) {
+          sendFlush()
+          context.become(doneReading)
+        }
+        else { context.become(doneReading) }
       }
```
helena: I'd break these case handlings out to private functions and just have

   case a => funcA(a)
   case b => funcB(b)

coordinator/src/main/scala/filodb.coordinator/RowSource.scala
```diff
@@ -79,27 +92,72 @@ trait RowSource extends Actor with StrictLogging {
...
+    case NodeCoordinatorActor.Ack(lastSequenceNo) =>
+      lastAckedSeqNo = lastSequenceNo
+      if (currentHiSeqNo == lastAckedSeqNo) sendFlush()
+
+    // Only until we get the Flushed signal do we know all rows are finished flushing.
+    case NodeCoordinatorActor.Flushed =>
+      finish()
   }
+
+  def sendFlush(): Unit = { coordinatorActor ! NodeCoordinatorActor.Flush(dataset, version) }
```
helena: def sendFlush(): Unit = coordinatorActor ! NodeCoordinatorActor.Flush(dataset, version)

coordinator/src/test/scala/filodb.coordinator/DatasetCoordinatorActorSpec.scala
```diff
@@ -0,0 +1,135 @@
...
+class DatasetCoordinatorActorSpec extends ActorTest(DatasetCoordinatorActorSpec.getNewSystem)
+with ScalaFutures {
+  import akka.testkit._
+  import SegmentSpec._
+  import DatasetCoordinatorActor._
+
+  implicit val defaultPatience =
+    PatienceConfig(timeout = Span(10, Seconds), interval = Span(50, Millis))
+
+  import scala.concurrent.ExecutionContext.Implicits.global
```
helena: Does ActorTest have an ActorSystem? I'm guessing yes because you call 'TestProbe()' which requires an implicit system. In which case I think you only need:

     import system.dispatcher

coordinator/src/test/scala/filodb.coordinator/DatasetCoordinatorActorSpec.scala
```diff
@@ -0,0 +1,135 @@
...
+
+  val config = ConfigFactory.parseString(
+                 """memtable.flush-trigger-rows = 100
+                    memtable.max-rows-per-table = 200""")
+                 .withFallback(ConfigFactory.load("application_test.conf"))
```
helena: At some point let's set up a common TestSettings class to handle config?

core/src/main/scala/filodb.core/columnstore/InMemoryColumnStore.scala
```diff
@@ -37,7 +37,11 @@ extends CachedMergingColumnStore with StrictLogging {
 
   def initializeProjection(projection: Projection): Future[Response] = Future.successful(Success)
 
-  def clearProjectionDataInner(projection: Projection): Future[Response] = ???
+  def clearProjectionDataInner(projection: Projection): Future[Response] = Future {
+    chunkDb.keys.filter { case (ds, _, _) => ds == projection.dataset }.foreach(chunkDb.remove)
+    rowMaps.keys.filter { case (ds, _, _) => ds == projection.dataset }.foreach(rowMaps.remove)
+    Success
+  }
```
helena: Maybe this and iterate once through each key collection using 'collect' versus twice using 'filter' + 'foreach'
 
     def clearProjectionDataInner(projection: Projection): Future[Response] = Future {
         chunkDb.keys.collect { case key if key._1 == projection.dataset => chunkDb remove key }
         rowMaps.keys.collect { case key if key._1 == projection.dataset => rowMaps remove key }
        Success
     }

core/src/main/scala/filodb.core/metadata/Dataset.scala
```diff
@@ -42,15 +49,19 @@ object DatasetOptions {
...
+ * Contains many helper functions especially pertaining to dataset partitioning.
+ */
 object Dataset {
   // If a partitioning column is not defined then this refers to a single global partition, and
   // the dataset must fit in one node.
   val DefaultPartitionColumn = ":single"
-  val DefaultPartitionKey: Types.PartitionKey = "/0"
+  val DefaultPartitionKey: PartitionKey = "/0"
 
   val DefaultOptions = DatasetOptions(chunkSize = 1000,
```
helena: are these all pubic API?

core/src/main/scala/filodb.core/metadata/Dataset.scala
```diff
@@ -93,4 +104,45 @@ object Dataset {
 
   def sortKeyHelper[K](dataset: Dataset, sortColumn: Column): Option[SortKeyHelper[K]] =
     sortKeyHelper[K](dataset.options, sortColumn)
+
+  case class BadPartitionColumn(reason: String) extends Exception("BadPartitionColumn: " + reason)
+
+  case class NullPartitionValue(partCol: String) extends Exception(s"Null partition value for col $partCol")
```
helena: I'd add a PartitionValueException as a base marker for these two

core/src/main/scala/filodb.core/metadata/Projection.scala
```diff
@@ -24,28 +27,59 @@ case class Projection(id: Int,
...
-  // TODO: error handling
-  def apply(dataset: Dataset, columns: Seq[Column], projectionId: Int = 0): RichProjection = {
+  case class BadSchema(reason: String) extends Exception("BadSchema: " + reason)
+
+  def apply[K](dataset: Dataset, columns: Seq[Column], projectionId: Int = 0): RichProjection[K] = {
+    val tryProj = make[K](dataset, columns, projectionId)
+    tryProj.recover {
+      case t: Throwable => throw t
+    }
+    tryProj.get
```
helena: Ouch, can't this throw an exception by calling .get in a bad state?

core/src/main/scala/filodb.core/metadata/Projection.scala
```diff
@@ -24,28 +27,59 @@ case class Projection(id: Int,
...
+      return fail(s"Unsupported sort column type ${sortColumn.columnType}")
+    }
+
+    val helper = Dataset.sortKeyHelper[K](dataset, sortColumn).get
+
+    for { partitionFunc <- Dataset.getPartitioningFunc(dataset, richColumns) } yield {
+      RichProjection[K](projectionId, dataset, sortColumn, sortColNo,
+                        normProjection.reverse, richColumns, helper, partitionFunc)
+    }
   }
```
helena: I'd break this out to smaller private functions and remove the need to call 'return'.

core/src/main/scala/filodb.core/metadata/Projection.scala
```diff
@@ -24,28 +27,59 @@ case class Projection(id: Int,
...
     val normProjection = dataset.projections(projectionId)
     val richColumns = {
       if (normProjection.columns.isEmpty) {
         columns
       } else {
         val columnMap = columns.map { c => c.name -> c }.toMap
+        val missing = normProjection.columns.toSet -- columnMap.keySet
+        if (missing.nonEmpty) return fail(s"Specified projection columns are missing: $missing")
         normProjection.columns.map(columnMap)
       }
```
helena: Would this work?    

     if (missing.nonEmpty) fail(s"Specified projection columns are missing: $missing")
     else normProjection.columns.map(columnMap)

core/src/main/scala/filodb.core/metadata/Projection.scala
```diff
@@ -24,28 +27,59 @@ case class Projection(id: Int,
...
+
     val normProjection = dataset.projections(projectionId)
     val richColumns = {
       if (normProjection.columns.isEmpty) {
         columns
       } else {
         val columnMap = columns.map { c => c.name -> c }.toMap
+        val missing = normProjection.columns.toSet -- columnMap.keySet
+        if (missing.nonEmpty) return fail(s"Specified projection columns are missing: $missing")
         normProjection.columns.map(columnMap)
```
helena: Would this work?    

     if (missing.nonEmpty) fail(s"Specified projection columns are missing: $missing")
     else normProjection.columns.map(columnMap)

core/src/main/scala/filodb.core/reprojector/MapDBMemTable.scala
```diff
@@ -22,32 +22,17 @@ import filodb.core.metadata.{Column, Dataset}
...
- *     max-rows-per-table = 1000000
- *     min-free-mb = 512
  *   }
  * }}}
  */
-class MapDBMemTable(config: Config) extends MemTable {
-  import MemTable._
-  import RowReader._
+class MapDBMemTable[K](val projection: RichProjection[K], config: Config) extends MemTable[K] {
   import collection.JavaConversions._
```
helena: import RowReader._

core/src/main/scala/filodb.core/reprojector/MapDBMemTable.scala
```diff
@@ -56,119 +41,53 @@ class MapDBMemTable(config: Config) extends MemTable {
...
+  val sortKeyFunc = projection.sortKeyFunc
+
+  implicit lazy val sortKeyOrdering: Ordering[K] = projection.helper.ordering
+  lazy val rowMap = db.createTreeMap("filo")
+                      .comparator(Ordering[(PartitionKey, K)])
+                      // Otherwise will pull out all values in a BTree Node at once
+                      .valuesOutsideNodesEnable()
+                      .counterEnable()
+                      .valueSerializer(Serializer.BYTE_ARRAY)
+                      .makeOrGet().asInstanceOf[BTreeMap[(PartitionKey, K), Array[Byte]]]
```
helena: Hrm, can we make makeOrGet accept a type param and have it maybe match on that internally and remove the need to clients calling it to have to do an InstanceOf reflection?

cli/src/main/scala/filodb.cli/CsvImportExport.scala
velvia: Hmm.   We probably do, it was more like the Cli was a quick and dirty thing.

cli/src/main/scala/filodb.cli/CsvImportExport.scala
velvia: Oh.  I think ScalaStyle should have caught this, but we might have turned off the style checker for parts of this file.

cli/src/main/scala/filodb.cli/CsvImportExport.scala
velvia: Yes, good point, will fix.

coordinator/src/main/scala/filodb.coordinator/DatasetCoordinatorActor.scala
velvia: Hmm.  Could group them under DatasetCoordinatorCommand or something.  Always tricky to know which actor/object to group commands and responses under.

cli/src/main/scala/filodb.cli/CsvImportExport.scala
helena: Yeah that's what I assumed - just a 'get it done' :-)
In my code reviews, just ignore any of my comments where I am wrong or didn't all all the info and my comment isn't applicable. Will save time :)

coordinator/src/main/scala/filodb.coordinator/DatasetCoordinatorActor.scala
velvia: This is not a bad scheme, though admittedly hard to do for me now.

coordinator/src/main/scala/filodb.coordinator/DatasetCoordinatorActor.scala
velvia: I think you'll like the upcoming solution  :)

coordinator/src/main/scala/filodb.coordinator/DatasetCoordinatorActor.scala
velvia: https://github.com/tuplejump/FiloDB/issues/33

coordinator/src/main/scala/filodb.coordinator/NodeCoordinatorActor.scala
velvia: I had all the responses extend NodeResponse as well.  The original `Response` was from earlier days, maybe get rid of it, though I like the idea of having a separate ErrorResponse thing.

coordinator/src/main/scala/filodb.coordinator/NodeCoordinatorActor.scala
velvia: Yeah will just delete.  There is no config now.

core/src/main/scala/filodb.core/columnstore/InMemoryColumnStore.scala
velvia: Great idea.

core/src/main/scala/filodb.core/metadata/Dataset.scala
velvia: public?  Depends on your definition of public.  Not external facing, like NodeCoordinatorActor, but public part of Dataset API, yes.

core/src/main/scala/filodb.core/metadata/Projection.scala
velvia: I know Scalaz has some validation stuff to help with that.  Have you used that before?

core/src/main/scala/filodb.core/metadata/Projection.scala
velvia: Sweet, this works!

core/src/main/scala/filodb.core/metadata/Projection.scala
velvia: No, cuz then richColumns would not have type `Seq[Column]` but something weird like Any.

core/src/main/scala/filodb.core/reprojector/MapDBMemTable.scala
velvia: Sorry, not sure what you mean exactly?  I prefer to place wildcard imports in a smaller scope, though I don't follow that rule universally myself.

velvia:
@helena Most of the CR comments have been addressed.

@satya noted your suggestions, will incorporate in subsequent work.

velvia:
BTW, found out why Travis CI is failing.  Apparently the Cassandra instance they have is configured differently, like with 256 virtual nodes or something, such that there are 256 token ranges, and we assume that there is 1 only in the tests.  :/

helena:
@velvia The overall code with changes LGTM, although I'm still getting familiar with the logic for several things ;)

velvia:
@helena thanks.  Will merge, so we can carry on with some work in parallel  :)
One area could definitely use some closer look is the ingestion protocol (`RowSource` and `NodeCoordinatorActor`, especially the back pressure.)

PR #34, opened 2015-10-28
fix scaladoc issues

darkjh:
Fix some scaladoc issues which prevent `sbt publish-local`. I could turn of scaladoc gen locally but IMO it's better to fix them.
However I didn't find a proper way to reference a project file, ideas are welcomed!


velvia:
@darkjh LGTM.  What do you mean "could not reference a project file"?

darkjh:
@velvia Thanks!
In fact the `[[doc/some_doc.md]]` should generate to a link to the file in scaladoc. But it does not seem to work. And I didn't find one thing in scaladoc that works for pointing an arbitrary  file in the project.

PR #35, opened 2015-10-29
Catch UnknownDataset error in truncateDataset op and correclty propag…

darkjh:
When trying to ingest data with the new changes I've got a scala match error over `UnknownDataset`. 
In fact `withDsCood` method sends back a `UnknownDataset` message if the correct dataset and version is not found. Here I catch the message and propagate an `NotFoundError` so that `saveAsFiloDataset` will call `createNewDataset` correctly.


velvia:
@darkjh LGTM, thanks.  What I don't get is that truncateDataset() should not even be called if the dataset is not found, because in `package.scala`:195, there is this code:
```scala
      try {
        val datasetObj = getDatasetObj(dataset)
        if (mode == SaveMode.Overwrite) truncateDataset(datasetObj, version)
      } catch {
        case e: NotFoundError =>
          createNewDataset(dataset, sortColumn, partCol, defaultPartitionKey, segmentSize, df1)
      }
```

PR #44, opened 2015-11-10
Added support for Cassandra Authentication

rstrickland:
This PR adds config keys for username and password to application.conf, as follows:

    cassandra {
        hosts = ["localhost"]
        port = 9042
        keyspace = "filodb"
        username = "cassandra"
        password = "cassandra"
    }

There are two tests that appear to be failing to pick up the auth settings.  After stepping through with the debugger I am unable to determine the root cause of this, as the session in scope always contains the settings.  My real-world tests don't exhibit this behavior, so I'm at a bit of a loss as to where the issue is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,8 @@
 [![Join the chat at https://gitter.im/velvia/FiloDB](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/velvia/FiloDB?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 Distributed.  Columnar.  Versioned.  Streaming.  SQL.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -3,6 +3,8 @@
 [![Join the chat at https://gitter.im/velvia/FiloDB](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/velvia/FiloDB?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 Distributed.  Columnar.  Versioned.  Streaming.  SQL.
 
+## NOTE: The repo has been relocated to [TupleJump](http://github.com/tuplejump/FiloDB).
```
velvia: Hmm, was this intended for our main repo?

velvia:
@rstrickland many thanks!    We can investigate the test issue later.   Going to merge.

Issue #27, opened 2015-10-05
Automatic segment length determination


velvia:
Currently, each segment has some fixed length, and it is hard wired and not configurable.  This obviously won't work for generic data ingestion.  For example, for time series ingestion you want to be able to fit a certain number of points per segment.  Or, maybe what you want is for each segment to correspond to a certain deviceId, or subpart of the deviceId time series.  So, this has to be

a.  Configurable, or
b. better yet, automatically determined.

velvia:
This change has been deprecated by the new segment key, so the above is not necessary anymore.

PR #45, opened 2015-11-15
Automatic/Dynamic Segmentation

velvia:
This PR implements #27 and has a large number of changes.  It also serves as the basis for changes to implement composite sort keys (well, without this change it would be really hard to configure segment length for composite sort keys).

User-facing changes:
- segment_size is no longer a dataset/configuration option.   Instead a new option will be coming, which is what most people really want: to configure the min and max segment size by # of rows, not by semantic meaning based on segment type.
- A new CLI command, `analyze` exists to analyze the distribution of segments and chunks in a dataset
- Reading efficiency of segment indexes has been improved
- `splits_per_node` has been fixed and defaults to 4 splits, improving most query times;  before it was limited to effectively 1 partition/thread per node

Ingestion changes:
- A new step is added: before the memtable is flushed, first, a new class `SegmentChopper` is used to determine and add new segments based on the rows in the memtable;  Compare and Swap is used to make sure that even in the presence of multiple writers, all writers agree on the segment boundaries.  The `Reprojector` now adheres strictly to the segment boundaries.
- A new table is used to record the segment start, end, and number of rows

Overall, core module testing has been expanded.

Status:
- Most tests pass except for the Spark tests; these will be worked on


velvia:
After discussion with @trulite, going to close this PR.  Instead, the initial changes will be spun out to a separate PR.

Issue #38, opened 2015-11-06
CLI does not ingest all rows


velvia:
From Gitter channel:

i found out some interesting things:
First: i used filo-cli to import data, check filo-cli.log i found:
```
[2015-11-06 19:47:03,113] INFO  f.c.DatasetCoordinatorActor [filo-cli-akka.actor.default-dispatcher-9] - MemTable active table rows: 44999
[2015-11-06 19:47:03,113] INFO  f.c.DatasetCoordinatorActor [filo-cli-akka.actor.default-dispatcher-9] - MemTable flushing rows: 55000
```
Then: i use spark-shell to check how many records in dataset gdelt and got 55000 records. It equals with 55k records was flushed by memTable. It seem memTable belongs to filo-cli process and it doesn't flush active table rows on exit. So we lost 45k records.

I'll bet it's more of a timing problem.  RowSource sends `Flush` message to `NodeCoordinatorActor`, but it may not flush everything.  The CLI is not persistent, and quits before everything is flushed.   Instead, let's implement a `FinishIngestion` message, which flushes everything at the end, including the active table....

PR #46, opened 2015-11-18
Performance and Ingestion Improvements

velvia:
This includes a bunch of important performance and ingestion improvements;  in short - ingestion is MUCH faster, stable, and exact;  query is faster due to concurrency working.

The biggest change, towards the end, is a new `FiloMemTable` which uses Filo vectors to store data in the memtable; it is significantly faster than the MapDB-based MemTable for both writing and reading.

Other changes:
- Added a `StringKeyHelper` and allow String as a sort key type.  The segment key is derived as the first `segment_size` characters of the sort key.
- A new CLI command, analyze exists to analyze the distribution of segments and chunks in a dataset
- Reading efficiency of segment indexes has been improved
- splits_per_node has been fixed and defaults to 4 splits, improving most query times; before it was limited to effectively 1 partition/thread per node
- Fix for #38 
- Two fixes for `RowSource` so that Spark ingestion no longer hangs at the end, and is at-least-once by design


velvia:
Latest change fixes issue #38

cli/src/main/scala/filodb.cli/CsvImportExport.scala
```diff
@@ -68,6 +68,17 @@ trait CsvImportExport {
...
+      case DatasetCoordinatorActor.Stats(_, _, _, activeRows, _) =>
+        if (activeRows > 0) {
+          println(s"Still $activeRows left to flush in active memTable, trigger another flush....")
+          actorAsk(coordinatorActor, NodeCoordinatorActor.Flush(dataset, version), timeout) {
+            case NodeCoordinatorActor.Flushed =>
+          }
+        }
+    }
+
     println("ingestCSV finished!")
```
helena: println(s"Ingestionof $csvPath completed.")

cli/src/main/scala/filodb.cli/CsvImportExport.scala
```diff
@@ -68,6 +68,17 @@ trait CsvImportExport {
       case RowSource.AllDone =>
     }
 
+    // There might still be rows left after the latest flush is done, so initiate another flush
```
helena: I might add a default explicitly

     private val DefaultTimeout: FiniteDuration = 5 seconds

    protected def parse[T, B](cmd: => Future[T], awaitTimeout: FiniteDuration = DefaultTimeout)(func: T => B): B = 
        func(Await.result(cmd, awaitTimeout))

Why explicitly execute a Future op if the op is run in an explicit Await on the thread?
     protected def actorAsk[B](actor: ActorRef, msg: Any,
                                          awaitTimeout: FiniteDuration = DefaultTimeout)(f: PartialFunction[Any, B]): B = {
           implicit val timeout = Timeout(awaitTimeout)
           parse(actor ? msg, askTimeout)(f)
     }

cli/src/main/scala/filodb.cli/CsvImportExport.scala
```diff
@@ -68,6 +68,17 @@ trait CsvImportExport {
...
+    actorAsk(coordinatorActor, NodeCoordinatorActor.GetIngestionStats(dataset, version)) {
+      case DatasetCoordinatorActor.Stats(_, _, _, activeRows, _) =>
+        if (activeRows > 0) {
+          println(s"Still $activeRows left to flush in active memTable, trigger another flush....")
+          actorAsk(coordinatorActor, NodeCoordinatorActor.Flush(dataset, version), timeout) {
+            case NodeCoordinatorActor.Flushed =>
```
helena: No return Type or validation response needed?

helena: All these println statements vs logging in the cli - curious why?

coordinator/src/main/scala/filodb.coordinator/DatasetCoordinatorActor.scala
```diff
@@ -121,7 +121,7 @@ private[filodb] class DatasetCoordinatorActor[K](projection: RichProjection[K],
   var flushesFailed = 0
 
   def makeNewTable(): MemTable[K] = {
-    new MapDBMemTable(projection, config)
+    new FiloMemTable(projection, config)
   }
```
helena: def makeNewTable: MemTable[K] =  new FiloMemTable(projection, config)

coordinator/src/main/scala/filodb.coordinator/DatasetCoordinatorActor.scala
```diff
@@ -168,6 +168,7 @@ private[filodb] class DatasetCoordinatorActor[K](projection: RichProjection[K],
       logger.warn("This should not happen. Not starting flush; $curReprojection; $flushingTable")
       return
     }
+    activeTable.forceCommit()
     flushingTable = Some(activeTable)
     activeTable = makeNewTable()
     val newTaskFuture = reprojector.reproject(flushingTable.get, version)
```
helena: What happens if 'flushingTable.get' is None? i.e. errors...

coordinator/src/main/scala/filodb.coordinator/DatasetCoordinatorActor.scala
```diff
@@ -168,6 +168,7 @@ private[filodb] class DatasetCoordinatorActor[K](projection: RichProjection[K],
       logger.warn("This should not happen. Not starting flush; $curReprojection; $flushingTable")
       return
     }
```
helena: The raw file doesn't show 'return' in the signature, but above it 'looks' as though you've got a 'return' for a Unit signature.

     private def reportStats(): Unit = {
        logger.info(s"MemTable active table rows: $activeRows")
        logger.info(s"MemTable flushing rows: $flushingRows")
    }

coordinator/src/main/scala/filodb.coordinator/RowSource.scala
```diff
@@ -23,24 +24,20 @@ object RowSource {
...
- * that we do not skip any incoming messages.
+ * Backpressure and at least once mechanism:  RowSource sends batches of rows to the NodeCoordinator,
+ * but does not send more than maxUnackedBatches before getting an ack back.  As long as it receives
+ * acks it will keep sending, but will stop once maxUnackedBatches is reached.  It then goes into a waiting
+ * state, waiting for Acks to come back.  Once waitingPeriod has elapsed, if it is allowed to ingest again,
+ * then it will replay unacked messages, hopefully get acks back, and once all messages are acked will go
+ * back into regular ingestion mode.  This ensures that we do not skip any incoming messages.
  */
 trait RowSource extends Actor with StrictLogging {
   import RowSource._
```
helena: Not part if this change set but in looking at the raw source I'd suggest this change:

     private var whoStartedMe: Option[ActorRef] = None

Then: whichever makes sense
 
     for (identity <- whoStartedMe) identity ! AllDone
     for (identity <- whoStartedMe) identity = sender
     whoStartedMe  map (_ ! AllDone)
     whoStartedMe  map (_ = sender)

coordinator/src/main/scala/filodb.coordinator/RowSource.scala
```diff
@@ -85,74 +75,86 @@ trait RowSource extends Actor with StrictLogging {
 
   def reading: Receive = {
     case GetMoreRows =>
```
helena: A lot of code to reside in a receive.. personally I'd make it clean, save it to a function and place the processing in that function.

coordinator/src/main/scala/filodb.coordinator/RowSource.scala
```diff
@@ -85,74 +75,86 @@ trait RowSource extends Actor with StrictLogging {
...
         }
       } else {
-        logger.debug(s" ==> doneReading: HiSeqNo = $currentHiSeqNo, lastAcked = $lastAckedSeqNo")
-        if (currentHiSeqNo == lastAckedSeqNo) sendFlush()
+        logger.info(s" ==> (${self.path.name}) doneReading: outstanding seqIds = ${outstanding.keys}")
+        if (outstanding.isEmpty)  { sendFlush() }
+        else {
+          schedule(waitingPeriod, CheckCanIngest)
+        }
         context.become(doneReading)
```
helena: I feel like we can become doneReading even if an error is thrown but I'd need to check out the code to see it better.

coordinator/src/main/scala/filodb.coordinator/RowSource.scala
```diff
@@ -85,74 +75,86 @@ trait RowSource extends Actor with StrictLogging {
...
           context.become(waiting)
         }
       } else {
-        logger.debug(s" ==> doneReading: HiSeqNo = $currentHiSeqNo, lastAcked = $lastAckedSeqNo")
-        if (currentHiSeqNo == lastAckedSeqNo) sendFlush()
+        logger.info(s" ==> (${self.path.name}) doneReading: outstanding seqIds = ${outstanding.keys}")
+        if (outstanding.isEmpty)  { sendFlush() }
+        else {
+          schedule(waitingPeriod, CheckCanIngest)
+        }
```
helena: if (outstanding.isEmpty) sendFlush() 
else schedule(waitingPeriod, CheckCanIngest)

coordinator/src/main/scala/filodb.coordinator/RowSource.scala
helena: No, probably not the case :)

coordinator/src/main/scala/filodb.coordinator/RowSource.scala
```diff
@@ -85,74 +75,86 @@ trait RowSource extends Actor with StrictLogging {
...
         context.become(doneReading)
       }
```
helena: If you did 2 cases, you can remove the need for an if/else

     case GetMoreRows if batchIterator.hasNext =>
     case GetMoreRows =>

coordinator/src/main/scala/filodb.coordinator/RowSource.scala
```diff
@@ -85,74 +75,86 @@ trait RowSource extends Actor with StrictLogging {
...
     case NodeCoordinatorActor.Flushed =>
       finish()
   }
 
+  // If we don't get acks back, need to keep trying
+  def doneReading: Receive = doneReadingAck orElse replay
+
   def sendFlush(): Unit = coordinatorActor ! NodeCoordinatorActor.Flush(dataset, version)
 
+  private var scheduledTask: Option[Cancellable] = None
```
helena: nice

core/src/main/scala/filodb.core/SortKeyHelper.scala
```diff
@@ -80,3 +80,24 @@ case class DoubleKeyHelper(segmentLen: Double) extends SingleSortKeyHelper[Doubl
...
+/**
+ * Right now, you have to specify a prefixLen for the string key helper.
+ * All string keys with identical prefixes (characters 0 to prefixLen - 1) will then be
+ * bucketed into the same segment.
+ * Thus one needs to look through your string sort key data and make sure to bucketize
+ * correctly.
+ *
+ * TODO: perhaps combine prefixLen with a range of chars the last char is allowed to vary on
+ */
+case class StringKeyHelper(prefixLen: Int) extends SingleSortKeyHelper[String] {
```
helena: Just to shy away from repetitive strings in the codebase... and also duplicate configurations:

     private final val Utf8Encoding = "UTF-8"

And then

     ByteVector(key.take(prefixLen).getBytes(Utf8Encoding))
     new String(bytes.toArray, Utf8Encoding)

helena: Might be re-usable around the filodb codbase though and belong outside this type.

core/src/main/scala/filodb.core/columnstore/Analyzer.scala
```diff
@@ -0,0 +1,92 @@
...
+      case x: Int if x < 0  =>  -x - 2
+      case x: Int if x >= bucketKeys.size =>  bucketKeys.size - 1
+      case x: Int => x
+    }
+    val key = bucketKeys(bucket)
+    Histogram(Math.min(elem, this.min),
+              Math.max(elem, this.max),
+              this.sum + elem,
+              this.numElems + 1,
+              buckets + (key -> (buckets.getOrElse(key, 0) + 1)))
```
helena: copy(
         min = Math.min(elem, this.min),
         max = Math.max(elem, this.max),
         sum = this.sum + elem,
         numElems = this.numElems + 1,
         buckets = buckets + (key -> (buckets.getOrElse(key, 0) + 1)))

core/src/main/scala/filodb.core/columnstore/Analyzer.scala
```diff
@@ -0,0 +1,92 @@
...
+                               chunksInSegment: Histogram,
+                               segmentsInPartition: Histogram) {
+  def prettify(): String = {
+    s"ColumnStoreAnalysis\n  numSegments: $numSegments\n  numPartitions: $numPartitions\n" +
+    rowsInSegment.prettify("# Rows in a segment") +
+    chunksInSegment.prettify("# Chunks in a segment") +
+    segmentsInPartition.prettify("# Segments in a partition")
+  }
+}
+
```
helena: Would it make more sense, since this creates a ColumnStoreAnalysis, that it is an apply?
     object ColumnStoreAnalysis  {
 
         def apply(cs: CachedMergingColumnStore, dataset: TableName, version: Int): ColumnStoreAnalysis = {
            var numSegments = 0
            var rowsInSegment: Histogram = Histogram.empty
            var chunksInSegment: Histogram = Histogram.empty
            var segmentsInPartition: Histogram = Histogram.empty
            val partitionSegments = (new collection.mutable.HashMap[PartitionKey, Int]).withDefaultValue(0)

           for { 
              param                        <- cs.getScanSplits(dataset)
              rowmaps                    = Await.result(cs.scanChunkRowMaps(dataset, version, (p: PartitionKey) => true, 
                                                        params = param), 5.minutes)
              (partKey, _, rowmap) <- rowmaps 
           } yield {
               // Figure out # chunks and rows per segment
               val numRows = rowmap.chunkIds.length
               val numChunks = rowmap.nextChunkId

               numSegments = numSegments + 1
               rowsInSegment = rowsInSegment.add(numRows, NumRowsPerSegmentBucketKeys)
               chunksInSegment = chunksInSegment.add(numChunks, NumChunksPerSegmentBucketKeys)
               partitionSegments(partKey) += 1
          }

          for { (partKey, numSegments) <- partitionSegments } {
              segmentsInPartition = segmentsInPartition.add(numSegments, NumSegmentsBucketKeys)
          }

          ColumnStoreAnalysis(numSegments, partitionSegments.size,
                rowsInSegment, chunksInSegment, segmentsInPartition)
        }
     }

core/src/main/scala/filodb.core/columnstore/Analyzer.scala
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    for { (partKey, numSegments) <- partitionSegments } {
+      segmentsInPartition = segmentsInPartition.add(numSegments, NumSegmentsBucketKeys)
+    }
+
+    ColumnStoreAnalysis(numSegments, partitionSegments.size,
+                        rowsInSegment, chunksInSegment, segmentsInPartition)
+  }
+}
```
helena: Would it make more sense, since this creates a ColumnStoreAnalysis, that it is an apply?
     object ColumnStoreAnalysis  {
 
         def apply(cs: CachedMergingColumnStore, dataset: TableName, version: Int): ColumnStoreAnalysis = {
            var numSegments = 0
            var rowsInSegment: Histogram = Histogram.empty
            var chunksInSegment: Histogram = Histogram.empty
            var segmentsInPartition: Histogram = Histogram.empty
            val partitionSegments = (new collection.mutable.HashMap[PartitionKey, Int]).withDefaultValue(0)

           for { 
              param                        <- cs.getScanSplits(dataset)
              rowmaps                    = Await.result(cs.scanChunkRowMaps(dataset, version, (p: PartitionKey) => true, 
                                                        params = param), 5.minutes)
              (partKey, _, rowmap) <- rowmaps 
           } yield {
               // Figure out # chunks and rows per segment
               val numRows = rowmap.chunkIds.length
               val numChunks = rowmap.nextChunkId

               numSegments = numSegments + 1
               rowsInSegment = rowsInSegment.add(numRows, NumRowsPerSegmentBucketKeys)
               chunksInSegment = chunksInSegment.add(numChunks, NumChunksPerSegmentBucketKeys)
               partitionSegments(partKey) += 1
          }

          for { (partKey, numSegments) <- partitionSegments } {
              segmentsInPartition = segmentsInPartition.add(numSegments, NumSegmentsBucketKeys)
          }

          ColumnStoreAnalysis(numSegments, partitionSegments.size,
                rowsInSegment, chunksInSegment, segmentsInPartition)
        }
     }

core/src/main/scala/filodb.core/columnstore/ColumnStore.scala
```diff
@@ -270,8 +270,8 @@ trait CachedMergingColumnStore extends ColumnStore with StrictLogging {
                                               schema: Seq[Column]): Seq[Segment[K]] = {
     val helper = implicitly[SortKeyHelper[K]]
     val segments = rowMaps.map { case (segmentId, rowMap) =>
-        val (segStart, segEnd) = helper.getSegment(helper.fromBytes(segmentId))
-        val segKeyRange = origKeyRange.copy(start = segStart, end = segEnd, endExclusive = true)
+        val segStart = helper.fromBytes(segmentId)
```
helena: Would we get an error if segmentId are null/empty?

core/src/main/scala/filodb.core/metadata/Dataset.scala
```diff
@@ -88,6 +89,8 @@ object Dataset {
                                     asInstanceOf[SortKeyHelper[K]]
       case java.lang.Double.TYPE => (new DoubleKeyHelper(options.segmentSize.toDouble)).
                                     asInstanceOf[SortKeyHelper[K]]
+      case StringClass         => (new StringKeyHelper(options.segmentSize.toInt)).
+                                    asInstanceOf[SortKeyHelper[K]]
     }
```
helena: Hrm, trying to envision why these StringKeyHelper, DoubleKeyHelper, LongKeyHelper need to be asInstanceOf-ed when they are already of type SortKeyHelper

core/src/main/scala/filodb.core/metadata/Dataset.scala
```diff
@@ -88,6 +89,8 @@ object Dataset {
                                     asInstanceOf[SortKeyHelper[K]]
       case java.lang.Double.TYPE => (new DoubleKeyHelper(options.segmentSize.toDouble)).
                                     asInstanceOf[SortKeyHelper[K]]
+      case StringClass         => (new StringKeyHelper(options.segmentSize.toInt)).
+                                    asInstanceOf[SortKeyHelper[K]]
```
helena: Would we ever get a match error with no case _ ? 

      case _ =>

core/src/main/scala/filodb.core/reprojector/FiloMemTable.scala
```diff
@@ -0,0 +1,216 @@
...
+      require(serializer == Column.Serializer.FiloSerializer)
+      VectorInfo(name, colType.clazz)
+  }
+  private val clazzes = filoSchema.map(_.dataType).toArray
+  private val colIdToNumber =
+    projection.columns.zipWithIndex.map { case (col, idx) => (col.name, idx) }.toMap
+
+  // Holds temporary rows before being flushed to chunk columnar storage
+  private val tempRows = new ArrayBuffer[RowReader]
+  private val callbacks = new ArrayBuffer[(Int, Int, () => Unit)]   // Start and end row index for each callback
```
helena: Hrm, wondering about the usage of ArrayBuffer and HashMap '.synchronized' - this is truly "initial cases only" ?

core/src/main/scala/filodb.core/reprojector/FiloMemTable.scala
```diff
@@ -0,0 +1,216 @@
...
+      readers.remove(readers.length - 1, 1)
+    }
+
+    // Add new rows to builder, adding to partKeyMap in the meantime
+    val baseLength = builder.builders.head.length   // Nonzero only if partial chunks added
+    val rowsToAdd = Math.min(chunkSize - baseLength, tempRows.length)
+    val nextChunkIndex = chunks.length
+    val sortKeyFunc = projection.sortKeyFunc
+
+    for { i <- 0 until rowsToAdd optimized } {
```
helena: cool - using optimized

core/src/main/scala/filodb.core/reprojector/FiloMemTable.scala
```diff
@@ -0,0 +1,216 @@
...
+  /**
+   * === Row ingest, read, delete operations ===
+   */
+  def ingestRows(rows: Seq[RowReader])(callback: => Unit): Unit = if (rows.nonEmpty) {
+    tempRows.synchronized {
+      val oldLen = tempRows.length
+      tempRows ++= rows
+      callbacks += ((oldLen, tempRows.length - 1, { () => callback }))
+    }
+    while (tempRows.length >= chunkSize) flushRowsToChunks()
```
helena: Maybe instead of java-like while loops, something like
 
    for {
        row <- tempRows
        if tempRows.length >= chunkSize
     } flushRowsToChunks()

core/src/main/scala/filodb.core/reprojector/FiloMemTable.scala
```diff
@@ -0,0 +1,216 @@
...
+    // If rows have not been flushed, schedule a task in the future to flush it
+    if (tempRows.nonEmpty && !flushTask.isDefined) {
+      val fut = scheduler.schedule(new Runnable { def run: Unit = { flushRowsToChunks() }},
+                                   flushInterval.toMillis,
+                                   TimeUnit.MILLISECONDS)
+      flushTask = Some(fut)
+    }
```
helena: Cleaner?

     if (tempRows.nonEmpty && flushTask.isEmpty) ...

core/src/main/scala/filodb.core/reprojector/FiloMemTable.scala
```diff
@@ -0,0 +1,216 @@
...
+      builder.addRow(row)
+      val keyMap = getKeyMap(projection.partitionFunc(row))
+      keyMap.put(sortKeyFunc(row), chunkRowIdToLong(nextChunkIndex, baseLength + i))
+    }
+
+    // Add chunks
+    val chunkAray = new Array[ByteBuffer](filoSchema.length)
+    builder.convertToBytes().foreach { case (colName, bb) => chunkAray(colIdToNumber(colName)) = bb }
+    chunks += chunkAray
+    readers += new FastFiloRowReader(chunkAray, clazzes)
```
helena: Wondering if this is all that's needed here? Again, not in an IDE so I'm probably way off :)

     val chunkArray = for ((cname, bb) <- builder.convertToBytes()) yield bb
     chunks += chunkArray

core/src/main/scala/filodb.core/reprojector/FiloMemTable.scala
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+  /**
+   * Forces all the rows in tempRows to be committed into the chunk store and disk storage / WAL.
+   * This might be necessary before flushing the memtable, or for testing.
+   * As a result, no background flush task needs to be scheduled since all rows will be flushed.
+   */
+  def forceCommit(): Unit = {
+    while (tempRows.nonEmpty) flushRowsToChunks()
+  }
```
helena: Or maybe..
     
     def forceCommit(): Unit = for (n <- tempRows) flushRowsToChunks()

core/src/main/scala/filodb.core/reprojector/MapDBMemTable.scala
```diff
@@ -84,6 +84,8 @@ class MapDBMemTable[K](val projection: RichProjection[K], config: Config) extend
 
   def numRows: Int = rowMap.size().toInt
 
+  def forceCommit(): Unit = {}
+
   def clearAllData(): Unit = {
     logger.info(s"MemTable: ERASING ALL TABLES!!")
```
helena: Seeing phrases in tools like Splunk with double exclamation points paired with words like 'erasing' can be frightening. Just sayin'.

core/src/main/scala/filodb.core/reprojector/Reprojector.scala
```diff
@@ -65,14 +65,13 @@ class DefaultReprojector(columnStore: ColumnStore)
...
       val segment = new RowWriterSegment(keyRange, memTable.projection.columns)
       logger.debug(s"Created new segment $segment for encoding...")
 
       // Group rows into chunk sized bytes and add to segment
-      segmentRowsIt.grouped(dataset.options.chunkSize).foreach { chunkRowsIt =>
-        val chunkRows = chunkRowsIt.toSeq
-        segment.addRowsAsChunk(chunkRows)
+      while (segmentRowsIt.nonEmpty) {
+        segment.addRowsAsChunk(segmentRowsIt.take(dataset.options.chunkSize))
       }
```
helena: Maybe a for comp instead?

     for (n <- segmentRowsIt) {
        segment.addRowsAsChunk(segmentRowsIt.take(dataset.options.chunkSize))
     }

core/src/test/resources/application_test.conf
```diff
@@ -26,6 +26,9 @@
 
     # Set the free memory requirement much lower for running tests
     min-free-mb = 10
+
+    filo.chunksize = 100
+    flush.interval = 5 s    # Make this high enough that tests will require manual flushing usually
```
helena: May be easier to read this way, and it's how the akka confs are written, just a thought.

     # Make this high enough that tests will require manual flushing usually
     flush.interval = 5 s

helena:
Mostly code style cleanup of while loops, 'return' statements and code compartmentalization thoughts.

cli/src/main/scala/filodb.cli/CsvImportExport.scala
velvia: Hmm yeah, that last one could definitely be logged.   And yes, need to add some other types ... MatchError is not that friendly of an error message  :)

coordinator/src/main/scala/filodb.coordinator/DatasetCoordinatorActor.scala
velvia: `flushingTable` cannot be None.... see line 172.

coordinator/src/main/scala/filodb.coordinator/DatasetCoordinatorActor.scala
velvia: Sorry, am confused what you're referencing.... those lines are from `startFlush` which has a signature return of `Unit`.

coordinator/src/main/scala/filodb.coordinator/RowSource.scala
velvia: Yeah, could defintiely use more eyes here.

coordinator/src/main/scala/filodb.coordinator/RowSource.scala
velvia: My current ScalaStyle settings have { } required after multiline if else.  I think this is a good idea in general because after code changes sometimes non obvious changes can lead to change of meaning.

core/src/main/scala/filodb.core/columnstore/Analyzer.scala
velvia: Hmmm, maybe.  However this does a lot of work, which takes an indefinitely amount of time, and those usually seem to be beyond the scope of an `apply`, which are usually for simple object creation.....

core/src/main/scala/filodb.core/columnstore/ColumnStore.scala
velvia: Yeah we would.  It should never be null though, so it would be a legit error.

core/src/main/scala/filodb.core/metadata/Dataset.scala
velvia: Yeah the Scala compiler doesn't seem to like it.   :/

core/src/main/scala/filodb.core/metadata/Dataset.scala
velvia: We shouldn't, because it would be outside the defined types, and in this case MatchError would probably be right.

core/src/main/scala/filodb.core/reprojector/FiloMemTable.scala
velvia: Hmmm ... the synchronized is used on an ongoing basis, because scheduled flushes happen on a different thread, and potentially at the same time as an insert which modifies the tempRows and callbacks.

core/src/main/scala/filodb.core/reprojector/FiloMemTable.scala
velvia: That's not the same thing though.  This is not a for loop, but truly a "flush while there are still rows" thing.... i.e. one of the few times when a while loop really makes sense.  :)

core/src/main/scala/filodb.core/reprojector/FiloMemTable.scala
velvia: Oh yeah, thanks.

core/src/main/scala/filodb.core/reprojector/FiloMemTable.scala
velvia: Well, convertToBytes yields a map of column name to ByteBuffer, and we need the `colIdToNumber` mapping to convert the map back to an ordered array.  I suppose we could use for {} instead of foreach, or possibly even better is to have an ordered list of column names, then we could directly convert to an array.

core/src/main/scala/filodb.core/reprojector/FiloMemTable.scala
velvia: Again, not the same thing, because each call to flushRowsToChunks() operates on a chunk of tempRows, or all of it.

core/src/main/scala/filodb.core/reprojector/MapDBMemTable.scala
velvia: I hope this never makes it into Splunk or a production environment.... it's really just for tests  :)

core/src/main/scala/filodb.core/reprojector/Reprojector.scala
velvia: Sorry, I know you don't like while's but this is another case where `segmentRowsIt.take(...)` takes a chunk of segmentRows at a time.  I did this instead of `grouped` because, for performance reasons, the objects in the iterator are not immutable, and `grouped` turns things into Seq's, and i wanted to keep it a pure Iterator. Also grouped is apparently really slow.  Will add a comment though.

velvia:
@helena addressed most CR comments.  Merge if it looks good  :)

coordinator/src/main/scala/filodb.coordinator/DatasetCoordinatorActor.scala
helena: Yeah it confused me, I saw in the code to review, that it ended with:

logger.warn("This should not happen. Not starting flush; $curReprojection; $flushingTable")
return

Odd.

Issue #30, opened 2015-10-15
Support composite sort and partition keys


velvia:
Today FiloDB only supports sort key from a single column.   Let's support multiple component sort keys.  This should be relatively contained - at least the core changes, yet be a good introduction to different parts of the code.  Change list that comes to mind:

* Create a new SortKeyHelper that can support multiple column sort keys.  Perhaps the following:
    * `TwoColumnSortKeyHelper extends SortKeyHelper[(A, B)]` 
    * `ThreeColumnSortKeyHelper` etc.
* Modify `MergeStrategy`, `Projection` to support multiple sort keys.  It should be a `Seq[String]`, and the earlier columns listed take priority.
* Figure out how to persist SegmentIDs etc. for multiple column sort keys
* Support multiple sort keys in Spark and CLIs

velvia:
Note: we will probably move to a model where there is a separate Segment key from the sort key.  The sort key is the primary key within a partition, whereas all records with the same segment key map into the same segment.  Within a segment, rows will be sorted by the sort key.

The segment key does not need to be a physically separate column.  It could, and may usually be, a computed function on another column.  For example, let's say for a time-series example, the sort key is a composite of ["timestamp", "uuid"].   Let's say we want to group all records into 30-second buckets.    Then the segment key would be `(timeslice timestamp 30s)` which is a function slicing the timestamp column (assumed to be milliseconds) into 30-second buckets.

velvia:
Proposed initial set of expressions:

* `(timeslice <ms-timestamp-col> <timeunitstr>)`
* `(round <int/long col> <number>)` -- the default 10000 segment len would be the function `(round colA 10000)`
* `(stringPrefix <str col> <numChars>)` -- gets the first numChars of the string

Some relevant code, for rounding:
```scala
  def getSegment(key: Long): (Long, Long) = {
    val segmentNum = key / segmentLen
    (segmentNum * segmentLen, (segmentNum + 1) * segmentLen)
  }
```

for string prefixing:
```scala
  def getSegment(key: String): (String, String) = {
    val start = key.take(prefixLen)
    val end = start.take(start.length - 1) + ((start(start.length - 1) + 1).toChar)
    (start, end)
  }
```

velvia:
Also:

* `(getOrElse <col name> <default value>)` -- provide a default value if the column value is null
* `(autoid)` -- generate an auto incrementing ID within each partition (need to think about how this works - might be a TimeUUID)

PR #59, opened 2016-02-09
Multi-column partition / row keys; separate segment key; computed columns; ++ Filtering (#30)

velvia:
Fixes: #30 

This is probably the biggest change to come to FiloDB, as it changes the data model.  Overall this change brings a huge bunch of features many folks have been waiting for.

Reviewers: I suggest starting by reading this entire description and the README, in particular the data modeling and computed column section first.  Have a look here: https://github.com/tuplejump/FiloDB/commit/c222f5d2fcf08f73ebb00433fa7716c60211fb17

## Split the sort key concept into row key and segment key

Before, records were grouped into segments using a difficult-to-configure "segmentLength" parameter, which was a function of the sort key.  Now, there are two separate keys for clarity: a *row key*, which acts as the primary key for insertion of rows, but has no sorting or segmenting function, and a separate *segment key*, which groups rows into segments, and partitions are sorted by segment key.   This makes it much easier for users to control segmentation (and later, filtering within a partition) separately from what makes each row unique.  Later on, separate projections could be created just by adjusting the segment key.

## Multi-column partition and row keys

Users can define partition and row keys using multiple columns.  The `RichProjection`, `Projection` and `Dataset` classes all contain provisions to support multiple columns for these keys.  What's more, a base `KeyType` type class is used instead of the previous `SortKeyHelper`, as any type of key can contain multiple columns and be serialized accordingly.  Note that the segment key cannot be multi column, because we have not figured out how to serialize the multi column keys in such a way that they could be binary compared for range scans.

Note the implications this has for filtering and computed columns, below.

Another change is that none of the partition, row, and segment keys can be null now.

## Computed Columns

To facilitate computation off the segment key, and get around the non-null requirement of keys, a new feature called Computed Columns allows users to compute values for the keys.  

* `:string <stringVal>` - constant value, used as a default for partition key
* `:getOrElse colName default` - used to provide a default if null encountered
* `:round colname #` - really useful for time slicing or creating segment keys from row numbers
* `:stringPrefix stringCol numChars` - can be used as a crude hashing mechanism for string columns to be used as partition keys

Nested computations are not yet supported, but computed columns could be used in place of a column name, and multiple computed columns could be specified to make up row or partition keys.

## Enhanced partition key filtering

Users can now use = or IN filters on any column of the partition key.  Unlike Cassandra, which has very strict requirements, FiloDB can filter down the number of partitions to scan significantly faster even if only one column is restricted.  This allows for far more flexible data models and lessens the number of aggregate/reporting tables needed compared to traditional Cassandra CQL.

## Other enhancements

- Support for java.sql.Timestamp / Spark SQL Timestamp columns
- MapDBMemTable is gone
- Changes in the options in filo-cli and Spark, per all the changes above.  Notably, default_partition_key is gone (use :getOrElse instead)
- Layout of C* tables are changed - thus this is a breaking change
- Deadlocks in ingestion logic (ColumnStore as well as FiloMemTable) were found and fixed; FiloMemTable logic is spread out and simplified
- Better error handling/supervision in coordinator
- Fix for computing splits for Cassandra instances with VNodes
- Config support for changing Cassandra read/connect timeouts, default consistency levels
- Less Futures, more efficient CassandraColumnStore
- Validation of RichProjection and ComputedColumns is now done using Scalactic


trulite:
Overall, amazing work on the refactor. Cleans up lot of paths really nicely. 
I am not going to do a complete code review. Design wise it does a lot of cleaning up and is paving the way for easily adding new features

trulite:
To elaborate - Since this is a big PR and overall we are satisfied small code review stuff can incorporated as issues. I would encourage us to baseline some branch so that we can start adding stuff as smaller PRs

PR #62, opened 2016-02-18
Two partition scan optimizations

velvia:
* When Spark predicates indicate the query is an exact match for one partition, do not scan all the partitions, just execute a single-partition query.
* Pass the scan splits preferred locations into Spark so Spark can colocate tasks on the right Cassandra node for local reads

To support the above, a minor refactor of the ColumnStoreScanner is done.  The `readSegments` method is gone, and instead the `scanSegments` and `scanRows` methods take in a `method` parameter allowing to choose between single partition and multiple partition filtered queries.

Also includes a minor bug fix.


velvia:
@helena thanks for CR, comments have been mostly addressed, going to merge.

Issue #41, opened 2015-11-09
Support Hive


rstrickland:
I really like this concept, but it's critical for us to be able to create permanent tables via a Hive metastore.  Would it be possible to support Hive at some point?

velvia:
@rstrickland welcome!    Would you mind painting a picture of your setup?   Hive metastore is part of your Hadoop installation?  Or DSE?   I don't quite understand - table meta stored in Hive, but actual tables stored in C*?  Thanks!

rstrickland:
We use a centralized Hive metastore that's shared by multiple Spark and EMR clusters that all serve different purposes.  The data is stored in multiple places, including C*.  However, with the lack of a good open source C*-Hive driver we've had to resort to creating temp tables every time we want to get at C* data.  It would be awesome if Filo supported legitimate Hive tables so we could bypass this step.

velvia:
Got it.   Would it be fine to do this through Spark — i.e. you cannot use the filo-cli but use Spark API to create tables etc.   (since Spark already can connect to the Hive metastore)

rstrickland:
We could, but we do have BI tools that use Hive proper (i.e. not the Spark SQL thrift server).  Ideally it would be great if that would work as well, but I know that's a bigger effort.

velvia:
Okay, I think I understand now.  You are looking for a proper FiloDB driver for HIVE that lets you query FiloDB from Hive itself.  Understood now.

velvia:
@rstrickland ok so to break this up into two steps:

1. Have a Hive driver (like DSE's) that automatically lets you query tables from Spark without having to do a CREATE EXTERNAL TABLE
2. Actually support queries directly from Hive without Spark.  Hmmmm....  I think this involves yucky input formats and Hive SerDes, etc.

velvia:
Ok, scoped out the work for Hive metastore support of FiloDB tables for querying in Spark.  Spark has a `HiveMetadataCatalog` class which has a `createDataSourceTable` method.  So one possibility is that when the FiloDB daemon/library spins up, it automatically resolves differences between FiloDB tables and the Hive catalog.  Other times this sync could in theory happen is when a user requests tables or schema, but this would then require a custom Hive plugin in Spark.  Need to think about how to automate the syncing.

[...]

PR #63, opened 2016-02-19
Semi-automatic Hive Metastore sync (#41)

velvia:
This PR adds a feature to optionally automatically sync FiloDB datasets into a Hive Metastore database.
* Uncomment/enable `filodb.hive.database-name` config to specify the Hive metastore database to sync the FiloDB tables to
* The automatic sync happens at the end of ingesting a dataset into FiloDb using the Spark data source write API
* Manual syncing can be triggered in Spark shell or in any Spark app that has the FiloDB jar linked in by calling `filodb.spark.syncToHive(sqlContext)`
* The tables registered are tied to the Spark data source, and thus only works through Spark, Spark SQL, the Spark Thrift server etc.

This change should allow BI tools to be used with Spark Thrift Server via JDBC connection without the need to create temporary tables every time.


velvia:
I tested it locally, but I'm not quite sure how to create a Hive database.

PR #66, opened 2016-03-01
Fix deprecation warnings.

adeandrade:


velvia:
@adeandrade thanks!

velvia:
@adeandrade so actually the Tyrpesafe Config library 1.0.0 is what is merged in, so there is a compile error in Travis:

```
[error] /home/travis/build/tuplejump/FiloDB/core/src/main/scala/filodb.core/metadata/SimpleComputations.scala:100: value getDuration is not a member of com.typesafe.config.Config
[error]         Good(config.getDuration("a", TimeUnit.MILLISECONDS))
[error]                     ^
```

In order to get this working, would need to force the newer version of Tyrpesafe Config (>1.2.0) to be used.

adeandrade:
I'm building this project with 2.11 and I had to bump Ficus, which brought a newer version of Typesafe Config. I'll include that commit.

adeandrade:
Ficus 1.2.X is not available for Scala 2.10. I'll just drop the PR then.

velvia:
Thanks. I'll add 2.11 support and include your changes then.    Will look into other deps like Ficus.  

-Evan
"Never doubt that a small group of thoughtful, committed citizens can change the world" - M. Mead

velvia:
Are there newer features in Ficus that you need?  Would keeping Ficus at current version and doing 2.11 work?

adeandrade:
Your current version of Ficus is not compiled against 2.11. That's why I had to bump it.

If you are interested in updating Scala take a look at my fork. I just committed those changes as well.

I also updated Phantom because I had some connectivity problems when using the latest cassandra driver (3.0.0). And I also updated Spark to 1.6.0.

Maybe you can cherry pick some commits from it.

velvia:
Hmmm… ok thanks.  I might just create a 2.11/Spark 1.6 branch and take your changes.   That’s a bummer that a library switched from 2.10 to 2.11 and didn’t maintain 2.10 for longer?   

If Spark moved to 2.11 by default it would be much easier to upgrade.  :/   Have to wait for Spark 2.0 for that.

PR #71, opened 2016-03-07
fixed regex to grab everything in version.sbt quotes

nhajratw:
The previous expression was dropping the initial '0.', resulting in 2.1-SNAPSHOT instead of 0.2.1-SNAPSHOT


velvia:
Thanks @nhajratw !

PR #80, opened 2016-03-11
added IntelliJ settings

nhajratw:
I know this can be a controversial topic, but I've found it helpful during collaboration if everyone involved uses the same IDE settings.

JetBrains provides details on how to safely share IntelliJ configurations here: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839 , which is what I used for this pull request.


velvia:
@nhajratw I'm OK merging this, but as I don't use IntelliJ myself, someone else will have to make sure those files are updated.

PR #81, opened 2016-03-11
Fix ide warnings

nhajratw:
IntelliJ is reporting a few 'improvements' to the code. Generic stuff as well as things configured via scalastyle. This is the first fix, which optimizes the import statements.


velvia:
Hmmm....  can't say I necessarily agree with the IDE import order.... I separated them out so that internal (e.g. FiloDB) imports are separate for easier visual identification, but I suppose everyone will have their own idea  :)   Not really a biggie either way.

nhajratw:
I'll check to see if there's a way to override the default ordering. (Probably is).

Issue #92, opened 2016-04-20
Lost rows during the same data ingesting into FiloDB


java8964:
I am testing FiloDB with latest trunk, include PR-91, and  can produce this problem that I saw the filodb lost raw after the data ingested. Below is the test case I run, and just want to make sure that it is a bug, not something I misunderstand how filodb is supposed to work.

I have this 1m records of unique emails, with domain. I will use the first 2 characters as the partition key, and use the domain as the segment key, and unique email address itself as the row key. I tested with in memory.
1) After I save the data from spark into filoDB, if I load the data back in one DF, the count is 1M.
2) Then I save the exactly same data from spark into filoDB again in the same spark shell, and load the data back in one DF, the count is less than 1M.

I enable the log at debug level for filoDB package, and the full log is attached.
From the log, you can see what is my steps, and the log output from filoDB, and the 2nd time, the count is not correct. 

I will also dig the log more, as want to understand more internal of filoDB. Hope the log is helpful to identify the root cause of this case.

Thanks
Yong

[filodb-lost-update-bug.log.gz](https://github.com/tuplejump/FiloDB/files/228873/filodb-lost-update-bug.log.gz)

velvia:
@java8964 so I'm looking at your logs now.  Your choice of segment key probably results in really small segments, btw.

I don't see anything unusual from the logs alone.

Repeating the steps from the other PR, including starting the Spark shell (except I added java options for G1GC), I first verified that all the emails are indeed unique.

I then discovered something in my new code, which breaks the InMemoryColumnStore, so I'll go back and fix that.  But in the meantime, you might want to issue a flush at the end manually and see if that fixes it.

`filodb.spark.FiloSetup.client.flushByName("poc1")`

(Note this API will change slightly in the future)

velvia:
@java8964 I have some changes in a new branch, which adds "clustering" support.
velvia/clustering-support-90

If you run in local / in-memory store, like your test in the other thread, you should find that it consistently returns 1 million rows now.   Basically one change is that it makes sure to flush any remaining rows it finds in the memtable at the end of ingestion.

If you test code in this branch on a cluster, be ware that it will require an extra port to be opened.  Right now the port assignment is random, and I'll need to add config to make sure it's not random.

It will be possible to connect in the future remotely to this port and get monitoring/stats out.

velvia:
Fix is merged now

Issue #90, opened 2016-04-08
Clustering support


velvia:
Currently, FiloDB is a jar/library used in Spark for both ingestion and reading.  For ingestion each node / worker in Spark holds a FiloDB memtable.  However, they are currently all independent and do not communicate with one another, and furthermore do not communicate with the driver.  This will have to change.

* We want to be able to poll each memtable for current status
* We want to be able to command each memtable to flush, and get errors back
* Certain operations really require some coordination.  For example, truncating or deleting a dataset cannot safely be done without coordinating with all nodes to ensure there aren't memtables for a given dataset still in session.
* Properly supporting a distributed in-memory column store will require coordination
* It would be nice for clients to be able to monitor the memtable state, externally or internally
* This is secondary concern, but we might want to implement our own mechanism so the user doesn't have to sort data themselves, but rather exchange incoming records such that data is partitioned by token range for local fast writes

Ideally, we want this clustering mechanism to be
* Independent of Spark, especially Spark internals (for example, in the future we can enable users who want to use FiloDB purely as a library to do that, without needing Spark... or for direct Kafka to FiloDB connection)
* Flexible for a variety of use cases
* Enable different communication patterns, as evidenced by the different use cases above

I'm leaning towards using Akka cluster, with the default TCP connectivity as a first cut.  This would require a separate port, but the huge advantage would be it would be able to easily meet all of the goals above.  Also, because all communication is designed around the NodeCoordinatorActor, an Actor-based mechanism is preferable.  Some other alternatives:

1. Spark has an internal Akka-based communication mechanism.  It only works for driver to worker communication,  and is being phased out in 2.0.
2. Layer Akka clustering/remote over Spark internal mechanism

In the future, we could explore using faster network mechanisms such as Aeron (https://github.com/real-logic/Aeron).

I'd love to hear feedback from @trulite @helena and others.

velvia:
@java8964 what do you think of requiring an extra port to be used between Spark workers and driver for this?

java8964:
Sorry, I kind of not sure what this "extra port" means. My understand is that current Spark and FiloDB both use "Akka", and Spark 2.0 is phasing out it.

[...]

PR #94, opened 2016-04-28
Clustering support and ensuring all rows ingested

velvia:
This addresses issues #90 and #92 

Adds an initial clustering implementation where every Filo NodeCoordinator on every Spark executor and the driver are connected together by an Akka cluster network.  Why use Akka cluster instead of Spark's own built in mechanisms?  Spark's Akka remoting is used privately and cannot be used by outside programs, and is going away in 2.0.  Using RDDs to distribute and get responses back is 1) inflexible, 2) not a good fit and does not allow async communications, and 3) forces one to rely on the Spark scheduler.  This gives us an extremely low latency, truly async communications channel between Spark components that can be leveraged.

Actual benefits:
- Now we can asynchronously (i.e. even while existing ingestion is happening) query the state of ingestion, in the future this can be done from an outside client for monitoring
- A fast cycle to ensure all rows have been flushed from the memtable, by first checking the current state, and issuing a flush command, to all executors is done.
- Some other bugs are fixed now.


velvia:
@helena have a look at Akka cluster support  :)

README.md
```diff
@@ -290,7 +290,7 @@ Some options must be configured before starting the Spark Shell or Spark applica
 
 1. Modify the `application.conf` and rebuild, or repackage a new configuration.
 2. Override the built in defaults by setting SparkConf configuration values, preceding the filo settings with `spark.filodb`.  For example, to change the default keyspace, pass `--conf spark.filodb.cassandra.keyspace=mykeyspace` to Spark Shell/Submit.  To use the fast in-memory column store instead of Cassandra, pass `--conf spark.filodb.store=in-memory`.
-3. It might be easier to pass in an entire configuration file to FiloDB.  Pass the java option `-Dconfig.file=/path/to/my-filodb.conf`, for example using `--java-driver-options`.
+3. It might be easier to pass in an entire configuration file to FiloDB.  Pass the java option `-Dfilodb.config.file=/path/to/my-filodb.conf`, for example using `--java-driver-options`.
```
helena: yeah - many akka apps might already have that set for something else

cli/src/main/scala/filodb.cli/CsvImportExport.scala
```diff
@@ -62,7 +62,7 @@ trait CsvImportExport extends StrictLogging {
     }
 
     // There might still be rows left after the latest flush is done, so initiate another flush
-    val activeRows = client.ingestionStats(dataset, version).numRowsActive
+    val activeRows = client.ingestionStats(dataset, version).head.numRowsActive
```
helena: Will a call to .head ever throw an error? Maybe prefer headOption getOrElse a default or whatever makes sense.

coordinator/src/main/resources/cluster-reference.conf
```diff
@@ -0,0 +1,244 @@
...
+    # of 'Joining' members to 'Up'. Typically used together with
+    # 'Cluster.registerOnMemberUp' to defer some action, such as starting actors,
+    # until the cluster has reached a certain size.
+    min-nr-of-members = 1
+
+    # Enable/disable info level logging of cluster events
+    log-info = on
+
+    # Enable or disable JMX MBeans for management of the cluster
+    jmx.enabled = on
```
helena: I'd make this configurable by the user and default to off

coordinator/src/main/resources/cluster-reference.conf
```diff
@@ -0,0 +1,244 @@
...
+    leader-actions-interval = 1s
+
+    # how often should the node move nodes, marked as unreachable by the failure
+    # detector, out of the membership ring?
+    unreachable-nodes-reaper-interval = 1s
+
+    # How often the current internal stats should be published.
+    # A value of 0s can be used to always publish the stats, when it happens.
+    # Disable with "off".
+    publish-stats-interval = off
```
helena: we may want to use this later on but yeah off for now is fine

coordinator/src/main/resources/cluster-reference.conf
```diff
@@ -0,0 +1,244 @@
...
+      # How often a node publishes metrics information.
+      gossip-interval = 3s
+
+      # How quickly the exponential weighting of past data is decayed compared to
+      # new data. Set lower to increase the bias toward newer values.
+      # The relevance of each data sample is halved for every passing half-life
+      # duration, i.e. after 4 times the half-life, a data sample’s relevance is
+      # reduced to 6% of its original relevance. The initial relevance of a data
+      # sample is given by 1 – 0.5 ^ (collect-interval / half-life).
+      # See http://en.wikipedia.org/wiki/Moving_average#Exponential_moving_average
```
helena: this was my idea, the EMA :)

coordinator/src/main/resources/cluster-reference.conf
```diff
@@ -0,0 +1,244 @@
...
+  # Default configuration for routers
+  actor.deployment.default {
+    # MetricsSelector to use
+    # - available: "mix", "heap", "cpu", "load"
+    # - or: Fully qualified class name of the MetricsSelector class.
+    #       The class must extend akka.cluster.routing.MetricsSelector
+    #       and have a public constructor with com.typesafe.config.Config
+    #       parameter.
+    # - default is "mix"
+    metrics-selector = mix
```
helena: yeah the mix is the best, i wrote this too, and the routers

coordinator/src/main/scala/filodb.coordinator/CoordinatorSetup.scala
```diff
@@ -1,24 +1,48 @@
...
+  //  - filodb-defaults.conf (resource / in jar)
+  //  - cluster-reference.conf
+  //  - all other reference.conf's
+  val systemConfig: Config = {
+    val customConfig = sys.props.get("filodb.config.file").orElse(sys.props.get("config.file"))
+                                .map { path => ConfigFactory.parseFile(new java.io.File(path)) }
+                                .getOrElse(ConfigFactory.empty)
+    // Don't ask me why, but ConfigFactory.parseResources() does NOT work in Spark 1.4.1 executors
+    // and only the below works.
+    val defaultsFromUrl = ConfigFactory.parseURL(getClass.getResource("/filodb-defaults.conf"))
```
helena: or https://github.com/typesafehub/config/blob/master/config/src/main/java/com/typesafe/config/ConfigFactory.java#L747

coordinator/src/main/scala/filodb.coordinator/CoordinatorSetup.scala
```diff
@@ -42,7 +66,7 @@ trait CoordinatorSetup {
     system.actorOf(NodeCoordinatorActor.props(metaStore, reprojector, columnStore, config),
                    "coordinator")
 
-  lazy val client = new Client(coordinatorActor)
+  lazy val cluster = Cluster(system)
```
helena: yay

coordinator/src/main/scala/filodb.coordinator/DatasetCoordinatorActor.scala
```diff
@@ -132,6 +133,7 @@ private[filodb] class DatasetCoordinatorActor(projection: RichProjection,
   var flushesStarted = 0
   var flushesSucceeded = 0
   var flushesFailed = 0
+  var rowsIngested = 0L
```
helena: Would be nice to move these vars to a case class for the coordinator actor

coordinator/src/main/scala/filodb.coordinator/NodeClusterActor.scala
```diff
@@ -0,0 +1,99 @@
...
+import akka.actor.{ActorRef, Props, RootActorPath}
+import akka.cluster.Cluster
+import akka.cluster.ClusterEvent._
+import akka.event.LoggingReceive
+import scala.collection.mutable.HashMap
+import scala.concurrent.duration._
+
+import filodb.core.ErrorResponse
+
+object NodeClusterActor {
```
helena: Shouldn't this be

      private[filodb] NodeClusterActor

coordinator/src/main/scala/filodb.coordinator/NodeClusterActor.scala
```diff
@@ -0,0 +1,99 @@
...
+
+  import context.dispatcher
+
+  // subscribe to cluster changes, re-subscribe when restart
+  override def preStart(): Unit = {
+    cluster.subscribe(self, initialStateMode = InitialStateAsEvents,
+      classOf[MemberEvent], classOf[UnreachableMember])
+  }
+
+  // TODO: leave cluster?
```
helena: Yes, that's all and how/where to leave the cluster, however if the owner of the cluster that is passing it into the constructor here is also an actor, I'd have that actor do it vs here in it's child.

coordinator/src/main/resources/cluster-reference.conf
velvia: Sorry, this config file was copied from Akka itself, because for some reason reference.conf gets overridden by Spark class path and the cluster settings are lost.  :/

cli/src/main/scala/filodb.cli/CsvImportExport.scala
velvia: Good point.

coordinator/src/main/scala/filodb.coordinator/NodeClusterActor.scala
```diff
@@ -0,0 +1,99 @@
...
+  override def postStop(): Unit = cluster.unsubscribe(self)
+
+  private def withRole(role: String)(f: Set[ActorRef] => Unit): Unit = roleToCoords.get(role) match {
+    case None       => sender ! NoSuchRole
+    case Some(refs) => f(refs)
+  }
+
+  def receive: Receive = LoggingReceive {
+    case MemberUp(member) =>
+      logger.info(s"Member is Up: ${member.address} with roles ${member.roles}")
```
helena: usually akka cluster logs this already, unless it's disabled

coordinator/src/main/scala/filodb.coordinator/NodeClusterActor.scala
helena: cluster unsubscribe self
    cluster leave selfAddress

coordinator/src/main/scala/filodb.coordinator/DatasetCoordinatorActor.scala
velvia: Hmmm.   So have a mutable case class for tracking stats, and an immutable version?

`case class MutableStats(var flushesStarted: Int, var flushesSucceeded: Int)`

coordinator/src/main/scala/filodb.coordinator/NodeClusterActor.scala
velvia: I think it doesn’t log the roles though.

coordinator/src/main/scala/filodb.coordinator/NodeClusterActor.scala
```diff
@@ -0,0 +1,99 @@
...
+      val memberCoordActor = RootActorPath(member.address) / "user" / "coordinator"
+      val fut = context.actorSelection(memberCoordActor).resolveOne(resolveActorTimeout)
+      fut.foreach { ref => self ! AddCoordActor(member.roles, ref) }
+      fut.recover {
```
helena: Or maybe
   
    for (actor <- context.actorSelection(path).resolveOne(timeout)  {
      self ! AddCoordActor(member.roles, ref)
    } recover ....

coordinator/src/main/scala/filodb.coordinator/NodeClusterActor.scala
```diff
@@ -0,0 +1,99 @@
...
+        case e: Exception =>
+          logger.warn(s"Unable to resolve coordinator at $memberCoordActor, ignoring. " +
+                      "Maybe NodeCoordinatorActor did not start up before node joined cluster.", e)
+      }
+
+    case AddCoordActor(roles, coordRef) =>
+      roles.foreach { role => roleToCoords(role) += coordRef }
```
helena: Are these roles == cluster.selfRoles ? If so you can get that globally vs passing a custom msg with it

velvia: They are the same, but we need to track the actorRefs for each role.

coordinator/src/main/scala/filodb.coordinator/client/IngestionOps.scala
```diff
@@ -1,38 +1,63 @@
...
+  /**
+   * Repeatedly checks the ingestion stats and flushes until everything is flushed for a given dataset.
+   */
+  def flushCompletely(dataset: DatasetRef, version: Int, timeout: FiniteDuration = 30.seconds): Int = {
+    var nodesFlushed: Int = 0
+
+    def anyRowsLeft: Boolean = {
+      val stats = ingestionStats(dataset, version, timeout)
+      logger.debug(s"Stats from all ingestors: $stats")
+      stats.map(_.numRowsActive).filter(_ >= 0).sum > 0
```
helena: You can iterate just once vs twice with

       stats collect { case s if s.numRowsActive >= 0 => s }.sum > 0

coordinator/src/main/scala/filodb.coordinator/client/IngestionOps.scala
```diff
@@ -1,38 +1,63 @@
...
     }
+
+    while(anyRowsLeft) {
+      nodesFlushed = flush(dataset, version, timeout)
+      Thread sleep 1000
```
helena: Ouch, Thead.sleep?

helena: If only actors use this trait we can do a better strategy in the future, I'd mark this as a TODO for now and tag a ticket to it soon.

spark/src/main/scala/filodb.spark/FiloSetup.scala
```diff
@@ -1,40 +1,89 @@
...
+      logger.info("Initializing FiloDriver clustering/coordination...")
+      role = "driver"
+      val filoConfig = configFromSpark(context)
+      _config = Some(filoConfig)
+      coordinatorActor
+
+      // Add in self cluster address, and join cluster ourselves
+      val selfAddr = cluster.selfAddress
+      cluster.join(selfAddr)
+      clusterActor
```
helena: this is duplicated on line 75 with no effect

coordinator/src/main/scala/filodb.coordinator/client/IngestionOps.scala
velvia: Yeah, let me think of a more reactive way to do this.  Maybe will have each node handle a FlushCompletely message, but that will probably require a state machine / FSM.  :/

spark/src/main/scala/filodb.spark/FiloSetup.scala
```diff
@@ -1,40 +1,89 @@
...
+      Thread sleep 1000   // Wait a little bit for cluster joining to take effect
```
helena: Typically I do an asynchronous impl w/out needing Thread.sleep, like this in an actor

     def receive: Actor.Receive = {
        case e: CurrentClusterState => initialize(e)
        case MemberUp(member)   => watch(member)
     }

And in initialize/watch functions I do any relevant work vs hold the thread.

spark/src/main/scala/filodb.spark/FiloSetup.scala
```diff
@@ -47,7 +96,27 @@ object FiloSetup extends CoordinatorSetup {
...
+  def init(filoConfig: Config): Unit = synchronized {
+    _config.getOrElse {
+      this.role = "executor"
+      _config = Some(filoConfig)
+      coordinatorActor       // force coordinator to start
+      // get address from config and join cluster.  note: it's ok to join cluster multiple times
+      val addr = AddressFromURIString.parse(filoConfig.getString("spark-driver-addr"))
+      logger.info(s"Initializing FiloExecutor clustering by joining driver at $addr...")
+      cluster.join(addr)
+      Thread sleep 1000
```
helena: Here too, we can in a new ticket address that.

coordinator/src/main/resources/cluster-reference.conf
helena: Ah.

spark/src/main/scala/filodb.spark/FiloSetup.scala
velvia: Yeah, I mean to get rid of this by having NodeClusterActor send a message back when some member is up, then from FiloSetup we can just do an actor ask.   The FiloSetup code has to be run in a Spark processing thread, so it can't be an actor itself.  I meant to investigate a better way to integrate actors into Spark processing, but no time right now.  :/      What I'm thinking is a way to run the actor receive loop inside another thread to simplify things.

coordinator/src/main/scala/filodb.coordinator/DatasetCoordinatorActor.scala
helena: private[filodb] case class DatasetStats(flushesStarted: Int, flushesSucceeded: Int, flushesFailed: Int, rowsIngested: Long)`

      private class DatasetCoordinatorActor ... {
        private var stats = DatasetStats.empty //initialize in the apply
     }

Issue #101, opened 2016-05-27
Count mismatch with FiloDB ingestion process


parekuti:
We are trying to load data from C* to FiloDB using 0.3 release and almost source and target record counts are not in sync.

velvia:
@parekuti did some awesome sleuthing and found the issue was due to token range wrapping.  The PR should fix this issue.

velvia:
This should take care of this issue at least due to token ranges not wrapping.

PR #105, opened 2016-06-08
Unwrap token ranges before querying by token range to fix read path issues

parekuti:
Unwrap token ranges before querying by token range to fix read path issues like count mismatch and unable to read few partitions. Issue #101

### DIFF
--- a/cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
+++ b/cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
@@ -6,9 +6,10 @@ import com.typesafe.config.Config
 import com.typesafe.scalalogging.slf4j.StrictLogging
 import java.net.InetSocketAddress
 import java.nio.ByteBuffer
+import java.util.HashSet
+
 import scala.concurrent.{ExecutionContext, Future}
 import spray.caching._
-
 import filodb.cassandra.FiloCassandraConnector
 import filodb.core._
 import filodb.core.store._
@@ -131,7 +132,8 @@ extends CachedMergingColumnStore with CassandraColumnStoreScanner with StrictLog
     val keyspace = clusterConnector.keySpaceName(dataset)
     require(splitsPerNode >= 1, s"Must specify at least 1 splits_per_node, got $splitsPerNode")
 
-    val tokensByReplica = metadata.getTokenRanges.asScala.toSeq.groupBy { tokenRange =>
+    val tokenRanges=unwrapTokenRanges(metadata.getTokenRanges.asScala.toSeq)
+    val tokensByReplica =tokenRanges.asScala.toSeq.groupBy { tokenRange =>
       metadata.getReplicas(keyspace, tokenRange)
     }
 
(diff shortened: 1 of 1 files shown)


cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
```diff
@@ -131,7 +132,8 @@ extends CachedMergingColumnStore with CassandraColumnStoreScanner with StrictLog
     val keyspace = clusterConnector.keySpaceName(dataset)
     require(splitsPerNode >= 1, s"Must specify at least 1 splits_per_node, got $splitsPerNode")
 
-    val tokensByReplica = metadata.getTokenRanges.asScala.toSeq.groupBy { tokenRange =>
+    val tokenRanges=unwrapTokenRanges(metadata.getTokenRanges.asScala.toSeq)
```
velvia: Can we add spaces around =?

Also, converting to a java hashset and then converting back to scala seems inefficient.  Let me see if there is a better way to do this.

cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
```diff
@@ -166,6 +168,13 @@ extends CachedMergingColumnStore with CassandraColumnStoreScanner with StrictLog
                                replicas.map(_.getSocketAddress).toSet)
     }
   }
+
+  def unwrapTokenRanges(wrappedRanges : Seq[TokenRange] ): HashSet[TokenRange] = {
```
velvia: How about this instead?

    def unwrapTokenRanges(wrappedRanges: Seq[TokenRange]): Seq[TokenRange] =
      wrappedRanges.flatMap(_.unwrap().asScala.toSeq)

parekuti:
Updated code based on Evan's comments....

cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
```diff
@@ -169,11 +170,8 @@ extends CachedMergingColumnStore with CassandraColumnStoreScanner with StrictLog
     }
   }
 
-  def unwrapTokenRanges(wrappedRanges : Seq[TokenRange] ): HashSet[TokenRange] = {
-    val tokenRanges = new HashSet[TokenRange]()
-    wrappedRanges.map(range => tokenRanges.addAll(range.unwrap()))
-    logger.debug(s"unwrapTokenRanges: ${tokenRanges.toString()}")
-    tokenRanges
+  def unwrapTokenRanges(wrappedRanges : Seq[TokenRange] ): Seq[TokenRange] = {
```
velvia: @parekuti two minor changes:  remove the space before ): and the braces { } is not needed since this is a one liner.  Not a big deal...

parekuti: @velvia Updated and committed changes.

velvia:
Thanks @parekuti looks good

PR #108, opened 2016-06-23
Change Cassandra hosts parameter from List to String

elppc:
This allows to be compatible with SparkConf.set("spark.filodb.cassandra.hosts","someIP,anotherIP")

### DIFF
--- a/cassandra/src/main/scala/filodb.cassandra/FiloCassandraConnector.scala
+++ b/cassandra/src/main/scala/filodb.cassandra/FiloCassandraConnector.scala
@@ -5,12 +5,16 @@ import com.typesafe.config.Config
 import net.ceedubs.ficus.Ficus._
 import scala.concurrent.duration.FiniteDuration
 import scala.concurrent.{ExecutionContext, Future}
+import scala.util.Try
 
 import filodb.core._
 
 trait FiloCassandraConnector {
   private[this] lazy val cluster =
-    Cluster.builder().addContactPoints(config.as[Seq[String]]("hosts") :_*)
+    Cluster.builder()
+                 .addContactPoints(
+                   Try(config.as[Seq[String]]("hosts")).getOrElse(config.getString("hosts").split(',').toSeq) :_*
+                  )
                  .withPort(config.getInt("port"))
                  .withAuthProvider(authProvider)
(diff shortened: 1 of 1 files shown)


velvia:
@elppc thx for the pull request.   This however would break compatibility with existing people who use a list, and there are good benefits to using a list (much better error checking, namely).   So, how about we make it parse both a list and a string?   You can do something like `Try(config.as[Seq[String]]("hosts")).getOrElse(config.getString("hosts").split(',').toSeq)`

This way no need to modify existing configs.

elppc:
I am not familiar with this synthax, could you explain me ?
config.as[Seq[String]]("hosts") :_*

velvia:
Sure.  So `config.as[Seq[String]]("hosts")` simply returns a Sequence of strings.  The entire line of code that I sent you above is saying, first try to get the config param "hosts" as a sequence of strings. If that fails / throws an exception, then do the getOrELse ...  basically try to get "hosts" as a string, then split it.
Finally, strings :_* the last few characters is for passing in variable length arguments....  I don't remember the equivalent in Java but it's like arg... or arg *

elppc:
My question is more on `as[Seq[String]]` which is not part of typesafe config and `:_*`

elppc:
Sorry I don't understand why you need `:_*` if I look in C* driver API, it only requires a collection of String

velvia:
@elppc thanks, would you mind rebasing and collapsing the commits into one?  Thx.

You are right the driver accepts collection as well.  I'm fine with that, just forgot about it.
The .as[Seq[String]] is a implicit provided by not the Typesafe Config library itself but a Scala wrapper on top called Ficus.

elppc:
Do you mean rebasing on tuplejump:master ? Can you explain me the steps ?

velvia:
@elppc yeah basically you want to squish all your commits into one.   In your branch do a `git rebase -i <sha>`, where the <sha> is the SHA of the commit just before the first one in this branch.   Then in vim or whatever editor you have configured, change all the second to last commits to "s" to squish them all into one.

elppc:
I think it's OK but let me know if I missed something

velvia:
@elppc many thanks again, merging now.

Issue #109, opened 2016-06-30
Add Multipartition scan feature via IN Operator


parekuti:

parekuti:
This should be fixed by PR #110

PR #110, opened 2016-06-30
Ability to scan multiple partitions using IN operator

parekuti:
This feature avoids full table scan when multiple partition keys are provided via IN clause in select query. Issue #109.

### DIFF
--- a/core/src/main/resources/filodb-defaults.conf
+++ b/core/src/main/resources/filodb-defaults.conf
@@ -35,6 +35,9 @@
 
       # Maximum number of chunks to flush at one time using one Unlogged Batch.
       chunk-batch-size = 16
+
+      # Maximum number of partitions can be fetched at once using IN query
+      inquery-partitions-limit = 12
     }
 
     memtable {
--- a/core/src/main/resources/filodb-defaults.conf
+++ b/core/src/main/resources/filodb-defaults.conf
@@ -36,7 +36,9 @@
       # Maximum number of chunks to flush at one time using one Unlogged Batch.
       chunk-batch-size = 16
 
-      # Maximum number of partitions can be fetched at once using IN query
+      # Maximum number of partitions that can be fetched at once that will still fit in
+      # one Spark partition/thread when using IN clause in query.
+      # If number of partitions are more than this limit then full table scan is performed.
       inquery-partitions-limit = 12
     }
 
--- a/core/src/main/scala/filodb.core/store/InMemoryColumnStore.scala
+++ b/core/src/main/scala/filodb.core/store/InMemoryColumnStore.scala
@@ -165,6 +165,26 @@ trait InMemoryColumnStoreScanner extends ColumnStoreScanner {
       toIterator
   }
 
+  def multiPartScan(projection: RichProjection, version: Int, partitions: Seq[Any]):
+  Iterator[(BinaryPartition, RowMapTreeLike)] = {
+    partitions.map{partition => {
+      logger.info(s"partition :${partition}")
+      val binPart = projection.partitionType.toBytes(partition.asInstanceOf[projection.PK])
+       (binPart, rowMaps.getOrElse((projection.datasetRef, binPart, version), EmptyRowMapTree))
+    }}.toIterator
+  }
+
+  def multiPartRangeScan(projection: RichProjection, version: Int, keyRanges: Seq[KeyRange[_, _]]):
+  Iterator[(BinaryPartition, RowMapTreeLike)] = {
+    keyRanges.map{range => {
+      val binKR = projection.toBinaryKeyRange(range)
+      val rowMapTree = rowMaps.getOrElse((projection.datasetRef, binKR.partition, version),
+        EmptyRowMapTree)
+      val subMap = rowMapTree.subMap(binKR.start, true, binKR.end, !binKR.endExclusive)
+      (binKR.partition, subMap)
+    }}.toIterator
+  }
+
   def singlePartRangeScan(projection: RichProjection, version: Int, k: KeyRange[_, _]):
     Iterator[(BinaryPartition, RowMapTreeLike)] = {
     val binKR = projection.toBinaryKeyRange(k)
@@ -193,7 +213,8 @@ trait InMemoryColumnStoreScanner extends ColumnStoreScanner {
       case SinglePartitionScan(partition) => singlePartScan(projection, version, partition)
       case SinglePartitionRangeScan(k)    => singlePartRangeScan(projection, version, k)
 
-      case MultiPartitionScan(partitions) => singlePartScan(projection, version, partitions)
+      case MultiPartitionScan(partitions) => multiPartScan(projection, version, partitions)
+      case MultiPartitionRangeScan(keyRanges) => multiPartRangeScan(projection, version, keyRanges)
 
       case FilteredPartitionScan(split, filterFunc) =>
         filteredPartScan(projection, version, split, filterFunc)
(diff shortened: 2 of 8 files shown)


core/src/main/resources/filodb-defaults.conf
```diff
@@ -35,6 +35,9 @@
 
       # Maximum number of chunks to flush at one time using one Unlogged Batch.
       chunk-batch-size = 16
+
+      # Maximum number of partitions can be fetched at once using IN query
```
velvia: I think we need to expand the explanation here.  This is the max # of FiloDB partitions that can be fetched at once that will still fit in one Spark partition/thread, right?

core/src/main/scala/filodb.core/store/InMemoryColumnStore.scala
```diff
@@ -165,6 +165,25 @@ trait InMemoryColumnStoreScanner extends ColumnStoreScanner {
       toIterator
   }
 
+  def multiPartScan(projection: RichProjection, version: Int, partitions: Seq[Any]):
+    Iterator[(BinaryPartition, RowMapTreeLike)] = {
+    partitions.map { partition => {
```
velvia: I think it's better to do `partitions.toIterator` rather than do a map and then convert everything to an Iterator.  That way, the chunkRowMap lookup is lazy, in case there are a lot of partitions this will save memory.

core/src/main/scala/filodb.core/store/InMemoryColumnStore.scala
```diff
@@ -165,6 +165,25 @@ trait InMemoryColumnStoreScanner extends ColumnStoreScanner {
...
+      val binPart = projection.partitionType.toBytes(partition.asInstanceOf[projection.PK])
+       (binPart, rowMaps.getOrElse((projection.datasetRef, binPart, version), EmptyRowMapTree))
+    }}.toIterator
+  }
+
+  def multiPartRangeScan(projection: RichProjection, version: Int, keyRanges: Seq[KeyRange[_, _]]):
+    Iterator[(BinaryPartition, RowMapTreeLike)] = {
+    keyRanges.map { range => {
```
velvia: Same here, make the keyRanges toIterator so the lookup is lazy.

spark/src/main/scala/filodb.spark/FiloRelation.scala
```diff
@@ -65,21 +65,38 @@ object FiloRelation extends StrictLogging {
...
+    // Are the filters EqualTo, In?
+    val predicateValues = filterStuff.collect {
+      case (pos, keyType, Seq(EqualTo(_, equalValue))) =>
+        Set(KeyFilter.parseSingleValue(keyType)(equalValue))
+      case (pos, keyType, Seq(In(_, inValues))) =>
+        (KeyFilter.parseValues(keyType)(inValues.toSet))
     }
-    if (equalPredValues.length == projection.partitionColumns.length) {
-      if (equalPredValues.length == 1) Some(equalPredValues.head) else Some(equalPredValues)
+    logger.info(s"Push down predicate values: ${predicateValues.toString()}")
```
velvia: I would make this very specific:  Push down partition predicates:

spark/src/main/scala/filodb.spark/FiloRelation.scala
```diff
@@ -65,21 +65,38 @@ object FiloRelation extends StrictLogging {
...
+    if (predicateValues.length == projection.partitionColumns.length) {
+      if(predicateValues.length == 1) {
+        predicateValues.flatten
+      }
```
velvia: minor: combine into one line } else {

spark/src/main/scala/filodb.spark/FiloRelation.scala
```diff
@@ -65,21 +65,38 @@ object FiloRelation extends StrictLogging {
...
+        val predList = combine(predicateValues)
+        if (predList.size <= inqueryPartitionsLimit) predList else Nil
+      }
     } else {
-      None
+      Nil
     }
   }
 
+  def combine[A](xs: Traversable[Traversable[A]]): Seq[Seq[A]] =
```
velvia: I don't quite understand what you're trying to do with combine.  Is this just to filter out the empty values?   Can't you just do `xs.filter(_.nonEmpty).toSeq` and get rid of the combine function?

spark/src/main/scala/filodb.spark/FiloRelation.scala
```diff
@@ -65,21 +65,38 @@ object FiloRelation extends StrictLogging {
...
-    if (equalPredValues.length == projection.partitionColumns.length) {
-      if (equalPredValues.length == 1) Some(equalPredValues.head) else Some(equalPredValues)
+    logger.info(s"Push down predicate values: ${predicateValues.toString()}")
+    if (predicateValues.length == projection.partitionColumns.length) {
+      if(predicateValues.length == 1) {
+        predicateValues.flatten
+      }
+      else {
+        val predList = combine(predicateValues)
+        if (predList.size <= inqueryPartitionsLimit) predList else Nil
```
velvia: Should probably explain what this is trying to do.  AFAICT if the predicates are above this size then it does a full table scan.  The limit of 12 seems a bit low though - but maybe one can experiment with this.

velvia:
@parekuti gave you lots of feedback, mostly minor.  Overall looks really awesome!   Thanks so much.

spark/src/main/scala/filodb.spark/FiloRelation.scala
parekuti: It's not just checking out for nulls. It's like creating list of all possible partition keys provided by the filters. For example pk1='FA' and pk2 IN ('123','345',678') then combine method will give the combinations like ('FA','123') , ('FA','345') and ('FA','678') => Full Partition Key values.

velvia: I'm not convinced this does what you want.  

First of all, predicateValues might have the incorrect order if the order of the filters is not the same as the partition keys. For example, lets say your partition keys are countryCode, actorName.  If you do WHERE actorName = "blah" and countryCode IN (a, b), I think your filter will not work because the predicateValues would be Seq(Set("blah"), Set(a, b)), but you actually need it in countryCode and actorName order.   So you need to sort the predicateValues by position first.

Second, I now understand what you mean by generating combinations, but the code isn't too obvious.  Maybe add a comment to explain what is meant by combine?

spark/src/main/scala/filodb.spark/FiloRelation.scala
```diff
@@ -92,6 +95,15 @@ object FiloRelation extends StrictLogging {
     }
   }
 
+  /**
+    * Method to get possible combinations of key values provided in partition key filters.
+    * For example filters provided in the query are actor2Code IN ('JPN', 'KHM') and year = 1979
+    * then this method returns all combinations like ('JPN',1979) , ('KHM',1979) which represents
+    * full partition key value.
+    * @param xs
```
velvia: I think you can remove this @param thing.... it's obvious from rest of description

spark/src/test/scala/filodb.spark/SaveAsFiloTest.scala
```diff
@@ -370,8 +370,8 @@ class SaveAsFiloTest extends SparkTestBase {
     df.agg(sum("numArticles")).collect().head(0) should equal (492)
 
     df.registerTempTable("gdelt")
-    sql.sql("select sum(numArticles) from gdelt where  actor2Code in ('JPN', 'KHM') " +
-      "and year=1979 and eventId >= 21 AND eventId <= 24").collect().
+    sql.sql("select sum(numArticles) from gdelt where year=1979 " +
```
velvia: Thanks.

velvia:
@parekuti this is looking pretty good.... just one minor comment.  Think it's otherwise ready for merge.
Thanks!

spark/src/main/scala/filodb.spark/FiloRelation.scala
parekuti: Should i just remove **@param** only or **@tparam A @return** as well.

velvia: I think remove both.  Thanks.

velvia:
@parekuti Thanks so much!  Merging.

PR #117, opened 2016-09-22
Guava shading

chernals:
Not sure if the explicit 17.0 in CassDep is useful. The other part really helped.


chernals:
Tested without it and it now "confirm" it is not needed. Added a commit.

PR #119, opened 2016-09-27
Fix double segment key rounding

StevenReitsma:
This fixes the bug where using a DoubleColumn as the segment key was not possible. Also added a test for using `:round` to round to non-integer values and fixed the URL to Travis. Tests pass on my end.


velvia:
Seems like the Typesafe repo is really having issues.  I had problems with this last night too.  :(   I think will merge this regardless.

PR #127, opened 2016-11-08
creating a function for checking java and another to check sbt

jenaiz:
Hi guys, 
I have found interesting to create a bash function to check java and one more to check sbt tool. As it is done to check JDK version greater that 1.8.


velvia:
@jenaiz thanks for your contribution!

jenaiz:
You are welcome.

Issue #120, opened 2016-09-30
No error is displayed when cassandra nodes are down for read path


parekuti:
No error is displayed when cassandra nodes are down while running query from thrift server.

PR #143, opened 2017-03-21
Modified to handle errors during read path

parekuti:
Related to issue: No error is displayed when cassandra nodes are down for read path #120

### DIFF
--- a/cassandra/src/main/scala/filodb.cassandra/Util.scala
+++ b/cassandra/src/main/scala/filodb.cassandra/Util.scala
@@ -30,6 +30,7 @@ object Util {
       // from invalid Enum strings, which should never happen, or some other parsing error
       case e: NoSuchElementException   => throw MetadataException(e)
       case e: IllegalArgumentException => throw MetadataException(e)
+      case e: Exception                => throw e
     }
   }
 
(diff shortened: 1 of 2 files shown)


cassandra/src/main/scala/filodb.cassandra/Util.scala
```diff
@@ -20,6 +20,7 @@ object Util {
         if (resultSet.wasApplied) Success else notAppliedResponse
       }.recover {
         case e: DriverException => throw StorageEngineException(e)
+        case e: Exception       => throw e
```
velvia: I think we should still wrap the error in StorageEngineException.

cassandra/src/main/scala/filodb.cassandra/Util.scala
```diff
@@ -30,6 +31,7 @@ object Util {
       // from invalid Enum strings, which should never happen, or some other parsing error
       case e: NoSuchElementException   => throw MetadataException(e)
       case e: IllegalArgumentException => throw MetadataException(e)
+      case e: Exception                => throw e
```
velvia: Same here, wrap in StorageEngineException

parekuti:
Added this general exception to handle just in case of other errors than DriverException. But this is not the code actually solving the issue. The changes made to the ChunkRowMap table helped to throw exception back to the client.

cassandra/src/main/scala/filodb.cassandra/Util.scala
parekuti: Done

cassandra/src/main/scala/filodb.cassandra/Util.scala
parekuti: Done.

PR #145, opened 2017-03-23
Modified to hanlde read path errors

parekuti:
No error is displayed when cassandra nodes are down for read path #120

### DIFF
--- a/cassandra/src/main/scala/filodb.cassandra/Util.scala
+++ b/cassandra/src/main/scala/filodb.cassandra/Util.scala
@@ -3,10 +3,11 @@ package filodb.cassandra
 import com.datastax.driver.core.exceptions.DriverException
 import com.datastax.driver.core._
 import java.nio.ByteBuffer
+
 import scala.concurrent.{Future, Promise}
 import scodec.bits.ByteVector
-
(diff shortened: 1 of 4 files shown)


cassandra/src/main/scala/filodb.cassandra/Util.scala
```diff
@@ -3,10 +3,11 @@ package filodb.cassandra
 import com.datastax.driver.core.exceptions.DriverException
 import com.datastax.driver.core._
 import java.nio.ByteBuffer
+
 import scala.concurrent.{Future, Promise}
 import scodec.bits.ByteVector
-
 import filodb.core._
+import monix.reactive.Observable
```
velvia: Can you put the monix import in the first section with everything else?  The order should be:

everything non-filo
internal filodb

parekuti: Done

parekuti:
Travis failed for the following reason:
[info] StreamingTest:
[info] - should ingest successive streaming RDDs as DataFrames... *** FAILED ***
[info]   1000 did not equal 1200 (StreamingTest.scala:106)

PR #156, opened 2018-10-04
feat(jmh): JMH Benchmark for Part Key Lucene Index

vishramachandran:
**Pull Request checklist**

### DIFF
--- a/jmh/src/main/scala/filodb.jmh/PartKeyIndexBenchmark.scala
+++ b/jmh/src/main/scala/filodb.jmh/PartKeyIndexBenchmark.scala
@@ -45,7 +45,7 @@ class PartKeyIndexBenchmark {
       partId += 1
     }
   }
-  partKeyBuilder.allContainers.head.consumeRecords(consumer)
+  partKeyBuilder.allContainers.foreach(_.consumeRecords(consumer))
   partKeyIndex.commitBlocking()
 
   @Benchmark
@@ -55,6 +55,7 @@ class PartKeyIndexBenchmark {
     for ( i <- 0 to 20 optimized) {
       partKeyIndex.partIdsFromFilters(
         Seq(ColumnFilter("job", Filter.Equals(s"App-$i")),
+            ColumnFilter("host", Filter.EqualsRegex("H0")),
             ColumnFilter("__name__", Filter.Equals("heap_usage"))),
         now,
         now + 1000)
@@ -93,7 +94,7 @@ class PartKeyIndexBenchmark {
   @BenchmarkMode(Array(Mode.Throughput))
   @OutputTimeUnit(TimeUnit.SECONDS)
   def startTimeLookupWithPartId(): Unit = {
-    for ( i <- 0 to 1000 optimized) {
+    for ( i <- 0 to 20 optimized) {
       partKeyIndex.startTimeFromPartId(i)
     }
   }
--- a/jmh/src/main/scala/filodb.jmh/PartKeyIndexBenchmark.scala
+++ b/jmh/src/main/scala/filodb.jmh/PartKeyIndexBenchmark.scala
@@ -89,4 +89,13 @@ class PartKeyIndexBenchmark {
     }
   }
 
+  @Benchmark
+  @BenchmarkMode(Array(Mode.Throughput))
+  @OutputTimeUnit(TimeUnit.SECONDS)
+  def startTimeLookupWithPartId(): Unit = {
+    for ( i <- 0 to 1000 optimized) {
+      partKeyIndex.startTimeFromPartId(i)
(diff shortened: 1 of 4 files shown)


vishramachandran:
Benchmark numbers:
```
[info] Benchmark                                                   Mode  Cnt     Score     Error  Units
[info] PartKeyIndexBenchmark.partIdsLookupWithEqualsFilters       thrpt   45  7618.440 ± 101.865  ops/s
[info] PartKeyIndexBenchmark.partIdsLookupWithPrefixRegexFilters  thrpt   45  1000.342 ±  77.338  ops/s
[info] PartKeyIndexBenchmark.partIdsLookupWithSuffixRegexFilters  thrpt   45   439.928 ±  21.094  ops/s
```

jmh/src/main/scala/filodb.jmh/PartKeyIndexBenchmark.scala
```diff
@@ -0,0 +1,101 @@
...
+  var partId = 1
+  val now = System.currentTimeMillis()
+  val consumer = new BinaryRegionConsumer {
+    def onNext(base: Any, offset: Long): Unit = {
+      val partKey = dataset.partKeySchema.asByteArray(base, offset)
+      partKeyIndex.addPartKey(partKey, partId, now)()
+      partId += 1
+    }
+  }
+  partKeyBuilder.allContainers.head.consumeRecords(consumer)
```
velvia: Note: you are only using the consumer for the first container.  You are adding a million records, which likely means many hundreds of containers.  So to make this correct you really need: `allContainers.foreach`

jmh/src/main/scala/filodb.jmh/PartKeyIndexBenchmark.scala
```diff
@@ -0,0 +1,101 @@
...
+
+  @Benchmark
+  @BenchmarkMode(Array(Mode.Throughput))
+  @OutputTimeUnit(TimeUnit.SECONDS)
+  def partIdsLookupWithPrefixRegexFilters(): Unit = {
+    for ( i <- 0 to 20 optimized) {
+      partKeyIndex.partIdsFromFilters(
+        Seq(ColumnFilter("job", Filter.Equals(s"App-$i")),
+          ColumnFilter("__name__", Filter.Equals("heap_usage")),
+          ColumnFilter("instance", Filter.EqualsRegex(".*2"))),
```
velvia: Might this benchmark be slower because it returns more results than the prefix one, for example?

jmh/src/main/scala/filodb.jmh/PartKeyIndexBenchmark.scala
```diff
@@ -0,0 +1,101 @@
...
+        now,
+        now + 1000)
+    }
+  }
+
+  @Benchmark
+  @BenchmarkMode(Array(Mode.Throughput))
+  @OutputTimeUnit(TimeUnit.SECONDS)
+  def startTimeLookupWithPartId(): Unit = {
+    for ( i <- 0 to 1000 optimized) {
```
velvia: I'd prefer for this to do the same number of iterations as the other tests .... ie 21 or to 20.  Makes it much easier to compare

vishramachandran: Done

vishramachandran:
Updated results after changes:

```
[info] Benchmark                                                   Mode  Cnt     Score     Error  Units
[info] PartKeyIndexBenchmark.partIdsLookupWithEqualsFilters       thrpt   45   206.647 ±  12.208  ops/s
[info] PartKeyIndexBenchmark.partIdsLookupWithPrefixRegexFilters  thrpt   45   178.463 ±  18.127  ops/s
[info] PartKeyIndexBenchmark.partIdsLookupWithSuffixRegexFilters  thrpt   45   217.995 ±   7.646  ops/s
[info] PartKeyIndexBenchmark.startTimeLookupWithPartId            thrpt   45  3861.372 ± 113.547  ops/s
```

jmh/src/main/scala/filodb.jmh/PartKeyIndexBenchmark.scala
vishramachandran: After adding keys from all containers, the results seem to be more intuitive. Queries with wildcards in suffix perform better. When prefixes are provided, there are good data structures like tries for the indexing.

velvia:
@vishramachandran hmmm.  So the actual # of lookups possible with Lucene is around 200 * 20 = 4000 QPS?  That seems low to me.  
StartTime lookups are also not very fast but at least not in the ingestion path I think....  around 80K QPS.

PR #158, opened 2018-10-09
feat(core,query,prom): LP, ExecPlan, and _filodb_chunkmeta_all function to debug chunk metadata info

velvia:
**Pull Request checklist**

**New behavior :**

This PR creates a new special function, parseable through PromQL, `_filodb_chunkmeta_all` which returns `ChunkSetInfo` records for matching time series.  It returns all the fields from the ChunkSetInfo plus additional data (# bytes and the Reader class) for chunks in one particular column.  This data can be used to debug many issues including potential missing chunks or missing data, as well as to debug encoding and compression issues.

other related changes:
- One line minor fix for index startTime
- Fixed the standalone IngestionAndRecoverySpec; switched TestTimeseriesProducer to use the new Gateway pipeline

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,8 @@ Now run the time series generator. This will ingest 20 time series (the default)
 java -cp gateway/target/scala-2.11/gateway-*-SNAPSHOT filodb.timeseries.TestTimeseriesProducer -c conf/timeseries-dev-source.conf
 ```
 
+NOTE: The `TestTimeseriesProducer` logs to logs/gateway-server.log.
+
 At this point, you should be able to confirm such a message in the server logs: `KAMON counter name=memstore-rows-ingested count=4999`
 
 Now you are ready to query FiloDB for the ingested data. The following command should return matching subset of the data that was ingested by the producer.
--- a/README.md
+++ b/README.md
@@ -39,6 +39,8 @@ See [architecture](doc/architecture.md) and [datasets and reading](doc/datasets_
   - [Predicate Pushdowns](#predicate-pushdowns)
   - [Sharding](#sharding)
 - [Using the FiloDB HTTP API](#using-the-filodb-http-api)
+- [PromQL Compatibility](#promql-compatibility)
+  - [FiloDB Extensions](#filodb-extensions)
 - [Using FiloDB Data Source with Spark](#using-filodb-data-source-with-spark)
   - [Configuring FiloDB](#configuring-filodb)
     - [Passing Cassandra Authentication Settings](#passing-cassandra-authentication-settings)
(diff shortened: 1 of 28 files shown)


README.md
```diff
@@ -379,6 +383,18 @@ TODO: add details about FiloDB sharding, the shard-key vs partition key mechanis
...
+
+Current status:
+
+### FiloDB Extensions
+
+Some special functions exist to aid debugging and for other purposes:
+
+| function | description    |
+|----------|----------------|
+| `_filodb_chunkmeta_all` | (CLI Only) Returns chunk metadata fields for all chunks matching the time range and filter criteria - ID, # rows, start and end time, as well as the number of bytes and type of encoding used for a particular column.  |
```
vishramachandran: Can we add example usage from CLI ?

core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
```diff
@@ -274,4 +274,51 @@ class FilteredChunkInfoIterator(base: ChunkInfoIterator, filter: ChunkSetInfo =>
...
+/**
+ * A RowReader that returns info about each ChunkSetInfo set via setInfo with a schema like this:
+ * 0:  ID (Long)
+ * 1:  NumRows (Int)
+ * 2:  startTime (Long)
+ * 3:  endTime (Long)
+ * 4:  numBytes(Int) of chunk
+ * 5:  readerclass of chunk
+ */
+class ChunkInfoRowReader(column: Column) extends RowReader {
```
vishramachandran: Following additional information will be useful:
* encoding format for the chunk
* whether it is a write-buffer

Eventually, these can be added:
* Block information (odp or regular) 
* Lock status (how many readers) for debugging locking issues.

core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
```diff
@@ -16,6 +16,8 @@ object TimeSeriesPartition extends StrictLogging {
   val nullChunks    = UnsafeUtils.ZeroPointer.asInstanceOf[AppenderArray]
   val nullInfo      = ChunkSetInfo(UnsafeUtils.ZeroPointer.asInstanceOf[BinaryRegion.NativePointer])
 
+  val _log = logger
```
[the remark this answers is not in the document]
velvia: It was needed for debug logging in TSPartition.

README.md
velvia: Sure.

core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
velvia: Yes definitely.   BTW the first two, encoding format for the chunk is what the "readerclass" is.  If the class is one of the uncompressed ones then it's a write buffer (there isn't really any other current way to tell, other than it being the first chunk).   

In future versions we can add a function which can derive the writebuffer-ness from the existing info...

agreed on block info

velvia:
@vishramachandran comments addressed in latest commit.  Will merge sometime after PRB.

PR #159, opened 2018-10-10
feat(core,gateway): Unify shardKeyHash algorithm and add copyTags feature to PromInputRecord

velvia:
**Pull Request checklist**

- InfluxRecord and PrometheusInputRecord have different shard key hash algorithms
- PrometheusInputRecord is missing some features

**New behavior :**

- InfluxRecord, PrometheusInputRecord and the QueryActor now all use the same hash algorithm and the number of methods is reduced down to one.   Namely the keys (which never changes) is not part of the calculation any longer
- PrometheusInputRecord now rejects "le" or whatever is defined in ignoreTagsForPartitionKeyHash
- `PrometheusInputRecord` can rename/copy tags that are missing based on new `copyTags` `DatasetOptions` config
- `PrometheusInputRecord` parsing is about 15% faster than before, due to just trimming the metric and not having to sort and process a second set of key-value pairs

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/queryengine/Utils.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryengine/Utils.scala
@@ -106,10 +106,10 @@ object Utils extends StrictLogging {
     }
 
   private def shardHashFromFilters(filters: Seq[ColumnFilter], shardColumns: Seq[String]): Option[Int] = {
-    val shardColValues = shardColumns.map { shardCol =>
+    val shardValMap = shardColumns.map { shardCol =>
       // So to compute the shard hash we need shardCol == value filter (exact equals) for each shardColumn
       filters.find(f => f.column == shardCol) match {
-        case Some(ColumnFilter(_, Filter.Equals(filtVal: String))) => filtVal
+        case Some(ColumnFilter(_, Filter.Equals(filtVal: String))) => shardCol -> filtVal
         case Some(ColumnFilter(_, filter)) =>
           logger.debug(s"Found filter for shard column $shardCol but $filter cannot be used for shard key routing")
           return None
@@ -117,9 +117,9 @@ object Utils extends StrictLogging {
           logger.debug(s"Could not find filter for shard key column $shardCol, shard key hashing disabled")
           return None
       }
-    }
-    logger.debug(s"For shardColumns $shardColumns, extracted filter values $shardColValues successfully")
-    Some(RecordBuilder.shardKeyHash(shardColumns, shardColValues))
+    }.toMap
+    val metric = shardValMap("__name__")
+    Some(RecordBuilder.shardKeyHash((shardValMap - "__name__").values.toSeq, metric))
   }
 
   /**
--- a/coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
@@ -56,16 +56,16 @@ class QueryEngine(dataset: Dataset,
 
   private def shardsFromFilters(filters: Seq[ColumnFilter],
                                 options: QueryOptions): Seq[Int] = {
-    val shardColumns = dataset.options.shardKeyColumns
+    val shardColumns = dataset.options.shardKeyColumns.sorted
     require(shardColumns.nonEmpty || options.shardOverrides.nonEmpty,
       s"Dataset ${dataset.ref} does not have shard columns defined, and shard overrides were not mentioned")
 
     options.shardOverrides.getOrElse {
-      val shardColValues = shardColumns.map { shardCol =>
+      val shardVals = shardColumns.map { shardCol =>
         // So to compute the shard hash we need shardCol == value filter (exact equals) for each shardColumn
         filters.find(f => f.column == shardCol) match {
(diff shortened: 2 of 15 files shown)


gateway/src/main/scala/filodb/gateway/conversion/InputRecord.scala
```diff
@@ -41,52 +44,79 @@ trait InputRecord {
...
+      val metric = metricTags.head._2
+      val transformedTags = transformTags(tags.filterNot(_._1 == dataset.options.metricColumn), dataset).toMap
+      (0 until tsProto.getSamplesCount).map { i =>
+        val sample = tsProto.getSamples(i)
+        PrometheusInputRecord(transformedTags, metric, dataset, sample.getTimestampMs, sample.getValue)
+      }
+    }
+  }
+
+  def transformTags(tags: Seq[(String, String)], dataset: Dataset): Seq[(String, String)] = {
```
vishramachandran: Can we document what we are transforming. Useful for quick understanding

gateway/src/main/scala/filodb/gateway/conversion/InputRecord.scala
```diff
@@ -41,52 +44,79 @@ trait InputRecord {
...
   }
 }
 
 object PrometheusInputRecord {
   val DefaultShardHash = -1
 
-  // Create PrometheusInputRecord from a TimeSeries protobuf object
-  def apply(tsProto: TimeSeries, dataset: Dataset): PrometheusInputRecord = {
+  // Create PrometheusInputRecords from a TimeSeries protobuf object
+  def apply(tsProto: TimeSeries, dataset: Dataset): Seq[PrometheusInputRecord] = {
```
vishramachandran: It appears this method along with transform tags is doing many iterations of tags. The scala map/filter constructs are not helping. Can we just do one (or as few) iterations of the tags and calculate everything in a mutable way ? That would be slightly more efficient and will create less garbage as well. Helps since it is in the ingestion path for each and every sample.

gateway/src/main/scala/filodb/timeseries/TestTimeseriesProducer.scala
```diff
@@ -189,7 +189,7 @@ object TestTimeseriesProducer extends StrictLogging {
 
             // Compute partition, shard key hashes and compute shard number
             // TODO: what to do if not all shard keys included?  Just return a 0 hash?  Or maybe random?
-            val shardKeyHash = RecordBuilder.combineHashIncluding(kvsForShardCalc, hashes, shardKeys).get
+            val shardKeyHash = 0 // TODO: this code is being replaced w/ the PrometheusInputRecord
```
vishramachandran: TODO for this PR, or TODO for later PR?
Will this break dev test cycle approach?

coordinator/src/main/scala/filodb.coordinator/queryengine/Utils.scala
```diff
@@ -106,20 +106,20 @@ object Utils extends StrictLogging {
...
         case _ =>
           logger.debug(s"Could not find filter for shard key column $shardCol, shard key hashing disabled")
           return None
       }
-    }
-    logger.debug(s"For shardColumns $shardColumns, extracted filter values $shardColValues successfully")
-    Some(RecordBuilder.shardKeyHash(shardColumns, shardColValues))
+    }.toMap
+    val metric = shardValMap("__name__")
+    Some(RecordBuilder.shardKeyHash((shardValMap - "__name__").values.toSeq, metric))
```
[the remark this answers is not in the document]
velvia: @vishramachandran this is actually the old QueryEngine code which should be decommissioned.   We cannot access the dataset in this method, unlike the newer QueryEngine.   I can add a TODO.

coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
```diff
@@ -56,16 +56,16 @@ class QueryEngine(dataset: Dataset,
 
   private def shardsFromFilters(filters: Seq[ColumnFilter],
                                 options: QueryOptions): Seq[Int] = {
-    val shardColumns = dataset.options.shardKeyColumns
+    val shardColumns = dataset.options.shardKeyColumns.sorted
```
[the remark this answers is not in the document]
velvia: ok

gateway/src/main/scala/filodb/gateway/conversion/InputRecord.scala
velvia: @vishramachandran let's leave this for later if this is really needed.   There is really no point to put any more effort into the Prometheus protobuf format, as we are moving away from it long term.  The code in here already speeds up the process by 15% actually.

gateway/src/main/scala/filodb/timeseries/TestTimeseriesProducer.scala
velvia: The other PR needs to be merged, as this code is going away due to the other PR.

coordinator/src/main/scala/filodb.coordinator/queryengine/Utils.scala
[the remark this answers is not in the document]
velvia: Fixed it

velvia:
@vishramachandran comments addressed, for most part.

PR #160, opened 2018-10-11
feat(query): implement query support for Resets, Deriv, IDelta and Irate Functions

sherali42:
**Pull Request checklist**

resets, deriv, idelta and irate functions are not supported in the query

**New behavior :**

- logical plan creation for resets, deriv, idelta and irate functions
- execution logic for resets, deriv, idelta and irate functions
- bug fix in the query parser for labels having special characters - "." and "-"

### DIFF
--- a/prometheus/src/main/scala/filodb/prometheus/parse/Parser.scala
+++ b/prometheus/src/main/scala/filodb/prometheus/parse/Parser.scala
@@ -13,7 +13,7 @@ trait BaseParser extends Expressions with JavaTokenParsers with RegexParsers wit
   }
 
   lazy val labelIdentifier: PackratParser[Identifier] = {
-    "[a-zA-Z_][a-zA-Z0-9_:]*".r ^^ { str => Identifier(str) }
+    "[a-zA-Z_][a-zA-Z0-9_:\\-\\.]*".r ^^ { str => Identifier(str) }
   }
 
   lazy val string: PackratParser[Identifier] = {
--- a/query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
+++ b/query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
@@ -67,7 +67,7 @@ trait RangeFunction {
             sampleToEmit: TransientRow,
             queryConfig: QueryConfig): Unit
 }
-
+// scalastyle:off
 object RangeFunction {
   def apply(func: Option[RangeFunctionId],
             funcParams: Seq[Any] = Nil): RangeFunction = {
@@ -76,6 +76,10 @@ object RangeFunction {
       case Some(Rate)             => RateFunction
       case Some(Increase)         => IncreaseFunction
       case Some(Delta)            => DeltaFunction
+      case Some(Resets)           => ResetsFunction
+      case Some(Irate)            => IRateFunction
+      case Some(Idelta)           => IDeltaFunction
+      case Some(Deriv)            => DerivFunction
       case Some(MaxOverTime)      => new MinMaxOverTimeFunction(Ordering[Double])
       case Some(MinOverTime)      => new MinMaxOverTimeFunction(Ordering[Double].reverse)
       case Some(CountOverTime)    => new CountOverTimeFunction()
@@ -87,6 +91,7 @@ object RangeFunction {
     }
   }
 }
+// scalastyle:on
 
 object LastSampleFunction extends RangeFunction {
 
--- a/query/src/main/scala/filodb/query/exec/rangefn/RateFunctions.scala
+++ b/query/src/main/scala/filodb/query/exec/rangefn/RateFunctions.scala
@@ -49,6 +49,71 @@ object RateFunctions {
       Some(result)
     }
   }
+
+  def derivFunction(window: Window): Option[Double] = {
+    if (window.size < 2) {
+      None  // cannot calculate result without 2 samples
+    } else {
+      // We pass in an arbitrary timestamp that is near the values in use
+      // to avoid floating point accuracy issues
+      Some(linearRegression(window, window(0).timestamp)._1)
+    }
+  }
+
+  /**
+    * Logic is kept consistent with Prometheus' linearRegression function in order to get consistent results.
+    * We can look at optimizations (if any) later.
+    */
+  def linearRegression(window: Window, interceptTime: Long): Tuple2[Double, Double] = {
+    var n = 0.0
+    var sumX = 0.0
+    var sumY = 0.0
+    var sumXY = 0.0
+    var sumX2 = 0.0
+    for (i <- 0 until window.size) {
+      var sample = window(i)
+      var x = (sample.timestamp-interceptTime).toDouble / 1000.0
+      n += 1.0
+      sumY += sample.value
+      sumX += x
+      sumXY += x * sample.value
+      sumX2 += x * x
+    }
+    val covXY = sumXY - sumX*sumY/n
+    val varX = sumX2 - sumX*sumX/n
+    val slope = covXY / varX
+    val intercept = sumY/n - slope*sumX/n
+    Tuple2 (slope, intercept)
+  }
+
+  def instantValue(startTimestamp: Long,
+                       endTimestamp: Long,
+                       window: Window,
+                       isRate: Boolean): Option[Double] = {
+    if (window.size < 2) {
+      None  // cannot calculate result without 2 samples
+    } else {
+      require(window.head.timestamp >= startTimestamp, "Possible internal error, found samples < startTimestamp")
+      require(window.last.timestamp <= endTimestamp, "Possible internal error, found samples > endTimestamp")
+      var lastSample = window.last.value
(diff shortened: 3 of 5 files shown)


query/src/main/scala/filodb/query/exec/rangefn/RateFunctions.scala
```diff
@@ -49,6 +49,71 @@ object RateFunctions {
...
+      // to avoid floating point accuracy issues
+      Some(linearRegression(window, window(0).timestamp)._1)
+    }
+  }
+
+  /**
+    * Logic is kept consistent with Prometheus' linearRegression function in order to get consistent results.
+    * We can look at optimizations (if any) later.
+    */
+  def linearRegression(window: Window, interceptTime: Long): Tuple2[Double, Double] = {
```
vishramachandran: Tuple[Double, Double] will generate a lot of garbage. Please see if you can avoid this

query/src/main/scala/filodb/query/exec/rangefn/RateFunctions.scala
```diff
@@ -49,6 +49,71 @@ object RateFunctions {
...
+    val varX = sumX2 - sumX*sumX/n
+    val slope = covXY / varX
+    val intercept = sumY/n - slope*sumX/n
+    Tuple2 (slope, intercept)
+  }
+
+  def instantValue(startTimestamp: Long,
+                       endTimestamp: Long,
+                       window: Window,
+                       isRate: Boolean): Option[Double] = {
```
vishramachandran: Per sample Option creates lot of garbage. Need to reconsider this.

query/src/main/scala/filodb/query/exec/rangefn/RateFunctions.scala
```diff
@@ -100,3 +165,74 @@ object DeltaFunction extends RangeFunction {
...
+  def removeFromWindow(row: TransientRow): Unit = {}
+
+  def apply(startTimestamp: Long,
+            endTimestamp: Long,
+            window: Window,
+            sampleToEmit: TransientRow,
+            queryConfig: QueryConfig): Unit = {
+    var resets = 0
+    var prev = window.head.value
+    for (i <- 1 until window.size) {
```
vishramachandran: This can be done more efficiently with O(n) complexity using the addToWindow and removeFromWindow functions. This implementation will result in O(n^w) complexity, where w is length of time window

query/src/main/scala/filodb/query/exec/rangefn/RateFunctions.scala
```diff
@@ -49,6 +49,71 @@ object RateFunctions {
       Some(result)
     }
   }
+
+  def derivFunction(window: Window): Option[Double] = {
```
vishramachandran: Avoid Option

query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
```diff
@@ -67,7 +67,7 @@ trait RangeFunction {
             sampleToEmit: TransientRow,
             queryConfig: QueryConfig): Unit
 }
-
+// scalastyle:off
```
broneill: What style was violated here? Perhaps we shouldn't use it all?

query/src/main/scala/filodb/query/exec/rangefn/RateFunctions.scala
broneill: What is the preferred choice here?

query/src/main/scala/filodb/query/exec/rangefn/RateFunctions.scala
broneill: What should be used instead here?

query/src/main/scala/filodb/query/exec/rangefn/RateFunctions.scala
```diff
@@ -49,6 +49,71 @@ object RateFunctions {
...
+    val intercept = sumY/n - slope*sumX/n
+    Tuple2 (slope, intercept)
+  }
+
+  def instantValue(startTimestamp: Long,
+                       endTimestamp: Long,
+                       window: Window,
+                       isRate: Boolean): Option[Double] = {
+    if (window.size < 2) {
+      None  // cannot calculate result without 2 samples
```
broneill: Is this an exception? Or perhaps NaN?

query/src/main/scala/filodb/query/exec/rangefn/RateFunctions.scala
vishramachandran: The caller does 
```
result.getOrElse(Double.NaN)
```
the NaN needs to be brought up here. Yes, a bit ugly but better to avoid the Option.

query/src/main/scala/filodb/query/exec/rangefn/RateFunctions.scala
vishramachandran: Same here. Bring the NaN up into this method and return Double.

query/src/main/scala/filodb/query/exec/rangefn/RateFunctions.scala
vishramachandran: This is not an exception. This essentially indicates that the time window is too small to calculate any rate. This should ideally result in a NoValue (NA or null) but we do not have such a capability in our result vectors. Hence we use an NaN. If the Option is removed and brought up into the method, this would be replaced with NaN

query/src/main/scala/filodb/query/exec/rangefn/RateFunctions.scala
vishramachandran: BTW, I should not have simply said "avoid Option".
I actually like the scala Option instead of passing around nulls (or other values). But using Option here is a hot code path is what needs to be reconsidered.

query/src/main/scala/filodb/query/exec/rangefn/RateFunctions.scala
sherali42: @vishramachandran "intercept" is needed for `predict_linear` function, but for now making it double. We can add it back when we implement it.

query/src/main/scala/filodb/query/exec/rangefn/RateFunctions.scala
sherali42: 👍

query/src/main/scala/filodb/query/exec/rangefn/RateFunctions.scala
sherali42: 👍 on it

query/src/main/scala/filodb/query/exec/rangefn/RateFunctions.scala
sherali42: 👍  will change it

query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
sherali42: @broneill no more than 11 cases i think, here we have 13

prometheus/src/main/scala/filodb/prometheus/parse/Parser.scala
```diff
@@ -13,7 +13,7 @@ trait BaseParser extends Expressions with JavaTokenParsers with RegexParsers wit
   }
 
   lazy val labelIdentifier: PackratParser[Identifier] = {
-    "[a-zA-Z_][a-zA-Z0-9_:]*".r ^^ { str => Identifier(str) }
+    "[a-zA-Z_][a-zA-Z0-9_:\\-\\.]*".r ^^ { str => Identifier(str) }
```
velvia: Is this to allow dashes and periods in labels?

query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
velvia: @sherali42 instead of turning off the style checker, can we just increase the # of cases allowed?  It's in scalastyle.xml

query/src/main/scala/filodb/query/exec/rangefn/RateFunctions.scala
```diff
@@ -46,16 +46,81 @@ object RateFunctions {
...
+    val varX = sumX2 - sumX*sumX/n
+    val slope = covXY / varX
+    val intercept = sumY/n - slope*sumX/n // keeping it, needed for predict_linear function = slope*duration + intercept
+    slope
+  }
+
+  def instantValue(startTimestamp: Long,
+                       endTimestamp: Long,
+                       window: Window,
+                       isRate: Boolean): Double = {
```
velvia: I wonder about this API.   Wouldn't it be better for there to be a separate rate function?  Why try to mix up the rate function and embed it inside of this?

prometheus/src/main/scala/filodb/prometheus/parse/Parser.scala
sherali42: @velvia yes

PR #162, opened 2018-10-19
fix(memory,core): Chunkmap locking

broneill:
**Pull Request checklist**

**New behavior :**

Changes to fix memory reclamation race conditions when accessing chunks. The OffheapLFSortedIDMap now includes a lock, which should be used everywhere within FiloDB where chunks are accessed. This might reduce some concurrency, but it should resolve the memory errors and corruption that we're seeing.


broneill:
Reviewers, please take a look. I've completed the changes as far as I can tell.

memory/src/main/scala/filodb.memory/data/OffheapLFSortedIDMap.scala
```diff
@@ -246,22 +294,151 @@ class OffheapLFSortedIDMapReader(memFactory: MemFactory, holderKlass: Class[_ <:
...
+          return true
+        }
+        // Scala doesn't support a simple break statement for some reason.
+        done = true
+      }
+    } while (!done)
+
+    // Wait for shared lock owners to release the lock.
+
+    val endNanos = System.nanoTime + timeoutNanos
```
velvia: Right now timeoutNanos defaults to 1ms, which seems way too low right?  Seems like we need to make it configurable

memory/src/main/scala/filodb.memory/data/OffheapLFSortedIDMap.scala
```diff
@@ -246,22 +294,151 @@ class OffheapLFSortedIDMapReader(memFactory: MemFactory, holderKlass: Class[_ <:
...
+    private def doClose(): Unit = {
+       closed = true;
+       releaseShared(inst)
+    }
+
     final def hasNext: Boolean = {
+      if (closed) return false
       nextElem = at(inst, curIdx + 1)
-      curIdx < (length(inst) - 1) && !isPtrNull(nextElem) && continue(nextElem)
+      val result = curIdx < (doLength(inst) - 1) && !isPtrNull(nextElem) && continue(nextElem)
```
velvia: I think we can probably get rid of the null check since locks protect all accesses?

memory/src/main/scala/filodb.memory/data/OffheapLFSortedIDMap.scala
```diff
@@ -246,22 +294,151 @@ class OffheapLFSortedIDMapReader(memFactory: MemFactory, holderKlass: Class[_ <:
...
   // curIdx has to be initialized to one less than the starting logical index
-  // NOTE: This always fetches items using the official API.  THis is slower, but guarantees that no matter
+  // NOTE: This always fetches items using the official API.  This is slower, but guarantees that no matter
   // how slowly the iterator user pulls, it will always be pulling the right thing even if state/mapPtr changes.
   private class SortedIDMapElemIterator(inst: MapHolder, var curIdx: Int, continue: NativePointer => Boolean)
   extends ElementIterator {
+    var closed: Boolean = false
     var nextElem: NativePointer = _
+
+    val caller: Exception = new Exception()
```
velvia: I don't see this used?  Not sure what the point of this is, and it uses up extra memory for each iteration.

memory/src/main/scala/filodb.memory/data/OffheapLFSortedIDMap.scala
```diff
@@ -360,31 +537,30 @@ extends OffheapLFSortedIDMapReader(memFactory, holderKlass) {
   //scalastyle:on
 
   /**
-   * Atomically calls the elementFunc and inserts the element it returns IF AND ONLY IF
-   * the item with the given key is not already in the map.  elementFunc is not called if key already exists.
+   * Atomically inserts the element it returns IF AND ONLY IF
```
velvia: Sorry, what was the rationale again for not being able to use a function?

memory/src/main/scala/filodb.memory/data/OffheapLFSortedIDMap.scala
```diff
@@ -442,13 +611,18 @@ extends OffheapLFSortedIDMapReader(memFactory, holderKlass) {
...
-      if (casState(mapPtr, curState, MapState.empty))
-        if (UnsafeUtils.unsafe.compareAndSwapLong(inst, mapPtrOffset, mapPtr, 0))
-          memFactory.freeMemory(mapPtr)
-      curState = state(inst)
+    acquireExclusive(inst)
+    try {
+      var curState = state(inst)
+      while (curState != MapState.empty) {
+        val mapPtr = inst.mapPtr
+        if (casState(mapPtr, curState, MapState.empty))
```
velvia: With an exclusive lock held, it seems likely unnecessary to do a CAS or check for a zero state, which in theory should never be possible with proper use of locks.  Though I guess extra safety doesn't hurt.

core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
```diff
@@ -244,34 +250,38 @@ trait ChunkInfoIterator { base: ChunkInfoIterator =>
...
 class ElementChunkInfoIterator(elIt: ElementIterator) extends ChunkInfoIterator {
+  def close(): Unit = elIt.close()
   def hasNext: Boolean = elIt.hasNext
   def nextInfo: ChunkSetInfo = ChunkSetInfo(elIt.next)
 }
 
 class FilteredChunkInfoIterator(base: ChunkInfoIterator, filter: ChunkSetInfo => Boolean) extends ChunkInfoIterator {
   var nextnext: ChunkSetInfo = ChunkSetInfo(0)
   var gotNext: Boolean = false
+  var calls: Int = 0
```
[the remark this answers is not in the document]
broneill: Probably a mistake.

memory/src/main/scala/filodb.memory/data/OffheapLFSortedIDMap.scala
broneill: It doubles each time, but the loop around it never gives up.

memory/src/main/scala/filodb.memory/data/OffheapLFSortedIDMap.scala
broneill: Oops. Mistake. It shouldn't be here.

memory/src/main/scala/filodb.memory/data/OffheapLFSortedIDMap.scala
broneill: If the function is called when the lock is held, and the function decides to allocate memory, that reclaim listener could go back to the chunk map and cause a deadlock.

memory/src/main/scala/filodb.memory/data/OffheapLFSortedIDMap.scala
broneill: I just decided to leave all the cas stuff in place for now, to keep the code change churn down.

core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
```diff
@@ -249,11 +290,21 @@ extends ReadablePartition with MapHolder {
     * not been paged into memory
     */
   final def timestampOfLatestSample: Long = {
-    if (numChunks == 0) -1
-    else infoLast.endTime
+    if (numChunks == 0) {
+      -1
+    } else {
+      // Acquire shared lock to safely access the native pointer.
+      offheapInfoMap.acquireShared(this)
```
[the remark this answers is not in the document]
broneill: It would cause an object to be allocated. A macro might work better.

core/src/test/scala/filodb.core/memstore/TimeSeriesPartitionSpec.scala
```diff
@@ -107,9 +107,10 @@ class TimeSeriesPartitionSpec extends MemFactoryCleanupTest with ScalaFutures {
     data1 shouldEqual (minData take 10)
 
     val blockHolder = new BlockMemFactory(blockStore, None, dataset1.blockMetaSize)
-    val flushFut = Future(part.makeFlushChunks(blockHolder))
+    // Task needs to fully iterate over the chunks, to release the shared lock.
+    val flushFut = Future(part.makeFlushChunks(blockHolder).toSeq)
```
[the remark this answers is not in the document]
broneill: What is the point of using the future here?

core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
```diff
@@ -236,11 +239,49 @@ extends ReadablePartition with MapHolder {
...
+      return new ChunkInfoIterator {
+        var closed = false
+
+        def close(): Unit = {
+          if (!closed) doClose()
+        }
+
+        private def doClose(): Unit = {
+          closed = true
+          offheapInfoMap.releaseShared(TimeSeriesPartition.this)
```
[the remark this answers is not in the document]
broneill: If in the same class, sure.

broneill:
Still seeing JVM crashes, this one from SerializationSpec:

# An error report file with more information is saved as:
# /home/travis/build/filodb/FiloDB/coordinator/hs_err_pid4989.log
Compiled method (nm)  284978 4482     n 0       com.kenai.jffi.Foreign::freeMemory (native)
 total in heap  [0x00007f675939dd90,0x00007f675939e0d8] = 840
 relocation     [0x00007f675939deb8,0x00007f675939df00] = 72
 main code      [0x00007f675939df00,0x00007f675939e0d0] = 464
 oops           [0x00007f675939e0d0,0x00007f675939e0d8] = 8
Compiled method (nm)  284978 4482     n 0       com.kenai.jffi.Foreign::freeMemory (native)
 total in heap  [0x00007f675939dd90,0x00007f675939e0d8] = 840
 relocation     [0x00007f675939deb8,0x00007f675939df00] = 72
 main code      [0x00007f675939df00,0x00007f675939e0d0] = 464
 oops           [0x00007f675939e0d0,0x00007f675939e0d8] = 8

core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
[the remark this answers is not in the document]
velvia: How would it result in an object?  Oh you mean to run some code inside.
Ok let me research some macro stuff and send it to you

core/src/test/scala/filodb.core/memstore/TimeSeriesPartitionSpec.scala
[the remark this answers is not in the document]
velvia: Not sure actually.  Maybe it was once needed and no longer needed.

memory/src/main/scala/filodb.memory/data/OffheapLFSortedIDMap.scala
```diff
@@ -246,22 +294,151 @@ class OffheapLFSortedIDMapReader(memFactory: MemFactory, holderKlass: Class[_ <:
...
+
+    var timeoutNanos = 1000000 // 1 millisecond
+    var warned = false
+
+    while (true) {
+      if (tryAcquireExclusive(inst, timeoutNanos)) {
+        return
+      }
+
+      timeoutNanos = Math.min(timeoutNanos << 1, 1000000000) // limit timeout to 1 second
```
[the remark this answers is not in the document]
velvia: Ah.  I'm fine with putting constants in a Java static class somewhere if it's faster.

memory/src/main/scala/filodb.memory/data/OffheapLFSortedIDMap.scala
```diff
@@ -246,22 +294,151 @@ class OffheapLFSortedIDMapReader(memFactory: MemFactory, holderKlass: Class[_ <:
...
+    do {
+      lockState = UnsafeUtils.getIntVolatile(inst, lockStateOffset)
+      if (lockState < 0) {
+        // Wait for exclusive lock to be released.
+        Thread.`yield`
+      } else if (UnsafeUtils.unsafe.compareAndSwapInt(inst, lockStateOffset, lockState, lockState | 0x80000000)) {
+        if (lockState == 0) {
+          return true
+        }
+        // Scala doesn't support a simple break statement for some reason.
```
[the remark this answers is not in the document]
velvia: Oh, I didn't know that.  :(

velvia:
@broneill what are the details of the SerializationSpec crash?    My guess is it might be cleanup related.  Somehow the nativeMemoryManager is being shut down before the free happens, maybe, but just a guess.

core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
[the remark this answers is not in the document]
broneill: I just implemented simple withShared and withExclusive methods, which generate two objects per usage. One for the code body, and one for the boxed return value.

broneill:
I just copied the crash info from the travis log which is linked from this PR.

core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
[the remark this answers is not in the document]
broneill: I find the syntax for using macros to be quite complicated, and it looks like it relies on reflection. This isn't going to perform any better.

broneill:
@velvia Did you try the develop branch or mine?

memory/src/main/scala/filodb.memory/data/OffheapLFSortedIDMap.scala
```diff
@@ -246,22 +286,173 @@ class OffheapLFSortedIDMapReader(memFactory: MemFactory, holderKlass: Class[_ <:
     innerBinSearch(0, mapLen)
   }
 
+  /**
+   * Acquire exclusive access to this map, spinning if necessary. Exclusive lock isn't re-entrant.
+   */
+  def acquireExclusive(inst: MapHolder): Unit = {
```
vishramachandran: Some metrics to track number of locks held?

core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
```diff
@@ -244,34 +250,36 @@ trait ChunkInfoIterator { base: ChunkInfoIterator =>
...
-
-  def single(info: ChunkSetInfo): ChunkInfoIterator = new ChunkInfoIterator {
-    var count = 0
-    def hasNext: Boolean = count == 0
-    def nextInfo: ChunkSetInfo = { count += 1; info }
-  }
 }
 
 class ElementChunkInfoIterator(elIt: ElementIterator) extends ChunkInfoIterator {
+  def close(): Unit = elIt.close()
```
vishramachandran: I dont see a call for the iterator's close() method from anywhere except the tests. Did I miss it? I was expecting for it to be invoked from PartitionTimeRangeReader class where we iterate across the chunks.

core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
```diff
@@ -244,34 +250,36 @@ trait ChunkInfoIterator { base: ChunkInfoIterator =>
...
   def hasNext: Boolean = elIt.hasNext
   def nextInfo: ChunkSetInfo = ChunkSetInfo(elIt.next)
```
vishramachandran: Do we handle exception thrown during next or hasNext? Will lock be released?

memory/src/main/scala/filodb.memory/data/OffheapLFSortedIDMap.scala
```diff
@@ -246,22 +286,173 @@ class OffheapLFSortedIDMapReader(memFactory: MemFactory, holderKlass: Class[_ <:
...
+    // Spin-lock implementation. Because the owner of the shared lock might be blocked by this
+    // thread as it waits for an exclusive lock, deadlock is possible. To mitigate this problem,
+    // timeout and retry, allowing shared lock waiters to make progress. The timeout doubles
+    // for each retry, up to a limit, but the retries continue indefinitely.
+
+    var timeoutNanos = 1.millisecond.toNanos
+    var warned = false
+
+    while (true) {
+      if (tryAcquireExclusive(inst, timeoutNanos)) {
```
vishramachandran: Wonder if it would make sense to have some metrics to measure and report contention (like wait times). This can help us correlate slowness in ingestion to contention in block acquisition.

core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
```diff
@@ -244,34 +250,36 @@ trait ChunkInfoIterator { base: ChunkInfoIterator =>
...
-  def single(info: ChunkSetInfo): ChunkInfoIterator = new ChunkInfoIterator {
-    var count = 0
-    def hasNext: Boolean = count == 0
-    def nextInfo: ChunkSetInfo = { count += 1; info }
-  }
 }
 
 class ElementChunkInfoIterator(elIt: ElementIterator) extends ChunkInfoIterator {
+  def close(): Unit = elIt.close()
   def hasNext: Boolean = elIt.hasNext
```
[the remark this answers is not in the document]
broneill: In general yes, although the implementation isn't expected to throw any exceptions.

core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
```diff
@@ -236,11 +239,47 @@ extends ReadablePartition with MapHolder {
...
     case LastSampleChunkScan        => if (numChunks == 0) ChunkInfoIterator.empty
-                                       else ChunkInfoIterator.single(infoLast)
+                                       else {
+      // Return a single element iterator which holds a shared lock.
+      offheapInfoMap.acquireShared(this)
+      try {
+        new OneChunkInfo(infoLast)
+      } catch {
+        case e: Throwable => offheapInfoMap.releaseShared(this); throw e;
+      }
```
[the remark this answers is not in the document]
broneill: Alignment of what?

memory/src/main/scala/filodb.memory/data/OffheapLFSortedIDMap.scala
broneill: Only one exclusive can be held at a time. Do you mean a counter of all locks? Or do you mean gauge of shared locks held? It can cause contention.

core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
broneill: All the uses of the iterator that I could see always iterate to the end. Are there cases I missed?

memory/src/main/scala/filodb.memory/data/OffheapLFSortedIDMap.scala
broneill: Most effective strategy would be to rely on JMX or something and sample all the threads for hotspots.

memory/src/main/scala/filodb.memory/data/OffheapLFSortedIDMap.scala
```diff
@@ -246,22 +286,173 @@ class OffheapLFSortedIDMapReader(memFactory: MemFactory, holderKlass: Class[_ <:
...
+        return
+      }
+
+      timeoutNanos = Math.min(timeoutNanos << 1, 1.second.toNanos) // limit timeout
+
+      if (!warned && timeoutNanos >= 1.second.toNanos) {
+        _logger.warn(s"Waiting for exclusive lock: $inst")
```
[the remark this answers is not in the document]
broneill: What is the preferred way to provide such a counter in the code today?

A mitigation strategy to simply release all the shared locks would cause more harm than good. The lock count could end up going negative as share locks are released.

core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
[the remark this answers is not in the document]
vishramachandran: code alignment. Convention is to align code in a case block to same indentation. :)

memory/src/main/scala/filodb.memory/data/OffheapLFSortedIDMap.scala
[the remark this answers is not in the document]
vishramachandran: We use Kamon to report metrics. One place to put the metric would be inside `object OffheapLFSortedIDMap`. You can see `TimeSeriesShardStats` for some examples.

memory/src/main/scala/filodb.memory/data/OffheapLFSortedIDMap.scala
vishramachandran: True. Wondering how we can provide some visibility to locking and contention via metrics effectively. Logging in and checking JMX is more of a debugging operation. If we can alert on this proactively, that would be good.

core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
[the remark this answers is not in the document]
vishramachandran: I realize that this may get ugly, but as I mentioned in another thread of this PR, exceptions can be thrown by upstream query processing logic. We may need to handle exceptions on next/hasNext at the topmost level and propagate the close() call down.

broneill:
I commented out all the code that frees memory, and now finally the tests pass. So something is corrupting the heap when it frees memory. Either a double free, or freeing of a bogus address.

broneill:
I'll add a global memory allocation address tracker, to help identify the problem.

broneill:
Coordinator test still randomly fails.

memory/src/main/scala/filodb.memory/data/OffheapLFSortedIDMap.scala
velvia: I'm OK with a counter to count shared and exclusive lock operations.  Gauge is too invasive and slow probably.  However a counter on something which is concurrent can definitely cause contention also.   Adding a counter to `TimeSeriesStats` makes it shard-local and might not be too bad.

memory/src/main/scala/filodb.memory/data/OffheapLFSortedIDMap.scala
```diff
@@ -300,19 +302,20 @@ class OffheapLFSortedIDMapReader(memFactory: MemFactory, holderKlass: Class[_ <:
...
+    var timeoutNanos = 1.millisecond.toNanos
     var warned = false
 
     while (true) {
       if (tryAcquireExclusive(inst, timeoutNanos)) {
         return
       }
 
-      timeoutNanos = Math.min(timeoutNanos << 1, 1000000000) // limit timeout to 1 second
+      timeoutNanos = Math.min(timeoutNanos << 1, 1.second.toNanos) // limit timeout
```
velvia: What I meant was to move this `1.second.toNanos` into a global val so that the cost of doing this conversion is not in every call.....

velvia:
@broneill thanks for changes, only looked at the one commit with initial feedback.  Gave you a comment about the `1.second.toNanos`.  Will finish looking later.

memory/src/main/scala/filodb.memory/data/OffheapLFSortedIDMap.scala
broneill: Ok.

memory/src/main/scala/filodb.memory/data/OffheapLFSortedIDMap.scala
broneill: Added some counters,

core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
broneill: Added some counters.

memory/src/main/scala/filodb.memory/data/OffheapLFSortedIDMap.scala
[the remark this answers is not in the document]
broneill: Added some counters.

core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
broneill: Added a releaseAllSharedLocks which runs when the query finishes. This covers all missed cases.

broneill:
I think I've resolved all issues. Please review again.

memory/src/main/scala/filodb.memory/data/OffheapLFSortedIDMap.scala
```diff
@@ -55,8 +72,82 @@ object OffheapLFSortedIDMap extends StrictLogging {
...
+
+  // Tracks all the shared locks held, by each thread.
+  val sharedLockCounts = new ThreadLocal[Map[MapHolder, Int]]
+
+  // Lock state memory offsets for all known MapHolder classes.
+  val lockStateOffsets = new ConcurrentHashMap[Class[_ <: MapHolder], Long]
+
+  // Updates the shared lock count, for the current thread.
+  //scalastyle:off
+  def adjustSharedLockCount(inst: MapHolder, amt: Int): Unit = {
```
velvia: I'm worried about this method and the next one, as well as the accompanying data structures.   Updating these structures are not cheap, and they could potentially use a ton of memory too (every Long and Int will be boxed, in both data structures).  When will they be updated?  When will the methods be called?  Do they grow based on the number of locks held?

core/src/main/scala/filodb.core/store/ReadablePartition.scala
```diff
@@ -58,13 +58,15 @@ trait ReadablePartition extends FiloPartition {
   /**
    * Obtains the native offheap pointer for the chunk id for data column columnID
    */
-  def dataChunkPointer(id: ChunkID, columnID: Int): BinaryVector.BinaryVectorPtr
+  // Disabled for now. Requires a shared lock on the chunk map.
+  //def dataChunkPointer(id: ChunkID, columnID: Int): BinaryVector.BinaryVectorPtr
```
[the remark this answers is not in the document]
broneill: I didn't find any callers, and so I commented it out for simplicity. Otherwise, native pointer leakage is a source of bugs.

memory/src/main/scala/filodb.memory/data/OffheapLFSortedIDMap.scala
broneill: A few extra object allocations is a bit small compared to all the others created during a query. I don't know how Scala int boxing works, but in Java, small values are cached.

velvia:
@broneill I'm fine with this moving ahead.   As for the test, if it is the ClusterRecoverySpec multi-JVM, as I suggested there is a patienceConfig which is an `implicit val` in the test which can be adjusted easily to see if it helps make it pass with a larger value.

PR #163, opened 2018-10-25
fix(core): Don't schedule an empty ODP task on the ingestion thread

velvia:
… Speeds up in-memory queries

**Pull Request checklist**

An ODP task is always created on the ingestion thread for every query, whether there are on-demand paging partitions or not.   This leads to higher query latencies and potential timeouts if the ingestion thread gets stuck.

**New behavior :**

If there are no new partitions to on-demand page (i.e. all partitions asked are already in memory, which is vast majority of times), an empty task with Nil is returned immediately rather than being scheduled on the ingestion thread.  Thus in most cases the ingestion thread will not block queries.

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/OnDemandPagingShard.scala
+++ b/core/src/main/scala/filodb.core/memstore/OnDemandPagingShard.scala
@@ -102,7 +102,8 @@ TimeSeriesShard(dataset, storeConfig, shardNum, rawStore, metastore, evictionPol
   // 3. Deal with partitions no longer in memory but still indexed in Lucene.
   //    Basically we need to create TSPartitions for them in the ingest thread -- if there's enough memory
   private def odpPartTask(indexIt: PartitionIterator, partKeyBytesToPage: ArrayBuffer[Array[Byte]],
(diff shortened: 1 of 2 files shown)


core/src/main/scala/filodb.core/memstore/OnDemandPagingShard.scala
```diff
@@ -102,44 +102,45 @@ TimeSeriesShard(dataset, storeConfig, shardNum, rawStore, metastore, evictionPol
...
+            partSet.add(part)
+            callback(partKeyBytesRef.bytes, unsafeKeyOffset)
+            part
+          }
+          // create the partition and update data structures (but no need to add to Lucene!)
+          // NOTE: if no memory, then no partition!
+        case p: TimeSeriesPartition => Some(p)
+      }
+    }.toVector.flatten
+  }
```
vishramachandran: Can we add a warning message if partIDs.isEmpty ? Bugs like this in the future will be brought to surface.

velvia: I did better than a warning, I added a require that partIDs cannot be empty  :)

PR #164, opened 2018-10-30
bug(query): Sliding window bug fixes

vishramachandran:
**Pull Request checklist**

Bug Fixes for:
1. Sliding window keeps a last sample even after time series stops ingesting,
   so functions like SumOverTime return incorrect value after stop.
2. Sliding window should not include values at window start time.
   Otherwise, values are double counted for the case where step==window

### DIFF
--- a/prometheus/src/main/scala/filodb/prometheus/query/PrometheusModel.scala
+++ b/prometheus/src/main/scala/filodb/prometheus/query/PrometheusModel.scala
@@ -67,7 +67,9 @@ object PrometheusModel {
       b.addLabels(LabelPair.newBuilder().setName(lv._1.toString).setValue(lv._2.toString))
     }
     srv.rows.foreach { row =>
-      b.addSamples(Sample.newBuilder().setTimestampMs(row.getLong(0)).setValue(row.getDouble(1)))
+      if (!row.getDouble(1).isNaN) {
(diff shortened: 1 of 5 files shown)


broneill:
See pull #165 which is out for review. It fixes the JVM crash issue.

prometheus/src/main/scala/filodb/prometheus/query/PrometheusModel.scala
```diff
@@ -67,7 +67,9 @@ object PrometheusModel {
       b.addLabels(LabelPair.newBuilder().setName(lv._1.toString).setValue(lv._2.toString))
     }
     srv.rows.foreach { row =>
-      b.addSamples(Sample.newBuilder().setTimestampMs(row.getLong(0)).setValue(row.getDouble(1)))
+      if (!row.getDouble(1).isNaN) {
```
velvia: Won't this prevent NaNs from being reported for raw sample queries?

vishramachandran: Good catch. I intended this only for HTTP API, not for raw time series. Applying this fix only for HTTP results.

Ideally, I would like to not apply this for HTTP results as well, but until we are able to set "NA" as a value in our vectors, we may have to continue to use NaN and filter them from results.

PR #166, opened 2018-10-31
feat(query,coordinator): Special __col PromQL filter for target column name

velvia:
**Pull Request checklist**

There is no way to specify the target column name for PromQL queries.  Instead, it is assumed the schema is timestamp/value, and the second column is always used as the value column.

**New behavior :**

1. A special PromQL tag, `__col`, is exposed and can be used to target a specific column for querying
2. `SelectRawPartitionsExec` now takes numeric ColumnIDs, and column name to ID validation is done during LP to EP materialize call.
3. `__name__` filter is transformed into the right filter if the dataset metricColumn is not __name__

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
@@ -16,6 +16,7 @@ import filodb.core.binaryrecord2.RecordBuilder
 import filodb.core.metadata.Dataset
 import filodb.core.query.{ColumnFilter, Filter}
 import filodb.core.Types
+import filodb.prometheus.ast.Vectors.PromMetricLabel
 import filodb.query._
 import filodb.query.exec._
 
@@ -179,10 +180,11 @@ class QueryEngine(dataset: Dataset,
                                    options: QueryOptions,
                                    lp: RawSeries): Seq[ExecPlan] = {
     val colIDs = getColumnIDs(dataset, lp.columns)
-    shardsFromFilters(lp.filters, options).map { shard =>
+    val renamedFilters = renameMetricFilter(lp.filters)
+    shardsFromFilters(renamedFilters, options).map { shard =>
       val dispatcher = dispatcherForShard(shard)
       SelectRawPartitionsExec(queryId, submitTime, options.itemLimit, dispatcher, dataset.ref, shard,
-        lp.filters, toRowKeyRange(lp.rangeSelector), colIDs)
+        renamedFilters, toRowKeyRange(lp.rangeSelector), colIDs)
     }
   }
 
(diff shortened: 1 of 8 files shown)


coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
```diff
@@ -191,13 +195,42 @@ class QueryEngine(dataset: Dataset,
...
-        lp.filters, toRowKeyRange(lp.rangeSelector), colID)
+        renamedFilters, toRowKeyRange(lp.rangeSelector), colID)
     }
   }
 
+  /**
+   * Renames Prom AST __name__ metric name filters to one based on the actual metric column of the dataset,
+   * if it is not the prometheus standard
+   */
+  private def renameMetricFilter(filters: Seq[ColumnFilter]): Seq[ColumnFilter] =
```
vishramachandran: IMO, the rename is more appropriate in the Plan materialization time, rather than at execution time. Can we move this to the QueryEngine class ?

prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
```diff
@@ -87,6 +93,11 @@ trait Vectors extends Scalars with TimeUnits with Base {
           case NotEqual(false) => ColumnFilter(labelMatch.label, query.Filter.NotEquals(labelMatch.value))
           case other: Any      => throw new IllegalArgumentException(s"Unknown match operator $other")
         }
+      }.filterNot(_.column == ColumnSelectorLabel)
```
vishramachandran: Can we document here why we are removing this column from the filters?

From my understanding, it is because the selected column names are part of the logical plan as a different input argument.

prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
```diff
@@ -87,6 +93,11 @@ trait Vectors extends Scalars with TimeUnits with Base {
...
+
+    protected def labelMatchesToColumnName(labels: Seq[LabelMatch]): Seq[String] =
+      labels.collect {
+        case LabelMatch(ColumnSelectorLabel, EqualMatch, col) => col
```
vishramachandran: EqualMatch only at the moment. We may support multiple columns with the regex filter or some other custom "in" operator that we can specify in PromQL.

query/src/main/scala/filodb/query/exec/SelectRawPartitionsExec.scala
```diff
@@ -24,24 +24,17 @@ final case class SelectRawPartitionsExec(id: String,
                                          shard: Int,
                                          filters: Seq[ColumnFilter],
                                          rowKeyRange: RowKeyRange,
-                                         columns: Seq[String]) extends LeafExecPlan {
+                                         colIds: Seq[Types.ColumnId]) extends LeafExecPlan {
```
vishramachandran: Can we add a `require(colIds.nonEmpty)`

prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
```diff
@@ -4,8 +4,15 @@ import filodb.core.query
 import filodb.core.query.ColumnFilter
 import filodb.query._
 
+object Vectors {
+  // The "tag" or key used to indicate the column to query to FiloDB.  A non-standard Prom extension.
+  val ColumnSelectorLabel = "__col"
```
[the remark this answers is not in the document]
vishramachandran: This is so that it is consistent with the "internal tags" convention.

coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
vishramachandran: I am mistaken! My eyes deceived me. For some reason, I thought this was in the SelectPartitionsExec class. :)

prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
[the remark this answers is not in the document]
velvia: ok fixed.

velvia:
@vishramachandran all feedback addressed.  Going to merge

PR #174, opened 2018-11-08
bug(core): Remove more eager iteration by ChunkIterators

vishramachandran:
**Pull Request checklist**

We filter out Range Vectors that are returned by the index, but do not have any data in memory. In order to do that we do a hasNext in the query data pipeline setup phase itself. This can cause unreleased locks in some cases. 

There is one other place where we materialize a SerializableRangeVector, and needs to be protected with the lock release logic.

**New behavior :**

Fix is to check if nonEmpty using a different chunk iterator than the one used to iterate through the data. We release lock for that check immediately.

### DIFF
--- a/query/src/main/scala/filodb/query/exec/AggrOverRangeVectors.scala
+++ b/query/src/main/scala/filodb/query/exec/AggrOverRangeVectors.scala
@@ -5,12 +5,14 @@ import java.nio.ByteBuffer
 import scala.collection.mutable
 
 import com.tdunning.math.stats.{ArrayDigest, TDigest}
+import com.typesafe.scalalogging.StrictLogging
 import monix.reactive.Observable
 
 import filodb.core.binaryrecord2.RecordBuilder
 import filodb.core.metadata.Column.ColumnType
 import filodb.core.metadata.Dataset
 import filodb.core.query._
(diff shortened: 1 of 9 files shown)


query/src/main/scala/filodb/query/exec/AggrOverRangeVectors.scala
```diff
@@ -513,18 +515,30 @@ class TopBottomKRowAggregator(k: Int, bottomK: Boolean) extends RowAggregator {
...
-        builder.startNewRecord()
-        builder.addLong(row.getLong(0))
-        builder.addDouble(row.getDouble(i + 1))
-        builder.endRecord()
-        i += 2
+    // Important TODO / TechDebt: We need to replace Iterators with cursors to better control
+    // the chunk iteration, lock acquisition and release. This is much needed for safe memory access.
+    try {
+      // validate no locks are held by the thread. If there are, quite possible a lock acquire or release bug exists
+      val numLocksReleased = OffheapLFSortedIDMap.releaseAllSharedLocks()
```
broneill: Add a common utility method in OffheapLFSortedIDMap, validateNoSharedLocks, which does the release and warning.

PR #183, opened 2018-11-27
bug(query): fix to get all assigned shards when there is no filter/ shardCol filters for label values query

sherali42:
**Pull Request checklist**

Fixed the following bug:

- For label values query, if the filter query doesn't contain any shard columns, API is throwing exception
- Fix is to get all the assigned shards for such queries to dispatch to all shards and gather results


coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
```diff
@@ -198,7 +198,17 @@ class QueryEngine(dataset: Dataset,
...
                                       lp: LabelValues): Seq[LabelValuesExec] = {
-    shardsFromFilters(lp.filters, options).map { shard =>
+    var shards: Seq[Int] = Seq.empty
+    try {
+      shards = shardsFromFilters(lp.filters, options)
+    } catch {
+      // expected - filters don't have enough info to derive shard mapping, send query to all the shards
+      // reusing the existing logic
+      case e: BadQueryException =>
+        logger.error("exception while materializing the plan", e)
```
broneill: If this is expected, is this an error or warn? Or info?

sherali42: @broneill ack, updated to warn

vishramachandran:
@sherali42 Why cant we require that the shard key columns be present in the query ?
We do not support queries without the shard key columns today anyway. That is a bridge we can cross later, not now.

sherali42:
> @sherali42 Why cant we require that the shard key columns be present in the query ?
> We do not support queries without the shard key columns today anyway. That is a bridge we can cross later, not now.

@vishramachandran This change is to handle the metadata queries such as fetching all Applications. There won't be any other filter criteria for fetching this list.

vishramachandran:
@sherali42 along with the FiloCLI SEGV fix, I will be appending a few more fixes that address the comments I added to this PR. Becomes easier since you are out on vacation right now. :)

coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
```diff
@@ -198,7 +198,17 @@ class QueryEngine(dataset: Dataset,
                                       submitTime: Long,
                                       options: QueryOptions,
                                       lp: LabelValues): Seq[LabelValuesExec] = {
-    shardsFromFilters(lp.filters, options).map { shard =>
+    var shards: Seq[Int] = Seq.empty
+    try {
+      shards = shardsFromFilters(lp.filters, options)
```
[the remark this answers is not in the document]
velvia: @vishramachandran @sherali42 I don't think the PR is needed at all.    Vish I suggest we back out this change.  If we really want to reach all shards just pass the shardOverrides into the query options.  This should be done by the upper layer query service.

vishramachandran:
@sherali42 This PR can be closed now without merge now that my other PR contains bug fix you provided here

PR #185, opened 2018-11-30
feat(coordinator): Custom downing provider and CoordinatedShutdown for split brains

velvia:
**Pull Request checklist**

Currently we are using Akka Cluster's auto-downing in several places, which is really not recommended for production and also leads to split brains with each side not coming down.

**New behavior :**

1. Switch to use akka-cluster-custom-downing provider, with a default strategy of majority-quorum.  The default settings cause the minority side to shut itself down automatically, thus preventing a lingering split brain situation, and leading to smooth resolution of split brains.
2. Switch to Akka's CoordinatedShutdown sequence instead of our custom one.  Add hooks for shutdown of the memstore, etc.  One advantage of this scheme is that the shutdown will first remove the QueryActors so that new queries will stop immediately before the rest of the shutdown commences.  Another advantage, besides coordination with the split brain resolver, is that the JVM will exit after shutdown, thus guaranteeing the minority side does not survive and new nodes can come up.
3. Minor bug fix for NodeClusterActor so that if ingestionconfig recovery errors out, it will continue to come up.  Otherwise there is no singleton in the system.

Have verified that the custom downing provider works:

```
[2018-12-03 17:03:51,136] INFO  t.a.c.a.MajorityLeaderAutoDown [filo-standalone-akka.actor.default-dispatcher-19] - Majority is auto-downing unreachable node [akka.tcp://filo-standalone@127.0.0.1:56723]
[2018-12-03 17:03:51,137] INFO  a.c.Cluster(akka://filo-standalone) [filo-standalone-akka.actor.default-dispatcher-19] - Cluster Node [akka.tcp://filo-standalone@127.0.0.1:2552] - Marking unreachable node [akka.tcp://filo-standalone@127.0.0.1:56723] as [Down]
[2018-12-03 17:03:53,095] INFO  a.c.Cluster(akka://filo-standalone) [filo-standalone-akka.actor.default-dispatcher-47] - Cluster Node [akka.tcp://filo-standalone@127.0.0.1:2552] - Leader can perform its duties again
[2018-12-03 17:03:53,103] INFO  a.c.Cluster(akka://filo-standalone) [filo-standalone-akka.actor.default-dispatcher-47] - Cluster Node [akka.tcp://filo-standalone@127.0.0.1:2552] - Leader is removing unreachable node [akka.tcp://filo-standalone@127.0.0.1:56723]
[2018-12-03 17:03:53,103] INFO  f.a.ClusterMembershipTracker [filo-standalone-akka.actor.default-dispatcher-18] - Member is Removed: akka.tcp://filo-standalone@127.0.0.1:56723 after Down. New member list Set(akka.tcp://filo-standalone@127.0.0.1:2552, akka.tcp://filo-standalone@127.0.0.1:56732)
```

### TODO

- Add an explicit split brain test (@helena are there examples of using the Akka multi-jVM test kit's facilities to inject a network partition?)

**BREAKING CHANGES**
None

**Other information**:

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/FilodbCluster.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/FilodbCluster.scala
@@ -3,12 +3,14 @@ package filodb.coordinator
 import java.util.concurrent.atomic.{AtomicBoolean, AtomicReference}
 
 import scala.collection.immutable
-import scala.concurrent.Await
+import scala.concurrent.{Await, Future}
+import scala.concurrent.duration._
 
 import akka.actor._
 import akka.cluster._
 import akka.cluster.ClusterEvent._
 import akka.util.Timeout
+import akka.Done
 import com.typesafe.config.{Config, ConfigFactory}
 import com.typesafe.scalalogging.StrictLogging
 import kamon.Kamon
@@ -172,50 +174,45 @@ final class FilodbCluster(val system: ExtendedActorSystem) extends Extension wit
       actor
     }
 
-  /** Begins node graceful shutdown:
-    *   A. Sets `isTerminating` to true and `isJoined` and `isInitialized` to false,
-    *   B. Calls `akka.cluster.Cluster.leave(selfAddress)`
-    *   C. Awaits the async graceful shutdown of the `filodb.coordinator.NodeGuardian`
-    *      supervised child actors via `filodb.coordinator.GracefulStopStrategy`
-    *   D. Shuts down `monix.execution.Scheduler.io` and `filodb.core.store.StoreFactory`'s MetaStore and MemStore
-    *   E. Calls the async terminate on `akka.actor.ActorSystem`
-    *
-    * Idempotent.
-    */
-  protected[filodb] def shutdown(): Unit =
-    if (!isTerminated) {
-      import akka.pattern.gracefulStop
+  /**
+   * Hook into Akka's CoordinatedShutdown sequence
+   * Please see https://doc.akka.io/docs/akka/current/actors.html#coordinated-shutdown for details
+   */
+  CoordinatedShutdown(system).addTask(CoordinatedShutdown.PhaseServiceUnbind, "queryShutdown") { () =>
+    implicit val timeout = Timeout(15.seconds)
+    // Reset shuts down all ingestion and query actors on this node
+    // TODO: be sure that status gets updated across cluster?
+    (coordinatorActor ? NodeProtocol.ResetState).map(_ ⇒ Done)
+  }
 
-      import NodeProtocol.GracefulShutdown
+  // TODO: hook into service-stop "forcefully kill connections?"  Maybe send each outstanding query "TooBad" etc.
 
-      logger.info("Leaving the cluster.")
-      _cluster.get foreach (_.leave(selfAddress))
+  CoordinatedShutdown(system).addTask(CoordinatedShutdown.PhaseBeforeClusterShutdown, "storeShutdown") { () =>
+    _isInitialized.set(false)
+    logger.info("Terminating: starting shutdown")
 
-      _isInitialized.set(false)
-      logger.info("Terminating: starting shutdown")
-
-      try {
-        if (clusterActor.isDefined) {
-          Await.result(gracefulStop(guardian, GracefulStopTimeout, GracefulShutdown), GracefulStopTimeout)
-        }
-
-        system.terminate foreach { _ =>
-          logger.info("Actor system was shut down")
-        }
-        metaStore.shutdown()
-        memStore.shutdown()
+    try {
+      metaStore.shutdown()
+      memStore.shutdown()
+      ioPool.shutdown()
+    } catch {
+      case NonFatal(e) =>
+        system.terminate()
         ioPool.shutdown()
-      } catch {
-        case NonFatal(e) =>
-          system.terminate()
-          ioPool.shutdown()
-      }
-      finally _isTerminated.set(true)
     }
+    finally _isTerminated.set(true)
 
-  Runtime.getRuntime.addShutdownHook(new Thread() {
-    override def run(): Unit = shutdown()
-  })
+    Future.successful(Done)
+  }
+
+  /**
+   * Invokes CoordinatedShutdown to shut down the ActorSystem, leave the Cluster, and our hooks above
+   * which shuts down the stores and threadpools.
+   * NOTE: Depending on the setting of coordinated-shutdown.exit-jvm, this might cause the JVM to exit.
+   */
+  protected[filodb] def shutdown(): Unit = {
+    CoordinatedShutdown(system).run(CoordinatedShutdown.UnknownReason)
+  }
 
   /** For usage when the `akka.cluster.Member` is needed in a non-actor.
     * Creates a temporary actor which subscribes to `akka.cluster.MemberRemoved`.
@@ -285,7 +282,5 @@ private[filodb] trait FilodbClusterNode extends NodeConfiguration with StrictLog
   def clusterSingleton(role: ClusterRole, watcher: Option[ActorRef]): ActorRef =
     cluster.clusterSingleton(role, watcher)
 
-  /** Only calls `akka.cluster.leave` akka.cluster.Cluster was called by user. */
   def shutdown(): Unit = cluster.shutdown()
-
 }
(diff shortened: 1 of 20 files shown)


coordinator/src/main/scala/filodb.coordinator/FilodbCluster.scala
```diff
@@ -172,50 +174,45 @@ final class FilodbCluster(val system: ExtendedActorSystem) extends Extension wit
...
+    Future.successful(Done)
+  }
+
+  /**
+   * Invokes CoordinatedShutdown to shut down the ActorSystem, leave the Cluster, and our hooks above
+   * which shuts down the stores and threadpools.
+   * NOTE: Depending on the setting of coordinated-shutdown.exit-jvm, this might cause the JVM to exit.
+   */
+  protected[filodb] def shutdown(): Unit = {
+    CoordinatedShutdown(system).run(CoordinatedShutdown.UnknownReason)
```
helena: If you haven't already, I think you can now remove [filodb.coordinator.NodeProtocol.GracefulShutdown](https://github.com/filodb/FiloDB/blob/develop/coordinator/src/main/scala/filodb.coordinator/NodeGuardian.scala#L167) and its usage in specs, and previous custom logic around `filodb.coordinator.GracefulStopAwareSupervisor.gracefulShutdown` 
- https://github.com/filodb/FiloDB/blob/develop/coordinator/src/main/scala/filodb.coordinator/BaseActor.scala#L80
- https://github.com/filodb/FiloDB/blob/develop/coordinator/src/main/scala/filodb.coordinator/BaseActor.scala#L86
- https://github.com/filodb/FiloDB/blob/develop/coordinator/src/main/scala/filodb.coordinator/BaseActor.scala#L94-L107 

I no longer see anything using [GracefulStopStrategy](https://github.com/filodb/FiloDB/blob/develop/coordinator/src/main/scala/filodb.coordinator/GracefulStopStrategy.scala#L17), and there's duplicate code, so perhaps that can be removed as well. I think originally one was for supervisors, like NodeGuardian, and one for child workers to handle graceful shutdown of their processing when possible.

coordinator/src/main/scala/filodb.coordinator/FilodbCluster.scala
```diff
@@ -172,50 +174,45 @@ final class FilodbCluster(val system: ExtendedActorSystem) extends Extension wit
...
-      } catch {
-        case NonFatal(e) =>
-          system.terminate()
-          ioPool.shutdown()
-      }
-      finally _isTerminated.set(true)
     }
+    finally _isTerminated.set(true)
 
-  Runtime.getRuntime.addShutdownHook(new Thread() {
```
helena: nice to remove this now ;)

coordinator/src/test/scala/filodb.coordinator/FilodbClusterSpec.scala
```diff
@@ -29,9 +29,9 @@ class FilodbClusterSpec extends AkkaSpec {
       cluster.clusterActor.isDefined shouldEqual true
       pathElementsExist(clusterActor) should be (true)
     }
+    // Maybe we should just take out this test as we now use Akka's CoordinatedShutdown, not our own shutdown
```
helena: Yes, and add a test asserting coordinated shutdown does what it should and I'd add one simulating problems to verify expected behavior/see any issues.

core/src/main/resources/filodb-defaults.conf
```diff
@@ -280,7 +280,11 @@ akka {
 
     # Auto downing is turned off by default.  See
     # http://doc.akka.io/docs/akka/2.3.16/scala/cluster-usage.html#Automatic_vs__Manual_Downing
-    # auto-down-unreachable-after = 120s
+    # Instead, we use this repo to provide smarter downing solutions:
+    # https://github.com/TanUkkii007/akka-cluster-custom-downing
+    # Note that this strategy should work well but for fixed size clusters QuorumLeader might be better
+    downing-provider-class = "tanukki.akka.cluster.autodown.MajorityLeaderAutoDowning"
```
vishramachandran: How is "majority" count of the cluster mutated? In other words, what if some nodes go down and never come up because of an infrastructure issue - how long does it take for majority to be updated ?

If there is a split brain before the majority is updated, the entire cluster could go down ?

velvia: The "majority" algorithm is not foolproof, it is not meant to be the production strategy, but a better strategy than auto-downing for dev environments.   For production we'd use the `StaticQuorum` strategy which is pretty foolproof for static cluster sizes.   I just wanted a default which was easy to configure.

Here is the code for the MajorityLeader thing though.

https://github.com/TanUkkii007/akka-cluster-custom-downing/blob/master/src/main/scala/tanukki/akka/cluster/autodown/MajorityAwareCustomAutoDownBase.scala

vishramachandran: Quorum is also not foolproof in the exact same scenario. A few nodes go down, and don't come back up. We then see a split brain with fewer than quorum number of nodes on both sides. Both sides are brought down.

I actually like the MajorityLeader better since it would tolerate this scenario better, but only if the "majority" is updated reliably on both split brain sides.

velvia: Is this the scenario you are saying?
a) A few nodes go down... in which case, the majority is still up.  Both algorithms should work
b) A few nodes go down and the remaining nodes are split by a partition. In which case there is no majority.

In b), the Quorum would still work in the sense that both sides go down.  This is fine, there is still no split brain.  A whole new cluster will be created by the cloud provider.

vishramachandran: In (b) we _do not_ want all nodes to go down. Only one side of the split brain should be brought down. Why rebuild the entire cluster unnecessarily ?

coordinator/src/main/scala/filodb.coordinator/FilodbCluster.scala
velvia: Yes, thanks

coordinator/src/test/scala/filodb.coordinator/IngestionStreamSpec.scala
```diff
@@ -58,14 +58,14 @@ class IngestionStreamSpec extends ActorTest(IngestionStreamSpec.getNewSystem) wi
     metaStore.newDataset(dataset6).futureValue shouldEqual Success
     metaStore.newDataset(dataset33).futureValue shouldEqual Success
     coordinatorActor ! NodeProtocol.ResetState
+    expectMsg(NodeProtocol.StateReset)
     clusterActor ! NodeProtocol.ResetState
     expectMsg(NodeProtocol.StateReset)
     Thread sleep 2000
   }
 
   override def afterAll(): Unit = {
```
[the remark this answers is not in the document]
velvia: 👍

core/src/main/resources/filodb-defaults.conf
velvia: If both are minorities, then there is no automatic way to choose which side is right, so some kind of manual intervention is required.  I think it is better for all nodes to go down, thus new nodes will come up, then to have a temporary split brain.  
However I think this is extremely unlikely.

velvia:
@helena so I'm trying to extend `NodeClusterSpec` to test split brain by using the TestConductor blackhole.  It seems it requires special config.  I tried adding `testTransport(on = true)` but it still doesn't help..... do you know what's up with this?

```
[JVM-3] - should update PartitionMap and shut down node 2 on network partition *** FAILED ***
[JVM-3]   akka.ConfigurationException: To use this feature you must activate the failure injector adapters (trttl, gremlin) by specifying `testTransport(on = true)` in your MultiNodeConfig.
[JVM-3]   at akka.remote.testconductor.Conductor$class.requireTestConductorTranport(Conductor.scala:133)
[JVM-3]   at akka.remote.testconductor.Conductor$class.throttle(Conductor.scala:105)
[JVM-3]   at akka.remote.testconductor.TestConductorExt.throttle(Extension.scala:53)
[JVM-3]   at akka.remote.testconductor.Conductor$class.blackhole(Conductor.scala:124)
[JVM-3]   at akka.remote.testconductor.TestConductorExt.blackhole(Extension.scala:53)
[JVM-3]   at filodb.coordinator.NodeClusterSpec$$anonfun$7$$anonfun$apply$mcV$sp$14.apply$mcV$sp(NodeClusterSpec.scala:214)
[JVM-3]   at akka.remote.testkit.MultiNodeSpec.runOn(MultiNodeSpec.scala:370)
[JVM-3]   at filodb.coordinator.NodeClusterSpec$$anonfun$7.apply$mcV$sp(NodeClusterSpec.scala:213)
```

velvia:
@helena I found if I manually injected the testConductor config then it works.  In Akka's MultiNodeSpec.scala, line 117:

        val configs = (_nodeConf get myself).toList ::: _commonConf.toList ::: transportConfig :: MultiNodeSpec.nodeConfig :: MultiNodeSpec.baseConfig :: Nil


See how _commonConf is ahead of transportConfig?  That means if commonConf contains a bunch of global config then it can override transportConfig.

PR #186, opened 2018-12-03
misc(cli,core,kafka,memory,query): Support building with JDK 11.

broneill:
**Pull Request checklist**

**Current behavior :** Building with JDK 11 fails with compilation errors.

**New behavior :** Made changes to support JDK 11 without breaking compatibility with older versions.


core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -1020,9 +1020,21 @@ class TimeSeriesShard(val dataset: Dataset,
   }
 
   def shutdown(): Unit = {
+    // Don't release memory if the VM is actually shutting down. This can cause a segv if
+    // other threads are still accessing the memory. Everything gets freed when the process
+    // has completely exited anyhow. Note that this check isn't elegant or perfect.
+    val t = new Thread()
+    try {
+      Runtime.getRuntime().addShutdownHook(t);
```
vishramachandran: This method can be run in the context of a shutdown hook. Need to find out if another hook can be setup from a running shutdown hook and what the implications are.

broneill: I didn't really want this change to be in this PR. Pushed to wrong branch...

kafka/src/main/scala/filodb/kafka/package.scala
```diff
@@ -33,7 +33,11 @@ package object kafka {
   implicit final class MapOps(immutable: Map[String, AnyRef]) {
     def asProps: JProperties = {
       val props = new JProperties()
-      props.putAll(immutable.filter(_._2 != null).asJava)
+      for ((k,v) <- immutable) {
```
velvia: Or:

    immutable.foreach { case (k, v) => if (k != null && v != null) props.put(k, v) }

velvia: The if condition can also be included in the for comprehension.

broneill: Which format is more efficient? In general, I try to avoid passing functions into iterators, since it creates additional objects and the introduction of a virtual call in a tight loop can slow things down as well.

broneill:
@vishramachandran I needed the () stuff to resolve errors that the compiler was emitting.

PR #188, opened 2018-12-03
feat(core, query): Full support for BinaryRecordColumn

vishramachandran:
**Pull Request checklist**

In metadata query results, we were assuming that partKeys returned was
always a map. This is not necessarily true.

**New behavior :**

* Partition Keys can now be sent across as a binary record column.
* The schema of the binary record column is sent in separately
  as part of the ResultSchema object itself.
* RecordSchema now also includes column names so that client can
  string/mapify the binary record that is returned in results
* Fixed isses in metadata API which returns partition keys.
  We dont assume that it is a map now.

**BREAKING CHANGES**

Has changes in RecordSchema object. May not be able to do rolling upgrades. Clients need to be upgraded to new version.

### DIFF
--- a/core/src/main/scala/filodb.core/binaryrecord2/RecordSchema.scala
+++ b/core/src/main/scala/filodb.core/binaryrecord2/RecordSchema.scala
@@ -1,5 +1,7 @@
 package filodb.core.binaryrecord2
 
+import scala.collection.mutable.ArrayBuffer
+
 import filodb.core.metadata.{Column, Dataset}
 import filodb.core.metadata.Column.ColumnType.{LongColumn, TimestampColumn}
 import filodb.core.query.ColumnInfo
@@ -158,8 +160,23 @@ final class RecordSchema(val columns: Seq[ColumnInfo],
    * EXPENSIVE to do at server side. Creates a easy-to-read string
    * representation of the contents of this BinaryRecord.
    */
-  def stringify(base: Any, offset: Long): String =
-    s"b2[${toStringPairs(base, offset).map(_.productIterator.mkString("=")).mkString(",")}]"
+  def stringify(base: Any, offset: Long): String = {
+    import Column.ColumnType._
+    val result = new ArrayBuffer[String]()
+    columnTypes.zipWithIndex.map {
+      case (IntColumn, i)    => result += s"${colNames(i)}= ${getInt(base, offset, i)}"
+      case (LongColumn, i)   => result += s"${colNames(i)}= ${getLong(base, offset, i)}"
+      case (DoubleColumn, i) => result += s"${colNames(i)}= ${getDouble(base, offset, i)}"
+      case (StringColumn, i) => result += s"${colNames(i)}= ${asJavaString(base, offset, i)}"
+      case (TimestampColumn, i) => result += s"${colNames(i)}= ${getLong(base, offset, i)}"
+      case (BitmapColumn, i) => result += s"${colNames(i)}= ${getInt(base, offset, i) != 0}"
+      case (MapColumn, i)    => val consumer = new StringifyMapItemConsumer
+                                consumeMapItems(base, offset, i, consumer)
+                                result += s"${colNames(i)}= ${consumer.prettyPrint}"
+      case (BinaryRecordColumn, i)  => result += s"${colNames(i)}= ${brSchema(i).stringify(base, offset)}"
+    }
+    s"b2[${result.mkString(",")}]"
+  }
 
   def stringify(address: NativePointer): String = stringify(UnsafeUtils.ZeroPointer, address)
   def stringify(bytes: Array[Byte]): String = stringify(bytes, UnsafeUtils.arayOffset)
@@ -295,7 +312,7 @@ trait MapItemConsumer {
  */
 class StringifyMapItemConsumer extends MapItemConsumer {
   val stringPairs = new collection.mutable.ArrayBuffer[(String, String)]
-  def prettyPrint: String = "[" + stringPairs.map { case (k, v) => s"$k: $v" }.mkString(", ") + "]"
+  def prettyPrint: String = "{" + stringPairs.map { case (k, v) => s"$k: $v" }.mkString(", ") + "}"
   def consume(keyBase: Any, keyOffset: Long, valueBase: Any, valueOffset: Long, index: Int): Unit = {
     stringPairs += (UTF8StringMedium.toString(keyBase, keyOffset) ->
                     UTF8StringMedium.toString(valueBase, valueOffset))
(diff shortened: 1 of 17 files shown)


core/src/main/scala/filodb.core/binaryrecord2/RecordSchema.scala
```diff
@@ -144,33 +152,35 @@ final class RecordSchema(val columnTypes: Seq[Column.ColumnType],
...
+  def stringify(base: Any, offset: Long): String =
+    s"b2[${mapify(base, offset).map(_.productIterator.mkString("=")).mkString(",")}]"
 
   def stringify(address: NativePointer): String = stringify(UnsafeUtils.ZeroPointer, address)
   def stringify(bytes: Array[Byte]): String = stringify(bytes, UnsafeUtils.arayOffset)
 
+  /**
+    * EXPENSIVE to do at server side. Creates a stringified map with contents of this BinaryRecord.
+    */
+  def mapify(base: Any, offset: Long): Map[String, String] = {
```
velvia: IMO mapify() is not the same output as stringify().  Once you do this, then non-map items will get flattened along with the map key-values, and it makes it harder to debug the layout.  Can we please keep both versions?  Or at the very least, delineate where the maps start and end so that it's easy to see which elements are part of the map, and which are not.

core/src/main/scala/filodb.core/binaryrecord2/RecordSchema.scala
```diff
@@ -24,12 +25,15 @@ import filodb.memory.format.{vectors => bv}
  *   the fact that all variable length fields after the partitionFieldStart are contiguous and can be binary compared
  *
  * @param columnTypes In order, the field or column type of each field in this schema
+ * @param brSchema schema of any binary record type column
  * @param partitionFieldStart Some(n) from n to the last field are considered the partition key.  A field number.
  * @param predefinedKeys A list of predefined keys to save space for the tags/MapColumn field(s)
  */
-final class RecordSchema(val columnTypes: Seq[Column.ColumnType],
+final class RecordSchema(val colNames: Seq[String],
```
velvia: IMO if you are going to have both the names and the types, it would be much simpler just to take in the columns, ie `val columns: Seq[Column]` instead of just names and types.

velvia: I'd prefer one less parameter

core/src/main/scala/filodb.core/binaryrecord2/RecordSchema.scala
```diff
@@ -288,6 +299,12 @@ class StringifyMapItemConsumer extends MapItemConsumer {
   }
 }
 
+class MapifyMapItemConsumer(toAdd: collection.mutable.Map[String, String]) extends MapItemConsumer {
```
velvia: You could just use the output of `StringifyMapItemConsumer` and do .toMap....  less code duplication

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -276,6 +276,10 @@ class PartKeyLuceneIndex(dataset: Dataset,
                            partKeyNumBytes: Int,
                            partId: Int, startTime: Long, endTime: Long): Document = {
     val document = new Document()
+    // TODO We can use RecordSchema.mapify to get the name/value pairs from partKey.
```
velvia: No we really don't want to do this.  It creates a huge amount of objects.  mapify should NOT be used in any ingestion paths at all, for memory reasons.  Please instead note that we cannot use mapify due to too much memory overhead.

core/src/main/scala/filodb.core/query/RangeVector.scala
```diff
@@ -143,23 +142,23 @@ final class SerializableRangeVector(val key: RangeVectorKey,
...
           } else {
             schema.columnTypes(0) match {
               // FIXME: underlying stringify method assumes that partition key is a map. Not necessarily true
-              case PartitionKeyColumn => schema.stringify(reader.getBlobBase(0), reader.getBlobOffset(0))
+              case BinaryRecordColumn => schema.stringify(reader.getBlobBase(0), reader.getBlobOffset(0))
               case _ => reader.getAny(0).toString
             }
           }
-          (firstCol +: (1 until schema1.length).map(reader.getAny(_).toString)).mkString("\t")
+          (firstCol +: (1 until schema.columnTypes.length).map(reader.getAny(_).toString)).mkString("\t")
```
velvia: There should still be a length method.  Or use `numFields`

core/src/main/scala/filodb.core/query/RangeVector.scala
```diff
@@ -143,23 +142,23 @@ final class SerializableRangeVector(val key: RangeVectorKey,
...
       rows.map {
         case br: BinaryRecord if br.isEmpty =>  "\t<empty>"
         case reader =>
-          val firstCol = if (formatTime && schema1.isTimeSeries) {
+          val firstCol = if (formatTime && schema.isTimeSeries) {
             val timeStamp = reader.getLong(0)
             s"${new DateTime(timeStamp).toString()} (${(curTime - timeStamp)/1000}s ago) $timeStamp"
           } else {
             schema.columnTypes(0) match {
               // FIXME: underlying stringify method assumes that partition key is a map. Not necessarily true
```
velvia: You can remove this now

core/src/main/scala/filodb.core/binaryrecord2/RecordSchema.scala
velvia: Another problem is that maps do not preserve order.  The entire point of prettyPrint is that it must print out fields in order and we must be able to tell what is in each field, for easy debugging.
I think what you can do is keep most of stringify().  For mapColumns, use mapify().  For PartitionKeyColumns, call stringify recursively on itself.  Then the layout is clear.

velvia: Sorry, not pretty print, stringify.

vishramachandran: The intention of this flattening together is for binary record processing details to be abstracted within. We don't want clients to be processing columns separately depending on the schema. The usage of `mapify` would for example be to resolve the partKey into a bunch of key-value pairs such as to respond to a prometheus api. We dont want caller to act situationally based on the schema. One piece of code should apply to all cases. If there is a use case for unflattened map, then we should have another API.

For preserving order, we can certainly do Seq[Tuple2] instead of map. That may be more appropriate.

core/src/main/scala/filodb.core/binaryrecord2/RecordSchema.scala
vishramachandran: Sure. Will try this out.

vishramachandran:
@velvia 

> I don't think mapify should be recursive, and I don't think the inner RecordSchema necessarily belongs in the RecordSchema. It is better for mapify() to leave out the inner BinaryRecordColumn, but the user can call mapify() on the schema for just the BinaryRecord in the BinaryRecordColumn itself

The intention here is to have clients use one code to convert the partKey binary record into string key-value pairs irrespective of the partKey schema. We dont want to write client code that is schema-aware. `mapify`, now renamed to `toStringPairs` is just a convenience method. This is a direct requirement for clients serving promQL API. If there are other requirements, we can have other convenience methods.

PR #189, opened 2018-12-06
 feat(core, query): Drop out of order samples at ingestion time, not query time

velvia:
**Pull Request checklist**

Currently new data samples are appended as they are.  During query time, out of order samples are dropped in an iterator.

**New behavior :**

A new field, `lastTime` is added to the `TimeSeriesPartition`.  During ingestion, timestamp of new sample is compared with lastTime and dropped if it is less than lastTime.  A new counter keeps track of dropped samples.

The iterator which dropped out of order samples is skipped during query.

Before:
```
[info] Benchmark                                Mode  Cnt     Score    Error  Units
[info] QueryInMemoryBenchmark.parallelQueries  thrpt   30  1347.586 ± 27.054  ops/s
```

After:
```
[info] Benchmark                                Mode  Cnt     Score    Error  Units
[info] QueryInMemoryBenchmark.parallelQueries  thrpt   30  1370.396 ± 85.834  ops/s
```

**BREAKING CHANGES**

None

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
+++ b/core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
@@ -61,7 +61,8 @@ class TimeSeriesPartition(val partID: Int,
                           // Shared class for mutating the infoMap / OffheapLFSortedIDMap given mapPtr above
                           offheapInfoMap: OffheapLFSortedIDMapMutator,
                           // Lock state used by OffheapLFSortedIDMap.
-                          var lockState: Int = 0)
+                          var lockState: Int = 0,
+                          var lastTime: Long = -1L)
 extends ReadablePartition with MapHolder {
   import TimeSeriesPartition._
 
@@ -108,13 +109,15 @@ extends ReadablePartition with MapHolder {
    * @param blockHolder the BlockMemFactory to use for encoding chunks in case of WriteBuffer overflow
    */
   final def ingest(row: RowReader, blockHolder: BlockMemFactory): Unit = {
+    val ts = dataset.timestamp(row)
+    if (ts < lastTime) {
+      shardStats.outOfOrderDropped.increment
+      return
+    }
     if (currentChunks == nullChunks) {
       // First row of a chunk, set the start time to it
-      initNewChunk(dataset.timestamp(row))
+      initNewChunk(ts)
     }
-
-    // Update the end time as well.  For now assume everything arrives in increasing order
-    ChunkSetInfo.setEndTime(currentInfo.infoAddr, dataset.timestamp(row))
 
     for { col <- 0 until numColumns optimized } {
       currentChunks(col).addFromReaderNoNA(row, col) match {
@@ -125,7 +128,10 @@ extends ReadablePartition with MapHolder {
         case other: AddResponse =>
       }
     }
+    lastTime = ts
     ChunkSetInfo.incrNumRows(currentInfo.infoAddr)
+    // Update the end time as well.  For now assume everything arrives in increasing order
+    ChunkSetInfo.setEndTime(currentInfo.infoAddr, ts)
   }
 
   private def nonEmptyWriteBuffers: Boolean = currentInfo != nullInfo && currentInfo.numRows > 0
@@ -354,7 +360,7 @@ extends ReadablePartition with MapHolder {
   private def infoGet(id: ChunkID): ChunkSetInfo = ChunkSetInfo(offheapInfoMap(this, id))
 
   // Caller must hold lock on offheapInfoMap.
-  private def infoLast(): ChunkSetInfo = ChunkSetInfo(offheapInfoMap.last(this))
+  private[core] def infoLast(): ChunkSetInfo = ChunkSetInfo(offheapInfoMap.last(this))
 
   private def infoPut(info: ChunkSetInfo): Unit = {
     offheapInfoMap.withExclusive(this, offheapInfoMap.put(this, info.infoAddr))
--- a/core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
+++ b/core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
@@ -62,6 +62,7 @@ class TimeSeriesPartition(val partID: Int,
                           offheapInfoMap: OffheapLFSortedIDMapMutator,
                           // Lock state used by OffheapLFSortedIDMap.
                           var lockState: Int = 0,
+                          // NOTE: this is a var because much faster to access this than lastInfo.endTime
                           var lastTime: Long = -1L)
 extends ReadablePartition with MapHolder {
   import TimeSeriesPartition._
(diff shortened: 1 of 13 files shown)


core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
```diff
@@ -61,7 +61,8 @@ class TimeSeriesPartition(val partID: Int,
                           // Shared class for mutating the infoMap / OffheapLFSortedIDMap given mapPtr above
                           offheapInfoMap: OffheapLFSortedIDMapMutator,
                           // Lock state used by OffheapLFSortedIDMap.
-                          var lockState: Int = 0)
+                          var lockState: Int = 0,
+                          var lastTime: Long = -1L)
```
vishramachandran: I assume you are not fetching the lastTime from writeBuffers or chunks due to performance reasons? Is that confirmed with a benchmark? Perhaps we can document it here.

vishramachandran: Looks like there may be an issue when a series is paged in. `lastTime` would be -1 and an out-of-order sample may potentially be introduced. You need to add logic to update lastTime once a partition is paged in.

query/src/test/scala/filodb/query/exec/SlidingWindowIteratorSpec.scala
```diff
@@ -133,7 +133,8 @@ class SlidingWindowIteratorSpec extends FunSpec with Matchers {
                     1538416644000L->210170299d,
                     1538416649000L->210172635d)
 
-  it ("should ignore out of order samples for RateFunction") {
+  // Out of order samples are now handled by ingestion
+  ignore("should ignore out of order samples for RateFunction") {
```
vishramachandran: FYI.. This test accepts out of order data above.

core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
vishramachandran: lastChunkInfo.lastTime should provide us this information. We shouldnt need to unwrap/decompress the chunk to get the lastTime.

query/src/test/scala/filodb/query/exec/SlidingWindowIteratorSpec.scala
velvia: Right, but the iterator for dropping out of order samples is disabled now, so the test doesn't make sense anymore.  Right?

Unless one were to redo all the query tests to be actually based on raw TimeSeriesPartitions, which is probably a good idea IMO.

core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
velvia: lastChunkInfo is potentially expensive is the problem.  Locks are involved and it has to consult the chunkMap.  Storing the lastTime, yes it uses an extra long but is much simpler and faster.  I can document this.

velvia: There are a couple of potential solutions, but they are all expensive.  We can retrieve the last ChunkSetInfo instead of the entire chunk, but even this is likely to add say 10-15 min to recovery.   We can also add a separate table which organizes the last time more efficiently.

However, if we are going to move to a more rigorous out of order solution, this might be moot.  So, perhaps we should just add a note and TODO that we have this edge case, and merge this to see how big of a problem the out of order is in the first place.

query/src/test/scala/filodb/query/exec/SlidingWindowIteratorSpec.scala
velvia: Reverted.

velvia:
@vishramachandran I reverted the disabling of the DropOutOfOrderIterator and added comments

core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
jfz: How about save the "lastTime" into CheckPoint? Should be easy to get that during ChunkSink time. It's at the group level rather than part level, but assumption is data comes in order anyways.

vishramachandran:
@velvia Were you able to measure the effects on ingestion benchmark if you change the lastTime to lastChunkInfo.endTime ? This may save some heap memory, but curious to know how it affects the throughput - small or big difference.

vishramachandran:
@velvia There is another possibility of out-of-order samples. If there is a failure between write of chunks and checkpoint to c*, then the entire chunk's data may be re-read from Kafka and this can cause the chunk to be re-written.

core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
velvia: The above would only work if most of the lastTimes in each TSPartition being flushed were about the same.  What if there is one with lagging timestamps?

velvia:
@vishramachandran so you've made a good point, that the checkpoint and the "last time" data basically need to be atomically written.  The checkpoint determines when the data is recovered and thus what data gets ingested again, they need to be in sync.

The problem is that dropping out of order stuff is inherently stateful.  :(

Also, as to how changing to lastChunkInfo.endTime affects ingestion performance.  I've finally fixed the ingestion benchmark so it is stable.   Current (keeping separate lastTime var):

```
[info] Benchmark                                Mode  Cnt        Score       Error  Units
[info] IngestionBenchmark.ingestEncodeRecords  thrpt   30  1977453.986 ± 52463.183  ops/s
```

It turns out that I cannot accurately measure the effect of changing to use `timestampOfLatestSample`.  The reason is that being able to measure an ingestion benchmark itself is not easy and requires resetting the lastTime state to be able to continuously ingest new data, otherwise we keep ingesting the same samples and they get dropped, thus you get an artificially higher ingestion rate (without the cost of encoding etc.)

core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
jfz: @velvia Timestamp to be check pointed should be the "lastTime" of all TSPartitions in that group for that time bucket. So "one with lagging timestamps" means that "lastTime of that TSParition" is <= "lastTime of the group time bucket", which is fine.
E.g. timeBucket cut time is 10:00, and lastTime of the timeBucket is also 10:00, and lastTime of TSPartition X is 9:30, which means there's no data for TSPartition X between 9:30 - 10:00. So when it starts ingestion(or recover) for newer data, we can safely assume that new data of TSPartition X should be >= 10:00.

vishramachandran:
@velvia The fix for the benchmark would be to not ingest the same data again. Also, resetLastTime is included in the benchmark. We could to prepare the containers to ingest outside of the benchmark method. You could try the [Setup JMH ](http://tutorials.jenkov.com/java-performance/jmh.html#state-setup-and-teardown) constructs for this.

Approving provided a reasonable attempt is made to achieve this.

velvia:
@vishramachandran Fixed benchmark to not ingest same data, but the downside is that it must use SingleShotMode, so it reports time instead of throughput, user must manually figure out the throughput.  

I tested using `lastSampleTime` instead of a separate lastTIme variable.  Initially it slowed down ingestion 2x but I optimized it so it was nearly invisible, and so went with that.  Thus now we are able to drop the lastTime variable.  

Will merge when the test passes.

PR #190, opened 2018-12-13
feat(query): Add support and tests for querying from LongColumns

velvia:
For now we just transform the Long values to Doubles.  Future Chunk-based iterations will take advantage of faster methods.

**Pull Request checklist**

One can query raw data from non-Double columns but cannot run any non-raw queries.

**New behavior :**

Non-raw queries from Long columns are now possible.

### DIFF
--- a/query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
+++ b/query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
@@ -3,11 +3,19 @@ package filodb.query.exec
 import monix.reactive.Observable
 import org.jctools.queues.SpscUnboundedArrayQueue
 
-import filodb.core.query.{IteratorBackedRangeVector, RangeVector, ResultSchema}
+import filodb.core.metadata.Column.ColumnType
+import filodb.core.metadata.Dataset
+import filodb.core.query.{ColumnInfo, IteratorBackedRangeVector, RangeVector, ResultSchema}
 import filodb.memory.format.RowReader
 import filodb.query.{Query, QueryConfig, RangeFunctionId}
 import filodb.query.exec.rangefn.{RangeFunction, Window}
 import filodb.query.util.IndexedArrayQueue
+
+object PeriodicSamplesMapper {
+  // A PeriodicSamplesMapper always outputs timestamp:Long/value:Double right now
+  val periodicSchema = ResultSchema(Seq(ColumnInfo("timestamp", ColumnType.TimestampColumn),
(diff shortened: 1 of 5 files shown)


query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
```diff
@@ -35,13 +43,25 @@ final case class PeriodicSamplesMapper(start: Long,
...
+      case ColumnType.DoubleColumn =>
+        source.map { rv =>
+          IteratorBackedRangeVector(rv.key,
+            new SlidingWindowIterator(rv.rows, start, step, end, window.getOrElse(0L),
+              RangeFunction(functionId, funcParams), queryConfig))
+        }
+      case ColumnType.LongColumn =>
+        source.map { rv =>
+          IteratorBackedRangeVector(rv.key,
+            new SlidingWindowIterator(new LongToDoubleIterator(rv.rows), start, step, end, window.getOrElse(0L),
```
broneill: It seems the bulk of this change is to convert longs to doubles automatically. That might cause a surprise if someone wanted the extra precision. When would a column type be defined as a long if only to convert it to double later?

velvia: Most of the range functions are going to be converted to type specific aggregations, but this will take time.  I agree the current state of this change doesn't offer any precision advantages, but it does unblock until over time we can add in the necessary query functionality.

query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
```diff
@@ -3,12 +3,20 @@ package filodb.query.exec
...
+import filodb.core.query.{ColumnInfo, IteratorBackedRangeVector, RangeVector, ResultSchema}
 import filodb.memory.format.RowReader
 import filodb.query.{Query, QueryConfig, RangeFunctionId}
 import filodb.query.exec.rangefn.{RangeFunction, Window}
 import filodb.query.util.IndexedArrayQueue
 
+object PeriodicSamplesMapper {
+  // A PeriodicSamplesMapper always outputs timestamp:Long/value:Double right now
+  val periodicSchema = ResultSchema(Seq(ColumnInfo("timestamp", ColumnType.TimestampColumn),
+                                        ColumnInfo("value", ColumnType.DoubleColumn)), 0)
```
vishramachandran: @velvia @tjackpaul  Number of row-key columns should be 1.
This is probably the reason for regression.

PR #192, opened 2018-12-14
feat(memory, core, query): ChunkedWindowIterator for up to 7.5x faster queries

velvia:
**Pull Request checklist**

Currently almost all queries (except raw ones) use the `SlidingWindowIterator` for windowing operations.  It goes through every "row" in each Time series, spanning the first row of the earliest time window to the last row of the last time window, regardless of the step size or window length.  This works but does not take advantage of various optimizations that are possible with columnar layout.

**New behavior :**

This PR introduces a couple new classes/concepts:
- A `ChunkedWindowIterator` which gives the parameters of each window plus each applicable chunkSet within that window, instead of a sliding window row-based abstraction, directly to a 
- `ChunkedRangeFunction` which calculates and updates its results one chunk * window at a time

### Improvements in Query Speed

The following functions have `ChunkedRangeFunctions` implemented:
- sum_over_time
- count_over_time - optimized to just use binary search on timestamp vectors, no need to look at the value vector at all
- avg_over_time
- stdvar_over_time
- dev_over_time
- LastSampleFunction
- min_over_time
- max_over_time

Using the `QueryInMemoryBenchmark` specifically on `sum_over_time`, we can see the following improvements depending on window size and step, varying from about 10% to almost 5x.  Other improvements:
- LastSample: 7.5x
- min_over_time: 6x

Step(s) | Window(s) | Overlap% | SlidingWindow (Row) | ChunkedWindow
-- | -- | -- | -- | --
150 | 300 | 50% | 1000 QPS | 4698 QPS
300 | 300 | 0% |   | 4880 QPS
100 | 300 | 66% |   |  
60 | 300 | 80% | 939 QPS | 2864 QPS
60 | 480 | 88% |   |  
10 | 300 | 97% | 1169 QPS | 1228 QPS
60 | 60 | 0% | 924 QPS | 3223 QPS
60 | 120 | 50% |   | 3227 QPS

In addition, optimizations which benefit even the `SlidingWindowIterator` are included:
- Moving vars from the body of each class to the constructors of inner iterators in various BinaryVectors, as well as the `FilteredChunkInfoIterator`.... saves about 10% 
- Some improvements to SerializableRangeVector
- A fix for the binarySearch functionality used in `PartitionTimeRangeReader` and new chunked iterators

### DIFF
--- a/core/src/main/scala/filodb.core/binaryrecord2/RecordContainer.scala
+++ b/core/src/main/scala/filodb.core/binaryrecord2/RecordContainer.scala
@@ -22,7 +22,8 @@ import filodb.memory.format.{RowReader, UnsafeUtils}
  * @param offset 64-bit native pointer or offset into Array[Byte] object memory
  * @param maxLength the maximum allocated size the container can grow to, including the initial length bytes
  */
-final class RecordContainer(val base: Any, val offset: Long, maxLength: Int) {
+final class RecordContainer(val base: Any, val offset: Long, maxLength: Int,
+                            var numRecords: Int = 0) {
   import RecordBuilder._
 
   @inline final def numBytes: Int =  UnsafeUtils.getInt(base, offset)
--- a/core/src/main/scala/filodb.core/binaryrecord2/RecordContainer.scala
+++ b/core/src/main/scala/filodb.core/binaryrecord2/RecordContainer.scala
@@ -23,6 +23,7 @@ import filodb.memory.format.{RowReader, UnsafeUtils}
  * @param maxLength the maximum allocated size the container can grow to, including the initial length bytes
  */
 final class RecordContainer(val base: Any, val offset: Long, maxLength: Int,
+                            // INTERNAL variable placed here to avoid Scalac using a bitfield and slowing things down
                             var numRecords: Int = 0) {
   import RecordBuilder._
 
--- a/core/src/main/scala/filodb.core/binaryrecord2/RecordContainer.scala
+++ b/core/src/main/scala/filodb.core/binaryrecord2/RecordContainer.scala
@@ -102,7 +102,7 @@ final class RecordContainer(val base: Any, val offset: Long, maxLength: Int) {
   final def allOffsets: Buffer[Long] = map { case (b, o) => o }
 
   class CountingConsumer(var count: Int = 0) extends BinaryRegionConsumer {
-    def onNext(base: Any, offset: Long): Unit = count += 1
+    final def onNext(base: Any, offset: Long): Unit = count += 1
   }
 
   /**
--- a/core/src/main/scala/filodb.core/query/RangeVector.scala
+++ b/core/src/main/scala/filodb.core/query/RangeVector.scala
@@ -170,14 +170,15 @@ object SerializableRangeVector extends StrictLogging {
             limit: Int): SerializableRangeVector = {
     var numRows = 0
     val oldContainerOpt = builder.currentContainer
-    val startRecordNo = oldContainerOpt.map(_.countRecords).getOrElse(0)
+    val startRecordNo = oldContainerOpt.map(_.numRecords).getOrElse(0)
     // Important TODO / TechDebt: We need to replace Iterators with cursors to better control
     // the chunk iteration, lock acquisition and release. This is much needed for safe memory access.
     try {
       OffheapLFSortedIDMap.validateNoSharedLocks()
-      rv.rows.take(limit).foreach { row =>
+      val rows = rv.rows
+      while (rows.hasNext && numRows < limit) {
         numRows += 1
-        builder.addFromReader(row)
+        builder.addFromReader(rows.next)
       }
     } finally {
       // When the query is done, clean up lingering shared locks caused by iterator limit.
--- a/core/src/main/scala/filodb.core/query/RangeVector.scala
+++ b/core/src/main/scala/filodb.core/query/RangeVector.scala
@@ -8,7 +8,7 @@ import filodb.core.binaryrecord.BinaryRecord
 import filodb.core.binaryrecord2.{MapItemConsumer, RecordBuilder, RecordContainer, RecordSchema}
 import filodb.core.metadata.Column
 import filodb.core.metadata.Column.ColumnType._
-import filodb.core.store.{ChunkInfoRowReader, ChunkScanMethod, ReadablePartition}
+import filodb.core.store._
 import filodb.memory.{MemFactory, UTF8StringMedium}
 import filodb.memory.data.OffheapLFSortedIDMap
 import filodb.memory.format.{RowReader, ZeroCopyUTF8String => UTF8Str}
@@ -104,12 +104,19 @@ trait RangeVector {
   }
 }
 
+// First column of columnIDs should be the timestamp column
 final case class RawDataRangeVector(key: RangeVectorKey,
                                     partition: ReadablePartition,
                                     chunkMethod: ChunkScanMethod,
                                     columnIDs: Array[Int]) extends RangeVector {
   // Iterators are stateful, for correct reuse make this a def
   def rows: Iterator[RowReader] = partition.timeRangeRows(chunkMethod, columnIDs)
+
+  // Obtain ChunkSetInfos from specific window of time from partition
+  def chunkInfos(windowStart: Long, windowEnd: Long): ChunkInfoIterator = partition.infos(windowStart, windowEnd)
+
+  def timestampColID: Int = partition.dataset.timestampColID
+  def valueColID: Int = columnIDs(1)
 }
 
 /**
--- a/core/src/main/scala/filodb.core/query/RangeVector.scala
+++ b/core/src/main/scala/filodb.core/query/RangeVector.scala
@@ -110,6 +110,7 @@ final case class RawDataRangeVector(key: RangeVectorKey,
   def chunkInfos(windowStart: Long, windowEnd: Long): ChunkInfoIterator = partition.infos(windowStart, windowEnd)
 
   def timestampColID: Int = partition.dataset.timestampColID
+  // the query engine is based around one main data column to query, so it will always be the second column passed in
   def valueColID: Int = columnIDs(1)
 }
 
--- a/core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
+++ b/core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
@@ -272,10 +272,11 @@ class ElementChunkInfoIterator(elIt: ElementIterator) extends ChunkInfoIterator 
   def nextInfo: ChunkSetInfo = ChunkSetInfo(elIt.next)
 }
 
-class FilteredChunkInfoIterator(base: ChunkInfoIterator, filter: ChunkSetInfo => Boolean) extends ChunkInfoIterator {
-  var nextnext: ChunkSetInfo = ChunkSetInfo(0)
-  var gotNext: Boolean = false
-
+class FilteredChunkInfoIterator(base: ChunkInfoIterator,
+                                filter: ChunkSetInfo => Boolean,
+                                // Move vars here for better performance -- no init field
+                                var nextnext: ChunkSetInfo = ChunkSetInfo(0),
+                                var gotNext: Boolean = false) extends ChunkInfoIterator {
   def close(): Unit = {
     base.close()
   }
--- a/core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
+++ b/core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
@@ -4,6 +4,7 @@ import java.nio.ByteBuffer
 
 import com.googlecode.javaewah.EWAHCompressedBitmap
 import com.typesafe.scalalogging.StrictLogging
+import debox.Buffer
 
 import filodb.core.Types._
 import filodb.core.metadata.{Column, Dataset}
@@ -297,6 +298,65 @@ class FilteredChunkInfoIterator(base: ChunkInfoIterator,
   def nextInfo: ChunkSetInfo = {
     gotNext = false   // reset so we can look for the next item where filter == true
     nextnext
+  }
+}
+
+/**
+ * A sliding window based iterator over the chunks needed to be read from for each window.
+ * Assumes the ChunkInfos are in increasing time order.
+ * The sliding window goes from (start-window, start) -> (end-window, end) in step increments, and for
+ * each window, this class may be used as a ChunkInfoIterator.
+ *
+ * @param infos the base ChunkInfoIterator to perform windowing over
+ * @param start the starting window end timestamp, must have same units as the ChunkSetInfo start/end times
+ * @param step the increment the window goes forward by
+ * @param end the ending window end timestamp.  If it does not line up on multiple of start + n * step, then
+ *            the windows will slide until the window end is beyond end.
+ * @param window the # of millis/time units that define the length of each window
+ */
+class WindowedChunkIterator(infos: ChunkInfoIterator, start: Long, step: Long, end: Long, window: Long,
+                            // internal vars, put it here for better performance
+                            var curWindowEnd: Long = -1L,
+                            var curWindowStart: Long = -1L,
+                            private var readIndex: Int = 0,
+                            windowInfos: Buffer[NativePointer] = Buffer.empty[NativePointer])
+extends ChunkInfoIterator {
+  final def close(): Unit = infos.close()
+  final def hasMoreWindows: Boolean = curWindowEnd < end
+  final def nextWindow(): Unit = {
+    // advance window pointers and reset read index
+    if (curWindowStart == -1L) {
+      curWindowEnd = start
+      curWindowStart = start - window
+    } else {
+      curWindowStart += step
+      curWindowEnd += step
+    }
+    readIndex = 0
+
+    // drop initial chunksets of window that are no longer part of the window
+    while (windowInfos.nonEmpty && ChunkSetInfo(windowInfos(0)).endTime < curWindowStart) {
+      windowInfos.remove(0)
+    }
+
+    var lastEndTime = if (windowInfos.isEmpty) -1L else ChunkSetInfo(windowInfos(windowInfos.length - 1)).endTime
+
+    // if new window end is beyond end of most recent chunkset, add more chunksets (if there are more)
+    while (curWindowEnd > lastEndTime && infos.hasNext) {
+      val next = infos.nextInfo
+      // Add if next chunkset is within window.  Otherwise keep going
+      if (curWindowStart <= next.endTime) {
+        windowInfos += next.infoAddr
+        lastEndTime = Math.max(next.endTime, lastEndTime)
+      }
+    }
+  }
+
+  def hasNext: Boolean = readIndex < windowInfos.length
+  def nextInfo: ChunkSetInfo = {
+    val next = windowInfos(readIndex)
+    readIndex += 1
+    ChunkSetInfo(next)
   }
 }
 
--- a/core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
+++ b/core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
@@ -270,8 +270,8 @@ object ChunkInfoIterator {
 
 class ElementChunkInfoIterator(elIt: ElementIterator) extends ChunkInfoIterator {
   def close(): Unit = elIt.close()
-  def hasNext: Boolean = elIt.hasNext
-  def nextInfo: ChunkSetInfo = ChunkSetInfo(elIt.next)
+  final def hasNext: Boolean = elIt.hasNext
+  final def nextInfo: ChunkSetInfo = ChunkSetInfo(elIt.next)
 }
 
 class FilteredChunkInfoIterator(base: ChunkInfoIterator,
@@ -283,7 +283,7 @@ class FilteredChunkInfoIterator(base: ChunkInfoIterator,
     base.close()
   }
 
-  def hasNext: Boolean = {
+  final def hasNext: Boolean = {
     try {
       while (base.hasNext && !gotNext) {
         nextnext = base.nextInfo
@@ -295,7 +295,7 @@ class FilteredChunkInfoIterator(base: ChunkInfoIterator,
     }
   }
 
-  def nextInfo: ChunkSetInfo = {
+  final def nextInfo: ChunkSetInfo = {
     gotNext = false   // reset so we can look for the next item where filter == true
     nextnext
   }
@@ -352,8 +352,8 @@ extends ChunkInfoIterator {
     }
   }
 
-  def hasNext: Boolean = readIndex < windowInfos.length
-  def nextInfo: ChunkSetInfo = {
+  final def hasNext: Boolean = readIndex < windowInfos.length
+  final def nextInfo: ChunkSetInfo = {
     val next = windowInfos(readIndex)
     readIndex += 1
     ChunkSetInfo(next)
--- a/core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
+++ b/core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
@@ -304,9 +304,9 @@ class FilteredChunkInfoIterator(base: ChunkInfoIterator,
 /**
  * A sliding window based iterator over the chunks needed to be read from for each window.
  * Assumes the ChunkInfos are in increasing time order.
- * The sliding window goes from (start-window, start) -> (end-window, end) in step increments, and for
- * each window, this class may be used as a ChunkInfoIterator.
- *
+ * The sliding window goes from (start-window, start] -> (end-window, end] in step increments, and for
+ * each window, this class may be used as a ChunkInfoIterator. Excludes start, includes end.
+
  * @param infos the base ChunkInfoIterator to perform windowing over
  * @param start the starting window end timestamp, must have same units as the ChunkSetInfo start/end times
  * @param step the increment the window goes forward by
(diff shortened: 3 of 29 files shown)


query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
```diff
@@ -59,15 +71,34 @@ class CountOverTimeFunction extends RangeFunction {
...
+  def addChunks(tsCol: Int, valueCol: Int, info: ChunkSetInfo, startTime: Long, endTime: Long): Unit = {
+    val timestampVector = info.vectorPtr(tsCol)
+    val tsReader = bv.LongBinaryVector(timestampVector)
+
+    // First row >= startTime, so we can just drop bit 31 (dont care if it matches exactly)
+    val startRowNum = tsReader.binarySearch(timestampVector, startTime) & 0x7fffffff
+    val endRowNum = tsReader.binarySearch(timestampVector, endTime) match {
+      // if endTime does not match, we want last row such that timestamp < endTime
+      // Note if we go past end of timestamps, it will never match, so this should make sure we don't go too far
+      case row if row < 0 => (row & 0x7fffffff) - 1
```
broneill: If row is 0x80000000, then this yields -1.

velvia: Yes, exactly.  That's the intended result, it indicates there is no such row satisfying the ceiling time constraint.

memory/src/main/scala/filodb.memory/format/vectors/DoubleVector.scala
```diff
@@ -144,15 +152,38 @@ trait DoubleVectorDataReader extends VectorDataReader {
...
+    var sum: Double = 0d
+    // Fetch and add multiple points at once for efficiency
+    while ((addr + 64) <= untilAddr) {
+      sum += UnsafeUtils.getDouble(addr) + UnsafeUtils.getDouble(addr + 8) +
+             UnsafeUtils.getDouble(addr + 16) + UnsafeUtils.getDouble(addr + 24) +
+             UnsafeUtils.getDouble(addr + 32) + UnsafeUtils.getDouble(addr + 40) +
+             UnsafeUtils.getDouble(addr + 48) + UnsafeUtils.getDouble(addr + 56)
+      addr += 64
+    }
+    while (addr < untilAddr) {
```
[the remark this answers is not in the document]
velvia: I might do this differently.  It turns out to simulate SIMD you really want to parallelize the adds, not add them serially.  Really doubt the compiler will know to do that.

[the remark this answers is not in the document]
broneill: I suspect that if you looked at the overall cost of the query, this loop is not likely to be the bottleneck. Adding numbers is super cheap, and the compiler knows how to unroll loop. The manual unrolling introduced a potential bug, as mentioned in the later comment.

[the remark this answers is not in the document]
velvia: BTW I did test the unrolled performance vs the normal while loop performance.  The unrolled performance is slightly faster.  Worth it?   Not sure.

query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
```diff
@@ -68,32 +71,105 @@ trait RangeFunction {
...
+  def addChunks(tsCol: Int, valueCol: Int, info: ChunkSetInfo,
+                startTime: Long, endTime: Long, queryConfig: QueryConfig): Unit = {
+    val timestampVector = info.vectorPtr(tsCol)
+    val tsReader = bv.LongBinaryVector(timestampVector)
+    val doubleVector = info.vectorPtr(valueCol)
+    val dblReader = bv.DoubleVector(doubleVector)
+
+    // First row >= startTime, so we can just drop bit 31 (dont care if it matches exactly)
+    val startRowNum = tsReader.binarySearch(timestampVector, startTime) & 0x7fffffff
+    val endRowNum = tsReader.ceilingIndex(timestampVector, endTime)
```
vishramachandran: I see that the above few lines will be a code pattern used in multiple places - for example, in downsampling. Would be nice to abstract this kind of a search further. Please see if possible. 

This is where cursors would really be beneficial.

core/src/main/scala/filodb.core/binaryrecord2/RecordContainer.scala
```diff
@@ -22,7 +22,8 @@ import filodb.memory.format.{RowReader, UnsafeUtils}
  * @param offset 64-bit native pointer or offset into Array[Byte] object memory
  * @param maxLength the maximum allocated size the container can grow to, including the initial length bytes
  */
-final class RecordContainer(val base: Any, val offset: Long, maxLength: Int) {
+final class RecordContainer(val base: Any, val offset: Long, maxLength: Int,
+                            var numRecords: Int = 0) {
```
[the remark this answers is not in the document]
velvia: Yes, will document this

core/src/main/scala/filodb.core/query/RangeVector.scala
```diff
@@ -104,12 +104,19 @@ trait RangeVector {
...
                                     chunkMethod: ChunkScanMethod,
                                     columnIDs: Array[Int]) extends RangeVector {
   // Iterators are stateful, for correct reuse make this a def
   def rows: Iterator[RowReader] = partition.timeRangeRows(chunkMethod, columnIDs)
+
+  // Obtain ChunkSetInfos from specific window of time from partition
+  def chunkInfos(windowStart: Long, windowEnd: Long): ChunkInfoIterator = partition.infos(windowStart, windowEnd)
+
+  def timestampColID: Int = partition.dataset.timestampColID
+  def valueColID: Int = columnIDs(1)
```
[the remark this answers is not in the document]
velvia: Yes, because the query engine is based around one main data column to query, so it will always be the second column passed in.  Will document this.

core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
```diff
@@ -292,12 +295,71 @@ class FilteredChunkInfoIterator(base: ChunkInfoIterator, filter: ChunkSetInfo =>
...
+      // Add if next chunkset is within window.  Otherwise keep going
+      if (curWindowStart <= next.endTime) {
+        windowInfos += next.infoAddr
+        lastEndTime = Math.max(next.endTime, lastEndTime)
+      }
+    }
+  }
+
+  final def hasNext: Boolean = readIndex < windowInfos.length
+  final def nextInfo: ChunkSetInfo = {
```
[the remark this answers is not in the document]
velvia: Done

jmh/src/main/scala/filodb.jmh/QueryInMemoryBenchmark.scala
```diff
@@ -133,4 +144,50 @@ class QueryInMemoryBenchmark extends StrictLogging {
     }
     Await.result(Future.sequence(futures), 60.seconds)
   }
+
+  // Now, simulate where step is larger than the window size.
```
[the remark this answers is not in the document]
velvia: Yeah, comment was not updated.  Will remove it.

query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
```diff
@@ -31,19 +32,69 @@ final case class PeriodicSamplesMapper(start: Long,
...
             sourceSchema: ResultSchema): Observable[RangeVector] = {
+    // Really, use the stale lookback window size, not 0 which doesn't make sense
     RangeVectorTransformer.requireTimeSeries(sourceSchema)
     source.map { rv =>
-      IteratorBackedRangeVector(rv.key,
-        new SlidingWindowIterator(rv.rows, start, step, end, window.getOrElse(0L),
-          RangeFunction(functionId, funcParams), queryConfig))
+      // TODO: move this out of the inner loop somehow
+      RangeFunction(functionId, funcParams, useChunked = true) match {
+        case c: ChunkedRangeFunction if rv.isInstanceOf[RawDataRangeVector] =>
```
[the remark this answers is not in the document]
velvia: Maybe.  Let me think about it.  I guess periodicSamplesMapper is only instantiated on top of raw vectors.   Doing the if gives us the option though to pass a non-raw RV in, with the restriction that chunked operations cannot be used.

memory/src/main/scala/filodb.memory/format/vectors/DoubleVector.scala
```diff
@@ -121,6 +121,14 @@ trait DoubleVectorDataReader extends VectorDataReader {
    */
   def iterate(vector: BinaryVectorPtr, startElement: Int = 0): DoubleIterator
 
+  /**
+   * Sums up the Double values in the vector from position start to position end.
+   * @param vector the BinaryVectorPtr native address of the BinaryVector
+   * @param start the starting element # in the vector to sum, 0 == first element
+   * @param end the ending element # in the vector to sum, inclusive
+   */
+  def sum(vector: BinaryVectorPtr, start: Int, end: Int): Double
```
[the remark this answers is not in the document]
velvia: Yes.  I'll probably add min/max right now, as that completes the set needed for downsampling of gauges.   If I'm really ambitious I'll play with counter correction too as I want to see what the potential improvement there is.

query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
```diff
@@ -68,32 +71,105 @@ trait RangeFunction {
...
+    case None                 => new LastSampleChunkedFunction()
+    case Some(CountOverTime)  => new CountOverTimeChunkedFunction()
+    case Some(SumOverTime)    => new SumOverTimeChunkedFunction()
+    case Some(AvgOverTime)    => new AvgOverTimeChunkedFunctionD()
+    case Some(StdDevOverTime) => new StdDevOverTimeChunkedFunctionD()
+    case Some(StdVarOverTime) => new StdVarOverTimeChunkedFunctionD()
+    case _                    => iterator(func, funcParams)
+  } else iterator(func, funcParams)
+
+  def iterator(func: Option[RangeFunctionId],
```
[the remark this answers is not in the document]
velvia: ok.

query/src/test/scala/filodb/query/exec/SlidingWindowIteratorSpec.scala
```diff
@@ -161,7 +161,7 @@ class SlidingWindowIteratorSpec extends FunSpec with Matchers {
     val end = 1000L
     val step = 5
     val slidingWinIterator = new SlidingWindowIterator(rawRows.iterator, start, step,
-      end,0, RangeFunction(None), queryConfig)
+      end, 0, RangeFunction(None, useChunked = false), queryConfig)
```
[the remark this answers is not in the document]
velvia: What is your advice here?  We can rename the SlidingWindowIteratorSpec into more of a PeriodicSampleMapperSpec, or keep it strictly for SlidingWindows and have another test for chunked iterator.  I'd probably prefer the former.

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
```diff
@@ -85,12 +111,25 @@ class AvgOverTimeFunction extends RangeFunction {
   }
 }
 
-class StdDevOverTimeFunction extends RangeFunction {
+class AvgOverTimeChunkedFunctionD(var sum: Double = 0d, var count: Int = 0) extends ChunkedDoubleRangeFunction {
```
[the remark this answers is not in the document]
velvia: D = Double.  Am going to add versions of these functions for longs.

jmh/src/main/scala/filodb.jmh/QueryInMemoryBenchmark.scala
```diff
@@ -98,14 +102,20 @@ class QueryInMemoryBenchmark extends StrictLogging {
...
    * They are designed to match all the time series (common case) under a particular metric and job
    */
-  val queries = Seq("heap_usage{app=\"App-2\"}",  // raw time series
+  val rawQuery = "heap_usage{app=\"App-2\"}"
+  val sumQuery = """sum_over_time(heap_usage{app="App-2"}[5m])"""
+  val queries = Seq(rawQuery,  // raw time series
                     """sum(rate(heap_usage{app="App-2"}[5m]))""",
                     """quantile(0.75, heap_usage{app="App-2"})""",
-                    """sum_over_time(heap_usage{app="App-2"}[5m])""")
+                    sumQuery)
```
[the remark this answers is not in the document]
velvia: Yeah.  I'd like to add benchmark methods for other things like min/max.

query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
velvia: I agree cursors are the right way to go.  I think it's better to leave the abstraction and effort to cursors as that will abstract all those things.

memory/src/main/scala/filodb.memory/format/vectors/DoubleVector.scala
```diff
@@ -144,15 +152,30 @@ trait DoubleVectorDataReader extends VectorDataReader {
...
       data
     }
   }
+
+  final def apply(vector: BinaryVectorPtr, n: Int): Double = UnsafeUtils.getDouble(vector + OffsetData + n * 8)
+  def iterate(vector: BinaryVectorPtr, startElement: Int = 0): DoubleIterator =
+    new Double64Iterator(vector + OffsetData + startElement * 8)
+
+  // end is inclusive
+  final def sum(vector: BinaryVectorPtr, start: Int, end: Int): Double = {
```
vishramachandran: One important concern. How is NaN handled ?
NaNs can occur for end-of-time-series markers. 
Can you please see how the performance is impacted because of any additional NaN check ?

memory/src/main/scala/filodb.memory/format/vectors/DoubleVector.scala
```diff
@@ -271,11 +297,15 @@ extends AppendableVectorWrapper[Long, Double] {
  * A wrapper to return Doubles from a Long vector... for when one can compress Double vectors as DDVs
  */
 object DoubleLongWrapDataReader extends DoubleVectorDataReader {
+  class DoubleLongWrapIterator(innerIt: LongIterator) extends DoubleIterator {
```
vishramachandran: How are NaNs handled in the Double data stored as Long vector ?

memory/src/main/scala/filodb.memory/format/vectors/IntBinaryVector.scala
```diff
@@ -274,34 +296,138 @@ trait IntVectorDataReader extends VectorDataReader {
...
+    sum
+  }
 }
 
 object OffheapUnsignedIntVector16 extends IntVectorDataReader {
   final def apply(vector: BinaryVectorPtr, n: Int): Int =
     (UnsafeUtils.getShort(vector + 8 + n * 2) & 0x0ffff).toInt
+  final def sum(vector: BinaryVectorPtr, start: Int, end: Int): Long = {
+    val startRoundedUp = (start + 3) & ~3
+    var sum = defaultSum(vector, start, Math.min(end, startRoundedUp - 1))
```
vishramachandran: How will end-of-time-series marker be done for a Long column type data ? With Double valued columns, NaN is used currently.

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
```diff
@@ -45,9 +44,32 @@ class SumOverTimeFunction extends RangeFunction {
...
+    sum += doubleReader.sum(doubleVect, startRowNum, endRowNum)
+  }
+}
+
+class SumOverTimeChunkedFunctionL extends SumOverTimeChunkedFunction() with ChunkedLongRangeFunction {
+  final def addTimeLongChunks(longVect: BinaryVector.BinaryVectorPtr,
+                              longReader: bv.LongVectorDataReader,
+                              startRowNum: Int,
+                              endRowNum: Int): Unit = {
+    sum += longReader.sum(longVect, startRowNum, endRowNum)
```
vishramachandran: Because of the absence of NA in our vectors, we use Double.NaN to identify no value in our result vectors today. Without this there would be some important holes in the query processing because we cannot identify "no value".

Since long values do not have a NaN, we need a replacement for "no-value" representation.

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
```diff
@@ -59,15 +81,29 @@ class CountOverTimeFunction extends RangeFunction {
...
+                startTime: Long, endTime: Long, queryConfig: QueryConfig): Unit = {
+    val timestampVector = info.vectorPtr(tsCol)
+    val tsReader = bv.LongBinaryVector(timestampVector)
 
-  var sum = 0d
-  var count = 0
+    // First row >= startTime, so we can just drop bit 31 (dont care if it matches exactly)
+    val startRowNum = tsReader.binarySearch(timestampVector, startTime) & 0x7fffffff
+    val endRowNum = tsReader.ceilingIndex(timestampVector, endTime)
+    val numRows = endRowNum - startRowNum + 1
```
vishramachandran: This short-cut will extra count the NaN values. Bummer. :(

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
```diff
@@ -85,12 +121,36 @@ class AvgOverTimeFunction extends RangeFunction {
...
+  }
+}
+
+class AvgOverTimeChunkedFunctionD extends AvgOverTimeChunkedFunction() with ChunkedDoubleRangeFunction {
+  final def addTimeDoubleChunks(doubleVect: BinaryVector.BinaryVectorPtr,
+                                doubleReader: bv.DoubleVectorDataReader,
+                                startRowNum: Int,
+                                endRowNum: Int): Unit = {
+    sum += doubleReader.sum(doubleVect, startRowNum, endRowNum)
+    count += (endRowNum - startRowNum + 1)
```
vishramachandran: NaN needs to be handled for both sum and count

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
```diff
@@ -85,12 +121,36 @@ class AvgOverTimeFunction extends RangeFunction {
...
-  var sum = 0d
-  var count = 0
-  var squaredSum = 0d
+class AvgOverTimeChunkedFunctionL extends AvgOverTimeChunkedFunction() with ChunkedLongRangeFunction {
+  final def addTimeLongChunks(longVect: BinaryVector.BinaryVectorPtr,
+                              longReader: bv.LongVectorDataReader,
+                              startRowNum: Int,
+                              endRowNum: Int): Unit = {
+    sum += longReader.sum(longVect, startRowNum, endRowNum)
+    count += (endRowNum - startRowNum + 1)
```
vishramachandran: End-Of-Time-Series marker needs to be allowed and handled for both sum and count

query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
```diff
@@ -111,3 +264,52 @@ object LastSampleFunction extends RangeFunction {
...
+    val timestampVector = info.vectorPtr(tsCol)
+    val tsReader = bv.LongBinaryVector(timestampVector)
+    val endRowNum = tsReader.ceilingIndex(timestampVector, endTime)
+
+    // update timestamp only if
+    //   1) endRowNum >= 0 (timestamp within chunk)
+    //   2) timestamp is within stale window; AND
+    //   3) timestamp is greater than current timestamp (for multiple chunk scenarios)
+    if (endRowNum >= 0) {
+      val ts = tsReader(timestampVector, endRowNum)
```
vishramachandran: This line may return timestamp of a end-of-timeseries marker, so NaN may be returned. The actual sample returned needs to be the one prior to this value.

memory/src/main/scala/filodb.memory/format/vectors/DoubleVector.scala
velvia: Sure thing.

memory/src/main/scala/filodb.memory/format/vectors/DoubleVector.scala
velvia: NaNs are not integral so cannot take advantage of Long based DDV compression.   Today any vectors with NaNs will not be compressed at all.

memory/src/main/scala/filodb.memory/format/vectors/IntBinaryVector.scala
velvia: There are multiple ways we can handle end of time series marker, but the general solution is out of scope for this PR.  Since only Prometheus clients send NaN for end of time series marker, and only for doubles, let's implement a solution to skip/detect NaNs for DoubleColumns only.

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
velvia: See the comment above.  Note that NaN is NOT the same thing as "no value".  A client can legitimately send NaN in theory, even for Prometheus which uses a very specific NaN value (there are billions of double values that can be interpreted as NaN).    We have NA capability in our vectors actually, but it is not used.  End of time series should be treated as its own problem though and using NA is not the most optimal solution.

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
velvia: I'll have a special version for Double columns which filters out the last value if its a NaN.  Does that work?

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
velvia: Yes, will do.

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
velvia: Again, we will punt on it for Long type data as there is no way the end of time series marker is generated for long columns today (since Prom doesn't output Long type data)

query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
velvia: Ok, will check for that.

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
vishramachandran: My comment was mis-understood.  This has no relation to End-of-Time-Series markers. This is a totally different topic which is relevant to our implementation of query processing.

In our result RangeVectors today, NaN means no-value. For example, if there is no last sample within 5 minutes, we set the value in result vector to NaN - reason being that NA is not supported in our result vectors.

My question is how are we going to set NaN on a Long based result vector ?

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
vishramachandran: I dont think so. This is count-over-time, not last sample. The count (number of samples) in the window should exclude NaN in the count. (BTW, I am not sure we are doing this today and it may be a bug).

velvia:
@vishramachandran NaN / EOTS marker handling has been added for all chunked range functions for DoubleColumns.   Here are the before/after comparisons for QueryInMemoryBenchmark.

Before:
```
[info] QueryInMemoryBenchmark.noOverlapQueries     thrpt   20   4341.321 ± 354.499  ops/s
[info] QueryInMemoryBenchmark.singleThreadedQuery  thrpt   20  18654.377 ± 428.580  ops/s
[info] QueryInMemoryBenchmark.someOverlapQueries   thrpt   20   3701.860 ± 380.050  ops/s
```

After:
```
[info] QueryInMemoryBenchmark.noOverlapQueries     thrpt   20   4722.348 ± 426.260  ops/s
[info] QueryInMemoryBenchmark.singleThreadedQuery  thrpt   20  19568.348 ± 410.880  ops/s
[info] QueryInMemoryBenchmark.someOverlapQueries   thrpt   20   4278.261 ± 458.860  ops/s
```

No negative effects.  :)
BTW the old sliding iterator I'm not sure handles all NaN / EOTS use cases correctly.  I could not find any test cases for it, I added one.

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
velvia: Oh, I see.  This is handled correctly for LastSampleChunkedFunction even for Long's.  Basically the value returned is always NaN unless there is an actual value found.

However we should think about how to properly handle these for other types like Histograms.  I think using NaN is probably not a viable long term solution, but that solution is also out of scope for this PR.

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
velvia: I thought you said NaN was used to mark end of a time series.  When would you have NaN in the middle of a series?

PR #196, opened 2019-01-05
fix(query): fix the assumption that PartitionKey will have only one column of type MapColumn

sherali42:
**Pull Request checklist**

Fix the assumption that PartitionKey will have only one column of type MapColumn, while serving the Label queries. With the fix, TimeSeriesParition will be decoded based on the PartitionKeySchema

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
+++ b/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
@@ -489,9 +489,11 @@ class TimeSeriesShard(val dataset: Dataset,
           // FIXME This is non-performant and temporary fix for fetching label values based on filter criteria.
           // Other strategies needs to be evaluated for making this performant - create facets for predefined fields or
           // have a centralized service/store for serving metadata
-          currVal = dataset.partKeySchema.toStringPairs(nextPart.partKeyBase, nextPart.partKeyOffset)
-            .find(_._1 equals labelName).map(pair => ZeroCopyUTF8String(pair._2)).head
-          foundValue = true
+          dataset.partKeySchema.toStringPairs(nextPart.partKeyBase, nextPart.partKeyOffset)
+            .find(_._1 equals labelName).map(pair => {
+            currVal = ZeroCopyUTF8String(pair._2)
+            foundValue = true
+          })
         } else {
           // FIXME partKey is evicted. Get partition key from lucene index
         }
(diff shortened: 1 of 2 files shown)


core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -478,12 +478,11 @@ class TimeSeriesShard(val dataset: Dataset,
       val partId = partIterator.next()
```
vishramachandran: can we make result generation lazy by wrapping partIterator ? This will reduce memory requirements.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -465,30 +463,46 @@ class TimeSeriesShard(val dataset: Dataset,
...
+    * Iterator for lazy traversal of partIdIterator, value for the given label will be extracted from the ParitionKey.
+    */
+  case class IndexValueResultIterator(partIterator: IntIterator, labelName: String, limit: Int)
+      extends Iterator[ZeroCopyUTF8String] {
+    var currVal: ZeroCopyUTF8String = _
+    var index = 0
+
+    override def hasNext: Boolean = {
+      var foundValue = false
+      while(partIterator.hasNext && index < limit && !foundValue) {
```
velvia: While this is better, have we investigated what it takes to take advantage of Lucene facets?  If its not too much work might as well do this as well.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -465,30 +463,46 @@ class TimeSeriesShard(val dataset: Dataset,
...
+        val nextPart = getPartitionFromPartId(partId)
+        if (nextPart != UnsafeUtils.ZeroPointer) {
+          // FIXME This is non-performant and temporary fix for fetching label values based on filter criteria.
+          // Other strategies needs to be evaluated for making this performant - create facets for predefined fields or
+          // have a centralized service/store for serving metadata
+          currVal = dataset.partKeySchema.toStringPairs(nextPart.partKeyBase, nextPart.partKeyOffset)
+            .find(_._1 equals labelName).map(pair => ZeroCopyUTF8String(pair._2)).head
+          foundValue = true
+        } else {
+          // FIXME partKey is evicted. Get partition key from lucene index
```
vishramachandran: @sherali42 If we find an evicted partition, iteration will stop.

Also, you are calling `head` directly. Will that never cause a NoSuchElementException ?

You need to move forward until end of the parent iterator before you return false.

sherali42: @vishramachandran 👍.  fixed the code, which was looking at `head` irrespective of whether label exists or not

vishramachandran: Looks good. I suggest you use `foreach` instead of `map`. map is supposed to return a result.

PR #197, opened 2019-01-08
doc(all): Add compression and predictive nibblepacking doc/proposal

velvia:
**Pull Request checklist**

**New behavior :**

Documentation and proposal of an encoding scheme for compression of both floating point and other types of data such as histograms.


velvia:
@broneill sure maybe we'll move some docs into Wiki.  I think the only problem is wiki is harder to move if we ever have to relocate the repo.  Can think about it then.

PR #200, opened 2019-01-10
feat(core,memory): Don't free the native memory used by TimeSeriesPartition too soon.

broneill:
**Pull Request checklist**

**Current behavior :**
Removing partitions can cause SEGVs.

**New behavior :**
Native memory is freed only after garbage collection.

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/PartitionSet.scala
+++ b/core/src/main/scala/filodb.core/memstore/PartitionSet.scala
@@ -333,6 +333,7 @@ final class PartitionSet(as: Array[FiloPartition], bs: Array[Byte], n: Int, u: I
       val j = i & mask
       val status = buckets(j)
       if (status == 3 && items(j) == item) {
+        items(j) = null // allow item to be garbage collected
         buckets(j) = 2
         len -= 1
         true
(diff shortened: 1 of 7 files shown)


vishramachandran:
@broneill what is the effect of this change if there is plenty of heap space available, but we are running out of non-heap (native) space? 

If GC is deferred, allocation for new incoming partitions may fail. Is this right? Could you explain the behavior in such a scenario?

vishramachandran:
@broneill I am not clear if this PR addresses the problem of readers being present (due to queries) at the time of freeing partition's (or shard's) native memory. 

Are we relying on the fact that the latency between removal of the partition to actual GC of the partition is sufficient for readers to finish their work on that partition's native memory ?

broneill:
> @broneill what is the effect of this change if there is plenty of heap space available, but we are running out of non-heap (native) space?
> 
> If GC is deferred, allocation for new incoming partitions may fail. Is this right? Could you explain the behavior in such a scenario?

If partitions cannot be allocated, then the current behavior is to drop the data and try again later. I can add an explicit GC before this, to help reduce the likelihood of this happening.

broneill:
> @broneill I am not clear if this PR addresses the problem of readers being present (due to queries) at the time of freeing partition's (or shard's) native memory.
> 
> Are we relying on the fact that the latency between removal of the partition to actual GC of the partition is sufficient for readers to finish their work on that partition's native memory ?

The readers will have a reference to the chunk map (OffheapLFSortedIDMap) indirectly through the thread-local reference to the shared lock state. The chunk map and the partition object are the same thing, and so the partition object remains non-reclaimable by the GC until after the readers finish.

PR #204, opened 2019-01-16
feat(core): Downsampling of raw time series data into another dataset

vishramachandran:
**Pull Request checklist**

Now able to downsample raw data into another dataset.
Downsampling happens at chunk flush time.

Every data column of the dataset can be associated with one more downsamplers. This configuration is done at the time of dataset creation. See documentation for more details. 

Downsample strategy for prometheus style counters will come in another PR.

Automatic column selection in query of downsampled data will also come in another PR.

For the local dev setup, we create another prometheus dataset for the downsample data in the same dev cluster.

Now that we get serious about deploying multiple datasets into same cluster, added copious logging of dataset field in all our log statements.


core/src/main/scala/filodb.core/binaryrecord2/RecordBuilder.scala
```diff
@@ -195,6 +249,8 @@ final class RecordBuilder(memFactory: MemFactory,
     recHash = combineHash(recHash, newHash)
   }
 
+  final def addFieldsFromBinRecord(base: Any, offset: Long, schema: RecordSchema): Unit = ???
```
broneill: Adding a final method that always fails? Why?

vishramachandran: Forgot to remove that. Will do.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -595,11 +603,12 @@ class TimeSeriesShard(val dataset: Dataset,
       System.currentTimeMillis() - storeConfig.demandPagedRetentionPeriod.toMillis)
     var numDeleted = 0
     InMemPartitionIterator(deletedParts).foreach { p =>
-      logger.debug(s"Purging partition with partId ${p.partID} from memory")
+      logger.debug(s"Purging partition with partId ${p.partID} from memoryin dataset=${dataset.ref} shard=$shardNum")
```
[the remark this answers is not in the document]
vishramachandran: Fixed.

core/src/main/scala/filodb.core/binaryrecord2/RecordBuilder.scala
```diff
@@ -132,6 +134,58 @@ final class RecordBuilder(memFactory: MemFactory,
...
+    addBlob(base, strDataOffset + 2, schema.blobNumBytes(base, strDataOffset, col))
+  }
+
+  /**
+    * IMPORTANT: Internal method, does not update hash values for the map key/values individually.
+    * If this method is used, then caller needs to also update the partitionHash manually.
+    */
+  private def addMapFromBr(base: Any, offset: Long, col: Int, schema: RecordSchema): Unit = {
+    val strDataOffset = schema.utf8StringOffset(base, offset, col)
+    addMap(base, strDataOffset + 4, BinaryRegionMedium.numBytes(base, strDataOffset))
```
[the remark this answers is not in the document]
velvia: Though it happens to work because on Intel processors one is little-endian and reading a short is the same as reading an int.  But it's better to change it; edge cases might not work otherwise.

core/src/main/scala/filodb.core/binaryrecord2/RecordBuilder.scala
```diff
@@ -132,6 +134,58 @@ final class RecordBuilder(memFactory: MemFactory,
...
+  private def addDoubleFromBr(base: Any, offset: Long, col: Int, schema: RecordSchema): Unit = {
+    addDouble(schema.getDouble(base, offset, col))
+  }
+
+  /**
+    * This also updates the hash for this record. OK since partKeys are added at the very end
+    * of the record.
+    */
+  final def addPartKeyFromBr(base: Any, offset: Long, partKeySchema: RecordSchema): Unit = {
+    var id = 0
```
velvia: Can we call this `fieldNo` or `colNo`?

core/src/main/scala/filodb.core/binaryrecord2/RecordBuilder.scala
```diff
@@ -132,6 +134,58 @@ final class RecordBuilder(memFactory: MemFactory,
...
+  private def addLongFromBr(base: Any, offset: Long, col: Int, schema: RecordSchema): Unit = {
+    addLong(schema.getLong(base, offset, col))
+  }
+
+  private def addDoubleFromBr(base: Any, offset: Long, col: Int, schema: RecordSchema): Unit = {
+    addDouble(schema.getDouble(base, offset, col))
+  }
+
+  /**
+    * This also updates the hash for this record. OK since partKeys are added at the very end
```
velvia: Let's add a more extensive ScalaDoc here.  Also I think a better name for the method is `addPartKeyRecordFields`; since we are adding all the fields from a partition key BR.

core/src/main/scala/filodb.core/binaryrecord2/RecordSchema.scala
```diff
@@ -142,6 +146,10 @@ final class RecordSchema(val columns: Seq[ColumnInfo],
   def utf8StringOffset(base: Any, offset: Long, index: Int): Long =
     offset + UnsafeUtils.getInt(base, offset + offsets(index))
 
+  def utf8StringNumBytes(base: Any, offset: Long, index: Int): Long =
```
velvia: This method is redundant.  It is exactly the same as `blobNumBytes`.

core/src/main/scala/filodb.core/downsample/ChunkDownsampler.scala
```diff
@@ -0,0 +1,272 @@
...
+      rowNum += 1
+    }
+    max
+  }
+}
+
+/**
+  * Downsamples by calculating average from average and count columns
+  */
+case class AvgAcDownsampler(override val colIds: Array[Int]) extends DoubleChunkDownsampler {
```
velvia: What is the use case here?  I don't quite understand it.  Maybe add docs for the use case.

core/src/main/scala/filodb.core/downsample/ChunkDownsampler.scala
```diff
@@ -0,0 +1,272 @@
...
+  val avgCol = colIds(0)
+  val countCol = colIds(1)
+  override def downsampleChunk(part: TimeSeriesPartition,
+                               chunkset: ChunkSetInfo,
+                               startRow: Int,
+                               endRow: Int): Double = {
+    val avgVecPtr = chunkset.vectorPtr(avgCol)
+    val avgColReader = part.chunkReader(avgCol, avgVecPtr)
+    val cntVecPtr = chunkset.vectorPtr(countCol)
+    val cntColReader = part.chunkReader(countCol, cntVecPtr)
```
velvia: You might want to check that the count column is in fact a DoubleColumn.  It might not be, or the user might make a mistake in configuration.

core/src/main/scala/filodb.core/downsample/ChunkDownsampler.scala
```diff
@@ -0,0 +1,272 @@
...
+    * "name(4@1)" where "name" is the downsampler name, 4 & 1 are the column IDs to be used by the function
+    */
+  def downsampler(strNotation: String): ChunkDownsampler = {
+    val parts = strNotation.split("[(@)]")
+    // TODO possibly better validation of string notation
+    require(parts.size >= 2, s"Downsampler '$strNotation' does not have downsampler name and column id. ")
+    val name = parts(0)
+    val colIds = parts.drop(1).map(_.toInt)
+    name match {
+      case "dMin"   => MinDownsampler(colIds)
```
velvia: I suggest you use the `Enum` / `EnumEntry` types/classes (like ColumnType has).  It automates / eliminates the code like you have below, makes it much nicer and less prone to error.

core/src/main/scala/filodb.core/downsample/ChunkDownsampler.scala
```diff
@@ -0,0 +1,272 @@
...
+  override val name: String = "dMin"
+  override def downsampleChunk(part: TimeSeriesPartition,
+                               chunkset: ChunkSetInfo,
+                               startRow: Int,
+                               endRow: Int): Double = {
+
+    val vecPtr = chunkset.vectorPtr(colIds(0))
+    val colReader = part.chunkReader(colIds(0), vecPtr)
+
+    var min = Double.MaxValue
```
velvia: This code overlaps with the code in the ChunkedMinRangeFunction thing (forgot exact name).  Maybe we should move it to a more common place, like in the DoubleReader.

core/src/main/scala/filodb.core/downsample/DownsamplePublisher.scala
```diff
@@ -0,0 +1,53 @@
...
+    * if there is any slowdown here.
+    *
+    * For now, there is a possibility of loss of samples in memory that have not been dispatched.
+    *
+    * @param shardNum
+    * @param resolution
+    * @param records each Array of Byte is a bunch of Binary Records that can be deserialized by the ingestion pipeline.
+    * @return Future of Success if all is good.
+    */
+  def publish(shardNum: Int, resolution: Int, records: Seq[Array[Byte]]): Future[Response]
```
velvia: I haven't gotten to how this is tied into the flush logic, but it seems to me that you want an Observable.  The original  flush chunks is an Observable, it seems you can then feed that into downsampling logic and emit and Observable of containers.  See how I do it in the GatewayServer.

core/src/main/scala/filodb.core/downsample/DownsampleOps.scala
```diff
@@ -0,0 +1,98 @@
...
+                 chunksets: ChunkInfoIterator,
+                 dsStates: Seq[DownsamplingState]): Unit = {
+    while (chunksets.hasNext) {
+      val chunkset = chunksets.nextInfo
+      val startTime = chunkset.startTime
+      val endTime = chunkset.endTime
+      // for each downsample resolution
+      dsStates.foreach { case DownsamplingState(res, builder) =>
+        var pStart = ( (startTime - 1) / res) * res + 1 // inclusive startTime for downsample period
+      var pEnd = pStart + res // end is inclusive
```
velvia: There's some weird indentation going on in these few lines.

core/src/main/scala/filodb.core/downsample/DownsampleOps.scala
```diff
@@ -0,0 +1,98 @@
...
+        // for each downsample period
+        while (pStart <= endTime) {
+          var startRowNum = 0
+          var endRowNum = 0
+          builder.startNewRecord()
+          // assume that first row key column is timestamp column, fix the row numbers for the downsample period
+          val timestampCol = dataset.rowKeyColumns.head.id
+          val vecPtr = chunkset.vectorPtr(timestampCol)
+          val colReader = part.chunkReader(timestampCol, vecPtr)
```
velvia: Rename this `timeReader` or `tsReader`?

core/src/main/scala/filodb.core/downsample/DownsampleOps.scala
```diff
@@ -0,0 +1,98 @@
...
+      dsStates.foreach { case DownsamplingState(res, builder) =>
+        var pStart = ( (startTime - 1) / res) * res + 1 // inclusive startTime for downsample period
+      var pEnd = pStart + res // end is inclusive
+        // for each downsample period
+        while (pStart <= endTime) {
+          var startRowNum = 0
+          var endRowNum = 0
+          builder.startNewRecord()
+          // assume that first row key column is timestamp column, fix the row numbers for the downsample period
+          val timestampCol = dataset.rowKeyColumns.head.id
```
velvia: Please just use `dataset.timestampColID`

core/src/main/scala/filodb.core/downsample/DownsampleOps.scala
```diff
@@ -0,0 +1,98 @@
...
+      val chunkset = chunksets.nextInfo
+      val startTime = chunkset.startTime
+      val endTime = chunkset.endTime
+      // for each downsample resolution
+      dsStates.foreach { case DownsamplingState(res, builder) =>
+        var pStart = ( (startTime - 1) / res) * res + 1 // inclusive startTime for downsample period
+      var pEnd = pStart + res // end is inclusive
+        // for each downsample period
+        while (pStart <= endTime) {
+          var startRowNum = 0
```
velvia: I don't see why these row numbers have to be vars.  They are only set once.

core/src/main/scala/filodb.core/downsample/DownsampleOps.scala
velvia: Also, the reader and vectorPtr is identical regardless of the downsampling resolution.  Move this outside and we save a tiny bit on initialization.

core/src/main/scala/filodb.core/memstore/DemandPagedChunkStore.scala
```diff
@@ -63,8 +63,7 @@ extends RawToPartitionMaker with StrictLogging {
    */
   def populateRawChunks(rawPartition: RawPartData): Task[ReadablePartition] = Task {
     // Find the right partition given the partition key
-    tsShard.getPartition(rawPartition.partitionKey).map { partition =>
-      val tsPart = partition.asInstanceOf[TimeSeriesPartition]
+    tsShard.getPartition(rawPartition.partitionKey).map { tsPart =>
```
velvia: This seems like an unnecessary rename

core/src/main/scala/filodb.core/store/IngestionConfig.scala
```diff
@@ -127,17 +128,19 @@ object IngestionConfig {
...
-      resolved  <- sourceConfig.resolveT
-      dataset   <- resolved.stringT(DatasetRefKey) // fail fast if missing
-      factory   <- resolved.stringT(SourceFactory) // fail fast if missing
-      numShards <- numShards(resolved)             // fail fast if missing
-      minNodes  <- minNumNodes(resolved)           // fail fast if missing
-      streamConfig = resolved.as[Option[Config]](IngestionKeys.SourceConfig).getOrElse(ConfigFactory.empty)
-      ref          = DatasetRef.fromDotString(dataset)
-      storeConf <- streamConfig.configT("store").map(StoreConfig.apply)
-    } yield IngestionConfig(ref, resolved, factory, streamConfig, storeConf)
+      resolved <- topLevelConfig.resolveT
```
velvia: Can you please line up the <- again?  It really looks much nicer with it like that.

doc/downsampling.md
```diff
@@ -0,0 +1,129 @@
...
+
+```
+./filo-cli -Dconfig.file=conf/timeseries-filodb-server.conf  --command create --dataset prometheus --dataColumns timestamp:ts,value:double --partitionColumns tags:map --shardKeyColumns __name__,app --downsamplers "tTime(0),dMin(1),dMax(1),dSum(1),dCount(1),dAvg(1)"
+```
+
+In the above example, the data column `value` with index 1 is configured with the dMin, dMax, sSum, dCount and dAvg downsamplers.
+
+Additional configuration is supplied via the ingestion config file for the dataset.
+
+```
```
velvia: I suggest we omit this and just have the user refer to (or supply a link to) the timeseries-dev-source file.  In the past when I tried maintaining this it always gets out of date.

kafka/src/main/scala/filodb/kafka/KafkaDownsamplePublisher.scala
```diff
@@ -0,0 +1,67 @@
...
+        Future.failed(new IllegalArgumentException(s"Unregistered resolution $resolution"))
+    }
+  }
+
+  def start(): Unit = {
+    logger.info(s"Starting Kafka Downsampling Publisher. Will be publishing to $topics ")
+    val producer = KafkaProducerSink[JLong, Array[Byte]](kafkaConfig, sched)
+    future = Observable.repeat(0).map { _: Int =>
+      val records = new ArrayBuffer[ProducerRecord[JLong, Array[Byte]]](consumeSize)
+      val consumer = new MessagePassingQueue.Consumer[ProducerRecord[JLong, Array[Byte]]] {
```
velvia: I think there is a whole bunch of code here that is also in the gateway module, you could probably reuse some of the gateway code.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -630,11 +639,20 @@ class TimeSeriesShard(val dataset: Dataset,
     indexRb.addLong(endTime)
     // Need to add 4 to include the length bytes
     indexRb.addBlob(p.partKeyBase, p.partKeyOffset, BinaryRegionLarge.numBytes(p.partKeyBase, p.partKeyOffset) + 4)
-    logger.debug(s"Added into timebucket partId ${p.partID} in shard=$shardNum partKey[${p.stringPartition}] with " +
-      s"startTime=$startTime endTime=$endTime")
+    logger.debug(s"Added into timebucket partId ${p.partID} in dataset=${dataset.ref} shard=$shardNum " +
```
velvia: I think next time, the changing of the logs should be a separate PR.

conf/timeseries-ds-1m-dev-source.conf
```diff
@@ -0,0 +1,59 @@
...
+      # This accepts both the standard kafka value of a comma-separated
+      # string and a Typesafe list of String values
+      # EXCEPT: do not populate value.deserializer, as the Kafka format is fixed in FiloDB to be messages of RecordContainer's
+      bootstrap.servers = "localhost:9092"
+      group.id = "filo-db-timeseries-ds-ingestion"
+
+      # Values controlling in-memory store chunking, flushing, etc.
+      store {
+        # Interval it takes to flush ALL time series in a shard.  This time is further divided by groups-per-shard
+        flush-interval = 1h
```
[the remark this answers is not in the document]
vishramachandran: Yes, but not in the dev-test configuration. Takes longer to test.

coordinator/src/main/scala/filodb.coordinator/ShardManager.scala
```diff
@@ -366,7 +367,7 @@ private[coordinator] final class ShardManager(settings: FilodbSettings,
         val metrics = new ShardHealthStats(setup.ref, _shardMappers(dataset.ref))
         val resources = setup.resources
         val source = setup.source
-        val state = DatasetInfo(resources, metrics, source, setup.storeConfig, dataset)
+        val state = DatasetInfo(resources, metrics, source, setup.downsample, setup.storeConfig, dataset)
```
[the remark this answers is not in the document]
vishramachandran: Yes. We need to do this sooner or later with all the new features (like histograms) flowing in.
It may just be easier to get this all in one document and place it either in configuration file. There is much scope for and need to simplify the dataset configuration loading part quite a bit.

Deferred to another PR.

core/src/main/scala/filodb.core/binaryrecord2/RecordBuilder.scala
vishramachandran: It literally is the column id of the data columns. Why call it field id ?
This is a column store format, so why not leave it as column?

core/src/main/scala/filodb.core/downsample/ChunkDownsampler.scala
vishramachandran: I tried this, but am possibly missing some information to cover all implementations. Can we leave the refactor out to a subsequent PR? I don't want to digress too much from the scope of this PR.

core/src/main/scala/filodb.core/downsample/DownsamplePublisher.scala
vishramachandran: See Step 3 - documentation just added in the doFlushSteps of TimeSeriesShard for details on how this is linked to the flush pipeline.

The kafka publish is indeed an observable inside the Kafka publisher, but it is not connected to the flush observable because of the checkpointing.

Today, I have queued things in memory for better throughput, but we may decide to change our mind and wait for an ack from Kafka before checkpointing.

kafka/src/main/scala/filodb/kafka/KafkaDownsamplePublisher.scala
vishramachandran: I did consider this seriously, but the requirements are slightly different with the resolutions etc. This was simple enough to implement in isolation. Moreover, we may decide to change it to include acks before checkpoint.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
vishramachandran: This was absolutely necessary for keeping sanity and debug my dev environment. With two datasets in same cluster, it is hard to distinguish which logs are for which dataset.

core/src/main/scala/filodb.core/downsample/ChunkDownsampler.scala
vishramachandran: This is where I think simplifying with just double column (for now) makes configuration as well as implementation simple. 

Here is where things gets murky:

* Can count downsample be a long or double?
* Should sum of count columns be a long or double?
* Should sum then have a long version too?
* Can sum and count types interchange ?

It is just simple to deal with one input type (at least for now) and cross the bridge of dealing with multiple types later. Let's discuss in person tomorrow. This thread may get long.

vishramachandran:
@velvia I have answered all of your comments. Please take another look. We can discuss in person if necessary.

coordinator/src/main/scala/filodb.coordinator/ShardManager.scala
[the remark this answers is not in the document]
velvia: Ok can we add a TODO and issue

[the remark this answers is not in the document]
velvia: What I was trying to get at was that there were so many changes to so many files due to adding downsampling config, that it would have been easier to just change it into one config.

core/src/main/scala/filodb.core/downsample/ShardDownsampler.scala
```diff
@@ -0,0 +1,134 @@
...
+      while (chunksets.hasNext) {
+        val chunkset = chunksets.nextInfo
+        val startTime = chunkset.startTime
+        val endTime = chunkset.endTime
+        val vecPtr = chunkset.vectorPtr(timestampCol)
+        val tsReader = part.chunkReader(timestampCol, vecPtr).asLongReader
+        // for each downsample resolution
+        records.foreach { case DownsampleRecords(resolution, builder) =>
+          var pStart = ((startTime - 1) / resolution) * resolution + 1 // inclusive startTime for downsample period
+        var pEnd = pStart + resolution // end is inclusive
```
velvia: Indentation here hasn't been fixed

PR #205, opened 2019-01-17
misc(core): unit test for testing extractTimebucket with multiple string partition columns

sherali42:
**Pull Request checklist**

Unit Test for validating the failure while extracting time buckets from cassandra. This is causing SEGV errors while instance restarts. It is happening if the schema has multiple string partition columns.


velvia:
@sherali42 @vishramachandran my proposal for making this code easier:
1) Create a special value class or small class just for pointing at strings or blobs (things which take a 2-byte length prefix).
2) This class has methods for extracting, converting to strings, etc. and also converting to/from BytesRef

The above will abstract over the need to deal directly with base, offset, etc. which should remove much of the confusion.

PR #207, opened 2019-01-22
bug(core): Improve safety of freeing native memory.

broneill:
**Pull Request checklist**

**Current behavior :**
Native memory can be freed too soon, due to various race conditions.

**New behavior :**
Locks are put in place to ensure that partitions are added/removed in a thread-safe fashion, and finalization is used to safely free the memory occupied by an entire shard. Shard removal is rare, and so delayed reclamation due to finalization isn't much of an issue.

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/OnDemandPagingShard.scala
+++ b/core/src/main/scala/filodb.core/memstore/OnDemandPagingShard.scala
@@ -134,7 +134,12 @@ TimeSeriesShard(dataset, storeConfig, shardNum, rawStore, metastore, evictionPol
                 unsafeKeyOffset = PartKeyLuceneIndex.bytesRefToUnsafeOffset(partKeyBytesRef.offset)
                 group = partKeyGroup(dataset.partKeySchema, partKeyBytesRef.bytes, unsafeKeyOffset, numGroups)
                 part <- Option(createNewPartition(partKeyBytesRef.bytes, unsafeKeyOffset, group, 4)) } yield {
-            partSet.add(part)
+            val stamp = partSetLock.writeLock()
+            try {
+              partSet.add(part)
+            } finally {
+              partSetLock.unlockWrite(stamp)
+            }
             callback(partKeyBytesRef.bytes, unsafeKeyOffset)
             part
           }
--- a/core/src/main/scala/filodb.core/memstore/PartitionSet.scala
+++ b/core/src/main/scala/filodb.core/memstore/PartitionSet.scala
@@ -333,6 +333,7 @@ final class PartitionSet(as: Array[FiloPartition], bs: Array[Byte], n: Int, u: I
       val j = i & mask
       val status = buckets(j)
       if (status == 3 && items(j) == item) {
+        items(j) = null // allow item to be garbage collected
         buckets(j) = 2
         len -= 1
         true
@@ -513,6 +514,7 @@ final class PartitionSet(as: Array[FiloPartition], bs: Array[Byte], n: Int, u: I
     if (lhs.size <= rhs.size) {
       cfor(0)(_ < buckets.length, _ + 1) { i =>
         if (buckets(i) == 3 && !rhs(items(i))) {
+          items(i) = null // allow item to be garbage collected
           buckets(i) = 2
           len -= 1
         }
@@ -561,6 +563,7 @@ final class PartitionSet(as: Array[FiloPartition], bs: Array[Byte], n: Int, u: I
     } else {
       cfor(0)(_ < buckets.length, _ + 1) { i =>
         if (buckets(i) == 3 && rhs(items(i))) {
+          items(i) = null // allow item to be garbage collected
           buckets(i) = 2
           len -= 1
         }
@@ -665,6 +668,7 @@ final class PartitionSet(as: Array[FiloPartition], bs: Array[Byte], n: Int, u: I
   def filterSelf(p: FiloPartition => Boolean): Unit =
     cfor(0)(_ < buckets.length, _ + 1) { i =>
       if (buckets(i) == 3 && !p(items(i))) {
+        items(i) = null // allow item to be garbage collected
         buckets(i) = 2
         len -= 1
       }
--- a/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
+++ b/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
@@ -901,40 +901,29 @@ class TimeSeriesShard(val dataset: Dataset,
     }
     part
   }
+  // scalastyle:on
 
   private def addPartition(recordBase: Any, recordOff: Long, group: Int, ingestOffset: Long) = {
-    val stamp = partSetLock.writeLock()
-
-    // Double check just in case multiple threads are calling this method.
-    val part = partSet.getWithIngestBR(recordBase, recordOff)
-    if (part != null) {
-      partSetLock.unlockWrite(stamp)
-      part
-    } else {
-      var newPart: TimeSeriesPartition = null
+    val partKeyOffset = recordComp.buildPartKeyFromIngest(recordBase, recordOff, partKeyBuilder)
+    val newPart = createNewPartition(partKeyArray, partKeyOffset, group)
+    if (newPart != OutOfMemPartition) {
+      val partId = newPart.partID
+      // NOTE: Don't use binRecordReader here.  recordOffset might not be set correctly
+      val startTime = dataset.ingestionSchema.getLong(recordBase, recordOff, timestampColId)
+      partKeyIndex.addPartKey(newPart.partKeyBytes, partId, startTime)()
+      timeBucketBitmaps.get(currentIndexTimeBucket).set(partId)
+      activelyIngesting.set(partId)
+      val stamp = partSetLock.writeLock()
       try {
-        val partKeyOffset = recordComp.buildPartKeyFromIngest(recordBase, recordOff, partKeyBuilder)
-        newPart = createNewPartition(partKeyArray, partKeyOffset, group)
-        if (newPart != OutOfMemPartition) {
-          val partId = newPart.partID
-          // NOTE: Don't use binRecordReader here.  recordOffset might not be set correctly
-          val startTime = dataset.ingestionSchema.getLong(recordBase, recordOff, timestampColId)
-          partKeyIndex.addPartKey(newPart.partKeyBytes, partId, startTime)()
-          timeBucketBitmaps.get(currentIndexTimeBucket).set(partId)
-          activelyIngesting.set(partId)
-          partSet.add(newPart)
-        }
+        partSet.add(newPart)
       } finally {
         partSetLock.unlockWrite(stamp)
       }
-      if (newPart != OutOfMemPartition) {
-        logger.trace(s"Created new partition ${newPart.stringPartition} on dataset=${dataset.ref} " +
-          s"shard $shardNum at offset $ingestOffset")
-      }
-      newPart
+      logger.trace(s"Created new partition ${newPart.stringPartition} on dataset=${dataset.ref} " +
+        s"shard $shardNum at offset $ingestOffset")
     }
+    newPart
   }
-  // scalastyle:on
 
   /**
     * Retrieves or creates a new TimeSeriesPartition, updating indices, then ingests the sample from record.
--- a/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
+++ b/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
@@ -233,6 +233,8 @@ class TimeSeriesShard(val dataset: Dataset,
     * PartitionSet - access TSPartition using ingest record partition key in O(1) time.
     */
   private[memstore] final val partSet = PartitionSet.ofSize(InitialNumPartitions, ingestSchema, recordComp)
+  // Use a StampedLock because it supports optimistic read locking. This means that no blocking
+  // occurs in the common case, when there isn't any contention reading from partSet.
   private[memstore] final val partSetLock = new StampedLock
 
   // The off-heap block store used for encoded chunks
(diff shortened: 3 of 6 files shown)


core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -957,10 +991,28 @@ class TimeSeriesShard(val dataset: Dataset,
...
+  /**
+   * Permanently removes the given partition ID from our in-memory data structures
+   * Also frees partition key if necessary
+   * @param locked true if caller holds partSetLock write lock
+   */
+  private def doRemovePartition(partitionObj: TimeSeriesPartition, locked: Boolean): Unit = {
+    if (locked) {
+      partSet.remove(partitionObj)
+    } else {
+      val stamp = partSetLock.writeLock()
```
vishramachandran: This statement is in the ingestion path. We should make sure this will exit and not "hang" or "block" ingestion.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -232,6 +233,7 @@ class TimeSeriesShard(val dataset: Dataset,
     * PartitionSet - access TSPartition using ingest record partition key in O(1) time.
     */
   private[memstore] final val partSet = PartitionSet.ofSize(InitialNumPartitions, ingestSchema, recordComp)
+  private[memstore] final val partSetLock = new StampedLock
```
vishramachandran: Can we document why a StampedLock was chosen here as opposed to any other lock?

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -537,7 +545,16 @@ class TimeSeriesShard(val dataset: Dataset,
 
   def numRowsIngested: Long = ingested
 
-  def numActivePartitions: Int = partSet.size
+  def numActivePartitions: Int = {
+    while (true) {
+      val stamp = partSetLock.tryOptimisticRead()
+      val size = partSet.size
+      if (partSetLock.validate(stamp)) {
+        return size
```
vishramachandran: This looks like a possibility for continuous looping. Need much caution since it is called from the ingestion path for update of gauges.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
broneill: Do you mean an exception? If an exception is thrown from this method, then the underlying data structures used here are corrupt. Ensuring the lock is released doesn't fix that.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
broneill: Yup.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
broneill: Will do.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
broneill: Not expected, but I changed the code and added a comment.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
vishramachandran: No. What I meant is that writeLock method call is a blocking call which wont return until lock is obtained. Did we evaluate the possibility of it hanging for a long time? Should there be a timeout ?

vishramachandran: I was referring to https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/StampedLock.html#writeLock--

> public long writeLock()
Exclusively acquires the lock, blocking if necessary until available.

broneill: Other than ingestion, the query acquires a read lock when calling getPartition. It should only be holding the lock on the order of a microsecond at most. No significant blocking should occur on the ingestion path.

core/src/main/scala/filodb.core/memstore/OnDemandPagingShard.scala
```diff
@@ -140,7 +140,12 @@ TimeSeriesShard(dataset, storeConfig, shardNum, rawStore, metastore, evictionPol
                 unsafeKeyOffset = PartKeyLuceneIndex.bytesRefToUnsafeOffset(partKeyBytesRef.offset)
                 group = partKeyGroup(dataset.partKeySchema, partKeyBytesRef.bytes, unsafeKeyOffset, numGroups)
                 part <- Option(createNewPartition(partKeyBytesRef.bytes, unsafeKeyOffset, group, 4)) } yield {
-            partSet.add(part)
+            val stamp = partSetLock.writeLock()
```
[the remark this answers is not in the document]
broneill: The write lock acquisition isn't a concern. Each loop iteration calls into Lucene and creates a new partition. I cannot infer the cost of these operations because to some extent, they're a black box. In fact, creating a new partition allocates memory, which might in turn block some more. So holding the partSetLock for a longer time stalls queries which need to read from partSet.

The usual trick I apply in cases like this doesn't work with a StampedLock:  Acquire the lock outside the loop and keep it. In the middle of the loop, check if there's a queue of waiters to gain access to the lock. If so, release and re-acquire to let them through. There's no method on StampedLock to ask if there's a queue.

core/src/main/scala/filodb.core/memstore/PartitionSet.scala
```diff
@@ -561,6 +563,7 @@ final class PartitionSet(as: Array[FiloPartition], bs: Array[Byte], n: Int, u: I
     } else {
       cfor(0)(_ < buckets.length, _ + 1) { i =>
         if (buckets(i) == 3 && rhs(items(i))) {
+          items(i) = null // allow item to be garbage collected
```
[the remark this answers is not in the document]
broneill: Yes, this allows the memory to be freed sooner. Leaving the object reference in the set means that it might live forever. There's no guarantee that this slot will ever be re-used.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -565,7 +575,19 @@ class TimeSeriesShard(val dataset: Dataset,
 
   def numRowsIngested: Long = ingested
 
-  def numActivePartitions: Int = partSet.size
+  def numActivePartitions: Int = {
+    var stamp = partSetLock.tryOptimisticRead()
+    var size = partSet.size
+    if (!partSetLock.validate(stamp)) {
+      // It's not expected that the optimistic lock will fail, considering that this method is
```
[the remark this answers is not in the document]
broneill: I'll use a weaker comment.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -915,10 +949,12 @@ class TimeSeriesShard(val dataset: Dataset,
                                              s"shard=$shardNum", e); disableAddPartitions()
     }
 
+  // Note: Caller must hold partSetLock write lock.
+  // scalastyle:off
```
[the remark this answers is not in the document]
broneill: I get this completely bizarre error for this line:

    if (addPartitionsDisabled() || !ensureFreeSpace(true)) { OutOfMemPartition }

Boolean expression can be simplified

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -1039,8 +1100,22 @@ class TimeSeriesShard(val dataset: Dataset,
     partKeyIndex.partIdsOrderedByEndTime(storeConfig.numToEvict, evictionWatermark, Long.MaxValue - 1)
   }
 
-  private[core] def getPartition(partKey: Array[Byte]): Option[TimeSeriesPartition] =
-    partSet.getWithPartKeyBR(partKey, UnsafeUtils.arayOffset).map(_.asInstanceOf[TimeSeriesPartition])
+  private[core] def getPartition(partKey: Array[Byte]): Option[TimeSeriesPartition] = {
+    var part = None: Option[FiloPartition]
```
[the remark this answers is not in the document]
broneill: Why did the original form compile?

broneill:
With respect to the other comment on locking once outside a loop, there's a Lucene call during each iteration. Holding a lock over a call with undefined internal behavior isn't a good idea. Better to release and reacquire and let queries through.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -876,24 +898,36 @@ class TimeSeriesShard(val dataset: Dataset,
...
-          s"shard $shardNum at offset $ingestOffset")
-      }
-      newPart
-    })
-    part
+    partSet.getWithIngestBR(recordBase, recordOff) match {
+      case null =>
+        val stamp = partSetLock.writeLock()
+        try {
+          val part = partSet.getOrAddWithIngestBR(recordBase, recordOff, {
```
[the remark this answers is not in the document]
velvia: I understand why the double check is there.  What I'm suggesting is to replace the second `getOrAddWithIngestBR` with simply `add`, because the check in `getOrAdd` is superfluous -- we already did the check with the first get call.  After all `getOrAdd` is simply the get followed by an add.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -876,24 +898,36 @@ class TimeSeriesShard(val dataset: Dataset,
 
   private[memstore] val addPartitionsDisabled = AtomicBoolean(false)
 
+  //scalastyle:off
```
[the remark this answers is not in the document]
velvia: Re: partitionSet that is true, but most of that is not our code, I can't vouch for that specific style.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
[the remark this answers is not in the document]
velvia: It compiles but is not considered good Scala form.  :)

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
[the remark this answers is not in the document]
broneill: That makes sense, although what I'm doing is double-checked locking. Although no other thread is likely to call this method, the second check prevents any odd issues.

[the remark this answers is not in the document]
broneill: I'm going fix the code such that an optimistic read lock is used first.

velvia:
@broneill so the main iteration through partitionSet by most queries does not hit the getPartition method.  See the `scanPartitions` method.  So that has to be fixed first.
I'm a bit worried about trying to acquire a lock every single time we get a partition object from ID though, that could slow down queries.  I would be sure to run benchmarks before and after....  maybe the `QueryInMemoryBenchmark` to be sure this isn't an issue.

broneill:
The scanPartitionsMethod doesn't (directly) touch partSet. The lock is there to guard access to partSet only. The other collection, partitions, is a NonBlockingHashMapLong, which is concurrent.

As far as the cost of getting a lock for each partition access, it uses an optimistic lock on the fast path, which is essentially free.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -898,33 +898,55 @@ class TimeSeriesShard(val dataset: Dataset,
...
+      val part = partSet.getWithIngestBR(recordBase, recordOff)
+      if (part != null) {
+        return part
+      }
+    } finally {
+      partSetLock.unlockRead(stamp)
+    }
+
+    stamp = partSetLock.writeLock()
+    var part = partSet.getWithIngestBR(recordBase, recordOff)
```
velvia: At this point you have proven that the record does not exist, and the ingestion is single threaded.  Why have to execute another get?

broneill: By the way, an optimistic read just reads a field twice. This costs virtually nothing.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -905,33 +893,26 @@ class TimeSeriesShard(val dataset: Dataset,
...
-    var part = partSet.getWithIngestBR(recordBase, recordOff)
-    var newPart: TimeSeriesPartition = null
+  private def addPartition(recordBase: Any, recordOff: Long, group: Int, ingestOffset: Long) = {
+    val stamp = partSetLock.writeLock()
 
-    try {
-      if (part == null) {
+    // Double check just in case multiple threads are calling this method.
+    val part = partSet.getWithIngestBR(recordBase, recordOff)
+    if (part != null) {
```
velvia: @broneill I'll trust your intuition that the double check takes care of diff race conditions.   :)

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -1051,8 +1079,22 @@ class TimeSeriesShard(val dataset: Dataset,
     partKeyIndex.partIdsOrderedByEndTime(storeConfig.numToEvict, evictionWatermark, Long.MaxValue - 1)
   }
 
-  private[core] def getPartition(partKey: Array[Byte]): Option[TimeSeriesPartition] =
-    partSet.getWithPartKeyBR(partKey, UnsafeUtils.arayOffset).map(_.asInstanceOf[TimeSeriesPartition])
+  private[core] def getPartition(partKey: Array[Byte]): Option[TimeSeriesPartition] = {
+    var part: Option[FiloPartition] = None
+    var stamp = partSetLock.tryOptimisticRead()
```
velvia: It might be good to add a comment here to explain what the double check for locking does.

PR #209, opened 2019-01-22
bug(core): bug fix in the rate function

sherali42:
**Pull Request checklist**

There is a bug in the helper method `extrapolatedRate`, which is causing a slight deviations for the rate/delta/increase calculations from Prometheus engine.

https://github.com/prometheus/prometheus/blob/c6399296dcc059bbc496f40e31aa2c7b7a341951/promql/functions.go (extrapolatedRate)

### DIFF
--- a/query/src/main/scala/filodb/query/exec/rangefn/RateFunctions.scala
+++ b/query/src/main/scala/filodb/query/exec/rangefn/RateFunctions.scala
@@ -22,7 +22,7 @@ object RateFunctions {
       var durationToStart = (window.head.timestamp - startTimestamp).toDouble / 1000
       val durationToEnd = (endTimestamp - window.last.timestamp).toDouble / 1000
       val sampledInterval = (window.last.timestamp - window.head.timestamp).toDouble / 1000
-      val averageDurationBetweenSamples = sampledInterval / window.size.toDouble
+      val averageDurationBetweenSamples = sampledInterval / (window.size.toDouble - 1)
       val delta = window.last.value - window.head.value
 
       if (isCounter && delta > 0 && window.head.value >= 0) {


query/src/main/scala/filodb/query/exec/rangefn/RateFunctions.scala
```diff
@@ -22,7 +22,7 @@ object RateFunctions {
       var durationToStart = (window.head.timestamp - startTimestamp).toDouble / 1000
       val durationToEnd = (endTimestamp - window.last.timestamp).toDouble / 1000
       val sampledInterval = (window.last.timestamp - window.head.timestamp).toDouble / 1000
-      val averageDurationBetweenSamples = sampledInterval / window.size.toDouble
+      val averageDurationBetweenSamples = sampledInterval / (window.size.toDouble - 1)
```
broneill: Is there a unit test to verify this?

vishramachandran:
@sherali42 If you can create a quick test with the failing test case into WindowIterationSpec, that would be nice.

query/src/main/scala/filodb/query/exec/rangefn/RateFunctions.scala
velvia: Is it possible for the window size to be 1, causing a DivideByZero to be thrown?

broneill: @velvia as a double, it will end up being infinity or NaN.

sherali42: @vishramachandran @broneill @velvia added a test case, which was failing earlier without the fix. RawData/ResultSet is taken from prometheus engine

PR #210, opened 2019-01-23
misc(query): allow metadata series queries with no shardcolumns in filters

sherali42:
**Pull Request checklist**

Metadata queries fails without sharwkey column filters. This change is to allow metadata queries without shard key columns in the column filters. Similar change is already in place for label queries. This is needed for integrating with graphana dashboards.

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
@@ -229,7 +229,13 @@ class QueryEngine(dataset: Dataset,
                                      options: QueryOptions,
                                      lp: SeriesKeysByFilters): Seq[PartKeysExec] = {
     val renamedFilters = renameMetricFilter(lp.filters)
-    shardsFromFilters(renamedFilters, options).map { shard =>
+    val filterCols = lp.filters.map(_.column).toSet
+    val shardsToHit = if (shardColumns.toSet.subsetOf(filterCols)) {
+                        shardsFromFilters(lp.filters, options)
+                      } else {
+                        shardMapperFunc.assignedShards
+                      }
+    shardsToHit.map { shard =>
       val dispatcher = dispatcherForShard(shard)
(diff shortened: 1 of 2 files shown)


coordinator/src/test/scala/filodb.coordinator/client/SerializationSpec.scala
```diff
@@ -230,6 +230,32 @@ class SerializationSpec extends ActorTest(SerializationSpecConfig.getNewSystem)
...
+      qParams)
+    val execPlan1 = engine.materialize(logicalPlan1, QueryOptions(0, 100))
+    roundTrip(execPlan1) shouldEqual execPlan1
+
+    // without shardcolumns in filters
+    val logicalPlan2 = Parser.metadataQueryToLogicalPlan(
+      "http_request_duration_seconds_bucket",
+      qParams)
+    val execPlan2 = engine.materialize(logicalPlan2, QueryOptions(0, 100))
+    roundTrip(execPlan2) shouldEqual execPlan2
```
velvia: Note: this doesn't actually test that with no shard columns all shards are hit.

coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
```diff
@@ -229,7 +229,13 @@ class QueryEngine(dataset: Dataset,
                                      options: QueryOptions,
                                      lp: SeriesKeysByFilters): Seq[PartKeysExec] = {
     val renamedFilters = renameMetricFilter(lp.filters)
-    shardsFromFilters(renamedFilters, options).map { shard =>
+    val filterCols = lp.filters.map(_.column).toSet
+    val shardsToHit = if (shardColumns.toSet.subsetOf(filterCols)) {
```
velvia: This is fine although you can also just all shardsFromFilters with the shardOverrides option if the shardColumns are not part of the set.
Also we should probably track how many such queries there are.  Please add a stat.

PR #212, opened 2019-01-24
feat(query): Apply limits on total number of samples

vishramachandran:
**Pull Request checklist**

Apply limit on total number of samples returned from query or sub-query instead
of sample limit per range vector paired with limit on number of range vectors.

New approach allows us to query high cardinality metrics across smaller time ranges.
This is necessary for query-based pre-aggregation.
It even allows us to query large number of samples for few range vectors.

Earlier approach could not handle these use cases.

**BREAKING CHANGES**
Semantics of limit has changed from samples-per-rangevector to total-number-of-samples
Query clients need to raise this limit.

### DIFF
--- a/core/src/main/resources/filodb-defaults.conf
+++ b/core/src/main/resources/filodb-defaults.conf
@@ -29,7 +29,7 @@ filodb {
     sample-limit = 1000000
 
     # Minimum step required for a query
-    min-step = 10 seconds
+    min-step = 5 seconds
   }
 
   shard-manager {
--- a/core/src/main/resources/filodb-defaults.conf
+++ b/core/src/main/resources/filodb-defaults.conf
@@ -27,6 +27,9 @@ filodb {
 
     # Maximum number of samples to return in a query
     sample-limit = 1000000
+
+    # Minimum step required for a query
+    min-step = 10 seconds
   }
 
   shard-manager {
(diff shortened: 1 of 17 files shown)


core/src/main/resources/filodb-defaults.conf
```diff
@@ -25,11 +25,11 @@ filodb {
...
 
-    # Maximum number of RangeVectors returned from any sub-plan QueryResult
-    # vectors-limit = 150
+    # Maximum number of samples to return in a query
+    sample-limit = 1000000
 
-    # Maximum number of samples to return per RangeVector/result time series
-    sample-limit = 200
+    # Minimum step required for a query
+    min-step = 10 seconds
```
tjackpaul: 5s default would be good. I have seen 5s scrape interval in monstor host prometheus configs

jmh/src/main/scala/filodb.jmh/QueryInMemoryBenchmark.scala
```diff
@@ -119,7 +119,7 @@ class QueryInMemoryBenchmark extends StrictLogging {
   val qParams = TimeStepParams(queryTime/1000, queryStep, (queryTime/1000) + queryIntervalMin*60)
   val logicalPlans = queries.map { q => Parser.queryRangeToLogicalPlan(q, qParams) }
   val queryCommands = logicalPlans.map { plan =>
-    LogicalPlan2Query(dataset.ref, plan, QueryOptions(1, 100))
+    LogicalPlan2Query(dataset.ref, plan, QueryOptions(1, 10000))
```
tjackpaul: Update it to 20k to match with what we have today(100*200) ?

PR #214, opened 2019-01-29
bug(akkabootstrapper): Retry fetching seeds endpoint before resorting…

vishramachandran:
… to creating new cluster

**Pull Request checklist**

Testing indicates that seeds endpoint retrieval can fail because of load balancer routing request to a node going down during a rolling upgrade.

This fix puts in place retry logic to ensure that we really solidify the observation of no cluster before creating one.

In addition, a rolling upgrade should not bring a node down while a new node is being brought up. Node downing and upping operations should be disjoint and carried out at different times.

### DIFF
--- a/akka-bootstrapper/src/main/scala/filodb/akkabootstrapper/ClusterSeedDiscovery.scala
+++ b/akka-bootstrapper/src/main/scala/filodb/akkabootstrapper/ClusterSeedDiscovery.scala
@@ -7,7 +7,7 @@ import scala.util.control.NonFatal
 import akka.actor.{Address, AddressFromURIString}
 import akka.cluster.Cluster
 import com.typesafe.scalalogging.StrictLogging
-import scalaj.http.Http
+import scalaj.http.{Http, HttpResponse}
 
 /** Seed node strategy. Some implementations discover, some simply read from immutable config. */
 abstract class ClusterSeedDiscovery(val cluster: Cluster,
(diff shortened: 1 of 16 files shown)


akka-bootstrapper/src/main/scala/filodb/akkabootstrapper/ClusterSeedDiscovery.scala
```diff
@@ -24,40 +24,44 @@ abstract class ClusterSeedDiscovery(val cluster: Cluster,
...
+    do {
+      try {
+        logger.info(s"Trying to fetch seeds from $seedsEndpoint ... $retriesRemaining retries remaining.")
+        response = Http(seedsEndpoint).timeout(2000, 2000).asString
+        logger.info(s"Seeds endpoint returned a ${response.code}. Response body was ${response.body}")
+      } catch {
+        case NonFatal(e) =>
+          logger.error(s"Seeds endpoint $seedsEndpoint call threw an exception", e)
+      }
+      Thread.sleep(settings.seedsHttpSleepBetweenRetries.toMillis)
```
tjackpaul: Isn't it better to add an exponential backoff ?

vishramachandran: This is during bootstrap, initial formation of cluster for the very first time. We should form the soon and not back-off exponentially.

akka-bootstrapper/src/main/scala/filodb/akkabootstrapper/ClusterSeedDiscovery.scala
```diff
@@ -24,40 +24,44 @@ abstract class ClusterSeedDiscovery(val cluster: Cluster,
...
 
   protected def discoverExistingCluster: Seq[Address] = {
 
     val seedsEndpoint = settings.seedsBaseUrl + settings.seedsPath
+    var response: HttpResponse[String] = null
+    var retriesRemaining = settings.seedsHttpRetries
+    do {
+      try {
+        logger.info(s"Trying to fetch seeds from $seedsEndpoint ... $retriesRemaining retries remaining.")
+        response = Http(seedsEndpoint).timeout(2000, 2000).asString
```
velvia: should this timeout be configurable also?

vishramachandran:
@velvia Currently there is a very serious bug/regression in the testing setup because of the _ns change.

We use the FormatConversion.prometheus dataset for shard key configuration which uses "_ns" as shard key, but we ingest and query with "app". 

Lets use one format and reduce confusion.

akka-bootstrapper/src/main/scala/filodb/akkabootstrapper/ClusterSeedDiscovery.scala
vishramachandran: Lets leave that out for now.

PR #218, opened 2019-01-30
 fix(memory): Check bounds of sum(), count() vector methods

velvia:
**Pull Request checklist**

It is easy to call methods like sum() and go beyond valid bounds, causing SEGVs

**New behavior :**

Bounds checks on the sum() and count() methods prevent out of bound errors.  Runtime should not be hugely affected as the bounds checks are done just one time per call and not per sample.


memory/src/main/scala/filodb.memory/format/vectors/DeltaDeltaVector.scala
```diff
@@ -231,8 +231,10 @@ object DeltaDeltaConstDataReader extends LongVectorDataReader {
   // let len = end - start + 1
   //   = initVal + start*slope + initVal + (start+1)*slope + .... + initVal + end*slope
   //   = len * initVal + len*start*slope + ((end-start)*len/2) * slope
-  final def sum(vector: BinaryVectorPtr, start: Int, end: Int): Double =
+  final def sum(vector: BinaryVectorPtr, start: Int, end: Int): Double = {
+    require(start >= 0 && end < length(vector), s"($start, $end) is out of bounds")
```
broneill: Should also include the length in the error message.

PR #219, opened 2019-01-31
bug(query): fix bug in ChunkedWindowIterator to include value present at time, timestamp - staleSampleAfterMs 

TanviBhavsar:
**Pull Request checklist**

**Current behavior :** 
Query server does not return value present at time: timestamp - staleSampleAfterMs 

**New behavior :**
Increase window by 1 to include value present at at time: timestamp - staleSampleAfterMs

### DIFF
--- a/query/src/test/scala/filodb/query/exec/WindowIteratorSpec.scala
+++ b/query/src/test/scala/filodb/query/exec/WindowIteratorSpec.scala
@@ -321,7 +321,6 @@ class WindowIteratorSpec extends RawDataWindowingSpec {
       100000, 600000L, 0,
       RangeFunction(None, ColumnType.DoubleColumn, useChunked = false),
       queryConfig)
-
     slidingWinIterator.map(r => (r.getLong(0), r.getDouble(1))).toList.filter(!_._2.isNaN) shouldEqual windowResults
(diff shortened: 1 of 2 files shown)


query/src/test/scala/filodb/query/exec/WindowIteratorSpec.scala
```diff
@@ -301,4 +301,32 @@ class WindowIteratorSpec extends RawDataWindowingSpec {
...
+      100000L -> 100d,
+      200000L -> 200d,
+      300000L -> 200d,
+      400000L -> 200d,
+      500000L -> 200d
+    )
+    val rv = timeValueRV(samples)
+
+    val slidingWinIterator = new SlidingWindowIterator(rv.rows, 100000L,
+      100000, 600000L, 0,
```
vishramachandran: Lets explicitly add a comment that SlidingWindowIterator requires window to be 0 whereas ChunkedWindowIterator requires window to be the staleSample config; 
It would be useful adding comment in both impl and test code.

TanviBhavsar: Added comment

PR #224, opened 2019-02-04
bug(coord): NPE seen in ShardManager when processing shard events

vishramachandran:
**Pull Request checklist**

NPE was seen in ShardManager when performing the following comparison:
`currentCoord.path.address == sender.path.address`

Essentially means that current coordinator for shard was null, and shard was unassigned, and yet we got a shard event. Could this be from the previous assignee of the shard?

This PR prevents the NPE and failure of the actor and logs the information for further debugging, but we still need to understand how this can happen.

@velvia Can you help verify/confirm that sender.path or currentCoord.path cannot be null too ?


vishramachandran:
@velvia Could the status actor keep sending IngestionError event to shard manager multiple times even after shard has been reassigned ?

PR #225, opened 2019-02-05
query: metadata query - enforce limit by returning truncated results

sherali42:
**Pull Request checklist**

For metadata queries, exception will not be thrown if the result size is bigger than the `limit`, instead resultset is truncated up to the limit.

### DIFF
--- a/core/src/main/scala/filodb.core/query/RangeVector.scala
+++ b/core/src/main/scala/filodb.core/query/RangeVector.scala
@@ -172,7 +172,7 @@ object SerializableRangeVector extends StrictLogging {
   def apply(rv: RangeVector,
             builder: RecordBuilder,
             schema: RecordSchema,
-            resultSize: Int = Int.MaxValue,
+            resultSize: Int = 0,
             limit: Int = Int.MaxValue,
             failOnLimit: Boolean = true): SerializableRangeVector = {
     var numRows = 0
--- a/core/src/main/scala/filodb.core/query/RangeVector.scala
+++ b/core/src/main/scala/filodb.core/query/RangeVector.scala
@@ -167,10 +167,12 @@ object SerializableRangeVector extends StrictLogging {
    * The most efficient option when you need to create multiple SRVs as the containers are automatically
    * shared correctly.
    * The containers are sent whole as most likely more than one would be sent, so they should mostly be packed.
+   * limit - to limit the rows added to the result
    */
   def apply(rv: RangeVector,
             builder: RecordBuilder,
-            schema: RecordSchema): SerializableRangeVector = {
+            schema: RecordSchema,
+            limit: Int = Int.MaxValue): SerializableRangeVector = {
     var numRows = 0
     val oldContainerOpt = builder.currentContainer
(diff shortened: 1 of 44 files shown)


core/src/main/scala/filodb.core/query/RangeVector.scala
```diff
@@ -179,7 +181,7 @@ object SerializableRangeVector extends StrictLogging {
     try {
       OffheapLFSortedIDMap.validateNoSharedLocks()
       val rows = rv.rows
-      while (rows.hasNext) {
+      while (numRows < limit && rows.hasNext) {
```
velvia: I think it's better to apply the limit not here in the SRV but in the MetadataExecPlan stuff, and I think it should be similar to the limit that Vish has in place.  In fact, why doesn't Vish's limits help here?

vishramachandran: @velvia  

1. timeseries queries need to have higher limits and metadata queries have lower limits.
2. Next, metadata queries should not throw an exception when limit is reached, whereas for timeseries data we should not need to throw exception.
3. Metadata queries just return one Range Vector, so limit has to be applied when constructing SRV.
3. The existing limit implementation for timeseries data can be slightly improved by taking remaining samples into the SRV factory and limiting the iteration there. The limit may happen sooner that current implementation.

core/src/main/scala/filodb.core/query/RangeVector.scala
```diff
@@ -167,22 +167,31 @@ object SerializableRangeVector extends StrictLogging {
...
       val rows = rv.rows
-      while (rows.hasNext) {
+      while (totalRows < limit && rows.hasNext) {
         numRows += 1
+        totalRows += 1
         builder.addFromReader(rows.next)
       }
+      if (failOnLimit && totalRows >= limit)
+        throw new ResponseTooLargeException(s"result size is more than $limit samples. " +
+          s"Try applying more filters or reduce time range.")
```
vishramachandran: Condition should be inside the loop.

vishramachandran: This should be ok since we have totalRows < limit check

core/src/main/scala/filodb.core/query/RangeVector.scala
```diff
@@ -167,22 +167,31 @@ object SerializableRangeVector extends StrictLogging {
...
    * The containers are sent whole as most likely more than one would be sent, so they should mostly be packed.
+   * limit - to limit the rows added to the result
    */
   def apply(rv: RangeVector,
             builder: RecordBuilder,
-            schema: RecordSchema): SerializableRangeVector = {
+            schema: RecordSchema,
+            resultSize: Int = 0,
+            limit: Int = Int.MaxValue,
+            failOnLimit: Boolean = true): SerializableRangeVector = {
```
velvia: Why does failOnLimit default to true?  From the description of the PR, it sounds like we don't want to fail on limit, just to truncate.

sherali42:
Closing this PR as I have raised another PR for this change #393

PR #229, opened 2019-02-12
feat(standalone): Include a simple profiler.

broneill:
**Pull Request checklist**

This PR defines a simple profiler, which runs as a background task and writes results periodically to a file. This makes it easier to run perf tests without having to set up remote connections from a profiling client.

### DIFF
--- a/standalone/src/main/java/filodb/standalone/SimpleProfiler.java
+++ b/standalone/src/main/java/filodb/standalone/SimpleProfiler.java
@@ -1,0 +1,468 @@
+package filodb.standalone;
+
+import java.io.FileOutputStream;
+import java.io.InterruptedIOException;
+import java.io.IOException;
+import java.io.OutputStream;
+
+import java.lang.management.ManagementFactory;
+import java.lang.management.ThreadInfo;
+import java.lang.management.ThreadMXBean;
+
+import java.lang.reflect.Method;
+
+import java.time.Instant;
+
+import java.nio.charset.StandardCharsets;
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Simple profiler which samples threads and periodically logs a report to a file. When the
+ * process is cleanly shutdown, the profiler stops and reports what it has immediately. This
+ * makes it possible to use a very long report interval without data loss.
+ */
+public class SimpleProfiler {
+    /**
+     * Launches a global profiler, based on system properties.
+     *
+     * -Dfilodb.SimpleProfiler.sampleRateMillis=<n>
+     * -Dfilodb.SimpleProfiler.reportIntervalSeconds=<n>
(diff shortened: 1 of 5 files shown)


standalone/src/main/java/filodb/standalone/SimpleProfiler.java
```diff
@@ -0,0 +1,468 @@
...
+     * -Dfilodb.SimpleProfiler.outFile=<filename>
+     *
+     * In order for profiling to be enabled, all of the above properies must be set.
+     *
+     * @throws IOException if file cannot be opened
+     */
+    public static boolean launch() throws IOException {
+        String prefix = "filodb.SimpleProfiler.";
+
+        int sampleRateMillis = getIntProperty(prefix, "sampleRateMillis");
```
vishramachandran: I suggest we use typesafe config instead of system properties here.
This way, we can enable this for specific environments.
Modifying command line arguments per environment may not work for everyone.

standalone/src/main/java/filodb/standalone/SimpleProfiler.java
```diff
@@ -0,0 +1,468 @@
...
+        if (reportIntervalSeconds <= 0) {
+            return false;
+        }
+
+        int topCount = getIntProperty(prefix, "topCount");
+        if (topCount <= 0) {
+            return false;
+        }
+
+        String outFile = System.getProperty(prefix + "outFile");
```
vishramachandran: Should we write to a tmp file and print out the path in logs?
Otherwise same file will be overwritten and this may not work in local environments with multiple servers.

Especially true if and when we use same server typesafe config for both servers.

standalone/src/main/java/filodb/standalone/SimpleProfiler.java
broneill: What do you suggest specifically? I don't want this profiler to turn into something much bigger, since it's just for debugging. I wouldn't expect anyone to run this in production.

standalone/src/main/java/filodb/standalone/SimpleProfiler.java
broneill: Temp file should work fine if the file isn't specified.

standalone/src/main/java/filodb/standalone/SimpleProfiler.java
vishramachandran: You could move the launch code into the class `FilodbCluster` and drive the start using Typesafe config that can be loaded into `FilodbSettings`.

Agree - this is not for prod. It could be enabled during load testing only.

broneill:
Please review again. I switched to typesafe config, and I added a little section to the README file.

vishramachandran:
We need to watch out for file size growth in long running instances. May be ok for now.

PR #234, opened 2019-02-14
fix(query): Limit chunked windowing range to ChunkSetInfo.numRows

velvia:
Sometimes timestamp vector can be longer than official chunk length.  Limit range of chunked window queries.

**Pull Request checklist**

binarySearches based on timestamp vector might return an upper bound that is out of range of other vectors if their lengths are not equal.   (The lengths not being equal is an artifact of how ingestion is capped by write buffers).

**New behavior :**

Limit the upper bound to the official length of a chunk based on ChunkSetInfo, not the length of the timestamp vector.


vishramachandran:
@velvia Can you apply same fix for downsampling logic too ? 
Approved after same fixes are applied to downsampling.

vishramachandran:
@velvia downsampling fixes need to go into separate PR for us to enable cherry-pick into integration branch.

PR #238, opened 2019-02-14
bug(query): fix bug in aggregate over time functions for NaN

TanviBhavsar:
**Pull Request checklist**

**Current behavior :** 
sum_over_time, avg_over_time and count_over_time return default values as 0 when there is no sample in the window.
min_over_time returns Double.MaxValue and max_over_time returns Double.MinValue when there is no sample in the window

**New behavior :**
Aggregate over time functions return NaN when no sample is present

Before:
image

After:
image


memory/src/main/scala/filodb.memory/format/vectors/LongBinaryVector.scala
```diff
@@ -195,7 +195,7 @@ object LongVectorDataReader64 extends LongVectorDataReader {
     require(start >= 0 && end < length(vector), s"($start, $end) is out of bounds, length=${length(vector)}")
     var addr = vector + OffsetData + start * 8
     val untilAddr = vector + OffsetData + end * 8 + 8   // one past the end
-    var sum: Double = 0d
+    var sum: Double = Double.NaN
```
velvia: Are you sure this is right?  There is no check for NaN when it adds the actual values.  I think this will lead to the sum always being NaN, which is not right.

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
```diff
@@ -41,7 +41,12 @@ class MinOverTimeChunkedFunctionD(var min: Double = Double.MaxValue) extends Chu
     val it = doubleReader.iterate(doubleVect, startRowNum)
     while (rowNum <= endRowNum) {
       val nextVal = it.next
-      if (!JLDouble.isNaN(nextVal)) min = Math.min(min, nextVal)  // cannot compare NaN, always < anything else
+      if (!JLDouble.isNaN(nextVal)) {
+        if (min.isNaN) {
+          min = Double.MaxValue
```
velvia: Slightly faster: instead of setting Double.MaxValue just set it to the nextVal because that becomes the max.  IE:

    min = if (min.isNan) nextval else Math.min(min, nextval)

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
```diff
@@ -78,7 +83,12 @@ class MaxOverTimeChunkedFunctionD(var max: Double = Double.MinValue) extends Chu
     val it = doubleReader.iterate(doubleVect, startRowNum)
     while (rowNum <= endRowNum) {
       val nextVal = it.next
-      if (!JLDouble.isNaN(nextVal)) max = Math.max(max, nextVal)  // cannot compare NaN, always < anything else
+      if (!JLDouble.isNaN(nextVal)) {
+        if (max.isNaN) {
```
velvia: Same here, can use the above if else to simplify and make it faster

query/src/test/scala/filodb/query/exec/WindowIteratorSpec.scala
```diff
@@ -330,4 +324,138 @@ class WindowIteratorSpec extends RawDataWindowingSpec {
         .asInstanceOf[ChunkedRangeFunction], queryConfig)()
     chunkedWinIt.map(r => (r.getLong(0), r.getDouble(1))).toList.filter(!_._2.isNaN) shouldEqual windowResults
   }
+
+  it("should calculate AvgOverTime correctly") {
```
velvia: how about "correctly even for windows with no values or NaN only"?  just a bit more info to make case precise

broneill:
Looks like a test fails, but it appears to pass each time without these changes:
LongVectorTest:
[info] - should not allow sum() with out of bound indices *** FAILED ***
[info]   NaN did not equal 2.1702486843044E13 (LongVectorTest.scala:129)

TanviBhavsar:
> @TanviBhavsar good change, just two things:
> 
> * make sure to look over LongBinaryVector.sum(), it does not look correct to start with NaN and not correct for it later
> * Something to think about for the chunked range functions.  I think it might be better, instead of setting the BinaryVector sum methods to NaN at the beginning then checking, to check in the ChunkedRangeFunction.addChunks methods if there is any data in the chunk, and just return NaN if there is no data.  That way you don't have to check NaN with every element.



> @TanviBhavsar good change, just two things:
> 
> * make sure to look over LongBinaryVector.sum(), it does not look correct to start with NaN and not correct for it later
> * Something to think about for the chunked range functions.  I think it might be better, instead of setting the BinaryVector sum methods to NaN at the beginning then checking, to check in the ChunkedRangeFunction.addChunks methods if there is any data in the chunk, and just return NaN if there is no data.  That way you don't have to check NaN with every element.

memory/src/main/scala/filodb.memory/format/vectors/LongBinaryVector.scala
TanviBhavsar: Sorry I missed this

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
TanviBhavsar: Good suggestion! Made the change

query/src/test/scala/filodb/query/exec/WindowIteratorSpec.scala
TanviBhavsar: Done

TanviBhavsar:
> Looks like a test fails, but it appears to pass each time without these changes:
> LongVectorTest:
> [info] - should not allow sum() with out of bound indices *** FAILED ***
> [info] NaN did not equal 2.1702486843044E13 (LongVectorTest.scala:129)

@broneill  I fixed it. However build is still failing :                                                               
[error] (coordinator/test:test) sbt.TestsFailedException: Tests unsuccessful

broneill:
Coordinator test failure is a separate issue, not caused by these changes.

velvia:
@TanviBhavsar so now that you have the changes to check if sum is NaN at beginning, I think you can revert the changes to DoubleVector and LongBinaryVector sum methods right?  They shouldn't be needed anymore

TanviBhavsar:
> @TanviBhavsar thanks, LGTM. Wait for the signal to merge....

Sure! Thanks!

PR #248, opened 2019-02-21
feat(memory): Support reentrant shared lock when exclusive lock is re…

broneill:
…quested and waiting.

**Pull Request checklist**

**Current behavior :**
When a thread attempts to acquire a shared lock it already owns, but an exclusive lock is waiting, the shared lock request is delayed until the exclusive locker waiter times out and retries.

**New behavior :**
A reentrant shared lock request is granted immediately.


broneill:
If a thread owns the shared lock, and then another thread requests an exclusive lock, this creates a stall in the queue preventing other threads from obtain locks of any kind. If a thread would be stalled because of this, but it already owns a shared lock, then it can barge ahead harmlessly.

PR #249, opened 2019-02-22
bug(memory, query): Window iteration mustn't release the shared chunk…

broneill:
… map lock too soon.

**Pull Request checklist**

**Current behavior :**
Window chunk iteration releases the chunk map look too soon, leading to observed data corruption.

**New behavior :**
Shared lock is retained until after the chunk data is safely copied.


core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
```diff
@@ -308,6 +308,9 @@ extends ChunkMap(memFactory, initMapSize) with ReadablePartition {
       }
       return info
     }
+
+    def lock(): Unit = chunkmapAcquireShared()
```
velvia: final def?

query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
```diff
@@ -99,24 +99,27 @@ final case class PeriodicSamplesMapper(start: Long,
...
                             step: Long,
                             end: Long,
                             window: Long,
                             rangeFunction: ChunkedRangeFunction,
-                            queryConfig: QueryConfig)
-                           (windowIt: WindowedChunkIterator =
-                              new WindowedChunkIterator(rv.chunkInfos(start - window, end), start, step, end, window)
-                           ) extends Iterator[TransientRow] with StrictLogging {
+                            queryConfig: QueryConfig) extends Iterator[TransientRow] with StrictLogging {
+  private var windowIt: WindowedChunkIterator = null
```
velvia: This is actually a case where we should just use lazy val here.  Then there is no need to use null and turn off the scalastyle.  I think it would be safer code too, and prevent others from accidentally changing the value (which all mutable variables have a risk of).

    lazy val windowIt = {
      val it = new WindowedChunkIterator(rv.chunkInfos(start - window, end), start, step, end, window)
      it.lock()
      it
    }

broneill: Ok.

core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
broneill: Ok.

query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
```diff
@@ -99,68 +99,64 @@ final case class PeriodicSamplesMapper(start: Long,
...
 
   override def next: TransientRow = {
     rangeFunction.reset()
     // TODO: detect if rangeFunction needs items completely sorted.  For example, it is possible
     // to do rate if only each chunk is sorted.  Also check for counter correction here
 
-    val windowIt = openWindowIt()
+    // Lazy variables have an extra lookup cost, due to a volatile bitmap field generated by
+    // the compiler. Copy to a local variable to reduce some overhead.
+    val wit = windowIt
```
velvia: Hmmm.  When you declare a val in a class, it is initialized in the constructor, thus rendering the original lazy val code useless.  Lazy Vals are fairly inexpensive, but if we are OK with initializing in the constructor, why not just use a val, then we don't need the extra 2 fields for the lazy val?  and I don't see why we wouldn't be OK with the initialization at constructor time (other than minor edge case where the windowed iterator is not used at all....)

broneill: I disassembled the code and verified that the internal iterator is not constructed until it's first used. The comment on the lazy val explains why it's done this way. The thread creating ChunkedWindowIterator isn't necessarily the thread which actually uses it.

PR #254, opened 2019-02-25
 bug(core): Add a lock to protect access to the activelyIngesting bitmap.

broneill:
**Pull Request checklist**

**Current behavior :**
Corruption of the actively ingesting bitmap has been observed, leading to permanent failures and stuck chunk map locks.

**New behavior :**
Guard access to the bitmap with a lock, and also ensure that stuck chunk map locks are released.


core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -761,7 +786,10 @@ class TimeSeriesShard(val dataset: Dataset,
        These keys are from earliest time bucket that are still ingesting */
       val earliestTimeBucket = cmd.timeBucket - numTimeBucketsToRetain
       if (earliestTimeBucket >= 0) {
-        val partIdsToRollOver = timeBucketBitmaps.get(earliestTimeBucket).and(activelyIngesting)
+        val bitmap = timeBucketBitmaps.get(earliestTimeBucket)
+        val stamp = activelyIngestingLock.readLock()
```
velvia: The read lock should be obtained before the `timeBucketBitmaps.get()`, which is the actual read.  The and operation below generates a new bitmap, it doesn't modify anything in place.  Furthermore, the read lock should cover down to line 793 where we do the `timeBucketBitmaps.get(cmd.timeBucket)` as that is also a read operation.

broneill: The call to timeBucketBitmaps.get() retrieves a pointer from a thread-safe map. No additional locking is required. The 'and' operation needs a read lock to ensure stability of the activelyIngesting bitmap being accessed. Line 793 is only accessing the thread-safe map and the locally accessible bitmap.

broneill:
My understanding is that the time bucket bitmaps are already isolated. The ingestion thread and flush thread(s) don't share these.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
velvia: The timeBucketBitmaps may be a thread safe map but the order of operations here is not atomic.  Couldn't you have a situation like this:

Flush task 1: timeBucketBitmaps.get()
Flush task 2: timeBucketBitmaps.get()
Flush task 1: timeBucketBitmaps.put() with modified map
Flush task 2: timeBucketBitmaps.put(). <-- here overwrites map modified by flush task 1

broneill: If this pattern is possible, then yes, all bitmaps must be protected.

vishramachandran:
@broneill We should also look at concurrent access of bitmaps held by the timeBucketBitmaps collection. I will look at this issue in more detail too.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
vishramachandran: @broneill  @velvia Ok - I now recall why this code is ok. Every flush task reads bits on the earliest time bucket bitmap and sets bits on the latest timeBucket, both of which are uniquely associated with the flush group. Since each flush group is associated with different latest and earliest time buckets, concurrent threads should not be reading or writing to same time bucket bitmaps, or even setting the same time bucket in the collection. This can in theory happen only if a flush task lasts more than the retention period (not possible).

I wish I had documented this clearly earlier in the code. 

@broneill if you could add these comments to the relevant lines, it would really help.

vishramachandran:
@broneill Repeating my comment on the timeBucketBitmap thread discussion...

@broneill  @velvia Ok - I now recall why the timeBucketBitmap mutation is ok. Every flush task reads bits on the earliest time bucket bitmap and sets bits on the latest timeBucket, both of which are uniquely associated with the flush group. Since each flush group is associated with different latest and earliest time buckets, concurrent threads should not be reading or writing to same time bucket bitmaps, or even setting the same time bucket in the collection. This can in theory happen only if a flush task lasts more than the retention period (not possible).

I wish I had documented this clearly earlier in the code. 

@broneill if you could add these comments to the relevant lines, it would really help. 

Should be ok to undo your a0ff390 commit.

broneill:
Reverted changes, comment added.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -666,12 +670,21 @@ class TimeSeriesShard(val dataset: Dataset,
     indexRb.endRecord(false)
   }
 
+  private def isActivelyIngesting(i: Integer): Boolean = {
```
vishramachandran: rename `i` to `partId`

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -666,12 +670,21 @@ class TimeSeriesShard(val dataset: Dataset,
...
   // scalastyle:off method.length
   private def doFlushSteps(flushGroup: FlushGroup,
                            partitionIt: Iterator[TimeSeriesPartition]): Task[Response] = {
     val tracer = Kamon.buildSpan("chunk-flush-task-latency-after-retries")
       .withTag("dataset", dataset.name)
       .withTag("shard", shardNum).start()
+
+    // Clean up lingering chunk map shared locks, left over from previous failures. The
+    // makeFlushChunks call below is what creates an iterator which holds a shared lock.
+    ChunkMap.validateNoSharedLocks()
```
vishramachandran: This call should happen on the thread that starts consuming the chunk iterator which happens in writeChunksFuture below. 

Whereas right now, it is running on the single threaded ingestion scheduler.

The lazy evaluation is quite tricky. I added some comments with "VERY IMPORANT". Please feel free to add more if necessary.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -824,6 +850,11 @@ class TimeSeriesShard(val dataset: Dataset,
         logger.error(s"Critical! Chunk persistence failed after retries and skipped in dataset=${dataset.ref} " +
           s"shard=$shardNum", e)
         shardStats.flushesFailedChunkWrite.increment
+
+        // Release the shared lock held by the chunk map iterator. There's no direct way to do
+        // this because the iterator has been wrapped and the close method cannot be accessed.
+        ChunkMap.releaseAllSharedLocks()
```
vishramachandran: This should be called in the writeChunksFuture and not writeTimeBucketsFuture.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
broneill: I've removed all this stuff for now. It should go into another commit.

vishramachandran:
LGTM now.

PR #256, opened 2019-02-26
misc: fix for On Demand Chunk page-in span

sherali42:
**Pull Request checklist**

Fix to accurately capture the latency for On Demand Paging of RawChunks. Span is created only if there are partitions that needs to be paged-in from datastore.

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/OnDemandPagingShard.scala
+++ b/core/src/main/scala/filodb.core/memstore/OnDemandPagingShard.scala
@@ -92,7 +92,7 @@ TimeSeriesShard(dataset, storeConfig, shardNum, rawStore, metastore, evictionPol
               // NOTE: this executes the partMaker single threaded.  Needed for now due to concurrency constraints.
               // In the future optimize this if needed.
               .mapAsync { rawPart => partitionMaker.populateRawChunks(rawPart).executeOn(singleThreadPool) }
-              .doOnComplete(span.finish())
+              .doOnComplete(() => span.finish())
               // This is needed so future computations happen in a different thread
               .asyncBoundary(strategy)
           } else { Observable.empty }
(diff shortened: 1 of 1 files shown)


broneill:
The change looks scarier than it actually is, but I don't understand it. Can you summarize how the change works?

sherali42:
> The change looks scarier than it actually is, but I don't understand it. Can you summarize how the change works?

- Span creation is moved inside the conditional block, so that it gets created only if there are chunks that needs to be paged in. 
- Span is closed on completion of the task as it is an observable. `.doOnComplete(() => span.finish())`
- Also moved `Observable.fromIterable(inMemoryPartitions) ++` outside, as we are doing this in both `if` and `else`

Earlier code was measuring the latency regardless of if we are actually ODPing.

core/src/main/scala/filodb.core/memstore/OnDemandPagingShard.scala
```diff
@@ -73,42 +73,48 @@ TimeSeriesShard(dataset, storeConfig, shardNum, rawStore, metastore, evictionPol
...
-    } else {
-      Observable.fromIterable(inMemoryPartitions) ++
-      Observable.fromTask(odpPartTask(indexIt, partKeyBytesToPage, methods, chunkMethod)).flatMap { odpParts =>
-        shardStats.partitionsQueried.increment(partKeyBytesToPage.length)
-        Observable.fromIterable(partKeyBytesToPage.zip(methods))
-          .mapAsync(storeConfig.demandPagingParallelism) { case (partBytes, method) =>
-            rawStore.readRawPartitions(dataset, allDataCols, SinglePartitionScan(partBytes, shardNum), method)
+    Observable.fromIterable(inMemoryPartitions) ++ {
+      if (storeConfig.multiPartitionODP) {
+        Observable.fromTask(odpPartTask(indexIt, partKeyBytesToPage, methods, chunkMethod)).flatMap { odpParts =>
```
vishramachandran: We also need a span in createODPPartitionsTask method to measure how much time it took to create the partitions in memory.

PR #259, opened 2019-02-27
bug(core): currentIndexTimeBucket should be atomically read and modified

vishramachandran:
**Pull Request checklist**

currentIndexTimeBucket is currently an int that is mutated in one thread, but can be read from another thread. Better to make it an AtomicInteger so the read happens safely.


vishramachandran:
ok. Let me double check volatile usage in scala.

PR #265, opened 2019-03-05
feat(memory, core): Row based, NibblePacked HistogramColumn

velvia:
**Pull Request checklist**

The v1 HistogramColumn was essentially a set of LongBinaryVectors each of which was compressed individually.  It had a number of problems, the biggest being that the base HistogramColumn blob contained pointers to the individual bucket vectors, and there was no easy way to persist and recover those pointers, and that problem was just punted.

Furthermore, the data had to be ingested in an increasing fashion, ie the user was responsible to make sure bucket values are increasing Prometheus-style.

**New behavior :**

This change consists of v2 of the histogram format: both the over-the-wire format as well as the write buffers and compressed vectors relies on histogram compression using the NibblePacking algorithm, with delta encoding of increasing values to save space.

- Histogram values are nibblepack-delta-compressed in the BinaryRecord and sent over the wire, saving significant space compared to the previously uncompressed 64-bit bucket values
- The write buffer now simply stores the compressed histograms as is, in a single contiguous buffer.  This also makes the chunks safe for persistence, with no pointer magic.
- Write buffers for HistogramColumns are now capped by size rather than # of elements.  The max size is configurable in the StoreConfig of each dataset.
- The encoded HistogramColumn chunk is simply the same as the write buffers.  This eliminates one of the biggest bottlenecks in ingestion (encoding).

#### Performance results

Performance and size are measured using 64-bucket histograms with real data and the `HistogramCompressor`, `HistogramIngestBenchmark` and `HistogramQueryBenchmark` apps.

* Ingestion rate: 25x faster than Prometheus schema and 10x faster than Histograms v1
* Over-the-wire/Kafka record size:  1/3rd size of Histograms v1 and 1/50th size of Prom schema
* Write buffer size: less than 1/10th of Histograms v1, 1/30th of Prom schema
* Compressed chunk size:   1/6th size of Histograms v1 (for busy/active histograms), 1/3rd for idle histograms
* Query speed: more than 100x faster than Prom-schema histogram quantiles

#### BREAKING CHANGES

The new binary histogram format is not compatible with the old one and has a different format code.  While there is code to reject incompatible histogram formats, it is best to wipe clean all data including Kafka partitions.


broneill:
This is a really huge PR, which also has appears to have independent changes. Is it possible that this can be broken down into smaller PRs?

velvia:
@broneill no the list of commits is huge because I merged from develop.  I'd have to rebase.  I can make it smaller by making a separate PR for the NibblePacking stuff itself.  Let me do that.

vishramachandran:
@velvia can you rebase from develop now that you have merged in the NibblePack code ?

velvia:
@vishramachandran please review https://github.com/filodb/FiloDB/pull/267 next.  After that is merged, then I will merge everything into this PR.  BTW rebase is probably not possible due to too many changes.

velvia:
@vishramachandran @broneill now this is ready to be reviewed with a smaller set of changes since NibblePacking and other PRs.  Thanks

PR #266, opened 2019-03-06
fix(coordinator): Use random Akka port in some tests to reduce failures

velvia:
**Pull Request checklist**

We use a long timeout to reduce failures between tests that run Akka Cluster on a port.

**New behavior :**

Randomize the port used by Akka Cluster in key tests.  Timeout should not be needed anymore.


broneill:
I've never found the trick of opening and closing a server socket to be very reliable. It always seems that there's some delay before the port becomes available. Ideally, you want to pass the ServerSocket instance you created into the server itself.

velvia:
@vishramachandran I tried using ServerSocket and ditched that approach here, because it kept erring out unexpectedly.  I don't believe it is reliable.

PR #268, opened 2019-03-06
 feat(query): Handling spread changes over time in Query Engine - Iteration 1

vishramachandran:
**Pull Request checklist**

**Current behavior :** 
If spread for a namespace/application changes over time, and query spans the time when the spread changed, we get two range vectors with same key that could possibility overlap in time depending on how ingestion is done. This can cause incorrect query results, affects join functionality and aggregation, time window result calculation etc.

**New Expectations from Gateway**
1. Require that spread changes be scheduled by the gateway for a specific time in the future to prevent overlap of the same time series on two shards.
2. Require that gateway ingest NaN into old shard to prevent staleness feature that "holds" a value for five minutes and can produce wrong aggregations.

**New behavior :**
1. spreadFunc in QueryOptions now receives all changes in spread as opposed to the current value
2. Query Planner schedules the RV stitching prior to a join, and finally just before sending out final result. We return an NaN instead of a wrong result if overlap is seen during stitching.

**Known Limitations in this First Iteration Solution**
Time windows aka lookback windows in queries may now span multiple shards due to spread change. This involves a more detailed solution and is out of the scope of this PR. As a first solution, we rely on spread change to be scheduled at the 0th minute of the hour so that the time window spanning across shards are minimized. 

We will progressively fix these limitations in subsequent iterations. 

**BREAKING CHANGES**

Logical Plan and Query Engine changes require redeployment of components using them.

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
@@ -23,10 +23,12 @@ import filodb.query.exec._
 
 
 /**
-  * This is a container for any state to be passed up to parent nodes for easier query planning.
+  * Plan Result includes the result exec plan(s) along with any state to be passed up the
+  * plan building call tree during query planning.
+  *
   * Not for runtime use.
   */
-case class PlanNotes(var stitch: Boolean = false)
+case class PlanResult(exec: Seq[ExecPlan], stitch: Boolean = false)
 
 /**
   * FiloDB Query Engine is the facade for execution of FiloDB queries.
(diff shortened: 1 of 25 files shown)


broneill:
Can you separate out the unrelated fixes into another PR?

query/src/main/scala/filodb/query/exec/RowKeyRange.scala
```diff
@@ -0,0 +1,12 @@
+package filodb.query.exec
+
+import filodb.core.binaryrecord.BinaryRecord
+
+sealed trait RowKeyRange
```
broneill: What is the point of this? It just seems like a comment to me.

query/src/main/scala/filodb/query/exec/StitchRvsExec.scala
```diff
@@ -0,0 +1,108 @@
...
+/**
+  * Range Vector Transformer version of StitchRvsExec
+  */
+final case class StitchRvsMapper() extends RangeVectorTransformer {
+
+  def apply(source: Observable[RangeVector],
+            queryConfig: QueryConfig,
+            limit: Int,
+            sourceSchema: ResultSchema): Observable[RangeVector] = {
+    qLogger.debug(s"StitchRvsMapper: Stitching results:")
```
broneill: spelling

query/src/main/scala/filodb/query/exec/StitchRvsExec.scala
```diff
@@ -0,0 +1,108 @@
...
+                               children: Seq[ExecPlan]) extends NonLeafExecPlan {
+  require(children.nonEmpty)
+
+  protected def args: String = ""
+
+  protected def schemaOfCompose(dataset: Dataset): ResultSchema = children.head.schema(dataset)
+
+  protected def compose(childResponses: Observable[(QueryResponse, Int)],
+                        queryConfig: QueryConfig): Observable[RangeVector] = {
+    qLogger.debug(s"StitchRvsExec: Stitching results:")
```
broneill: Spelling

coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
```diff
@@ -21,6 +21,13 @@ import filodb.prometheus.ast.Vectors.PromMetricLabel
 import filodb.query.{exec, _}
 import filodb.query.exec._
 
+
+/**
+  * This is a container for any state to be passed up to parent nodes for easier query planning.
```
broneill: Ew. This design needs a bit more work if it's expected to evolve into a general purpose collection of "hints" and "backdoor" state propagation.

vishramachandran: Introduced PlanResult.

PR #269, opened 2019-03-06
feat(memory): Implementation of NibblePacking compression/storage scheme

velvia:
**Pull Request checklist**

Just the NibblePack.scala by itself for easy reviewing.

### DIFF
--- a/memory/src/main/scala/filodb.memory/format/NibblePack.scala
+++ b/memory/src/main/scala/filodb.memory/format/NibblePack.scala
@@ -1,0 +1,283 @@
+package filodb.memory.format
+
+import java.nio.ByteOrder.LITTLE_ENDIAN
+
+import org.agrona.{DirectBuffer, MutableDirectBuffer}
+import scalaxy.loops._
+
+/**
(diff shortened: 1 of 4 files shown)


memory/src/main/scala/filodb.memory/format/NibblePack.scala
```diff
@@ -0,0 +1,283 @@
...
+
+  trait Sink {
+    def process(data: Long): Unit
+  }
+
+  final case class DeltaSink(outArray: Array[Long]) extends Sink {
+    private var current: Long = 0L
+    private var pos: Int = 0
+    def process(data: Long): Unit = {
+      if (pos < outArray.size) {
```
broneill: Why check and throw data away? Just remove the check and have an exception be thrown instead?

velvia: I remember why this is necessary now.  It's because NibblePack always works in multiples of 8.  So, if a user packed 12 values, they will pass in an output array of 12 values and expect it to work, but in reality NibblePack demands 16.   I will add a comment.

velvia:
@broneill added random tests, found and fixed a bug.  :)

PR #271, opened 2019-03-07
bug(core): Disable flush failure recovery step which has race conditions

broneill:
**Pull Request checklist**

**Current behavior :**
Potential race condition when trying to recover after failed flushes.

**New behavior :**
Don't bother attempting to recover lost flush attempts. We already tried a few times, and we favor availability over durability in cases like this.

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
+++ b/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
@@ -850,9 +850,19 @@ class TimeSeriesShard(val dataset: Dataset,
           s"shard=$shardNum", e)
         shardStats.flushesFailedChunkWrite.increment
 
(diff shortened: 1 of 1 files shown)


broneill:
As the code is implemented today (based on my understanding), if a failure is reported to this block of code, iteration isn't stopped. This means that all remaining buffers will get encoded as chunks. If the line of code I commented out was restored, and a long sleep placed in front of it, the normal iteration would finish and the backdoor iteration would see nothing. Hence commenting it out has the same effect.

If we want to abort all remaining writes, then the iterator needs to have a lock, and the call back into the partition to update the appenders list also needs to be made thread safe.

To me, this extra work seems unnecessary since we're already accepting the possibility that chunks will be lost forever when the Cassandra write fails. This extra recovery step is just added complexity.

PR #275, opened 2019-03-11
feat/fix(core,query): Remove binaryrecord v1 and associated SEGVs

velvia:
**Pull Request checklist**

BinaryRecord v1 was still used in a few places, namely in selecting the time range (called "RowKeyRange") during queries.  However, today we only support time ranges, not any other kind of row keys, in queries.  The extra complexity of using BinaryRecord v1's was designed to support any kind of query range, not just time, and it is not needed today.  It also led to some SEGVs during performance testing.

**New behavior :**

filodb.core.binaryrecord (v1) is completely removed.   
`RowKeyChunkScan` and `RowKeyRange` are now replaced by `TimeRangeChunkScan` which just takes Long timestamps.
Dependency of binaryrecord is removed everywhere.

**BREAKING CHANGES**

Since the leaf ExecPlans now take `TimeRangeChunkScan` (or rather ChunKScanMethod) rather than `RowKeyRange`, and ExecPlans transmit over the wire, this change will require updates on all Akka clients talking to FiloDB.


coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
```diff
@@ -344,14 +344,13 @@ class QueryEngine(dataset: Dataset,
...
     rangeSelector match {
-      case IntervalSelector(from, to) => RowKeyInterval(BinaryRecord(dataset, Seq(from)),
-                                                        BinaryRecord(dataset, Seq(to)))
-      case AllChunksSelector          => AllChunks
-      case EncodedChunksSelector      => EncodedChunks
-      case WriteBufferSelector        => WriteBuffers
-      case InMemoryChunksSelector     => InMemoryChunks
+      case IntervalSelector(from, to) => TimeRangeChunkScan(from, to)
+      case AllChunksSelector          => AllChunkScan
+      case EncodedChunksSelector      => ???
```
broneill: Why leave this line here if not supported?

velvia: It's for future support.  We'd like to be able to query just the encoded chunks for debugging.

Also re spark.  No it's not used.  I don't want to get rid of it as surely we will have to support some kind of ad hoc / Spark based querying of long term data, so it could still be useful in the future.

We should think about getting rid of the stress module though.  Or rewriting stress tests for current environment.

PR #276, opened 2019-03-12
feat(query): Include partId in query result on verbose option

vishramachandran:
**Pull Request checklist**

**Current behavior :** 

Only shard number is included in result range vector

**New behavior :**

if "verbose" option is enabled as query parameter 
* partId is included in range vector result
* partId and shardNum is included

### DIFF
--- a/prometheus/src/main/scala/filodb/prometheus/query/PrometheusModel.scala
+++ b/prometheus/src/main/scala/filodb/prometheus/query/PrometheusModel.scala
@@ -76,8 +76,8 @@ object PrometheusModel {
     b.build()
   }
 
-  def toPromSuccessResponse(qr: filodb.query.QueryResult): SuccessResponse = {
-    SuccessResponse(Data(toPromResultType(qr.resultType), qr.result.map(toPromResult(_))))
+  def toPromSuccessResponse(qr: filodb.query.QueryResult, verbose: Boolean): SuccessResponse = {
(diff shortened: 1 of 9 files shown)


prometheus/src/main/scala/filodb/prometheus/query/PrometheusModel.scala
```diff
@@ -91,8 +91,12 @@ object PrometheusModel {
   /**
     * Used to send out HTTP response
     */
-  def toPromResult(srv: SerializableRangeVector): Result = {
-    Result(srv.key.labelValues.map { case (k, v) => (k.toString, v.toString)},
+  def toPromResult(srv: SerializableRangeVector, verbose: Boolean): Result = {
+    val tags = srv.key.labelValues.map { case (k, v) => (k.toString, v.toString)} ++
+                (if (verbose) Map("shards" -> srv.key.sourceShards.toString(), "partIds" -> srv.key.partIds.toString())
```
tjackpaul: Should we follow the reserved keyword label syntax.

`_shards_`
`_partids_`

This way, it will standout clearly and will not conflict with any other user labels.

vishramachandran: Agreed. Fixed.

vishramachandran:
Proceeding to merge since earlier build passed, and the delta is just the tag name.

PR #278, opened 2019-03-13
feat(query): histogram_bucket function

velvia:
**Pull Request checklist**

**New behavior :**

Implements a `histogram_bucket` instant function to extract one histogram bucket from a HistogramColumn.

### DIFF
--- a/query/src/main/scala/filodb/query/exec/rangefn/InstantFunction.scala
+++ b/query/src/main/scala/filodb/query/exec/rangefn/InstantFunction.scala
@@ -265,7 +265,7 @@ case class HistogramBucketImpl(funcParams: Seq[Any]) extends HistToDoubleIFuncti
 
   final def apply(value: Histogram): Double = {
     for { b <- 0 until value.numBuckets optimized } {
-      if (value.bucketTop(b) == bucket) return value.bucketValue(b)
+      if (Math.abs(value.bucketTop(b) - bucket) <= 0.00001) return value.bucketValue(b)
     }
     Double.NaN
   }
--- a/query/src/main/scala/filodb/query/exec/rangefn/InstantFunction.scala
+++ b/query/src/main/scala/filodb/query/exec/rangefn/InstantFunction.scala
@@ -1,5 +1,7 @@
 package filodb.query.exec.rangefn
 
+import scalaxy.loops._
+
 import filodb.memory.format.vectors.Histogram
 import filodb.query.InstantFunctionId
 import filodb.query.InstantFunctionId.{Log2, Sqrt, _}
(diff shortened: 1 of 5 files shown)


query/src/main/scala/filodb/query/exec/rangefn/InstantFunction.scala
```diff
@@ -248,4 +251,22 @@ case class HistogramQuantileImpl(funcParams: Seq[Any]) extends HistToDoubleIFunc
...
+ *                     to any existing bucket then NaN is returned.
+ */
+case class HistogramBucketImpl(funcParams: Seq[Any]) extends HistToDoubleIFunction {
+  require(funcParams.length == 1, "Bucket/le required for histogram bucket")
+  require(funcParams(0).isInstanceOf[Number], "histogram_bucket parameter must be a number")
+  val bucket = funcParams(0).asInstanceOf[Number].doubleValue()
+
+  final def apply(value: Histogram): Double = {
+    for { b <- 0 until value.numBuckets optimized } {
+      if (value.bucketTop(b) == bucket) return value.bucketValue(b)
```
vishramachandran: Comparing doubles is tricky. This may or may not work. You will need to use a different strategy to identify the bucket.

velvia: When would this not work?

vishramachandran: For some detail 
https://howtodoinjava.com/java/basics/correctly-compare-float-double/

velvia: @vishramachandran fixed.

query/src/main/scala/filodb/query/exec/rangefn/InstantFunction.scala
```diff
@@ -248,4 +251,22 @@ case class HistogramQuantileImpl(funcParams: Seq[Any]) extends HistToDoubleIFunc
...
+ *                     to any existing bucket then NaN is returned.
+ */
+case class HistogramBucketImpl(funcParams: Seq[Any]) extends HistToDoubleIFunction {
+  require(funcParams.length == 1, "Bucket/le required for histogram bucket")
+  require(funcParams(0).isInstanceOf[Number], "histogram_bucket parameter must be a number")
+  val bucket = funcParams(0).asInstanceOf[Number].doubleValue()
+
+  final def apply(value: Histogram): Double = {
+    for { b <- 0 until value.numBuckets optimized } {
+      if (Math.abs(value.bucketTop(b) - bucket) <= 0.00001) return value.bucketValue(b)
```
vishramachandran: Would 0.00001 as threshold be enough ?
If buckets are 1ms, 2ms, 4ms etc and all the bucket limits are in seconds, multiple buckets may match.
10 ^ -10 may be safer.

velvia: Sure we can use that.

velvia: @vishramachandran ok changed to 1E-10

PR #279, opened 2019-03-13
feat(gateway): Add --gen-prom-data and --gen-hist-data to dev gateway; multiple schema support

velvia:
**Pull Request checklist**

Users use `TestTimeseriesProducer` to generate Prometheus-schema data to Kafka for testing.  No support for generating data for other schemas.

**New behavior :**

Users now will use `./dev-gateway.sh` and the `GatewayServer` to generate test data.   
Prometheus format data can be generated using `--gen-prom-data`.
Histogram schema and source files are included and test data can be generated using `--gen-hist-data`.
The `TestTimeseriesProducer` is deprecated as an app for generating test data.
Furthermore, both the `GatewayServer` and the `TestConsumer` both load dataset definitions from the MetaStore, they are no longer hard wired to the Prom dataset.


http/src/test/scala/filodb/http/ClusterApiRouteSpec.scala
```diff
@@ -83,21 +81,9 @@ class ClusterApiRouteSpec extends FunSpec with ScalatestRouteTest with AsyncTest
...
-        println(s"Current statuses = $statuses")
-        info(s"Current statuses = $statuses")
-        if (statuses.exists(_ == ShardStatusError)) {
-          info(s"ERROR in status, breaking")
-          throw new RuntimeException(s"Got error in statuses $statuses")
-        }
-      } while (statuses.take(2) != Seq(ShardStatusActive, ShardStatusActive))
+
+      // Give the coordinator nodes some time to get started
+      Thread sleep 1000
```
broneill: This seems much simpler than what was done before, but is this sleep long enough? Can both strategies be used?

velvia: The problem is that the wait loop is not really reliable, and in any case we don't really care about the exact status in this test.  We just want to check that the results are properly formed in HTTP.

dev-gateway.sh
```diff
@@ -1,2 +1,8 @@
 #!/usr/bin/env bash
-java -cp gateway/target/scala-2.11/gateway-* filodb.gateway.GatewayServer conf/timeseries-dev-source.conf &
\ No newline at end of file
+#
+# Starts a local Gateway for ingesting data into FiloDB (run with no options)
+# Type --help to see options - options include generating random test data and exiting.
+args=${@:-"conf/timeseries-dev-source.conf"}
+java -Dconfig.file=conf/timeseries-filodb-server.conf  \
+     -Dkamon.prometheus.embedded-server.port=9097  \
```
tjackpaul: Why do we need an embedded prometheus server in gateway ?

gateway/src/main/scala/filodb/gateway/GatewayServer.scala
```diff
@@ -52,21 +69,29 @@ object GatewayServer extends StrictLogging {
...
   val containersSize = Kamon.histogram("containers-size-bytes")
 
+  // Most options are for generating test data
+  class GatewayOptions(args: Seq[String]) extends ScallopConf(args) {
+    val samplesPerSeries = opt[Int](short = 'n', default = Some(100),
+                                    descr="# of samples per time series")
+    val numSeries = opt[Int](short='p', default = Some(20), descr="# of total time series")
+    val sourceConfigPath = trailArg[String](descr="Path to source config, eg conf/timeseries-dev-source.conf")
+    val genHistData = toggle(noshort=true, descrYes="Generate histogram-schema test data and exit")
+    val genPromData = toggle(noshort=true, descrYes="Generate Prometheus-schema test data and exit")
```
vishramachandran: I think the options should take some config file such as `--genData some-config-file-with-schema` that has details about the schema of data to send. This could be one or multiple entries (in the future when we have multiple schemas per dataset).

Having a dataset definition config file is also useful for dataset creation purposes. We dont need to remember commands, instead create dataset by pointing to a definition file.

Data generation for two types of data columns (double, hist) would be supported, both of which we have now.

Test data generation based on dataset definition also has other benefits where we can test other schemas easily without much setup.

We may need a quick and dirty partKey generator too.

velvia: I have thought about configurable schema generation.  That would be a good feature, I'd like to save it for a separate PR though and not blow up the scope of this.

velvia:
@vishramachandran will do.

PR #280, opened 2019-03-14
feat(core): Add key=value optional parameters to each Column definition

velvia:
**Pull Request checklist**

Currently, the only way to distinguish different columns besides the name is the columnType attribute.  There is no way to annotate, for example, if a DoubleColumn is a counter or a gauge.

**New behavior :**

This PR adds the capability to add one or more key=value attributes to a Column definition.  They can be added at definition time like this:

    value:double:counter=true
    timestamp:ts:interval=10

The attributes are made available as a new params field in the Column trait and properly serialized in the column and dataset definitions.  Old column definitions are still backwards compatible and parse to empty params.

The intention is to use this as a foundation for a couple uses:
- Whether a histogram column is increasing (Prom-counter like) with each sample or not
- Whether a value is a gauge or counter.   Counters might result in different kind of encoding or counter correction strategy
- expected interval of samples for timestamp columns.  This might lead to smaller or more efficient write buffers in the future.

The parameters are validated at dataset definition time.  HistogramColumns now require a counter=true/false.

#### Breaking changes

Datasets including histogram columns now need to have a `counter=true/false` in the params.

### DIFF
--- a/core/src/main/scala/filodb.core/metadata/Column.scala
+++ b/core/src/main/scala/filodb.core/metadata/Column.scala
@@ -61,12 +61,20 @@ object DataColumn {
 }
 
 object Column extends StrictLogging {
+  import Dataset._
+  import TrySugar._
+
   sealed trait ColumnType extends EnumEntry {
     def typeName: String
     // NOTE: due to a Spark serialization bug, this cannot be a val
     // (https://github.com/apache/spark/pull/7122)
     def clazz: Class[_]
     def keyType: SingleKeyTypeBase[_]
+
+    /**
+     * Validates the params found in the column definition.  By default no checking is done.
+     */
+    def validateParams(params: Config): Unit Or One[BadSchema] = Good(())
   }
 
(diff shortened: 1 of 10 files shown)


vishramachandran:
Shouldn't the type be included explicitly for a histogram when dataset is defined?
I dont see that validation.

core/src/main/scala/filodb.core/metadata/Column.scala
```diff
@@ -31,12 +34,17 @@ trait Column {
  */
 case class DataColumn(id: Int,
                       name: String,
-                      columnType: Column.ColumnType) extends Column {
+                      columnType: Column.ColumnType,
+                      params: Config = ConfigFactory.empty) extends Column {
+  import collection.JavaConverters._
```
vishramachandran: We should add some validation for histogram columns to ensure presence of the metadata.

velvia: ok I'll add a validation API.

velvia: @vishramachandran added, have a look.

PR #282, opened 2019-03-15
feat(core): Correct support for intermittently publishing time series

vishramachandran:
**Pull Request checklist**

We were assigning duplicate partId when an evicted partKey starts re-ingesting. When this partKey is queried, ODP causes corruption of the shard data structures since partSet cannot hold duplicate TimeSeriesPartition objects for same partKey.

**New behavior :**

Prior to creating a new partition in the ingestion pipeline, the idea here is to convert the partKey from ingestion record to a Seq[ColumnFilter], essentially a Lucene query and look up matching partIds (typically just zero or one). If one is found, use the partId. If more than one is found, then it means that there is another PK that includes the same set of tags as the one in question, but has additional tags. We need to find out the most specific partKey by iterating through the partKey doc-value for each matching partId and then resolve the correct partKey.

However, this may be an expensive lookup that needs to happen in the ingestion pipeline. We could optimize this by placing evicted partKeys in a bloom filter, and do Lucene query only if really necessary.

Other bugs fixed:
* Increased akka serialized message limitation
* ODPed partition was being assigned a new partId incorrectly. Now ODPed partition uses the same partId.

Testing Approach: The changes were manually tested by reproducing the issue locally in an E2E test, and retesting after fix. Added unit test to verify assignment of partId on re-ingestion.

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
+++ b/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
@@ -319,16 +319,13 @@ class TimeSeriesShard(val dataset: Dataset,
   private final val shardDownsampler = new ShardDownsampler(dataset, shardNum, downsampleConfig.enabled,
     downsampleConfig.resolutions, downsamplePublisher, shardStats)
 
-  /* Items below are used for bloom filter of evicted time series */
   private case class PartKey(base: Any, offset: Long)
-  private val expectedElements = 5000000
-  private val falsePositiveRate = 0.01
-  private val genHash = new CanGenerateHashFrom[PartKey] {
-    override def generateHash(from: PartKey): Long = {
-      dataset.partKeySchema.partitionHash(from.base, from.offset)
-    }
-  }
-  private val evictedTimeSeries = BloomFilter[PartKey](expectedElements, falsePositiveRate)(genHash)
+  private val evictedPartKeys =
+    BloomFilter[PartKey](storeConfig.evictedPkBfCapacity, falsePositiveRate = 0.01)(new CanGenerateHashFrom[PartKey] {
+      override def generateHash(from: PartKey): Long = {
+        dataset.partKeySchema.partitionHash(from.base, from.offset)
+      }
+    })
 
   case class InMemPartitionIterator(intIt: IntIterator) extends PartitionIterator {
     var nextPart = UnsafeUtils.ZeroPointer.asInstanceOf[TimeSeriesPartition]
@@ -946,7 +943,7 @@ class TimeSeriesShard(val dataset: Dataset,
   // scalastyle:on
 
   private def lookupPreviouslyAssignedPartId(partKeyBase: Array[Byte], partKeyOffset: Long): Option[Int] = {
-    if (evictedTimeSeries.mightContain(PartKey(partKeyBase, partKeyOffset))) {
+    if (evictedPartKeys.mightContain(PartKey(partKeyBase, partKeyOffset))) {
       val filters = dataset.partKeySchema.toStringPairs(partKeyBase, partKeyOffset)
         .map { pair => ColumnFilter(pair._1, Filter.Equals(pair._2)) }
       val matches = partKeyIndex.partIdsFromFilters2(filters, 0, Long.MaxValue)
@@ -963,7 +960,8 @@ class TimeSeriesShard(val dataset: Dataset,
                               // find the most specific match for the given ingestion record
                               val nextPartId = iter.next
                               partKeyIndex.partKeyFromPartId(nextPartId).foreach { candidate =>
-                                if (partKeyEquals(partKeyBase, partKeyOffset, candidate)) {
+                                if (dataset.partKeySchema.equals(partKeyBase, partKeyOffset,
+                                      candidate.bytes, PartKeyLuceneIndex.bytesRefToUnsafeOffset(candidate.offset))) {
                                   partId = nextPartId
                                   logger.debug(s"There is already a partId $partId assigned for " +
                                     s"${dataset.partKeySchema.stringify(partKeyBase, partKeyOffset)}")
@@ -976,11 +974,6 @@ class TimeSeriesShard(val dataset: Dataset,
                             } else Some(partId)
       }
     } else None
-  }
-
-  private def partKeyEquals(partKeyBase: Array[Byte], partKeyOffset: Long, partKeyHeap: BytesRef): Boolean = {
-    dataset.partKeySchema.equals(partKeyBase, partKeyOffset,
-      partKeyHeap.bytes, PartKeyLuceneIndex.bytesRefToUnsafeOffset(partKeyHeap.offset))
   }
 
   /**
@@ -1134,9 +1127,7 @@ class TimeSeriesShard(val dataset: Dataset,
               logger.warn(s"endTime $endTime was not correct. how?", new IllegalStateException())
             } else {
               logger.debug(s"Evicting partId=${partitionObj.partID} from dataset=${dataset.ref} shard=$shardNum")
-              // IMPORTANT: Below step assumes that partKey in TimeSeriesPartition is off-heap to avoid allocation
-              // of an object.
-              evictedTimeSeries.add(PartKey(partitionObj.partKeyBase, partitionObj.partKeyOffset))
+              evictedPartKeys.add(PartKey(partitionObj.partKeyBase, partitionObj.partKeyOffset))
               removePartition(partitionObj)
               partsRemoved += 1
               maxEndTime = Math.max(maxEndTime, endTime)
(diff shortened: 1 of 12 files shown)


core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -1060,9 +1124,10 @@ class TimeSeriesShard(val dataset: Dataset,
             if (isActivelyIngesting(partitionObj.partID))
               logger.warn(s"Partition ${partitionObj.partID} is ingesting, but it was eligible for eviction. How?")
             if (endTime == PartKeyLuceneIndex.NOT_FOUND || endTime == Long.MaxValue) {
-              logger.warn(s"endTime ${endTime} was not correct. how?", new IllegalStateException())
+              logger.warn(s"endTime $endTime was not correct. how?", new IllegalStateException())
             } else {
               logger.debug(s"Evicting partId=${partitionObj.partID} from dataset=${dataset.ref} shard=$shardNum")
+              evictedPartKeys.add(PartKey(partitionObj.partKeyBase, partitionObj.partKeyOffset))
```
broneill: You're stashing a pointer to memory which gets freed (indirectly) the next line below. You'll need to make a copy of the part key in a JVM byte array or something.

broneill: If this just generates a hashcode and then tosses away reference to the key, that's fine. Just document this because the next line deletes the key, and it tripped me up.

vishramachandran: fixed

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -923,31 +936,72 @@ class TimeSeriesShard(val dataset: Dataset,
...
+        .map { pair => ColumnFilter(pair._1, Filter.Equals(pair._2)) }
+      val matches = partKeyIndex.partIdsFromFilters2(filters, 0, Long.MaxValue)
+      matches.result.cardinality() match {
+        case 1 =>           val partId = matches.result.intIterator().next()
+                            logger.debug(s"There is already a partId $partId assigned for " +
+                              s"${dataset.partKeySchema.stringify(partKeyBase, partKeyOffset)}")
+                            Some(partId)
+        case 0 =>           shardStats.evictedPartKeyBloomFilterFalsePositives.increment()
+                            None
+        case c if c > 1 =>  val iter = matches.result.intIterator()
```
velvia: Can we have a counter for the # of times that the match returns more than 1?   This is an especially expensive path so we want to see how often it's invoked.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -973,22 +1027,32 @@ class TimeSeriesShard(val dataset: Dataset,
         s"shard=$shardNum", e); disableAddPartitions()
     }
 
+  /**
+    * Creates new partition and adds them to the shard data structures. DOES NOT update
+    * lucene index. It is the caller's responsibility to add or skip that step depending on the situation.
+    */
   protected def createNewPartition(partKeyBase: Array[Byte], partKeyOffset: Long,
-                                   group: Int, initMapSize: Int = initInfoMapSize): TimeSeriesPartition =
+                                   group: Int, usePartId: Option[Int],
```
velvia: Instead of an Option[Int], can we pass two params, an Int and a reuseFlag, to avoid allocation of Option?

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -1060,9 +1124,14 @@ class TimeSeriesShard(val dataset: Dataset,
...
               logger.warn(s"Partition ${partitionObj.partID} is ingesting, but it was eligible for eviction. How?")
             if (endTime == PartKeyLuceneIndex.NOT_FOUND || endTime == Long.MaxValue) {
-              logger.warn(s"endTime ${endTime} was not correct. how?", new IllegalStateException())
+              logger.warn(s"endTime $endTime was not correct. how?", new IllegalStateException())
             } else {
               logger.debug(s"Evicting partId=${partitionObj.partID} from dataset=${dataset.ref} shard=$shardNum")
+              // add the evicted partKey to a bloom filter so that we are able to quickly
+              // find out if a partId has been assigned to an ingesting partKey before a more expensive lookup.
+              evictedPartKeys.add(PartKey(partitionObj.partKeyBase, partitionObj.partKeyOffset))
+              shardStats.evictedPkBloomFilterSize.set(evictedPartKeys.approximateElementCount())
```
velvia: The approx element count of a bloom filter is like O(n) operation.  I suggest we don't set this every time, but on a minutely schedule (like all the other expensive stats we update at the end of flush cycle).

gateway/src/main/scala/filodb/timeseries/TestTimeseriesProducer.scala
```diff
@@ -97,20 +97,20 @@ object TestTimeseriesProducer extends StrictLogging {
       val endQuery = startQuery + 300
       val query =
         s"""./filo-cli '-Dakka.remote.netty.tcp.hostname=127.0.0.1' --host 127.0.0.1 --dataset prometheus """ +
-        s"""--promql 'heap_usage{dc="DC0",_ns="App-0"}' --start $startQuery --end $endQuery --limit 15"""
+        s"""--promql 'heap_usage{_ns="App-0"}' --start $startQuery --end $endQuery --limit 15"""
```
velvia: Please do not update the TimeSeriesProducer as this is changing significantly in my other PR

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
velvia: Basically, what we want to see:
- When we need to add partitions, how often do we need to actually search Lucene (hopefully bloom filter hit rate is low)
- When we search Lucene, how often we need to compare multiple values

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -653,7 +666,7 @@ class TimeSeriesShard(val dataset: Dataset,
     shardStats.numActivelyIngestingParts.set(cardinality)
   }
 
-  private def addPartKeyToTimebucket(indexRb: RecordBuilder, p: TimeSeriesPartition) = {
+  private def addPartKeyToTimebucketRb(indexRb: RecordBuilder, p: TimeSeriesPartition) = {
```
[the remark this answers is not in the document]
vishramachandran: RecordBuilder

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
vishramachandran: But it would not be thread-safe to do a read on another thread while writes are happening to the bloom filter. Technically speaking, it is constant time since the size of the bloom filter is constant. But I get your reasoning. Let me think about other ways too...

gateway/src/main/scala/filodb/timeseries/TestTimeseriesProducer.scala
vishramachandran: OK. I'll remove the changes

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
vishramachandran: will fix

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
vishramachandran: will add.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
vishramachandran: I am thinking it is acceptable since we set this only in the case where a partition is evicted. This should happen rarely.

vishramachandran:
We cannot cleanup the bloom filter without cleaning up lucene index - it would lead to wrong partId assignments. The system should be operationalized by monitoring the false positive rate for bloom filter. If it spikes up above a specific threshold, consider one or both these remedial measures
1. Increase capacity of bloom filter
2. Restart node

We also piggy back on the fact that nodes will be restarted on a regular cadence for maintenance and upgrades. The size of the bloom filter should be set based on that cadence.

vishramachandran:
> how many evictions can the bloom filter handle before getting saturated and being of no use

I have already answered this question in my previous comment. We also piggy back on the fact that nodes will be restarted on a regular cadence for maintenance and upgrades. The size of the bloom filter should be set based on that cadence. The estimated capacity of the bloom filter is part of storeConfig. 

Bloom filter sizes:
```
scala> BloomFilter[String](5000000, falsePositiveRate = 0.01).numberOfBits
res9: Long = 47925292
// about 6MB

scala> BloomFilter[String](50000000, falsePositiveRate = 0.01).numberOfBits
res10: Long = 479252919
// about 60MB
```

> We need to monitor how much this slows down the ingestion of new metrics

Look up of bloom filter is fast and it is done only when the partKey is not in the heap. So in theory, it should not affect ingestion.

velvia:
Based on the sizes posted, it seems size of bloom filter will be reasonable.

PR #284, opened 2019-03-15
feat(memory, core): Add logs/stats for block manager; reclaim blocks even if data dropped

velvia:
**Pull Request checklist**

Adds more logging and one counter for the rare occurrence when the BlockManager cannot find blocks to reclaim and is out of new blocks to return.

Also, even when data is dropped, mark all blocks as reclaimable and flush/commit checkpoints, as we prefer availability.

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
+++ b/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
@@ -731,10 +731,6 @@ class TimeSeriesShard(val dataset: Dataset,
     }.flatMap {
       case Success           => blockHolder.markUsedBlocksReclaimable()
         commitCheckpoint(dataset.ref, shardNum, flushGroup)
-      // Even if data is dropped we still mark blocks as reclaimable and flush checkpoints.  We prioritize the
-      // availability of the latest metrics/data for alerting.
-      case DataDropped       => blockHolder.markUsedBlocksReclaimable()
-        commitCheckpoint(dataset.ref, shardNum, flushGroup)
       case er: ErrorResponse => Future.successful(er)
     }.recover { case e =>
       logger.error(s"Internal Error when persisting chunks in dataset=${dataset.ref} shard=$shardNum - should " +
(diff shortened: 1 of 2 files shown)


core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -731,6 +731,10 @@ class TimeSeriesShard(val dataset: Dataset,
     }.flatMap {
       case Success           => blockHolder.markUsedBlocksReclaimable()
         commitCheckpoint(dataset.ref, shardNum, flushGroup)
+      // Even if data is dropped we still mark blocks as reclaimable and flush checkpoints.  We prioritize the
+      // availability of the latest metrics/data for alerting.
+      case DataDropped       => blockHolder.markUsedBlocksReclaimable()
+        commitCheckpoint(dataset.ref, shardNum, flushGroup)
```
vishramachandran: It is too early to make this move. We still do not know that the "Could not obtain block" error is because each and every block in the system becomes non-reclaimable. I suspect there is some other bug lurking around.

This move will result in premature data loss that is avoidable. 

Even if we decide to move forward with such an approach it should be done only after multiple flush intervals have passed without success - not immediately after one round.

memory/src/main/scala/filodb.memory/BlockManager.scala
```diff
@@ -186,6 +189,10 @@ class PageAlignedBlockManager(val totalMemorySizeInBytes: Long,
       if (entry.getValue.isEmpty) timeOrderedListIt.remove()
     }
     if (reclaimed < num) reclaimFrom(usedBlocks, stats.blocksReclaimedMetric)
+    if (reclaimed < num) {
```
vishramachandran: Worth adding a comment that goes like " // if we do not get required blocks even after reclaim call"
There would be a tendency to combine both if blocks to one.

velvia:
@vishramachandran feedback addressed

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
velvia: That move is independent of the bug.  But it can be separated out to a different PR.

PR #288, opened 2019-03-19
feat(query): Histogram sum() over RangeVectors

velvia:
**Pull Request checklist**

**New behavior :**

Adds the ability to do sum() over HistogramColumn range vectors.
Monotonicity correction and ability to sum histograms of different schemas is also added for proper histogram summing functionality.

### DIFF
--- a/memory/src/main/scala/filodb.memory/format/vectors/Histogram.scala
+++ b/memory/src/main/scala/filodb.memory/format/vectors/Histogram.scala
@@ -132,6 +132,13 @@ trait HistogramWithBuckets extends Histogram {
   def buckets: HistogramBuckets
   final def numBuckets: Int = buckets.numBuckets
   final def bucketTop(no: Int): Double = buckets.bucketTop(no)
+  final def valueArray: Array[Double] = {
+    val values = new Array[Double](numBuckets)
+    for { b <- 0 until numBuckets optimized } {
+      values(b) = bucketValue(b)
+    }
+    values
+  }
 }
 
 final case class LongHistogram(buckets: HistogramBuckets, values: Array[Long]) extends HistogramWithBuckets {
@@ -190,6 +197,11 @@ final case class MutableHistogram(buckets: HistogramBuckets, values: Array[Doubl
 object MutableHistogram {
   def empty(buckets: HistogramBuckets): MutableHistogram =
     MutableHistogram(buckets, Array.fill(buckets.numBuckets)(Double.NaN))
+
+  def apply(h: Histogram): MutableHistogram = h match {
+    case hb: HistogramWithBuckets => MutableHistogram(hb.buckets, hb.valueArray)
+    case other: Histogram         => ???
+  }
 }
 
 /**
--- a/memory/src/main/scala/filodb.memory/format/vectors/Histogram.scala
+++ b/memory/src/main/scala/filodb.memory/format/vectors/Histogram.scala
@@ -179,8 +179,10 @@ final case class MutableHistogram(buckets: HistogramBuckets, values: Array[Doubl
   final def copy: Histogram = MutableHistogram(buckets, values.clone)
 
   /**
-   * Adds the values from another MutableHistogram having the same bucket schema.  If it does not, then
-   * an exception is thrown -- for now.  Modifies itself.
+   * Adds the values from another Histogram.
+   * If the other histogram has the same bucket scheme, then the values are just added per bucket.
+   * If the scheme is different, then an approximation is used so that the resulting histogram has
+   * an approximate sum of the individual distributions, with the original scheme.  Modifies itself.
    */
   final def add(other: HistogramWithBuckets): Unit =
     if (buckets == other.buckets) {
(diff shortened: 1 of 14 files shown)


query/src/main/scala/filodb/query/exec/AggrOverRangeVectors.scala
```diff
@@ -12,7 +12,7 @@ import filodb.core.metadata.Column.ColumnType
 import filodb.core.metadata.Dataset
 import filodb.core.query._
 import filodb.memory.data.ChunkMap
-import filodb.memory.format.{RowReader, UnsafeUtils, ZeroCopyUTF8String}
+import filodb.memory.format.{vectors => bv, RowReader, UnsafeUtils, ZeroCopyUTF8String}
```
broneill: Style request: Aliasing makes the code harder to read, but it looks like 'bv' is only used later. Can you put the import closer to its first usage? It makes things easier to read, in my opinion.

query/src/test/scala/filodb/query/exec/AggrOverRangeVectorsSpec.scala
```diff
@@ -4,21 +4,17 @@ import scala.annotation.tailrec
...
 import monix.execution.Scheduler.Implicits.global
 import monix.reactive.Observable
-import org.scalatest.{FunSpec, Matchers}
 import org.scalatest.concurrent.ScalaFutures
 
 import filodb.core.metadata.Column.ColumnType
 import filodb.core.query._
-import filodb.memory.format.{RowReader, ZeroCopyUTF8String}
-import filodb.query.{AggregationOperator, QueryConfig}
+import filodb.memory.format.{vectors => bv, RowReader, ZeroCopyUTF8String}
```
broneill: Ditto here on the 'bv' alias.

query/src/test/scala/filodb/query/exec/AggrOverRangeVectorsSpec.scala
```diff
@@ -353,7 +331,26 @@ class AggrOverRangeVectorsSpec extends FunSpec with Matchers with ScalaFutures {
     result6b.size shouldEqual 1
     result6b(0).key shouldEqual ignoreKey
     compareIter(result6b(0).rows.map(_.getDouble(1)), Seq(5.4d,5.6d).iterator)
+  }
+
+  it("should sum histogram RVs") {
```
vishramachandran: Can we have a few test cases where we add histograms with buckets that do not match?

memory/src/test/scala/filodb.memory/format/vectors/HistogramTest.scala
```diff
@@ -91,5 +91,40 @@ class HistogramTest extends NativeVectorTest {
...
+    }
+
+    it("should add histogram w/ diff bucket scheme and result in monotonically increasing histogram") {
+      val hist1 = mutableHistograms(0).copy.asInstanceOf[MutableHistogram]
+
+      val scheme2 = GeometricBuckets(2.0, 6.0, 3)
+      val hist2 = LongHistogram(scheme2, Array(10L, 20L, 40L))
+
+      hist1.add(hist2)
+      hist1.makeMonotonic()
```
vishramachandran: add() function should itself take care of calling makeMonotonic() if buckets are different.
I don't see any scenario where we wouldn't do that.

memory/src/main/scala/filodb.memory/format/vectors/Histogram.scala
```diff
@@ -172,8 +179,10 @@ final case class MutableHistogram(buckets: HistogramBuckets, values: Array[Doubl
...
   /**
-   * Adds the values from another MutableHistogram having the same bucket schema.  If it does not, then
-   * an exception is thrown -- for now.  Modifies itself.
+   * Adds the values from another Histogram.
+   * If the other histogram has the same bucket scheme, then the values are just added per bucket.
+   * If the scheme is different, then an approximation is used so that the resulting histogram has
+   * an approximate sum of the individual distributions, with the original scheme.  Modifies itself.
    */
   final def add(other: HistogramWithBuckets): Unit =
     if (buckets == other.buckets) {
```
vishramachandran: Bucket comparison may be an O(n) operation depending on the type of Histogram buckets and that will be done very often. We should make this a constant time operation for fully enumerated buckets.

I am seeing that GeometricBuckets is the only type of histogram bucketing we have now. 
1. Are the optimizations holding good because we have only one implementation of this trait ?
2. At some point, we when we have enumerated buckets this code will be invalidated.

memory/src/test/scala/filodb.memory/format/vectors/HistogramTest.scala
velvia: Yes there is a scenario.  add() is used in sum_over_time within a chunk where we are guaranteed that buckets are the same, so no need for makeMonotonic().   However I can make the default add call that, and have an inner one which doesn't.

query/src/test/scala/filodb/query/exec/AggrOverRangeVectorsSpec.scala
velvia: Sure.

memory/src/main/scala/filodb.memory/format/vectors/Histogram.scala
velvia: It's probably impossible to make this constant time for all types of buckets.  We can possibly use some kind of hash for non-geometric buckets, which would make it mostly constant time.  

However, even if it was not constant time, adding buckets of the same scheme is still way faster than dealing with different schemes.

memory/src/main/scala/filodb.memory/format/vectors/Histogram.scala
```diff
@@ -183,13 +192,43 @@ final case class MutableHistogram(buckets: HistogramBuckets, values: Array[Doubl
         values(b) += other.bucketValue(b)
       }
     } else {
-      throw new UnsupportedOperationException(s"Cannot add other with buckets ${other.buckets} to myself $buckets")
+      // For now, if the bucket does not match, add it to the next highest matching bucket in our scheme.
+      // This should approximately work since buckets contain everything less than or equal.
```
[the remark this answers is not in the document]
velvia: It's not incorrect, but it does result in loss of granularity.   However, the existing approach is not correct either; this approach is an improvement.

query/src/main/scala/filodb/query/exec/AggrOverRangeVectors.scala
velvia: Sure.

memory/src/main/scala/filodb.memory/format/vectors/Histogram.scala
```diff
@@ -172,24 +179,64 @@ final case class MutableHistogram(buckets: HistogramBuckets, values: Array[Doubl
...
+  final def addNoCorrection(other: HistogramWithBuckets): Unit =
     if (buckets == other.buckets) {
       // If it was NaN before, reset to 0 to sum another hist
       if (values(0).isNaN) java.util.Arrays.fill(values, 0.0)
       for { b <- 0 until numBuckets optimized } {
         values(b) += other.bucketValue(b)
       }
     } else {
-      throw new UnsupportedOperationException(s"Cannot add other with buckets ${other.buckets} to myself $buckets")
+      throw new UnsupportedOperationException(s"Cannot add histogram of scheme ${other.buckets} to $buckets")
```
vishramachandran: This exception needs to be handled during downsampling, if it occurs.

memory/src/main/scala/filodb.memory/format/vectors/Histogram.scala
```diff
@@ -172,24 +179,64 @@ final case class MutableHistogram(buckets: HistogramBuckets, values: Array[Doubl
...
+      // TODO: In the future, support adding buckets of different scheme.  Below is an example
+      // var ourBucketNo = 0
+      // for { b <- 0 until other.numBuckets optimized } {
+      //   // Find our first bucket greater than or equal to their bucket
+      //   while (ourBucketNo < numBuckets && bucketTop(ourBucketNo) < other.bucketTop(b)) ourBucketNo += 1
```
vishramachandran: If you plan to leave the code commented, please also document the problems with this approach.
IMO, it is not viable for the reasons we discussed. Primarily because it is not commutative, and it can be performed with better granularity.

velvia: I'm adding a note.

PR #290, opened 2019-03-21
misc(core): Reduce zipkin message batchsize

tjackpaul:
**Pull Request checklist**

**Current behavior :** 
Zipkin message batch size is 5mb today. This sends more spans in a single request, adding load on zipkin collectors.

**New behavior :**
Reduce the zipkin message batch size to 128kb. I do not think it should be part of a config at the moment. We will add it to config later if required.

### DIFF
--- a/core/src/main/scala/filodb.core/zipkin/Zipkin.scala
+++ b/core/src/main/scala/filodb.core/zipkin/Zipkin.scala
@@ -2,13 +2,15 @@ package filodb.core.zipkin
 
 import java.net.InetAddress
 
+import scala.util.Try
+
 import com.typesafe.config.Config
 import kamon.{Kamon, SpanReporter}
 import kamon.trace.Span.{FinishedSpan => KamonSpan, Mark, TagValue}
 import kamon.util.Clock
 import org.slf4j.LoggerFactory
-import scala.util.Try
 import zipkin2.{Endpoint, Span => ZipkinSpan}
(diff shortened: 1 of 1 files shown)


core/src/main/scala/filodb.core/zipkin/Zipkin.scala
```diff
@@ -126,15 +128,19 @@ class ZipkinReporter extends SpanReporter {
...
+    var url = s"$zipkinEndpoint/api/v2/spans"
     if (zipkinEndpoint == null || zipkinEndpoint.trim.isEmpty) {
-      AsyncReporter.create(
-        OkHttpSender.create(s"http://$zipkinHost:$zipkinPort/api/v2/spans")
-      )
-    } else {
-      AsyncReporter.create(
-        OkHttpSender.create(s"$zipkinEndpoint/api/v2/spans")
-      )
+      url = s"http://$zipkinHost:$zipkinPort/api/v2/spans"
```
velvia: Rather than use a var:

    val url = if (zipkinEndpoint == null || zipkinEndpoint.trim.isEmpty) s"http://$zipkinHost:$zipkinPort/api/v2/spans"
                   else s"$zipkinEndpoint/api/v2/spans"

tjackpaul: Ok. This is taken care of.

PR #291, opened 2019-03-21
bug(core): ODP from Cassandra only if Lucene index indicates that there is data

vishramachandran:
**Pull Request checklist**

If earliestTime in memory for a partition is more than queryStart time, we decide to page from cassandra.

**New behavior :**

We page from cassandra only if Lucene index indicates that partition's startTime is earlier than the earliest time in memory, and the query is for earlier than the earliest time in memory.

The code organization for ODP imo has been less than ideal because there is a lot of things happening and leak in abstraction with non-functional side effects. I do understand everything has been done to reduce object allocation and double iteration of objects. I haven't really changed all of that, but my thinking is that the changes in the PR may add a bit to the chaos.

If you guys have any suggestions on how to make this better without affecting performance, please call it out.

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/OnDemandPagingShard.scala
+++ b/core/src/main/scala/filodb.core/memstore/OnDemandPagingShard.scala
@@ -2,6 +2,7 @@ package filodb.core.memstore
 
 import scala.collection.mutable.ArrayBuffer
 import scala.concurrent.ExecutionContext
+import scala.util.control.NonFatal
 
 import debox.Buffer
 import kamon.Kamon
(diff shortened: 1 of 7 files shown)


vishramachandran:
@velvia OnDemandPagingShard.scanPartitions still needs to iterate across all partitions because there may be other partitions whose earliestTime is earlier than the queryStartTime. Those partIds will *not* be in the startTime map. Those that are in memory will be classified as inMemoryPartitions and not ODPed.

Lets talk in person if necessary.

core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
```diff
@@ -322,7 +322,7 @@ extends ChunkMap(memFactory, initMapSize) with ReadablePartition {
 
   final def earliestTime: Long = {
     if (numChunks == 0) {
-      Long.MinValue
+      Long.MaxValue
```
broneill: Given that you simply flipped this from min to max confuses me. Can you add a doc regarding the method contract?

vishramachandran: Added doc.
We use earliestTime method in new contexts now which need an alternate default value if chunks are not present.
Hence I had to change the contract.

vishramachandran:
@broneill Yes - existing tests needed to change. The index needed to be flushed now in various places since it is actively used now.

core/src/main/scala/filodb.core/memstore/OnDemandPagingShard.scala
```diff
@@ -61,35 +61,41 @@ TimeSeriesShard(dataset, storeConfig, shardNum, rawStore, metastore, evictionPol
...
-      }.getOrElse {
-        // add it to partitions which do not need to be ODP'ed, send these directly and first
-        inMemoryPartitions += p
+    val pagingMethods = new ArrayBuffer[ChunkScanMethod]
+    val inMemOdp = debox.Set.empty[Int]
+
+    iterResult.partIdsInMemoryMayNeedOdp.foreach { case (pId, startTime) =>
+      val p = partitions.get(pId)
+      if (p != null) {
+        checkIfReallyNeedToOdp(p, chunkMethod, pagingEnabled, startTime).map { rawChunkMethod =>
```
velvia: Minor point- I think `chunksToFetch` or `chunksToODP` is still a better name for this method.  `checkIf....` seems to imply it returns a Boolean, but it actually returns the ChunkMethod.

PR #292, opened 2019-03-22
 feat(coordinator): Resync shard mapping with a full shard state command

broneill:
**Pull Request checklist**

**Current behavior :**
Updates to shard ingestion assignment are sent incrementally via start/stop commands, but this can cause the state to get out of sync due to failovers. When the state is out of sync among the cluster nodes, a full restart is required to fix the state.

**New behavior :**
The start/stop commands are removed, and a new command is sent which propagates the entire shard assignment state to all nodes, when anything changes. Although this is less efficient, the number of times this message is expected to be sent is relatively small.

**BREAKING CHANGES**
This is an incompatible change, in that the start/stop ingestion commands are gone. During a deployment, ingestion might stall but when all nodes are updated, a final resync is expected to be performed by all nodes. If this doesn't happen, a second deployment might be required.

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/IngestionActor.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/IngestionActor.scala
@@ -114,7 +114,7 @@ private[filodb] final class IngestionActor(dataset: Dataset,
 
     for (shard <- 0 until state.map.numShards) {
       if (state.map.coordForShard(shard) == context.parent) {
-        if (state.map.shouldBeIngesting(shard)) {
+        if (state.map.isAnIngestionState(shard)) {
           if (shardsToStop.contains(shard)) {
             // Is aready ingesting, and it must not be stopped.
             shardsToStop.remove(shard)
--- a/coordinator/src/main/scala/filodb.coordinator/IngestionActor.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/IngestionActor.scala
@@ -36,7 +36,7 @@ object IngestionActor {
 
 /**
   * Oversees ingestion and recovery process for a single dataset.  The overall process for a single shard:
-  * 1. Resync shard command is received and start() called
+  * 1. Shard state command is received and start() called
   * 2. MemStore.setup() is called for that shard
   * 3. IF no checkpoint data is found, THEN normal ingestion is started
   * 4. IF checkpoints are found, then recovery is started from the minimum checkpoint offset
@@ -60,7 +60,7 @@ private[filodb] final class IngestionActor(dataset: Dataset,
 
   final val streamSubscriptions = new HashMap[Int, CancelableFuture[Unit]]
   final val streams = new HashMap[Int, IngestionStream]
-  private var resyncVersion: Long = 0
+  private var shardStateVersion: Long = 0
 
   // Params for creating the default memStore flush scheduler
   private final val numGroups = storeConfig.groupsPerShard
@@ -89,7 +89,7 @@ private[filodb] final class IngestionActor(dataset: Dataset,
   def receive: Receive = LoggingReceive {
     case GetStatus               => status(sender())
     case e: IngestRows           => ingest(e)
-    case e: ResyncShardIngestion => resync(e, sender())
+    case e: ShardIngestionState  => resync(e, sender())
   }
 
   /**
@@ -97,14 +97,14 @@ private[filodb] final class IngestionActor(dataset: Dataset,
     * reconciles any differences. It does so by stopping ingestion for shards that aren't mapped
     * to this node, and it starts ingestion for those that are.
     */
-  private def resync(resync: ResyncShardIngestion, origin: ActorRef): Unit = {
-    if (invalid(resync.ref)) {
-      logger.error(s"$resync is invalid for this ingester '${dataset.ref}'.")
+  private def resync(state: ShardIngestionState, origin: ActorRef): Unit = {
+    if (invalid(state.ref)) {
+      logger.error(s"$state is invalid for this ingester '${dataset.ref}'.")
       return
     }
 
-    if (resync.version != 0 && resync.version <= resyncVersion) {
-      logger.info(s"Ignoring old ResyncShardIngestion version: ${resync.version} <= $resyncVersion")
+    if (state.version != 0 && state.version <= shardStateVersion) {
+      logger.info(s"Ignoring old ShardIngestionState version: ${state.version} <= $shardStateVersion")
       return
     }
 
@@ -112,8 +112,8 @@ private[filodb] final class IngestionActor(dataset: Dataset,
     // which must continue being ingested.
     val shardsToStop = HashSet() ++ streams.keySet
 
-    for (shard <- 0 until resync.map.numShards) {
-      if (resync.map.coordForShard(shard) == context.parent) {
+    for (shard <- 0 until state.map.numShards) {
+      if (state.map.coordForShard(shard) == context.parent) {
         // Must be ingesting from the shard.
         if (shardsToStop.contains(shard)) {
           // Is aready ingesting, and it must not be stopped.
@@ -130,8 +130,8 @@ private[filodb] final class IngestionActor(dataset: Dataset,
       stopIngestion(shard)
     }
 
-    if (resync.version != 0) {
-      resyncVersion = resync.version
+    if (state.version != 0) {
+      shardStateVersion = state.version
     }
   }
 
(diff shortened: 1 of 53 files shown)


coordinator/src/main/scala/filodb.coordinator/IngestionActor.scala
```diff
@@ -60,6 +60,7 @@ private[filodb] final class IngestionActor(dataset: Dataset,
 
   final val streamSubscriptions = new HashMap[Int, CancelableFuture[Unit]]
   final val streams = new HashMap[Int, IngestionStream]
+  private var shardStateVersion: Long = 0
```
velvia: Just a note: in Actor classes everything is essentially private, because the direct reference to an instance of an Actor class is NEVER shared.  The only way to create an actor is via `actorOf`, which returns strictly an `ActorRef` -- thus Actor classes, methods, and fields are never directly accessible.  I'm fine with "private" to communicate intent I guess.

coordinator/src/main/scala/filodb.coordinator/IngestionActor.scala
```diff
@@ -82,66 +83,101 @@ private[filodb] final class IngestionActor(dataset: Dataset,
...
 
-      val ingestion = for {
-        _ <- memStore.recoverIndex(dataset.ref, e.shard)
-        checkpoints <- memStore.metastore.readCheckpoints(dataset.ref, e.shard) }
-      yield {
-        if (checkpoints.isEmpty) {
-          // Start normal ingestion with no recovery checkpoint and flush group 0 first
-          normalIngestion(e.shard, None, 0, storeConfig.diskTTLSeconds)
+    if (state.version != 0 && state.version <= shardStateVersion) {
+      logger.info(s"Ignoring old ShardIngestionState version: ${state.version} <= $shardStateVersion")
```
velvia: Are invalid versions things that we should track via metrics?  Maybe add some Kamon metric here?

coordinator/src/main/scala/filodb.coordinator/IngestionActor.scala
```diff
@@ -82,66 +83,101 @@ private[filodb] final class IngestionActor(dataset: Dataset,
...
+      return
+    }
+
+    // Start with the full set of all shards being ingested, and remove shards from this set
+    // which must continue being ingested.
+    val shardsToStop = HashSet() ++ streams.keySet
```
velvia: It's probably adequate to just do `streams.keySet` as keySet is supposed to return some immutable Set.

coordinator/src/main/scala/filodb.coordinator/IngestionActor.scala
```diff
@@ -82,66 +83,101 @@ private[filodb] final class IngestionActor(dataset: Dataset,
...
+
+    for (shard <- 0 until state.map.numShards) {
+      if (state.map.coordForShard(shard) == context.parent) {
```
velvia: This code relies on the list of coordinators in a `ShardMapper`.   This _should_ work but coordinators can have different states.  Namely, we probably want to filter out shards with a "Down" or "Stopped" `ShardStatus`.  Possibly `Error` ones, but I guess `Error` ones can be left alone.

velvia: I would also add debug log statements so we can figure out, if needed, what this code block is regarding as keeping or stopping.

coordinator/src/main/scala/filodb.coordinator/ShardManager.scala
```diff
@@ -508,59 +518,66 @@ private[coordinator] final class ShardManager(settings: FilodbSettings,
...
         for {
           subscription <- _subscriptions.subscription(ref)
         } subscription.subscribers foreach (_ ! snapshot)
+
+        // Also send a complete state command to all ingestion actors. Most are expected to
+        // actually be ingesting something, so also send the dataset info. This might be
+        // redundant, but sending it again is harmless.
+
+        val state = _datasetInfo(snapshot.ref)
+        val setupMsg = client.IngestionCommands.DatasetSetup(state.dataset.asCompactString,
```
velvia: SetupMessage is supposed to be a one-time thing, but this method is going to be called from multiple places and definitely more than once in the lifetime of a coordinator for that dataset.   Creating the DatasetSetup message and serializing it is kinda expensive.   That is why a warning is logged when it is received and the dataset is already setup.  Let's chat and think about ways to not send this with every publishChanges()....

coordinator/src/main/scala/filodb.coordinator/ShardManager.scala
```diff
@@ -508,59 +518,66 @@ private[coordinator] final class ShardManager(settings: FilodbSettings,
...
+          state.storeConfig, state.source, state.downsample)
+
+        // TODO: Need to provide a globally consistent version, incremented when anything
+        //       changes, for any dataset.
+        val resync = ShardIngestionState(0, snapshot.ref, snapshot.map)
```
velvia: Hmmm.   Seems like this will always send a 0 version....

broneill: Yes. See TODO comment.

coordinator/src/main/scala/filodb.coordinator/ShardManager.scala
broneill: How frequently are changes to be expected?

coordinator/src/main/scala/filodb.coordinator/IngestionActor.scala
broneill: The goal here is to have a mutable copy of the set.

coordinator/src/main/scala/filodb.coordinator/IngestionActor.scala
broneill: The more interesting metric should be added when the leader attempts to modify the mappings, but it gets a version mismatch. Such a failure indicates dueling leaders. This code is to be added later.

coordinator/src/main/scala/filodb.coordinator/IngestionActor.scala
broneill: I can switch it to be consistent. The default Scala behavior is for everything to be public by default. It would be nice it everything was private by default. Oh well.

coordinator/src/main/scala/filodb.coordinator/IngestionActor.scala
```diff
@@ -82,66 +83,105 @@ private[filodb] final class IngestionActor(dataset: Dataset,
...
```
vishramachandran: Does the == work reliably for local and remote actor comparison?

coordinator/src/main/scala/filodb.coordinator/ShardManager.scala
```diff
@@ -120,7 +120,11 @@ private[coordinator] final class ShardManager(settings: FilodbSettings,
     origin ! mapperOpt(ref).getOrElse(DatasetUnknown(ref))
 
   private def mapperOpt(ref: DatasetRef): Option[CurrentShardSnapshot] =
-    _shardMappers.get(ref).map(m => CurrentShardSnapshot(ref, m))
+    // Although a copy of the ShardMapper isn't typically required, it is required for the
+    // tests to work properly. This is because the TestProbe provides access to the local
+    // ShardMapper instance, and so any observation of the snapshot would expose the latest
+    // mappings instead.
+    _shardMappers.get(ref).map(m => CurrentShardSnapshot(ref, m.copy()))
```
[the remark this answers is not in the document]
broneill: Yes, it's additional copies, but how frequently is this expected? I'm favoring simplicity here. Also note that an early copy ensures that the snapshot is in fact a snapshot, and we don't have to worry about any concurrent modifications that might occur by the time the message is serialized and send to the actors. I don't think any such concurrent modification is possible, but this is safer.

coordinator/src/main/scala/filodb.coordinator/IngestionActor.scala
broneill: I believe this was the same comparison as before. Equality checking for ActorRef considers the path and uid (according to the docs), so this comparison should work. Unlike Java ==, Scala calls the equals method.

broneill:
@vishramachandran It should be possible to read the dataset state at startup without making any of these other changes. Perhaps that should be done first, in a separate PR?

coordinator/src/main/scala/filodb.coordinator/ShardManager.scala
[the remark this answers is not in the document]
vishramachandran: Can we rename the method then to mapperCopyOpt ?

coordinator/src/main/scala/filodb.coordinator/ShardManager.scala
velvia: I'm not sure having a large number of actors simultaneously reading from C* is a good idea.   I think having the `NodeClusterActor` send the setup message one time (when the datasets are setup) to all coordinators is probably better.

Some of this might need to be rethought also when we do more schema changes.

broneill:
I'm sending the setup messages when a shard gets assigned to a node. This might be redundant, but it's pretty much the same as before.

PR #293, opened 2019-03-22
 feat(query): Parallelize scanPartitions and all QE work in query threadpool

velvia:
**Pull Request checklist**

The setup of each query, including `scanPartitions`, Lucene partId lookup, and setting up RangeVectorTransformers, happens single-threaded in the Actor receive loop's threadpool.   The overhead of this setup being run in serial can be significant.

**New behavior :**

1.  The `ExecPlan.execute` method is run entirely inside of a Monix Task, including all setup, `scanPartitions`, etc.  This means these tasks can now be parallelized and execute outside of the Actor loop.
2. All of the QueryEngine tasks and Observables are now run in a separate, configurable threadpool, named query-<dataset>.  This will make it easier to do profiling later too.

### Before

```
[info] Benchmark                                               Mode  Cnt      Score     Error  Units
[info] QueryInMemoryBenchmark.noOverlapQueries                thrpt   40   3724.422 ±  94.726  ops/s
[info] QueryInMemoryBenchmark.singleThreadedMinOverTimeQuery  thrpt   40  13859.802 ± 197.079  ops/s
[info] QueryInMemoryBenchmark.singleThreadedRawQuery          thrpt   40  24375.510 ± 385.174  ops/s
[info] QueryInMemoryBenchmark.singleThreadedSumRateCCQuery    thrpt   40   4626.947 ±  57.162  ops/s
[info] QueryInMemoryBenchmark.someOverlapQueries              thrpt   40   3112.174 ±  88.544  ops/s
```

### After

```
[info] Benchmark                                               Mode  Cnt      Score     Error  Units
[info] QueryInMemoryBenchmark.noOverlapQueries                thrpt   40   9521.884 ± 142.717  ops/s
[info] QueryInMemoryBenchmark.singleThreadedMinOverTimeQuery  thrpt   40  14027.576 ± 128.490  ops/s
[info] QueryInMemoryBenchmark.singleThreadedRawQuery          thrpt   40  23583.630 ± 553.107  ops/s
[info] QueryInMemoryBenchmark.singleThreadedSumRateCCQuery    thrpt   40   4627.667 ±  37.934  ops/s
[info] QueryInMemoryBenchmark.someOverlapQueries              thrpt   40   5756.704 ± 221.029  ops/s
```

### DIFF
--- a/core/src/main/resources/filodb-defaults.conf
+++ b/core/src/main/resources/filodb-defaults.conf
@@ -246,6 +246,11 @@ akka {
       "filodb.query.LogicalPlan" = kryo
     }
 
+    default-dispatcher.fork-join-executor {
+      parallelism-factor = 2.0
+      parallelism-max = 32
+    }
+
     # Just the defaults to start with. TODO optimize and pick the executor needed.
     shard-status-dispatcher {
       type = Dispatcher
@@ -304,27 +309,6 @@ akka {
       threshold = 12.0
       expected-response-after = 5s
     }
-  }
-
-  # Just the defaults to start with. TODO optimize and pick the executor needed.
-  shard-status-dispatcher {
-    # Dispatcher is the name of the event-based dispatcher
-    type = Dispatcher
-    # What kind of ExecutionService to use
-    executor = "fork-join-executor"
-    # Configuration for the fork join pool
-    fork-join-executor {
-      # Min number of threads to cap factor-based parallelism number to
-      parallelism-min = 2
-      # Parallelism (threads) ... ceil(available processors * factor)
-      parallelism-factor = 2.0
-      # Max number of threads to cap factor-based parallelism number to
-      parallelism-max = 10
-    }
-    # Throughput defines the maximum number of messages to be
-    # processed per actor before the thread jumps to the next actor.
-    # Set to 1 for as fair as possible.
-    throughput = 100
   }
 
   # Be sure to terminate/exit JVM process after Akka shuts down.  This is important for the
--- a/core/src/main/resources/filodb-defaults.conf
+++ b/core/src/main/resources/filodb-defaults.conf
@@ -38,6 +38,9 @@ filodb {
 
     # Minimum step required for a query
     min-step = 5 seconds
+
+    # Number of threads in query threadpool.  There is one threadpool per dataset.
+    num-threads = 8
   }
 
   shard-manager {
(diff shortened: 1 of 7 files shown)


query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
```diff
@@ -24,7 +24,7 @@ final case class PeriodicSamplesMapper(start: Long,
                                        end: Long,
                                        window: Option[Long],
                                        functionId: Option[RangeFunctionId],
-                                       funcParams: Seq[Any] = Nil) extends RangeVectorTransformer {
+                                       funcParams: Seq[Any] = Nil) extends RangeVectorTransformer with StrictLogging {
```
vishramachandran: Should not add logger here - it will affect serialization. Use query logger instead.

core/src/main/resources/filodb-defaults.conf
```diff
@@ -243,6 +246,11 @@ akka {
       "filodb.query.LogicalPlan" = kryo
     }
 
+    default-dispatcher.fork-join-executor {
+      parallelism-factor = 2.0
+      parallelism-max = 32
```
vishramachandran: Why are we hard coding these values ? These should  be chosen based on the hardware the process is running on.

core/src/main/resources/filodb-defaults.conf
```diff
@@ -38,6 +38,9 @@ filodb {
 
     # Minimum step required for a query
     min-step = 5 seconds
+
+    # Number of threads in query threadpool.  There is one threadpool per dataset.
+    num-threads = 8
```
vishramachandran: I dont understand this. Each dataset has a queryActor. Why thread per dataset?
Why are we hard-coding this value ?

vishramachandran:
@velvia  @broneill  A thought for the future.
I don't see a need for all queries to be routed through a single actor (QueryActor). There is no need for serialization at all whatsoever here.
We could simply expose a grpc based HTTP api and route queries directly from there.

core/src/main/resources/filodb-defaults.conf
velvia: We don't need as many threads for the default fork-join pool since it is now doing next to nothing.  Will document.

core/src/main/resources/filodb-defaults.conf
velvia: Just to make sure we are on the same page - so there is now an explicit threadpool for all QueryEngine work for each dataset.  This parameter controls that threadpool size.  The QueryActor is just handling the API and passing off the work - it is doing next to nothing now. 
The whole point of this change is to offload work from the QueryActor receiving thread to a separate threadpool that we can easily profile and control.

PR #294, opened 2019-03-23
misc(core): Make zipkin message maxBytesOut configurable

tjackpaul:
**Pull Request checklist**

Added following two configs:
```
kamon.zipkin.max.requests
kamon.zipkin.message.max.bytes
```

Also, updated default values for those 2 configs as follows:
```
kamon.zipkin.max.requests=128
kamon.zipkin.message.max.bytes=256kb/s
```

### DIFF
--- a/core/src/main/scala/filodb.core/zipkin/Zipkin.scala
+++ b/core/src/main/scala/filodb.core/zipkin/Zipkin.scala
@@ -127,6 +127,8 @@ class ZipkinReporter extends SpanReporter {
     val zipkinEndpoint = Kamon.config().getString(ZipkinEndpoint)
     val zipkinHost = Kamon.config().getString(HostConfigKey)
     val zipkinPort = Kamon.config().getInt(PortConfigKey)
+    val maxRequests = Kamon.config().getInt(MaxRequests)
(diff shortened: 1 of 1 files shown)


core/src/main/scala/filodb.core/zipkin/Zipkin.scala
```diff
@@ -137,8 +139,8 @@ class ZipkinReporter extends SpanReporter {
       OkHttpSender.newBuilder()
         .encoding(Encoding.JSON)
         .endpoint(url)
-        .maxRequests(64)
-        .messageMaxBytes(1024 * 128)
+        .maxRequests(if (maxRequests > 0) maxRequests else 128)
+        .messageMaxBytes(if (messageMaxBytes > 0) messageMaxBytes else 1024 * 256)
```
vishramachandran: Can we add the default to filodb-defaults.conf ?

tjackpaul: Even then we might need this check.. Because the value can be 0 ?

velvia: Ah that's true.  Probably need to check regardless

PR #296, opened 2019-03-26
feat(core): Optimize index recovery for long term retention

vishramachandran:
**Pull Request checklist**

**Current behavior :** 
We attempt to create a TimeSeriesPartition for every partKey in cass index time buckets.
With this reverse lookup of partKey to partId could also be done using partSet.

This was ok for 3 day retention given most partitions are ingesting, but does not scale for long term retention where majority of partitions are not ingesting. If left as is, it would lead to eviction of nodes from the beginning. 

**New behavior :**
* We now have a new partKey->partId on heap map which is used for reverse lookup.
* we create a TimeSeriesPartition object only for currently ingesting partitions.
* We load time buckets in reverse order since we know that last record wins. This way lucene documents do not need to be updated.

The goal of the PR is to
* Reduce number of entries in `partitions` and `partSet` data structures, thereby reducing memory requirements especially for long term datasets. We pay the cost of a temporary map during recovery that holds all partKeys on heap, but is discarded before ingestion starts.
* Optimize the Lucene recovery by scanning through records in reverse order of buckets. This avoids the expensive Lucene upserts - adds become sufficient.

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
+++ b/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
@@ -2,7 +2,6 @@ package filodb.core.memstore
 
 import java.util.concurrent.locks.StampedLock
 
-import scala.collection.mutable
 import scala.concurrent.{Await, ExecutionContext, Future}
 import scala.concurrent.duration._
 import scala.util.{Random, Try}
@@ -434,7 +433,7 @@ class TimeSeriesShard(val dataset: Dataset,
     /* We need this map to track partKey->partId because lucene index cannot be looked up
        using partKey efficiently, and more importantly, it is eventually consistent.
         The map and contents will be garbage collected after we are done with recovery */
-    val partIdMap = new mutable.HashMap[BytesRef, Int]()
+    val partIdMap = debox.Map.empty[BytesRef, Int]
 
     val earliestTimeBucket = Math.max(0, currentIndexTimeBucket - numTimeBucketsToRetain)
     logger.info(s"Recovering timebuckets $earliestTimeBucket until $currentIndexTimeBucket " +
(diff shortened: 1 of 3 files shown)


core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -459,44 +468,63 @@ class TimeSeriesShard(val dataset: Dataset,
...
+        val partId = if (endTime == Long.MaxValue) {
+          // this is an actively ingesting partition
+          val group = partKeyGroup(dataset.partKeySchema, partKeyBaseOnHeap, partKeyOffset, numGroups)
           val part = createNewPartition(partKeyBaseOnHeap, partKeyOffset, group, CREATE_NEW_PARTID, 4)
           // In theory, we should not get an OutOfMemPartition here since
           // it should have occurred before node failed too, and with data sropped,
           // index would not be updated. But if for some reason we see it, drop data
           if (part == OutOfMemPartition) {
-            logger.error("Could not accommodate partKey while recovering index. " +
+            logger.error("Could not accommodate partKey $while recovering index. " +
```
broneill: Uh, $while??

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -459,44 +468,63 @@ class TimeSeriesShard(val dataset: Dataset,
...
+          // but no need to create a new TSPartition heap object
+          val id = nextPartitionID
+          incrementPartitionID()
+          Some(id)
         }
+
+        // add newly assigned partId to lucene index
+        partId.foreach { partId =>
+          partIdMap.put(partKeyBytesRef, partId)
+          partKeyIndex.addPartKey(partKeyBaseOnHeap, partId, startTime, endTime,
```
velvia: If the partID exists already, why are we using addPartKey?   ie why is this add instead of the upsert?

vishramachandran:
@velvia partIdMap is only used during index recovery time. It is thrown away and GCed after recovery is done. partKeyBaseOnHeap is also GCed. Hope that answers your question.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
vishramachandran: partId is an Option. It is Some if a new partId was created for the partKey.In this case, we need to add it into the lucene index.  It is None it means the partId creation was not done, so no action is taken.

It is an add because we iterate through the buckets in reverse order, and first one wins. We dont need to do an upsert because we do not try to action on same partId twice. Last one wins.

vishramachandran:
@velvia I have updated goal of PR in the PR description.

vishramachandran:
@velvia addressed your comments. Yes - The multiple records per partKey was tested in an E2E run.

PR #297, opened 2019-03-26
 feat(query): aggr-over-vectors: Up to 2x speedup for no grouping

velvia:
**Pull Request checklist**

`RangeVectorAggregator` aggregates one window at a time; for each window it iterates through all the source RangeVectors.   When the number of RangeVectors is large, this uses up a huge amount of memory as various iterators need to be held in memory for each RangeVector, plus locks held for the entire duration of the query.

**New behavior :**

An optimized `RangeVectorAggregator.fastReduce` for the major use case where there is no grouping needed and the source is a `PeriodicSamplesMapper` is used.  `fastReduce` aggregates one time series at a time, each time going through all the windows.   The speedup (up to 2x) comes from preserving cache locality of scans (by iterating through each Time series) and simplifying hasNext logic.  In addition, for high cardinality queries this uses significantly less memory.  Also this design minimizes lock contention as each Time series chunkMap only needs to be locked when iterating through that time series, as opposed to the previous design which would have locked all the ChunkMaps for the duration of the entire query.

Original benchmark:
```
info] Benchmark                                                   Mode  Cnt      Score     Error  Units
[info] QueryHiCardInMemoryBenchmark.scanSumOfRateBenchmark         avgt   40  14907.748 ± 582.749  us/op
[info] QueryHiCardInMemoryBenchmark.scanSumOfSumOverTimeBenchmark  avgt   40   1382.022 ±  42.530  us/op
```

After this change:
```
[info] Benchmark                                                   Mode  Cnt     Score     Error  Units
[info] QueryHiCardInMemoryBenchmark.scanSumOfRateBenchmark         avgt   40  9996.798 ± 372.992  us/op
[info] QueryHiCardInMemoryBenchmark.scanSumOfSumOverTimeBenchmark  avgt   40   638.851 ±  28.911  us/op
```


query/src/main/scala/filodb/query/exec/AggrOverRangeVectors.scala
```diff
@@ -164,6 +171,54 @@ object RangeVectorAggregator {
...
+
+  /**
+   * A fast reduce method intended specifically for the case when no grouping needs to be done AND
+   * the previous transformer is a PeriodicSampleMapper with fixed output lengths.
+   * It's much faster than mapReduce() since it iterates through each vector first and then from vector to vector.
+   * Time wise first iteration also uses less memory for high-cardinality use cases and reduces the
+   * time window of holding chunk map locks to each time series, instead of the entire query.
+   */
+  def fastReduce(rowAgg: RowAggregator,
+                 skipMapPhase: Boolean,
```
broneill: It looks like skipMapPhase is always false for now. When would it be true?

core/src/main/scala/filodb.core/query/RangeVector.scala
```diff
@@ -81,6 +81,8 @@ object CustomRangeVectorKey {
   }
 
   val emptyAsZcUtf8 = toZcUtf8(CustomRangeVectorKey(Map.empty))
+
+  val empty = CustomRangeVectorKey(Map.empty)
```
vishramachandran: Can place this on line 82 and use in line 85

core/src/main/scala/filodb.core/query/ResultTypes.scala
```diff
@@ -27,10 +27,12 @@ final case class ColumnInfo(name: String, colType: Column.ColumnType)
 /**
  * Describes the full schema of result types, including how many initial columns are for row keys.
  * The first ColumnInfo in the schema describes the first vector in Vectors and first field in Tuples, etc.
- * @param brSchemas if any of the columns is a binary record, thsi
+ * @param brSchemas if any of the columns is a BinaryRecord: map of colNo -> inner BinaryRecord schema
+ * @param fixedVectorLen if defined, each vector is guaranteed to have exactly this many output elements
```
vishramachandran: Not clear to me in what situations it is defined and not defined? Can we document known scenarios.

query/src/main/scala/filodb/query/exec/AggrOverRangeVectors.scala
```diff
@@ -64,14 +65,21 @@ final case class AggregateMapReduce(aggrOp: AggregationOperator,
...
       val groupBy: Map[ZeroCopyUTF8String, ZeroCopyUTF8String] =
         if (by.nonEmpty) rv.key.labelValues.filter(lv => byLabels.contains(lv._1))
         else if (without.nonEmpty) rv.key.labelValues.filterNot(lv =>withoutLabels.contains(lv._1))
         else Map.empty
       CustomRangeVectorKey(groupBy)
     }
-    RangeVectorAggregator.mapReduce(aggregator, skipMapPhase = false, source, grouping)
+
+    // IF no grouping is done AND prev transformer is Periodic (has fixed length), use optimal path
+    sourceSchema.fixedVectorLen.filter { _ => without.isEmpty && by.isEmpty }.map { fixedLen =>
```
vishramachandran: The filter usage here isnt readable because you are ignoring the func param.
Can we rephrase the expression using simple if else?  Below is much more readable and less cryptic.

```
if (without.isEmpty && by.isEmpty && sourceSchema.fixedVectorLen.isDefined)
```

query/src/main/scala/filodb/query/exec/AggrOverRangeVectors.scala
```diff
@@ -164,6 +171,54 @@ object RangeVectorAggregator {
...
+   * It's much faster than mapReduce() since it iterates through each vector first and then from vector to vector.
+   * Time wise first iteration also uses less memory for high-cardinality use cases and reduces the
+   * time window of holding chunk map locks to each time series, instead of the entire query.
+   */
+  def fastReduce(rowAgg: RowAggregator,
+                 skipMapPhase: Boolean,
+                 source: Observable[RangeVector],
+                 outputLen: Int): Observable[RangeVector] = {
+    // Can't use an Array here because rowAgg.AggHolderType does not have a ClassTag
+    val accs = collection.mutable.ArrayBuffer.fill(outputLen)(rowAgg.zero)
```
vishramachandran: Memory used is O(steps in the query).
What if use requests a long time duration. Shouldnt this optimization be applied only if small time ranges are chosen?

jmh/src/main/scala/filodb.jmh/QueryHiCardInMemoryBenchmark.scala
```diff
@@ -125,7 +125,7 @@ class QueryLeafScanInMemoryBenchmark extends StrictLogging {
   @Benchmark
   @BenchmarkMode(Array(Mode.AverageTime))
   @OutputTimeUnit(TimeUnit.MICROSECONDS)
-  @OperationsPerInvocation(500)
+  @OperationsPerInvocation(100)
   def scanSumOfRateBenchmark(): Unit = {
     (0 until numQueries).foreach { _ =>
       Await.result(scanSumOfRate.execute(store, dataset, queryConfig).runAsync, 60.seconds)
```
vishramachandran: You should compare results for other aggregations too. Not just sum.

Note that top-k and quantile aggregations use more memory for the accumulators.

jmh/src/main/scala/filodb.jmh/QueryHiCardInMemoryBenchmark.scala
```diff
@@ -36,9 +37,9 @@ class QueryLeafScanInMemoryBenchmark extends StrictLogging {
   import NodeClusterActor._
 
   val numShards = 1
-  val numSeries = 200
-  val samplesDuration = 2.hours
-  val numSamples = numSeries * (samplesDuration / 10.seconds).toInt
+  val numSeries = 8000   // NOTE: num of time series queried is this / 4
+  val samplesDuration = 15.minutes
+  val numSamples = (samplesDuration / 10.seconds).toInt   // # samples PER TIME SERIES
```
vishramachandran: Please add additional benchmarks for longer query intervals with very significantly larger number of steps.

query/src/main/scala/filodb/query/exec/AggrOverRangeVectors.scala
velvia: I believe `skipMapPhase` is true in higher level reduces - ie the phase that takes intermediate aggregates from each leaf/shard and aggregates them further.

query/src/main/scala/filodb/query/exec/AggrOverRangeVectors.scala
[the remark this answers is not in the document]
velvia: Doing that would require more work and you would need more memory for intermediate group results too, so benefits are less clear.  So that's left to another PR.

core/src/main/scala/filodb.core/query/ResultTypes.scala
velvia: ok

velvia:
@vishramachandran  I ran the `QueryInMemoryBenchmark` which tests a variety of queries, with the setting of 120 steps (2 hours at 60 second steps).   In develop:

```
[info] Benchmark                                               Mode  Cnt     Score     Error  Units
[info] QueryInMemoryBenchmark.noOverlapQueries                thrpt   40  5611.396 ± 561.730  ops/s
[info] QueryInMemoryBenchmark.singleThreadedMinOverTimeQuery  thrpt   40  2823.361 ± 150.729  ops/s
[info] QueryInMemoryBenchmark.singleThreadedRawQuery          thrpt   40  6245.082 ±  80.362  ops/s
[info] QueryInMemoryBenchmark.singleThreadedSumRateCCQuery    thrpt   40  1721.710 ±  65.876  ops/s
[info] QueryInMemoryBenchmark.someOverlapQueries              thrpt   40  3119.790 ± 192.175  ops/s
```

On this branch:
```
[info] Benchmark                                               Mode  Cnt     Score     Error  Units
[info] QueryInMemoryBenchmark.noOverlapQueries                thrpt   40  6405.271 ± 279.468  ops/s
[info] QueryInMemoryBenchmark.singleThreadedMinOverTimeQuery  thrpt   40  2735.856 ±  23.708  ops/s
[info] QueryInMemoryBenchmark.singleThreadedRawQuery          thrpt   40  6131.428 ± 158.662  ops/s
[info] QueryInMemoryBenchmark.singleThreadedSumRateCCQuery    thrpt   40  2075.833 ±  24.181  ops/s
[info] QueryInMemoryBenchmark.someOverlapQueries              thrpt   40  3420.781 ±  92.350  ops/s
``` 

So, even at many more steps, it remains an improvement (and rate is slow, which lowers the overall improvement).

I agree that the more steps the query has, the more memory this uses and likely decreases the advantage.  Thus I will put in a query config option and only activate it when the # steps is below a configurable limit.

query/src/main/scala/filodb/query/exec/AggrOverRangeVectors.scala
velvia: Added a configurable max # of windows when the fastReduce path is chosen instead of the normal one.

velvia:
@vishramachandran comments addressed.  New configuration option added to choose when `fastReduce` path is chosen.

velvia:
@vishramachandran I'm reducing the default and merging.  Note that under no circumstance should `fastReduce` be slower than the other path, though it might use more memory.   However, we can tune more later.

PR #316, opened 2019-04-11
feat(http, coordinator): explainPlan support in FiloDB HTTP

TanviBhavsar:
**Pull Request checklist**

**New behavior :**
 Adds ability to get ExecPlan from FiloDB HTTP

### DIFF
--- a/query/src/main/scala/filodb/query/exec/ExecPlan.scala
+++ b/query/src/main/scala/filodb/query/exec/ExecPlan.scala
@@ -1,21 +1,20 @@
 package filodb.query.exec
 
-import scala.collection.mutable.ArrayBuffer
-import scala.concurrent.duration.FiniteDuration
-import scala.util.control.NonFatal
-
-import kamon.Kamon
-import monix.eval.Task
-import monix.execution.Scheduler
-import monix.reactive.Observable
-
 import filodb.core.DatasetRef
 import filodb.core.metadata.Dataset
 import filodb.core.query.{RangeVector, ResultSchema, SerializableRangeVector}
 import filodb.core.store.ChunkSource
 import filodb.memory.format.RowReader
-import filodb.query._
 import filodb.query.Query.qLogger
+import filodb.query._
+import kamon.Kamon
+import monix.eval.Task
+import monix.execution.Scheduler
+import monix.reactive.Observable
+
+import scala.collection.mutable.ArrayBuffer
+import scala.concurrent.duration.FiniteDuration
+import scala.util.control.NonFatal
 
 /**
   * This is the Execution Plan tree node interface.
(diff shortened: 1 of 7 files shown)


TanviBhavsar:
> 👍
> 
> Can you document in your PR how you tested this change? Was it local E2E test?

Yes I tested it locally using explainPlan = true

velvia:
Oh @TanviBhavsar  it would be great to have a unit test as well.

vishramachandran:
@velvia The reason it was not a separate route was because we'd have one route for each type of query
* query
* query_range
* metadata
etc.

One could argue that the response schema is not vastly different. It just has an additional field "debuginfo". We could also aim for printing the plan along with the result in the future.

velvia:
@TanviBhavsar ok if we go with param instead of separate route, then please rename the param to `explainOnly` as this is more accurate - implies there is no query

TanviBhavsar:
> @TanviBhavsar ok if we go with param instead of separate route, then please rename the param to `explainOnly` as this is more accurate - implies there is no query



> @TanviBhavsar ok if we go with param instead of separate route, then please rename the param to `explainOnly` as this is more accurate - implies there is no query

@velvia Changed explainPlan to explainOnly. Also updated doc/http_api.md

query/src/main/scala/filodb/query/exec/ExecPlan.scala
```diff
@@ -198,6 +198,16 @@ trait ExecPlan extends QueryCommand {
     ((transf :+ curNode) ++ childr).mkString(if (useNewline) "\n" else " @@@ ")
   }
 
+  final def getPlan(level: Int = 0): Seq[String] = {
+    val transf = rangeVectorTransformers.reverse.zipWithIndex.map { case (t, i) =>
+      s"${"-"*(level + i)}T~${t.getClass.getSimpleName}(${t.args})"
```
velvia: Might want to put this part which is shared in a function

velvia:
@TanviBhavsar good work and thanks for the test and changes!   Will approve after explain param is changed back to explainOnly .... not sure if it was accidentally changed back in the last commit.

query/src/main/scala/filodb/query/exec/ExecPlan.scala
TanviBhavsar: Missed this before. Added function to print RangeVectorTransformers

TanviBhavsar:
> @TanviBhavsar good work and thanks for the test and changes! Will approve after explain param is changed back to explainOnly .... not sure if it was accidentally changed back in the last commit.

Thanks Evan! I have made the changes

velvia:
@TanviBhavsar thanks, I already approved but one more request ...

The http_api.md doc changes imply only the `query_range` route supports the param, but query also right?  Instead of creating a separate entry, I'd add a note to both the query and query_range sections that the param `explainOnly` returns exec plan only.  Thanks

TanviBhavsar:
> @TanviBhavsar thanks, I already approved but one more request ...
> 
> The http_api.md doc changes imply only the `query_range` route supports the param, but query also right? Instead of creating a separate entry, I'd add a note to both the query and query_range sections that the param `explainOnly` returns exec plan only. Thanks

Removed separate entry and added param to  query and query_range section

PR #322, opened 2019-04-16
bug(query) : comparison operator returns metric even when no values are present

TanviBhavsar:
**Pull Request checklist**

**Current behavior :** 
Output has RangeVector with metric even when there are no rows present
 ```
{
        "metric": {
          "job": "Test-data-1554428389670",
          "instance": "inst-8",
          "service": "service-9",
          "_ns": "Test-data-1554428389670",
          "code": "566",
          "version": "3.1.86",
          "__name__": "my_counter",
          "mode": "user"
        },
        "values": [
          
        ]
 }
```

**New behavior :**
Remove range vectors which don't have any values
```
{
  "status": "success",
  "data": {
    "resultType": "matrix",
    "result": [
      
    ]
  }
}
```

### DIFF
--- a/prometheus/src/main/scala/filodb/prometheus/query/PrometheusModel.scala
+++ b/prometheus/src/main/scala/filodb/prometheus/query/PrometheusModel.scala
@@ -78,7 +78,7 @@ object PrometheusModel {
 
   def toPromSuccessResponse(qr: filodb.query.QueryResult, verbose: Boolean): SuccessResponse = {
     SuccessResponse(Data(toPromResultType(qr.resultType),
-      qr.result.map(toPromResult(_, verbose)).filter(_.values.size > 0)))
+      qr.result.map(toPromResult(_, verbose)).filter(_.values.nonEmpty)))
   }
 
   def toPromResultType(r: QueryResultType): String = {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: prometheus/src/main/scala/filodb/prometheus/query/PrometheusModel.scala [hunk 1]
```diff
@@ -77,7 +76,8 @@ object PrometheusModel {
   }
 
   def toPromSuccessResponse(qr: filodb.query.QueryResult, verbose: Boolean): SuccessResponse = {
-    SuccessResponse(Data(toPromResultType(qr.resultType), qr.result.map(toPromResult(_, verbose))))
+    SuccessResponse(Data(toPromResultType(qr.resultType),
+      qr.result.map(toPromResult(_, verbose)).filter(_.values.size > 0)))
```
vishramachandran: Can we filter earlier in the transformation pipeline as follows:

```
qr.result.filter(_.rows.nonEmpty).map(toPromResult(_, verbose)))
```

vishramachandran: Ignore earlier comment. We would never have empty results. It would have NaNs.

That said, it is better to use `_.values.nonEmpty` instead of  `_.values.size > 0`

PR #324, opened 2019-04-16
feat(config): Better default buckets for Kamon-Prometheus

velvia:
Better default buckets for Kamon-prometheus.  Larger range for time tracing and more buckets for other histograms.

### DIFF
--- a/core/src/main/resources/filodb-defaults.conf
+++ b/core/src/main/resources/filodb-defaults.conf
@@ -358,13 +358,16 @@ kamon {
       16384,
       32768,
       65536,
-      128000,
-      256000,
-      512000
+      131072,
+      262144,
+      524288
     ]
 
-    # Start at 0.1ms so we can measure smaller timings
+    # Start at 0.01ms so we can measure smaller timings
     time-buckets = [
+      0.00001,
+      0.000025,
+      0.00005,
       0.0001,
       0.00025,
       0.0005,
--- a/core/src/main/resources/filodb-defaults.conf
+++ b/core/src/main/resources/filodb-defaults.conf
@@ -338,3 +338,53 @@ kamon.zipkin {
   max.requests = 128
   message.max.bytes = 262144
 }
+
+kamon {
+  prometheus.buckets {
+    # Have more buckets, better resolution really helps esp with heatmaps
+    default-buckets = [
+      4,
+      8,
+      16,
+      32,
+      64,
+      128,
+      256,
+      512,
+      1024,
+      2048,
+      4096,
+      8192,
+      16384,
+      32768,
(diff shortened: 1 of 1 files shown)


core/src/main/resources/filodb-defaults.conf
```diff
@@ -338,3 +338,53 @@ kamon.zipkin {
...
+      32768,
+      65536,
+      128000,
+      256000,
+      512000
+    ]
+
+    # Start at 0.1ms so we can measure smaller timings
+    time-buckets = [
+      0.0001,
```
vishramachandran: Shouldnt we go lower since we measure sub-ms latencies like actor mailbox time, etc. 
Our overall query latency is a few ms, and we have several spans within. So we should start lower.

core/src/main/resources/filodb-defaults.conf
```diff
@@ -338,3 +338,53 @@ kamon.zipkin {
...
+      256,
+      512,
+      1024,
+      2048,
+      4096,
+      8192,
+      16384,
+      32768,
+      65536,
+      128000,
```
broneill: The last 3 buckets aren't proper doubles. It should be 65536 * 2, ...

core/src/main/resources/filodb-defaults.conf
```diff
@@ -338,3 +338,53 @@ kamon.zipkin {
...
+      0.001,
+      0.0025,
+      0.005,
+      0.01,
+      0.025,
+      0.05,
+      0.1,
+      0.25,
+      0.5,
+      1,
```
broneill: These buckets don't seem to have uniform ratios.

velvia: It's OK to not have uniform ratios.  We can make them uniform I guess for slightly better distribution, ultimately it's what makes sense to us.

broneill: What makes "sense" is bit subjective. We don't have any data to know what the "best" defaults are. My thought is that the ratios should at least be uniform, so it seems more obvious that the defaults aren't that well thought out. The numbers aren't as magical.

core/src/main/resources/filodb-defaults.conf
velvia: Yes this starts at 0.1ms, but I can make it go even lower.

@vishramachandran BTW the overall lesson is that useful buckets often need more than 20 buckets.  We should expect 30-40 buckets for histograms.

velvia:
@vishramachandran  @broneill  most comments addressed, other than proper ratios for time buckets, which I think are reasonable right now.  Going to merge

PR #333, opened 2019-04-26
feat(core,kafka): Detailed ingestion logging for matching subset of records

velvia:
**Pull Request checklist**

**New behavior :**

- Filter for `TestConsumer` to dump out raw Kafka ingestion records when a string partition key field matches a given value
- A new dataset/store config that when enabled allows detailed logging of each sample ingested when the partition key has fields matching the config

### DIFF
--- a/conf/timeseries-dev-source.conf
+++ b/conf/timeseries-dev-source.conf
@@ -71,6 +71,11 @@
         # Thats about 6MB
         evicted-pk-bloom-filter-capacity = 50000
 
+        # Uncomment to log at DEBUG ingested samples matching certain filters
+        # Only works for StringColumn fields in the partition key
+        # trace-filters = {
+        #   metric = "bad-metric-to-log"
+        # }
       }
       downsample {
         # can be disabled by setting this flag to false
--- a/conf/timeseries-dev-source.conf
+++ b/conf/timeseries-dev-source.conf
@@ -71,8 +71,8 @@
         # Thats about 6MB
         evicted-pk-bloom-filter-capacity = 50000
 
-        # Uncomment to log at DEBUG ingested samples matching certain filters
-        # Only works for StringColumn fields in the partition key
+        # Uncomment to log at DEBUG ingested samples matching these filters on Partition Key columns
+        # Only works for StringColumn fields in the partition key. Scope may be expanded later.
         # trace-filters = {
         #   metric = "bad-metric-to-log"
         # }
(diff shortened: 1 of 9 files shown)


core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
```diff
@@ -392,6 +392,34 @@ extends ChunkMap(memFactory, initMapSize) with ReadablePartition {
...
+ */
+class TracingTimeSeriesPartition(partID: Int,
+                                 dataset: Dataset,
+                                 partitionKey: BinaryRegion.NativePointer,
+                                 shard: Int,
+                                 bufferPool: WriteBufferPool,
+                                 shardStats: TimeSeriesShardStats,
+                                 memFactory: MemFactory,
+                                 initMapSize: Int) extends
+TimeSeriesPartition(partID, dataset, partitionKey, shard, bufferPool, shardStats, memFactory, initMapSize) {
```
vishramachandran: Prefer composition /  decorator pattern over inheritance.  
Inheritance leads to bad problems later on (remember Stack extends from Vector in java) ? One such place we already have  in code is OnDemandPagingShard which should have not extended TimeSeriesShard.

However, I dont have an alternate solution with less code changes without adding more memory usage for TSP. :(

Please add a comments so readers understand.

conf/timeseries-dev-source.conf
```diff
@@ -72,6 +72,11 @@
         # Thats about 6MB
         evicted-pk-bloom-filter-capacity = 50000
 
+        # Uncomment to log at DEBUG ingested samples matching certain filters
+        # Only works for StringColumn fields in the partition key
+        # trace-filters = {
+        #   metric = "bad-metric-to-log"
```
vishramachandran: Should we include _ns in the filter as well?
If the filter needs to be on a popular metric name, too much debug  can happen.

core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
velvia: The reason why I used inheritance was not so much memory but just ease of implementation.  With composition we'd need to add in tons of methods and clutter things up quite a bit.  If it simply implemented `ReadablePartition` that might break things in a bunch of places.  So best way to keep changes small and balance out different needs.

conf/timeseries-dev-source.conf
velvia: For the current debug case, `StringColumn` filtering is fine, we don't need filtering by Map/labels.  We can extend it later if that is needed.  Let's keep it simple.

PR #334, opened 2019-04-26
feat(core, config): Simplification of Dataset Setup via config

vishramachandran:
**Pull Request checklist**

Require following steps for dataset creation
* initDataset
* createDataset
* setupDataset 

In addition, cluster singleton would send out setup dataset messages on start shard ingestion events.

Third, there was a limitation that spare nodes cannot handle queries since the Query actors are not setup. There is a workaround in place for that, but it is temporary

**New behavior :**

* Remove dataset creation/setup from CLI
* Dataset is now configured via config supplied as part of application.conf
* Ingestion source config also contains the dataset definition. The source path is included in the server config.
* Every NodeCoordinator will setup every dataset from server config on startup.
* IngestionConfig table and DatasetTable have been removed from C* schema.

New Sequence of Events:
* Node Cluster Actor loads all the datasets from this config and initializes ShardManager.
* Node Cluster Actor sends out a CoordinatorRegistered to the registering coodinator.
* Upon receiving that message coordinator initializes all the datasets on the node.
* ShardManager now does not send SetupDataset messages during shard assignment.

Unit Testing:
* SetupDataset case class is now used only for unit testing
* Some unit tests need to initialize both NodeCoodinatorActor and NodeClusterActor with the setup message if the dataset config is not present in the server config.

**BREAKING CHANGES**

* Significant changes in this PR.
* Need config changes for deployment
* Cannot do rolling upgrades

Benchmarks are still working:

Before
```
[info] Benchmark                                                        Mode  Cnt      Score     Error  Units
[info] QueryAndIngestBenchmark.parallelQueries                         thrpt   45    298.498 ±  17.089  ops/s
[info] QueryAndIngestBenchmark.parallelQueries:JFR                     thrpt             NaN              N/A
[info] QueryInMemoryBenchmark.noOverlapQueries                         thrpt   45   9470.662 ± 200.798  ops/s
[info] QueryInMemoryBenchmark.noOverlapQueries:JFR                     thrpt             NaN              N/A
[info] QueryInMemoryBenchmark.singleThreadedMinOverTimeQuery           thrpt   45   9290.641 ±  97.687  ops/s
[info] QueryInMemoryBenchmark.singleThreadedMinOverTimeQuery:JFR       thrpt             NaN              N/A
[info] QueryInMemoryBenchmark.singleThreadedRawQuery                   thrpt   45  12795.824 ± 246.735  ops/s
[info] QueryInMemoryBenchmark.singleThreadedRawQuery:JFR               thrpt             NaN              N/A
[info] QueryInMemoryBenchmark.singleThreadedSumRateCCQuery             thrpt   45   3896.019 ± 109.958  ops/s
[info] QueryInMemoryBenchmark.singleThreadedSumRateCCQuery:JFR         thrpt             NaN              N/A
[info] QueryInMemoryBenchmark.someOverlapQueries                       thrpt   45   8372.649 ± 192.382  ops/s
[info] QueryInMemoryBenchmark.someOverlapQueries:JFR                   thrpt             NaN              N/A
[info] QueryOnDemandBenchmark.parallelQueries                          thrpt   45    683.805 ± 114.539  ops/s
[info] QueryOnDemandBenchmark.parallelQueries:JFR                      thrpt             NaN              N/A
[info] QueryHiCardInMemoryBenchmark.scanSumOfRateBenchmark              avgt   45  15968.221 ± 926.313  us/op
[info] QueryHiCardInMemoryBenchmark.scanSumOfRateBenchmark:JFR          avgt             NaN              N/A
[info] QueryHiCardInMemoryBenchmark.scanSumOfSumOverTimeBenchmark       avgt   45   1953.259 ± 294.274  us/op
[info] QueryHiCardInMemoryBenchmark.scanSumOfSumOverTimeBenchmark:JFR   avgt             NaN              N/A
[info]
```

After

```
[info] Benchmark                                                        Mode  Cnt      Score      Error  Units
[info] QueryAndIngestBenchmark.parallelQueries                         thrpt   45    289.074 ±   16.577  ops/s
[info] QueryAndIngestBenchmark.parallelQueries:JFR                     thrpt             NaN               N/A
[info] QueryInMemoryBenchmark.noOverlapQueries                         thrpt   45   8653.319 ±  803.506  ops/s
[info] QueryInMemoryBenchmark.noOverlapQueries:JFR                     thrpt             NaN               N/A
[info] QueryInMemoryBenchmark.singleThreadedMinOverTimeQuery           thrpt   45   9000.817 ±  369.030  ops/s
[info] QueryInMemoryBenchmark.singleThreadedMinOverTimeQuery:JFR       thrpt             NaN               N/A
[info] QueryInMemoryBenchmark.singleThreadedRawQuery                   thrpt   45  12921.314 ±  221.268  ops/s
[info] QueryInMemoryBenchmark.singleThreadedRawQuery:JFR               thrpt             NaN               N/A
[info] QueryInMemoryBenchmark.singleThreadedSumRateCCQuery             thrpt   45   4136.640 ±   23.669  ops/s
[info] QueryInMemoryBenchmark.singleThreadedSumRateCCQuery:JFR         thrpt             NaN               N/A
[info] QueryInMemoryBenchmark.someOverlapQueries                       thrpt   45   8769.869 ±  117.805  ops/s
[info] QueryInMemoryBenchmark.someOverlapQueries:JFR                   thrpt             NaN               N/A
[info] QueryOnDemandBenchmark.parallelQueries                          thrpt   45    949.192 ±   51.742  ops/s
[info] QueryOnDemandBenchmark.parallelQueries:JFR                      thrpt             NaN               N/A
[info] QueryHiCardInMemoryBenchmark.scanSumOfRateBenchmark              avgt   45  15807.801 ± 1566.297  us/op
[info] QueryHiCardInMemoryBenchmark.scanSumOfRateBenchmark:JFR          avgt             NaN               N/A
[info] QueryHiCardInMemoryBenchmark.scanSumOfSumOverTimeBenchmark       avgt   45   1887.235 ±   90.744  us/op
[info] QueryHiCardInMemoryBenchmark.scanSumOfSumOverTimeBenchmark:JFR   avgt             NaN               N/A


```


http/src/main/scala/filodb/http/FiloHttpServer.scala
```diff
@@ -39,9 +42,14 @@ class FiloHttpServer(actorSystem: ActorSystem) extends StrictLogging {
...
       settings.httpServerBindHost,
       settings.httpServerBindPort)
-    val bind = Await.result(bindingFuture,
+    binding = Await.result(bindingFuture,
       scala.concurrent.duration.Duration.fromNanos(settings.httpServerStartTimeout.toNanos))
-    logger.info("FiloDB HTTP server is live at http:/{}/", bind.localAddress)
+    logger.info("FiloDB HTTP server is live at http:/{}/", binding.localAddress)
+  }
+
+  def shutdown(hardDeadline: FiniteDuration): Future[Http.HttpTerminated] = {
```
broneill: Is this an unrelated change? Why is it needed?

standalone/src/main/scala/filodb.standalone/FiloServer.scala
```diff
@@ -72,6 +75,11 @@ class FiloServer(watcher: Option[ActorRef]) extends FilodbClusterNode {
     }
   }
 
+  override def shutdown(): Unit = {
```
broneill: Why is this needed?

http/src/main/scala/filodb/http/FiloHttpServer.scala
vishramachandran: This is part of some trial and errors to get IngestionAndRecoverySpec multi-jvm spec to run.
I am finding that the ports are not being released in the restart step of the tests.

standalone/src/main/scala/filodb.standalone/FiloServer.scala
vishramachandran: Same as above
> This is part of some trial and errors to get IngestionAndRecoverySpec multi-jvm spec to run.
I am finding that the ports are not being released in the restart step of the tests.

http/src/main/scala/filodb/http/FiloHttpServer.scala
velvia: Rather than shutting down previous ports we should just try to allocate a different, unique port for each test.

cli/src/main/scala/filodb.cli/CsvImportExport.scala
```diff
@@ -38,16 +36,16 @@ trait CsvImportExport extends StrictLogging {
...
-    client.setupDataset(dataset,
-                        DatasetResourceSpec(1, 1),
-                        IngestionSource(classOf[CsvStreamFactory].getName, config),
-                        StoreConfig(storeConf)).foreach {
-      case e: ErrorResponse =>
-        println(s"Errors setting up ingestion: $e")
-        exitCode = 2
-        return
-    }
+// FIXME ****** Need to fix this
```
velvia: Why don't you just get rid of the entire CSV ImportExport thing.  It was for old FiloDB and really doesn't make sense anymore.

velvia: Especially since you are already doing lots of CLI cleanup.  :)

conf/timeseries-128shards-source.conf
```diff
@@ -1,4 +1,21 @@
     dataset = "prometheus"
+
+    definition {
+      partition-columns = ["tags:map"]
+      data-columns = ["timestamp:ts", "value:double"]
+      row-key-columns = [ "timestamp" ]
```
velvia: Can we have a default for this so it doesn't need to be included every time?

conf/timeseries-filodb-server.conf
```diff
@@ -5,6 +5,10 @@ filodb {
     port = 9042
     partition-list-num-groups = 1
   }
+  dataset-configs = [
```
velvia: Add comments that this defines each dataset and ingestion source

coordinator/src/main/scala/filodb.coordinator/NodeClusterActor.scala
```diff
@@ -48,32 +48,32 @@ object NodeClusterActor {
...
+   * @param dataset the dataset to start ingesting
    * @param resources the sharding and number of nodes for ingestion and querying
    * @param source the IngestionSource on each node.  Use noOpSource to not start ingestion and
    *               manually push records into NodeCoordinator.
    * @param storeConfig a StoreConfig for the MemStore.
    * @return DatasetVerified - meaning the dataset and columns are valid.  Does not mean ingestion is
    *                           setup on all nodes - for that, subscribe to ShardMapUpdate's
    */
-  final case class SetupDataset(ref: DatasetRef,
+  final case class SetupDataset(dataset: Dataset,
```
velvia: Be sure to test serialization of this.  We had trouble with serializing the Dataset class before.  It might be preferable to serialize using dataset.toString

coordinator/src/main/scala/filodb.coordinator/NodeClusterActor.scala
```diff
@@ -314,14 +311,10 @@ private[filodb] class NodeClusterActor(settings: FilodbSettings,
...
-  // The initial recovery handler: recover dataset setup/ingestion config first
   def datasetHandler: Receive = LoggingReceive {
     case e: SetupDataset =>
-      setupDataset(e, sender()) map { _ => self ! SetupDatasetFinished(e.ref) }
-    case e: SetupDatasetFinished => {
-      initDatasets -= e.ref
-      if (initDatasets.isEmpty) initiateShardStateRecovery()
-    }
+      // used by unit testing only
+      initializeDataset(e.dataset, e.config, Some(sender()))
```
velvia: Why do we even need the `SetupDataset` anymore?  Can we get rid of it entirely?  Tests should be using config too.  We shouldn't have a test-only path here.  Please make the tests use the same init mechanism as everyone else....

core/src/main/scala/filodb.core/metadata/Column.scala
```diff
@@ -27,6 +28,11 @@ trait Column {
   // More type safe than just using ==, if we ever change the type of ColumnId
   // TODO(velvia): remove this and just use id
   def hasId(_id: ColumnId): Boolean = name == _id
+
+  def toStringNotation: String = {
```
velvia: This method is really not necessary, it duplicates code from DataColumn.toString which should be used instead.  Just make Dataset column definitions all DataColumns and you'll be done.

core/src/main/scala/filodb.core/metadata/Dataset.scala
```diff
@@ -204,19 +212,17 @@ object DatasetOptions {
...
-    val Array(database, name, partColStr, dataColStr, rowKeyIndices, downsamplersStr, optStr) =
-      compactStr.split('\u0001')
-    val partitionColumns = partColStr.split(':').toSeq.map(raw => DataColumn.fromString(raw))
-    val dataColumns = dataColStr.split(':').toSeq.map(raw => DataColumn.fromString(raw))
-    val rowKeyIDs = rowKeyIndices.split(':').toSeq.map(_.toInt)
-    val downsamplers = downsamplersStr.split(':').filterNot(_.isEmpty).toSeq.map(ChunkDownsampler.downsampler(_))
-    val databaseOption = if (database == "") None else Some(database)
-    val options = DatasetOptions.fromString(optStr)
-    Dataset(name, partitionColumns, dataColumns, rowKeyIDs, downsamplers, databaseOption, options)
+  def fromConfig(config: Config): Dataset = {
```
velvia: Note: Config is much more expensive to serialize than the String.  Though we probably won't be doing this that often is my guess.

http/src/test/scala/filodb/http/PrometheusApiRouteSpec.scala
```diff
@@ -38,19 +38,12 @@ class PrometheusApiRouteSpec extends FunSpec with ScalatestRouteTest with AsyncT
...
-
-  before {
-    probe.send(cluster.coordinatorActor, NodeProtocol.ResetState)
-    probe.expectMsg(NodeProtocol.StateReset)
-    // Note: at this point all ingestor actors are shut down
-    cluster.metaStore.clearAllData().futureValue
-    cluster.metaStore.newDataset(FormatConversion.dataset).futureValue shouldEqual Success
-    probe.send(clusterProxy, NodeProtocol.ResetState)
-    probe.expectMsg(NodeProtocol.StateReset)
+    val command = SetupDataset(FormatConversion.dataset, DatasetResourceSpec(2, 1), noOpSource, TestData.storeConf)
```
velvia: Again we should just rely on a config file for tests

velvia:
Oh, I would say that Coordinators also should not wait for the CoordinatorRegistered message.  They should just initialize themselves at startup.

coordinator/src/main/scala/filodb.coordinator/NodeClusterActor.scala
vishramachandran: SetupDataset is never transmitted over the wire anymore. We use it only during unit tests. I will add comments as such

coordinator/src/main/scala/filodb.coordinator/NodeClusterActor.scala
vishramachandran: Fixing unit tests to use config was a big change and introduced possibility of regressions. Hence retained the SetupDataset capability there. 

Multi-JVM test like IngestionAndRecoverySpec use the same path as the real servers. They do not use SetupDataset.

core/src/main/scala/filodb.core/metadata/Dataset.scala
vishramachandran: This is only executed at boot time when we load datasets from config.

vishramachandran:
> Oh, I would say that Coordinators also should not wait for the CoordinatorRegistered message. They should just initialize themselves at startup.

There are race conditions with doing it that way. For one, the cluster Singleton may not have been initialized. Remember that statusActor needs reference to the clusterSingleton. Hence this approach is necessary.

http/src/test/scala/filodb/http/PrometheusApiRouteSpec.scala
vishramachandran: I'll am using config files where ever possible like multi-jvm tests. But where we inject datasets manually, it is harder to inject config. I'll try.

conf/timeseries-128shards-source.conf
vishramachandran: I suggest that we not have defaults for dataset configuration. This typically does not change and need to be carefully set at the beginning. Defaults are error prone, and do not help the person understand every configuration fully.

core/src/main/scala/filodb.core/metadata/Column.scala
vishramachandran: The `toStringNotation` is used serialization and deserialization from/to `Config`. I considered using `toString` but that was one direction only. Deserialization was not possible really. At the same time, I didnt want to break any toString invocations happening today.

If you think it is safe to consolidate the two and replace `toString` with implementation of `toStringNotation` I can do so.

cli/src/main/scala/filodb.cli/CsvImportExport.scala
vishramachandran: Done

conf/timeseries-filodb-server.conf
vishramachandran: Done

velvia:
@vishramachandran  hmmm.  I wonder if we should just change the NodeCoordinator to take in an ActorRef to the cluster actor, ie make sure ClusterActor is initialized first.   That would simplify a lot of things including the whole workflow.  The only reason why things are done the current way is that in the CLI, the NodeCoordinator is started up without the ClusterActor, but that was for a use case (CSV ingestion and export) which we aren't going to support anymore.

vishramachandran:
> I wonder if we should just change the NodeCoordinator to take in an ActorRef to the cluster actor, ie make sure ClusterActor is initialized first.

@velvia The cyclic dependency needs to be broken. Shard Manager on cluster singleton depends on NodeCoord registration. NodeCoord cannot wait for singleton to have started. 

Moreover, we have to address the case where singleton and NodeCoord are on different nodes.

core/src/main/scala/filodb.core/metadata/Column.scala
velvia: When does one need to write out Config from a Column?

README.md
```diff
@@ -144,22 +144,24 @@ Build the required projects
 sbt standalone/assembly cli/assembly gateway/assembly
 ```
 
-First set up the dataset. This should create the keyspaces and tables in Cassandra. 
+First initialize the dataset. This should create the keyspaces and tables in Cassandra.
```
velvia: There is no more dataset to initialize.  Maybe you mean "First initialize the C* keyspaces" ?

conf/timeseries-dev-source.conf
```diff
@@ -1,5 +1,27 @@
     dataset = "prometheus"
+
+    # Should not change once dataset has been set up on the server and data has been persisted to cassandra
+    definition {
```
velvia: More comments should be added here.  That this is for defining the schema used for ingesting data and persistence.  Also it is not just when data has been persisted to Cassandra, also once data is written into Kafka.
That changing the definition after data is ingested will lead to corruption and inability to ingest or query data.

Also please add comment for what partition-columns and data-columns are.

README.md
vishramachandran: That is right. This is already clarified in the following sentence.

PR #348, opened 2019-05-02
fix(memory): Fix DeltaDelta bug where slope(int) multiply overflowed

velvia:
**Pull Request checklist**

For some chunks encoded using Delta-Delta, there is a bug resulting from integer overflow where reading the chunk starting at a certain chunk index would lead to incorrect numbers being returned.  Basically the slope, which is a 32-bit integer, was multiplied by an integer and the result was overflow.

**New behavior :**

The slope calculation for DDV is now done using Long's so it won't overflow in the same way.


velvia:
@broneill  because slope cannot be more than a 32-bit number, that is how it is stored.  DDV encoding ensures the slope is no more than 32 bits.  32bit x 32bit will not overflow a Long.

broneill:
It looks a bit odd now, because before the types returned by the field extraction functions match the documentation. Why not just call toLong where necessary? At least it's consistent and doesn't give the odd impression that these functions can actually accept a long slope.

velvia:
@broneill it's easier than doing it in every single place.  I guess I can change it the other way too.

velvia:
@broneill changed to have toLong's conversion everywhere

PR #349, opened 2019-05-03
bug(query): topk and bottomk has trailing value 

TanviBhavsar:
**Pull Request checklist**

**Current behavior :** 
If value exists at timestamp t1 and topk query is fired for time (t1-500) to (t1+500), value at t1 continues till t1+500.  Value at t1 should be there only till t1 + staleSampleAfterMs which is 300 by default. This happens because row is not reset and previous value is carried over.

**New behavior :**
Reset remaining values of row to overwrite previous row value

### DIFF
--- a/query/src/main/scala/filodb/query/exec/AggrOverRangeVectors.scala
+++ b/query/src/main/scala/filodb/query/exec/AggrOverRangeVectors.scala
@@ -580,6 +580,12 @@ class TopBottomKRowAggregator(k: Int, bottomK: Boolean) extends RowAggregator {
         row.setDouble(i + 1, el.value)
         i += 2
       }
+      // Reset remaining values of row to overwrite previous row value
+      while (i < numRowReaderColumns) {
+        row.setString(i, CustomRangeVectorKey.emptyAsZcUtf8)
+        row.setDouble(i + 1, if (bottomK) Double.MaxValue else Double.MinValue)
+        i += 2
+      }
       row
     }
     def resetToZero(): Unit = { heap.clear() }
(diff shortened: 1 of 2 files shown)


vishramachandran:
@TanviBhavsar Can you add a unit test that reproduces the bug: fails before and passes after your fix.

TanviBhavsar:
> @TanviBhavsar Can you add a unit test that reproduces the bug: fails before and passes after your fix.

I could not reproduce the bug in the existing unit test. I have been able to reproduce it locally only after increasing shards to 256. I could not reproduce it even on dev2, have seen it in perf.

TanviBhavsar:
@vishramachandran  I have added unit test. The bug is reproducable only when mapReduce is called on single range vector which has Double.NaN after a non NaN value. 
For example:
val samples: Array[RangeVector] = Array(
      toRv(Seq((1556744143L, 42d), (1556744158L, 42d),(1556744173L, Double.NaN )))
    )
It is not reproducible in existing unit test which has multiple RangeVectors even after NaN is moved to end:
val samples: Array[RangeVector] = Array(
      toRv(Seq((2L, 5.6d),(3L, Double.NaN) )),
      toRv(Seq( (2L, 4.4d),(3L, Double.NaN))),
      toRv(Seq((2L, 5.4d),(3L, Double.NaN)))
    )
The reason is that in first case same row is used for all timestamp without resetting.

TanviBhavsar:
> Nice work getting to root cause, and unit test case!
> Please document the bug a bit in the test case for posterity.

Thanks Vish! I have added comment to the test case

PR #351, opened 2019-05-07
bug(core): Recovery reporting was delayed for final 100% mark

vishramachandran:
**Pull Request checklist**

* Recovery reporting was not happening exactly when offset was reached. Fixed that.
* Added more consistent logging format for shards and partIds

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/IngestionActor.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/IngestionActor.scala
@@ -171,8 +171,8 @@ private[filodb] final class IngestionActor(dataset: Dataset,
                                        checkpoints) }
         yield {
           // Start reading past last offset for normal records; start flushes one group past last group
-          normalIngestion(shard, Some(lastOffset + 1), (lastFlushedGroup + 1) % numGroups,
-                          storeConfig.diskTTLSeconds)
+          normalIngestion(shard, Some(lastOffset.getOrElse(endRecoveryWatermark) + 1),
+                          (lastFlushedGroup + 1) % numGroups, storeConfig.diskTTLSeconds)
         }
       }
     }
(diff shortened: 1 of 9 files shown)


coordinator/src/main/scala/filodb.coordinator/IngestionActor.scala
```diff
@@ -265,9 +265,10 @@ private[filodb] final class IngestionActor(dataset: Dataset,
           off }
         .until(_ >= endOffset)
         // TODO: move this code to TimeSeriesShard itself.  Shard should control the thread
-        .lastL.runAsync(shardInstance.ingestSched)
+        .lastOptionL.runAsync(shardInstance.ingestSched)
```
velvia: When would this ever return None?   Not too clear on semantics of this

core/src/main/scala/filodb.core/memstore/TimeSeriesMemStore.scala
```diff
@@ -140,16 +140,22 @@ extends MemStore with StrictLogging {
...
                     reportingInterval: Long): Observable[Long] = {
     val shard = getShardE(dataset, shardNum)
-    val endOffset = checkpoints.values.max
     shard.setGroupWatermarks(checkpoints)
-    var targetOffset = checkpoints.values.min + reportingInterval
-    stream.map(shard.ingest(_)).collect {
-      case offset: Long if offset > targetOffset || offset >= endOffset =>
-        targetOffset += reportingInterval
-        offset
+    if (endOffset < startOffset) Observable.empty
```
velvia: You don't need to pass in endOffset and startOffset, they are calculatable from checkpoints.

coordinator/src/main/scala/filodb.coordinator/IngestionActor.scala
vishramachandran: See the other changes in commit. None would be returned when endOffset < startOffset.
This is required to eagerly move state from Recovering to Active if there is no data in kafka.

core/src/main/scala/filodb.core/memstore/TimeSeriesMemStore.scala
vishramachandran: I know - I wanted one place where the calculation was done. Hence the parameter is being passed in.

PR #356, opened 2019-05-09
bug(core): Not purging partition if it has unflushed data

vishramachandran:
**Pull Request checklist**

We saw during testing that an intermittently ingesting partition which didnt ingest for retentionPeriod and then ingests could be purged prior to data being flushed because we dont check if unflushed data. 

**New behavior :**

Check if there are unflushed chunks prior to deleting the partition

Also using TermInSetQuery instead of batching - it is more efficient and uses less memory.

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
+++ b/core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
@@ -149,8 +149,8 @@ class PartKeyLuceneIndex(dataset: Dataset,
   }
 
   /**
-    * Use to delete partitions that were ingesting before retention period
-    * @return partIds of deleted partitions
+    * Find partitions that ended ingesting before a given timestamp. Used to identify partitions that can be purged.
+    * @return matching partIds
     */
   def partIdsEndedBefore(endedBefore: Long): EWAHCompressedBitmap = {
     val collector = new PartIdCollector()
@@ -159,9 +159,14 @@ class PartKeyLuceneIndex(dataset: Dataset,
     collector.result
   }
 
-  def removePartKeys(partIds: Array[Int]): Unit = {
+  /**
+    * Delete partitions with given partIds
+    */
+  def removePartKeys(partIds: EWAHCompressedBitmap): Unit = {
     val terms = new util.ArrayList[BytesRef]()
-    partIds.foreach { pId =>
+    val iter = partIds.intIterator()
+    while (iter.hasNext) {
+      val pId = iter.next()
       terms.add(new BytesRef(pId.toString.getBytes))
     }
     indexWriter.deleteDocuments(new TermInSetQuery(PART_ID, terms))
@@ -337,6 +342,8 @@ class PartKeyLuceneIndex(dataset: Dataset,
     partIds.foreach { pId =>
       terms.add(new BytesRef(pId.toString.getBytes))
     }
+    // dont use BooleanQuery which will hit the 1024 term limit. Instead use TermInSetQuery which is
+    // more efficient within Lucene
     searcherManager.acquire().search(new TermInSetQuery(PART_ID, terms), collector)
     span.finish()
     collector.startTimes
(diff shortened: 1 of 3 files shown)


core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -330,13 +333,11 @@ class PartKeyLuceneIndex(dataset: Dataset,
...
     val collector = new PartIdStartTimeCollector()
-    partIds.grouped(512).foreach { batch => // limit on query clause count is 1024, hence batch
-      val booleanQuery = new BooleanQuery.Builder
-      batch.foreach { pId =>
-        booleanQuery.add(new TermQuery(new Term(PART_ID, pId.toString)), Occur.SHOULD)
-      }
-      searcherManager.acquire().search(booleanQuery.build(), collector)
+    val terms = new util.ArrayList[BytesRef]()
+    partIds.foreach { pId =>
+      terms.add(new BytesRef(pId.toString.getBytes))
```
velvia: Is there a way to not need a huge amount of heap for each Int (4 bytes)?
This requires, per partition, an Array[Byte], which is >= 28 bytes, plus another BytesRef which is another 28+ bytes.  Wowzas.

velvia: Also, is it safe to remove the grouping of 512 terms?  How is this code not requiring the previous grouping?

velvia: (please add docs to explain why it's OK)

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -147,21 +148,23 @@ class PartKeyLuceneIndex(dataset: Dataset,
...
+  def partIdsEndedBefore(endedBefore: Long): EWAHCompressedBitmap = {
     val collector = new PartIdCollector()
     val deleteQuery = LongPoint.newRangeQuery(PartKeyLuceneIndex.END_TIME, 0, endedBefore)
     searcherManager.acquire().search(deleteQuery, collector)
-    indexWriter.deleteDocuments(deleteQuery)
-    collector.intIterator()
+    collector.result
+  }
+
+  def removePartKeys(partIds: Array[Int]): Unit = {
```
velvia: Consider passing in another EWAHCompressedBitmap instead of Array[Int], may save space.  I know this is probably for filtering out partitions.  This could be easily done with bitmap math.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -674,16 +674,21 @@ class TimeSeriesShard(val dataset: Dataset,
...
+    val removedParts = new EWAHCompressedBitmap()
+    InMemPartitionIterator(partsToPurge.intIterator()).foreach { p =>
+      if (!p.infosToBeFlushed.hasNext) {
+        logger.debug(s"Purging partition with partId=${p.partID} from memory in dataset=${dataset.ref} shard=$shardNum")
+        removePartition(p)
+        removedParts.set(p.partID)
+        numDeleted += 1
+      }
     }
+    if (!removedParts.isEmpty) partKeyIndex.removePartKeys(removedParts.toArray)
```
velvia: Again, just pass in the `removedParts` and work with it directly.  You can avoid the allocation of the extra array

vishramachandran: will do

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
vishramachandran: Evan, unfortunately we are limited by the capabilities Lucene offers. The TermInSetQuery requires BytesRef.

I have confirmed that TermInSetQuery does not limit number of terms. The unit  test confirms that it can exceed the 1024 limit of BooleanQuery.

I will add docs.

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
vishramachandran: Will do.

PR #357, opened 2019-05-10
bug(core): endTime update for re-ingesting time series should happen immediately

vishramachandran:
**Pull Request checklist**

For intermittently ingesting time series, the endTime gets updated as Long.MaxValue only during index time bucket flush event that happens once in a flush interval. This is too late and queries prior to this event skip this time series in result.

Fix is to update the index immediately after re-ingestion.


core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -1068,7 +1063,17 @@ class TimeSeriesShard(val dataset: Dataset,
     try {
       val part: FiloPartition = getOrAddPartition(recordBase, recordOff, group, ingestOffset)
       if (part == OutOfMemPartition) { disableAddPartitions() }
-      else { part.asInstanceOf[TimeSeriesPartition].ingest(binRecordReader, overflowBlockFactory) }
+      else {
+        part.asInstanceOf[TimeSeriesPartition].ingest(binRecordReader, overflowBlockFactory)
+        activelyIngesting.synchronized {
```
broneill: Instead of full synchronization, the use of a StampedLock is cheaper in the common case. This does require more changes, however.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -933,19 +925,12 @@ class TimeSeriesShard(val dataset: Dataset,
...
-        updatePartEndTimeInIndex(p, Long.MaxValue)
-        timeBucketBitmaps.get(timeBucket).set(p.partID)
-        activelyIngesting.set(p.partID)
-      }
+    if (partFlushChunks.isEmpty && activelyIngesting.get(p.partID)) {
+      var endTime = p.timestampOfLatestSample
+      if (endTime == -1) endTime = System.currentTimeMillis() // this can happen if no sample after reboot
+      updatePartEndTimeInIndex(p, endTime)
+      timeBucketBitmaps.get(timeBucket).set(p.partID)
+      activelyIngesting.clear(p.partID)
```
broneill: Lock should probably held over the get and clear calls together. Perhaps a combined get/clear method? remove?

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -1068,7 +1053,15 @@ class TimeSeriesShard(val dataset: Dataset,
...
       val part: FiloPartition = getOrAddPartition(recordBase, recordOff, group, ingestOffset)
       if (part == OutOfMemPartition) { disableAddPartitions() }
-      else { part.asInstanceOf[TimeSeriesPartition].ingest(binRecordReader, overflowBlockFactory) }
+      else {
+        part.asInstanceOf[TimeSeriesPartition].ingest(binRecordReader, overflowBlockFactory)
+        if (!activelyIngesting.get(part.partID)) {
+          // time series was inactive and has just started re-ingesting
+          updatePartEndTimeInIndex(part.asInstanceOf[TimeSeriesPartition], Long.MaxValue)
+          timeBucketBitmaps.get(currentIndexTimeBucket).set(part.partID)
+          activelyIngesting.set(part.partID)
```
broneill: Need to hold the lock here too, over the two calls. Or an atomic get/set method: insert? setIfNotExists?

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -932,19 +924,18 @@ class TimeSeriesShard(val dataset: Dataset,
...
-        // Partition started re-ingesting.
-        updatePartEndTimeInIndex(p, Long.MaxValue)
-        timeBucketBitmaps.get(timeBucket).set(p.partID)
-        activelyIngesting.set(p.partID)
+    if (partFlushChunks.isEmpty && activelyIngesting.get(p.partID)) {
+      // Below is coded to work concurrently with logic in getOrAddPartitionAndIngest
+      // where we try to activate an idle time series
+      activelyIngesting.withWriteLock {
+        // do not use get/set since activelyIngesting locks are not reentrant
+        if (!activelyIngesting.getWithoutLock(p.partID)) {
```
broneill: Not not? (no ! operator)

vishramachandran: Ugh! copy/paste error. Thanks for the catch

broneill:
@velvia A StampedLock optimistic read lock is effectively free. The performance difference cannot be attributed solely to that. All it does is read a field twice. However, checking the bitmap index does perform some extra steps. One thing to keep in mind however, is that the overall cost of ingestion within the system is fairly tiny. Keeping an activelyIngesting state in the TSPartition is much simpler, and so it's a good idea for that reason alone. However, does the flush task ever access or modify the activelyIngesting state?

vishramachandran:
@velvia Are you sure?
Before:  0.034 ±      error of  0.026   s/op
After:  0.025 ±    error of   0.001   s/op

Can you really draw conclusions from that ? The error possibility is too high from the first run.

There is a reason activelyIngesting needs to be a bitmap. We perform logical operations on the entire bitmap to find out partitions to roll over.

For what it is worth, I ran the benchmark from develop branch again, and here are the results

```
[info] Benchmark                                           Mode  Cnt          Score         Error  Units
[info] IngestionBenchmark.compareBRv2IngestPartition      thrpt   45  228663445.702 ± 1435168.415  ops/s
[info] IngestionBenchmark.compareBRv2IngestPartition:JFR  thrpt                 NaN                  N/A
[info] QueryAndIngestBenchmark.parallelQueries            thrpt   45        280.862 ±      16.202  ops/s
[info] QueryAndIngestBenchmark.parallelQueries:JFR        thrpt                 NaN                  N/A
[info] IngestionBenchmark.ingestEncodeRecords                ss   45          0.026 ±       0.001   s/op
[info] IngestionBenchmark.ingestEncodeRecords:JFR            ss                 NaN                  N/A
```

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -461,6 +461,10 @@ class TimeSeriesShard(val dataset: Dataset,
   def completeIndexRecovery(): Unit = {
     commitPartKeyIndexBlocking()
     startFlushingIndex() // start flushing index now that we have recovered
+    // sleep until all the recovered index entries are readable. Otherwise, duplicate partIds
+    // could be created for same partKey during ingestion. This happens when an intermittent
+    // non-ingesting partition which is not in heap re-ingests before index entry is readable.
+    Thread.sleep(storeConfig.partIndexFlushMaxDelaySeconds.seconds.toMillis + 2.seconds.toMillis)
```
broneill: This is a bit dodgy. Is it possible to sleep in a loop, testing some condition to verify that recovery has completed?

vishramachandran: We could verify that the last record written is readable and loop until then. There is no other way to verify really. 
But if Lucene is adhering to its contract, it should make the records readable within the maxTime we configure it with.

We have a flush thread configured as such:

```
  private val flushThread = new ControlledRealTimeReopenThread(indexWriter,
                                                               searcherManager,
                                                               storeConfig.partIndexFlushMaxDelaySeconds,
                                                               storeConfig.partIndexFlushMinDelaySeconds)
```

broneill: If record based verification is possible, then that's what should be done. The ControlledRealTimeReopenThread is documented as "experimental", and so I wouldn't expect it to correctly honor any contract.

vishramachandran: fixed

velvia:
@vishramachandran  let's discuss offline.  A design that does not require checking of a separate data structure, compared to one that does, is still a win.

PR #364, opened 2019-05-15
feat(core): Return extra WriteBuffer memory back for reuse

velvia:
**Pull Request checklist**

The `WriteBufferPool` allocates more WriteBuffers from native memory upon request.  When a surge of TSPartitions dies down however, and the WriteBuffers are released and returned to the pool, the memory is never released.  The pool retains a very large number of unused WriteBuffers. After a while, we run out of WriteBuffer memory, and this condition does not change.

**New behavior :**

There is a new `max-buffer-pool-size` setting.  When the `WriteBufferPool` is at this size, any buffers that are returned are released back to native memory instead of being retained in the pool.  This allows the native memory usage to shrink as ingesting partitions shrinks and allows memory to be reused for other purposes, such as for other shards.

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
+++ b/core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
@@ -132,7 +132,6 @@ extends ChunkMap(memFactory, initMapSize) with ReadablePartition {
       ChunkSetInfo.setChunkID(infoAddr, currentChunkID)
       ChunkSetInfo.resetNumRows(infoAddr)    // Must reset # rows otherwise it keeps increasing!
       ChunkSetInfo.setStartTime(infoAddr, ts)
-      ChunkSetInfo.setEndTime(infoAddr, ts)
       currentInfo = ChunkSetInfo(infoAddr)
       currentChunks = newAppenders
       // Don't publish the new chunk just yet. Wait until it has one row.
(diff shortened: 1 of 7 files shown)


core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
```diff
@@ -132,6 +132,7 @@ extends ChunkMap(memFactory, initMapSize) with ReadablePartition {
       ChunkSetInfo.setChunkID(infoAddr, currentChunkID)
       ChunkSetInfo.resetNumRows(infoAddr)    // Must reset # rows otherwise it keeps increasing!
       ChunkSetInfo.setStartTime(infoAddr, ts)
+      ChunkSetInfo.setEndTime(infoAddr, ts)
```
broneill: What's this? An unrelated mistake>

velvia: It's an unrelated but simple fix.  Well, now that I'm looking closer at it, it's probably not necessary and can just be removed.

broneill:
With all these special configs, can you write a doc outlining guidelines for setting and measuring them? It doesn't need to be included in this submission.

velvia:
@broneill sure.  We try to add comments in the example configs explaning what they are.

PR #366, opened 2019-05-15
bug(core): Guard access to the evicted partition keys bloom filter, to prevent it from being freed concurrently.

broneill:
**Pull Request checklist**

**Current behavior :**
SEGVs have been observed from the bloom filter, and one possible cause is concurrent shutdown. Only the ingestion thread is expected to access it, but shutdown can run from another one.

**New behavior :**
Guard all access with a synchronized block. This won't affect performance because the bloom filter isn't accessed that often, and it's almost always from the ingestion thread. There's no contention.


broneill:
There's no guarantee that any threads (ingestion) are no longer accessing the bloom filter when shutdown is called. This can free the off heap memory used by the bloom filter too soon. Something else can then allocate it and corrupt it.

PR #379, opened 2019-05-20
bug(core): Don't call isEmpty on chunk map iterator from the ingestion thread.

broneill:
bug(core): Don't call isEmpty on chunk map iterator from the ingestion thread. It acquires the shared lock when the flush task thread was supposed to.

**Pull Request checklist**


broneill:
Is there some way of enabling a code coverage report? Then it would be pretty easy to determine if an "is empty" test exists.

broneill:
I added back the isEmpty check to blow up if it returns true. This test failed: TimeSeriesMemStoreSpec: should ingestStream and flush on interval *** FAILED ***

PR #393, opened 2019-05-28
fix(Query): metadata query - enforce limit by returning truncated res…

sherali42:
…ults

**Pull Request checklist**

For metadata queries, exception will not be thrown if the result size is bigger than the limit, instead resultset is truncated up to the limit.


velvia:
@sherali42  I don't see any changes at all to this PR.  As a result my opinion of it still has not changed.  Let's chat offline if needed but I don't think this is the right direction nor is it needed.

PR #396, opened 2019-05-31
fix(query): resets function should not return 0 when there is not data for the window

sherali42:
**Pull Request checklist**

Change `resets` function implementation to reset value back to Double.NaN when window size is zero, making it consistent with Prometheus implementation


velvia:
One more thought: the resets function will be much much faster after using my changes from the counter correction PR.  This is due to the auto drop/reset detection and most of the time, there would not be any resets and this is indicated by the chunk metadata.  Also this kind of bug would not be there as the ChunkedWindow approach is much less stateful and easier to debug.

velvia:
@sherali42  can we close this PR? Can reopen later if we want.

PR #398, opened 2019-06-03
fix(coordinator) : spread-assignment should be optional

TanviBhavsar:
**Pull Request checklist**

**Current behavior :** 
Query actor does not check whether spread-assignment  is present in config. So spread-assignment is mandatory in config

**New behavior :**
Query Actor checks whether spread-assignment is present in config and reads only when it is presents

### DIFF
--- a/README.md
+++ b/README.md
@@ -383,6 +383,8 @@ Some special functions exist to aid debugging and for other purposes:
 | function | description    |
 |----------|----------------|
 | `_filodb_chunkmeta_all` | (CLI Only) Returns chunk metadata fields for all chunks matching the time range and filter criteria - ID, # rows, start and end time, as well as the number of bytes and type of encoding used for a particular column.  |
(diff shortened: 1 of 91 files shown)


vishramachandran:
@TanviBhavsar Can you create a new branch off of develop with these changes? I suspect you are probably using an old branch.

Second, I dont understand why we saw the exception even though you had `spread-assignment = []` in filodb-defaults.conf. Why didnt that take effect? Why are you removing it ?

TanviBhavsar:
> @TanviBhavsar Can you create a new branch off of develop with these changes? I suspect you are probably using an old branch.
> 
> Second, I dont understand why we saw the exception even though you had `spread-assignment = []` in filodb-defaults.conf. Why didnt that take effect? Why are you removing it ?

It is not throwing exception now. I made the change so that spread-assignment is not mandatory in config.

PR #399, opened 2019-06-03
fix(coordinator) : spread-assignment should be optional

TanviBhavsar:
**Pull Request checklist**

**Current behavior :** 
Query actor does not check whether spread-assignment is present in config. So spread-assignment is mandatory in config

**New behavior :**
Query Actor checks whether spread-assignment is present in config and reads only when it is presents


TanviBhavsar:
> I do not think it is a great idea to remove default conf.

PR #407, opened 2019-06-17
feat(query): Multi-Level aggregation for queries touching many shards

vishramachandran:
**Pull Request checklist**

All aggregations happen at one level, no matter how many shards the query touches.

**New behavior :**

If the metric touches more than a threshold number of shards, we do aggregations at multiple levels.

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
@@ -190,7 +190,7 @@ class QueryEngine(dataset: Dataset,
                                    submitTime: Long,
                                    options: QueryOptions,
                                    lp: Aggregate, spreadProvider : SpreadProvider): PlanResult = {
-    val toReduce = walkLogicalPlanTree(lp.vectors, queryId, submitTime, options, spreadProvider )
+    val toReduceLevel1 = walkLogicalPlanTree(lp.vectors, queryId, submitTime, options, spreadProvider )
     // Now we have one exec plan per shard
     /*
      * Note that in order for same overlapping RVs to not be double counted when spread is increased,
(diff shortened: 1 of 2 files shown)


coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
```diff
@@ -203,10 +203,22 @@ class QueryEngine(dataset: Dataset,
...
-    toReduce.plans.foreach(_.addRangeVectorTransformer(AggregateMapReduce(lp.operator, lp.params, lp.without, lp.by)))
-    // One could do another level of aggregation per node too. Ignoring for now
-    val reduceDispatcher = pickDispatcher(toReduce.plans)
-    val reducer = ReduceAggregateExec(queryId, reduceDispatcher, toReduce.plans, lp.operator, lp.params)
+    toReduceLevel1.plans.foreach(
+      _.addRangeVectorTransformer(AggregateMapReduce(lp.operator, lp.params, lp.without, lp.by))
+    )
+
+    val toReduceLevel2 =
+      if (toReduceLevel1.plans.size >= 16) {
```
tjackpaul: Should it be configurable ? So that it can be dynamically adjusted ?

coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
```diff
@@ -203,10 +203,22 @@ class QueryEngine(dataset: Dataset,
      *
      * Starting off with solution 1 first until (2) or some other approach is decided on.
      */
-    toReduce.plans.foreach(_.addRangeVectorTransformer(AggregateMapReduce(lp.operator, lp.params, lp.without, lp.by)))
-    // One could do another level of aggregation per node too. Ignoring for now
-    val reduceDispatcher = pickDispatcher(toReduce.plans)
-    val reducer = ReduceAggregateExec(queryId, reduceDispatcher, toReduce.plans, lp.operator, lp.params)
+    toReduceLevel1.plans.foreach(
```
velvia: Please use brace syntax for multi-line foreach

coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
velvia: +1 to make configurable

coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
```diff
@@ -203,10 +203,22 @@ class QueryEngine(dataset: Dataset,
...
+      _.addRangeVectorTransformer(AggregateMapReduce(lp.operator, lp.params, lp.without, lp.by))
+    )
+
+    val toReduceLevel2 =
+      if (toReduceLevel1.plans.size >= 16) {
+        // If number of children is above a threshold, parallelize aggregation
+        val groupSize = Math.sqrt(toReduceLevel1.plans.size).ceil.toInt
```
velvia: How was sqrt determined to be a good function?

vishramachandran: Right now it is just hypothesis, to perform equal number of reductions at every node, thereby distributing work equally. We would need to test this hypothesis in a real environment. Network latency is also at play here.

coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
vishramachandran: We could make it configurable but with more changes - QueryEngine does not have server configuration right now. 

I didn't see the value in making it configurable yet. This value doesn't change from environment to environment. Also, default config change has the same process as code change.

velvia:
It's easy to add optional config without affecting other services.  

    config.as[Option[Int]]("max-shards-per-aggregation-level").getOrElse(DefaultMaxShardsPerAggLevel)

PR #418, opened 2019-06-25
feat(query) : Binary Join Many to Many - "and", "or" and "unless"

TanviBhavsar:
Support for Binary Join Many to Many with "and", "or" and "unless" operators


vishramachandran:
@TanviBhavsar after looking at the PR a bit more, I think I understand your motivation of colocating the two. Let me think a bit more.

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
```diff
@@ -0,0 +1,142 @@
...
+  */
+final case class SetOperatorExec(id: String,
+                                dispatcher: PlanDispatcher,
+                                lhs: Seq[ExecPlan],
+                                rhs: Seq[ExecPlan],
+                                binaryOp: BinaryOperator,
+                                cardinality: Cardinality,
+                                on: Seq[String],
+                                ignoring: Seq[String]) extends NonLeafExecPlan {
+  require(cardinality == Cardinality.ManyToMany, "set operations must only use many-to-many matching")
```
vishramachandran: IS `cardinality` even needed now ? Can be removed.

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
```diff
@@ -0,0 +1,142 @@
...
+      if (!jk.isEmpty)
+        rhsKeysSet += jk
+    }
+
+    lhsRvs.foreach { lhs =>
+      val jk = joinKeys(lhs.key)
+      // Add range vectors from lhs which are present in lhs and rhs both
+      // Result should also have range vectors for which rhs does not have any keys
+      if (rhsKeysSet.contains(jk) || rhsKeysSet.isEmpty) {
+        result += IteratorBackedRangeVector(lhs.key, lhs.rows)
```
vishramachandran: No need to create new RV. Can simply used `lhs`. Same comment on a few other places.

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
```diff
@@ -0,0 +1,142 @@
...
+  private def setOpUnless(lhsRvs: List[SerializableRangeVector]
+                          , rhsRvs: List[SerializableRangeVector]): List[IteratorBackedRangeVector] = {
+    val rhsKeysSet = new mutable.HashSet[Map[Utf8Str, Utf8Str]]()
+    var result = new ListBuffer[IteratorBackedRangeVector]()
+    rhsRvs.foreach { rv =>
+      val jk = joinKeys(rv.key)
+      rhsKeysSet += jk
+    }
+    // Add range vectors which are not present in rhs
+    lhsRvs.foreach { lhs =>
```
vishramachandran: consider using map/filter instead of foreach.

TanviBhavsar:
> @TanviBhavsar Can we simply use another ExecPlan for Set Operators ?
> They are completely different.
> 
> 1. We are trying to fit in Set operators as ManyToMany which is a bit kludgy
> 2. I am not seeing any logic that needs to be shared/reused between the two - the flow is fully disjoint.
> 
> I dont see why these need to be co-located.
> 
> Let me know your thoughts, and lets discuss in person if needed.

Have added Exec plan for Set Operator

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
TanviBhavsar: Prometheus has this validation that set operations must only use many-to-many matching

TanviBhavsar: Removed cardinality. Validation is done in vectorMatch validate

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
TanviBhavsar: I missed that. Made the change

TanviBhavsar:
> Thanks for the changes.
> Looks great. Some minor changes requested. Then we will be ready to go.
@vishramachandran  Addressed your review comments

query/src/main/scala/filodb/query/exec/BinaryJoinExec.scala
```diff
@@ -42,8 +41,7 @@ final case class BinaryJoinExec(id: String,
                                 on: Seq[String],
                                 ignoring: Seq[String],
                                 include: Seq[String]) extends NonLeafExecPlan {
-  require(cardinality != Cardinality.ManyToMany || binaryOp.isInstanceOf[ComparisonOperator],
-    "Many To Many cardinality is supported only for comparison binary operators")
```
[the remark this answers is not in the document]
TanviBhavsar: Modified the check as discussed

PR #419, opened 2019-06-25
bug(core): Queries have been executing in ingestSched

vishramachandran:
**Pull Request checklist**

Looks like executeOn causes subsequent map/foreach lambdas to run on the specified executor than the default one assigned for the whole pipeline. Call to `asyncBoundary` is needed to fix this.

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/OnDemandPagingShard.scala
+++ b/core/src/main/scala/filodb.core/memstore/OnDemandPagingShard.scala
@@ -7,8 +7,7 @@ import debox.Buffer
 import kamon.Kamon
 import kamon.trace.Span
 import monix.eval.Task
-import monix.execution.Scheduler
-import monix.reactive.{Observable, OverflowStrategy}
+import monix.reactive.Observable
 
 import filodb.core.Types
 import filodb.core.downsample.{DownsampleConfig, DownsamplePublisher}
@@ -35,10 +34,6 @@ class OnDemandPagingShard(dataset: Dataset,
 TimeSeriesShard(dataset, storeConfig, shardNum, bufferMemoryManager, rawStore, metastore, evictionPolicy,
                 downsampleConfig, downsamplePublisher)(ec) {
   import TimeSeriesShard._
-
-  private val singleThreadPool = Scheduler.singleThread(s"make-partition-${dataset.ref}-$shardNum")
-  // TODO: make this configurable
-  private val strategy = OverflowStrategy.BackPressure(1000)
 
   private def startODPSpan(): Span = Kamon.buildSpan(s"odp-cassandra-latency")
     .withTag("dataset", dataset.name)
@@ -102,9 +97,7 @@ TimeSeriesShard(dataset, storeConfig, shardNum, bufferMemoryManager, rawStore, m
           if (partKeyBytesToPage.nonEmpty) {
             val span = startODPSpan()
             rawStore.readRawPartitions(dataset, allDataCols, multiPart, computeBoundingMethod(pagingMethods))
-              // NOTE: this executes the partMaker single threaded.  Needed for now due to concurrency constraints.
-              // In the future optimize this if needed.
-              .mapAsync { rawPart => partitionMaker.populateRawChunks(rawPart) }
+              .map { rawPart => partitionMaker.populateRawChunks(rawPart) }
               .doOnTerminate(ex => span.finish())
           } else { Observable.empty }
         }
@@ -116,7 +109,7 @@ TimeSeriesShard(dataset, storeConfig, shardNum, bufferMemoryManager, rawStore, m
             Observable.fromIterable(partKeyBytesToPage.zip(pagingMethods))
               .mapAsync(storeConfig.demandPagingParallelism) { case (partBytes, method) =>
                 rawStore.readRawPartitions(dataset, allDataCols, SinglePartitionScan(partBytes, shardNum), method)
-                  .mapAsync { rawPart => partitionMaker.populateRawChunks(rawPart) }
+                  .map { rawPart => partitionMaker.populateRawChunks(rawPart) }
                   .defaultIfEmpty(getPartition(partBytes).get)
                   .headL
               }
--- a/core/src/main/scala/filodb.core/memstore/OnDemandPagingShard.scala
+++ b/core/src/main/scala/filodb.core/memstore/OnDemandPagingShard.scala
@@ -104,7 +104,9 @@ TimeSeriesShard(dataset, storeConfig, shardNum, bufferMemoryManager, rawStore, m
             rawStore.readRawPartitions(dataset, allDataCols, multiPart, computeBoundingMethod(pagingMethods))
               // NOTE: this executes the partMaker single threaded.  Needed for now due to concurrency constraints.
               // In the future optimize this if needed.
-              .mapAsync { rawPart => partitionMaker.populateRawChunks(rawPart).executeOn(singleThreadPool) }
+              .mapAsync { rawPart =>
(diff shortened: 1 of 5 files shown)


core/src/main/scala/filodb.core/memstore/OnDemandPagingShard.scala
```diff
@@ -104,7 +104,9 @@ TimeSeriesShard(dataset, storeConfig, shardNum, bufferMemoryManager, rawStore, m
             rawStore.readRawPartitions(dataset, allDataCols, multiPart, computeBoundingMethod(pagingMethods))
               // NOTE: this executes the partMaker single threaded.  Needed for now due to concurrency constraints.
               // In the future optimize this if needed.
-              .mapAsync { rawPart => partitionMaker.populateRawChunks(rawPart).executeOn(singleThreadPool) }
+              .mapAsync { rawPart =>
+                partitionMaker.populateRawChunks(rawPart).executeOn(singleThreadPool).asyncBoundary
+              }
               .doOnTerminate(ex => span.finish())
               // This is needed so future computations happen in a different thread
               .asyncBoundary(strategy)
```
broneill: Is this needed anymore? Should the strategy be passed to the new async calls too?

core/src/main/scala/filodb.core/memstore/OnDemandPagingShard.scala
```diff
@@ -104,7 +104,9 @@ TimeSeriesShard(dataset, storeConfig, shardNum, bufferMemoryManager, rawStore, m
             rawStore.readRawPartitions(dataset, allDataCols, multiPart, computeBoundingMethod(pagingMethods))
               // NOTE: this executes the partMaker single threaded.  Needed for now due to concurrency constraints.
               // In the future optimize this if needed.
-              .mapAsync { rawPart => partitionMaker.populateRawChunks(rawPart).executeOn(singleThreadPool) }
+              .mapAsync { rawPart =>
+                partitionMaker.populateRawChunks(rawPart).executeOn(singleThreadPool).asyncBoundary
```
velvia: Can we add a comment explaining why asyncBoundary is needed?  Thx

core/src/main/scala/filodb.core/memstore/OnDemandPagingShard.scala
```diff
@@ -118,7 +120,9 @@ TimeSeriesShard(dataset, storeConfig, shardNum, bufferMemoryManager, rawStore, m
             Observable.fromIterable(partKeyBytesToPage.zip(pagingMethods))
               .mapAsync(storeConfig.demandPagingParallelism) { case (partBytes, method) =>
                 rawStore.readRawPartitions(dataset, allDataCols, SinglePartitionScan(partBytes, shardNum), method)
-                  .mapAsync { rawPart => partitionMaker.populateRawChunks(rawPart).executeOn(singleThreadPool) }
+                  .mapAsync { rawPart =>
+                    partitionMaker.populateRawChunks(rawPart).executeOn(singleThreadPool).asyncBoundary
```
velvia: Actually, now that we have locks, I'm not sure this needs to be run on the ingestion thread anymore....  I think this was designed before the lock stuff was really in place.... Brian should comment... @broneill

core/src/main/scala/filodb.core/store/ChunkSource.scala
```diff
@@ -146,7 +146,7 @@ trait DefaultChunkSource extends ChunkSource {
       // NOTE: this executes the partMaker single threaded.  Needed for now due to concurrency constraints.
       // In the future optimize this if needed.
       .mapAsync { rawPart => partMaker(dataset, partMethod.shard).populateRawChunks(rawPart)
-                               .executeOn(singleThreadPool) }
+                               .executeOn(singleThreadPool).asyncBoundary }
```
velvia: Same here.  Evaluate whether populateRawChunks really needs to be on the singleThreadPool

core/src/main/scala/filodb.core/memstore/OnDemandPagingShard.scala
vishramachandran: @velvia It is needed. The locks take care of chunk-map operations.
Partition creation modifies the shard state, namely PartitionSet, PartitionMap etc which need to be altered by ingestion thread only (unless we add additional locking).

core/src/main/scala/filodb.core/memstore/OnDemandPagingShard.scala
vishramachandran: @velvia It is needed. The locks take care of chunk-map operations.
Partition creation modifies the shard state, namely PartitionSet, PartitionMap etc which need to be altered by ingestion thread only (unless we add additional locking).

core/src/main/scala/filodb.core/store/ChunkSource.scala
vishramachandran: @velvia It is needed. The locks take care of chunk-map operations.
Partition creation modifies the shard state, namely PartitionSet, PartitionMap etc which need to be altered by ingestion thread only (unless we add additional locking).

core/src/main/scala/filodb.core/memstore/OnDemandPagingShard.scala
vishramachandran: I already added in one place. I can add in all places.

core/src/main/scala/filodb.core/memstore/OnDemandPagingShard.scala
vishramachandran: @velvia can you comment here? Looks like there is an `.asyncBoundary(strategy)` for multiPart == true. Will adding this when multiPart == false also solve the issue ?

vishramachandran: Can we remove the `.asyncBoundary` after `.executeOn(...)`  ?

velvia: In theory should be equivalent to the earlier `.asyncBoundary`.  IE, you should need just one `asyncBoundary`, not two like the earlier code.  AsyncBoundaries force thread switches btw if the thread pool has multiple threads or you are scheduling executions on a different pool/scheduler.

core/src/main/scala/filodb.core/memstore/OnDemandPagingShard.scala
velvia: @vishramachandran  populateRawChunks does not modify `PartitionSet` etc. or any ingestion thread data structures.  The code only modifies ChunkMap, it's function is solely to copy the chunks to off heap.

vishramachandran: ok removing executeOn

core/src/main/scala/filodb.core/memstore/OnDemandPagingShard.scala
```diff
@@ -104,12 +104,8 @@ TimeSeriesShard(dataset, storeConfig, shardNum, bufferMemoryManager, rawStore, m
             rawStore.readRawPartitions(dataset, allDataCols, multiPart, computeBoundingMethod(pagingMethods))
               // NOTE: this executes the partMaker single threaded.  Needed for now due to concurrency constraints.
               // In the future optimize this if needed.
-              .mapAsync { rawPart =>
-                partitionMaker.populateRawChunks(rawPart).executeOn(singleThreadPool).asyncBoundary
-              }
+              .mapAsync { rawPart => partitionMaker.populateRawChunks(rawPart) }
```
velvia: One more thing, mapAsync (and the fact of using a Task for populateRawChunks) is not needed, just use map.

PR #426, opened 2019-06-27
feat(query) : operator precedence

TanviBhavsar:
**Pull Request checklist**

**Current behavior :**
Queries having multiple operators are not executed according to precedence
For example Logical Plan for query:  10 % http_requests_total + 5

`ScalarVectorBinaryOperation(MOD,10.0,ScalarVectorBinaryOperation(ADD,5.0,PeriodicSeries(RawSeries(IntervalSelector(1524855688000,1524855988000),List(ColumnFilter(__name__,Equals(http_requests_total))),List()),1524855988000,1000000,1524855988000),false),true)]"`

**New behavior :**
Logical plan  and thereby ExecPlan is generated according to Prometheus operator precedence
https://prometheus.io/docs/prometheus/latest/querying/operators/
The following list shows the precedence of binary operators in Prometheus, from highest to lowest:

1. ^
2. *, /, %
3. +, -
4. ==, !=, <=, <, >=, >
5. and, unless
6. or

Operators on the same precedence level are left-associative. For example, 2 * 3 % 2 is equivalent to (2 * 3) % 2. However ^ is right associative, so 2 ^ 3 ^ 2 is equivalent to 2 ^ (3 ^ 2).

Logical Plan after applying precedence for above query:
`ScalarVectorBinaryOperation(ADD,5.0,ScalarVectorBinaryOperation(MOD,10.0,PeriodicSeries(RawSeries(IntervalSelector(1524855688000,1524855988000),List(ColumnFilter(__name__,Equals(http_requests_total))),List()),1524855988000,1000000,1524855988000),true),false)`


TanviBhavsar:
> The usual way in which operator precedence implemented is at the parser level. This also makes it easy to support arbitrary nesting of groups that have parenthesis around them. Is our parser and tree not capable of doing this?
> 
> This:
> 
> lazy val arithmeticOp = pow | mul | div | mod | add | sub
> 
> Would need to be broken down into more levels. Same with setOp.

 I tried, however it is getting tricky to handle all cases with grammar

TanviBhavsar:
> Approved with some minor comments!

Thanks Vish! I have addressed the review comments

PR #430, opened 2019-06-27
feat(core): reject samples with Nano/Micro second timestamps

TanviBhavsar:
**Pull Request checklist**

Reject samples before 1 month and after 1 day from current time

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
+++ b/core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
@@ -1,5 +1,7 @@
 package filodb.core.memstore
 
+import java.time.Instant
+
 import com.typesafe.scalalogging.StrictLogging
 import scalaxy.loops._
 
@@ -119,8 +121,17 @@ extends ChunkMap(memFactory, initMapSize) with ReadablePartition {
   def ingest(row: RowReader, blockHolder: BlockMemFactory): Unit = {
     // NOTE: lastTime is not persisted for recovery.  Thus the first sample after recovery might still be out of order.
     val ts = dataset.timestamp(row)
+    val currentEpochTime : Long = Instant.now.getEpochSecond()
+    // Reject samples before 1 month and after 1 day from current time
+    val lowerLimit = currentEpochTime - (31 * 24 * 60 * 60)
+    val upperLimit = currentEpochTime + (24 * 60 * 60 )
+
     if (ts < timestampOfLatestSample) {
       shardStats.outOfOrderDropped.increment
+      return
+    }
+     else if (((ts/1000) < lowerLimit) || ((ts/1000) > upperLimit)) {
+      _log.error(s"Rejecting row as timestamp: ${row.getLong(0)} is not within limits" )
       return
     }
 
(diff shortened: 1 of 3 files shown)


core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
```diff
@@ -119,10 +121,19 @@ extends ChunkMap(memFactory, initMapSize) with ReadablePartition {
   def ingest(row: RowReader, blockHolder: BlockMemFactory): Unit = {
     // NOTE: lastTime is not persisted for recovery.  Thus the first sample after recovery might still be out of order.
     val ts = dataset.timestamp(row)
+    val currentEpochTime : Long = Instant.now.getEpochSecond()
+    // Reject samples before 1 month and after 1 day from current time
+    val lowerLimit = currentEpochTime - (31 * 24 * 60 * 60)
+    val upperLimit = currentEpochTime + (24 * 60 * 60 )
```
vishramachandran: Need to make the 1 month, and 1 day configurable; and dont calculate each time.

core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
```diff
@@ -119,10 +121,19 @@ extends ChunkMap(memFactory, initMapSize) with ReadablePartition {
...
+
     if (ts < timestampOfLatestSample) {
       shardStats.outOfOrderDropped.increment
       return
     }
+     else if (((ts/1000) < lowerLimit) || ((ts/1000) > upperLimit)) {
```
vishramachandran: incoming timestamp is already in millisecond. why divide by 1000 ?

core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
```diff
@@ -119,10 +121,19 @@ extends ChunkMap(memFactory, initMapSize) with ReadablePartition {
   def ingest(row: RowReader, blockHolder: BlockMemFactory): Unit = {
     // NOTE: lastTime is not persisted for recovery.  Thus the first sample after recovery might still be out of order.
     val ts = dataset.timestamp(row)
+    val currentEpochTime : Long = Instant.now.getEpochSecond()
```
vishramachandran: It should be in millis. System.currentTimeMillis() ?

core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
```diff
@@ -119,10 +121,19 @@ extends ChunkMap(memFactory, initMapSize) with ReadablePartition {
...
+    // Reject samples before 1 month and after 1 day from current time
+    val lowerLimit = currentEpochTime - (31 * 24 * 60 * 60)
+    val upperLimit = currentEpochTime + (24 * 60 * 60 )
+
     if (ts < timestampOfLatestSample) {
       shardStats.outOfOrderDropped.increment
       return
     }
+     else if (((ts/1000) < lowerLimit) || ((ts/1000) > upperLimit)) {
+      _log.error(s"Rejecting row as timestamp: ${row.getLong(0)} is not within limits" )
```
vishramachandran: Dont log. Just increase a kamon counter. Too many log messages will affect performance.

velvia:
1) We don't need a lower limit.  There are legitimate use cases for lower time stamps, and a lower limit does not prevent ns/microsecond timestamps, which will always be higher.  
2) The check needs to be super super fast.  This is one of the most performance critical paths.  Absolutely cannot compute the time limit every single time.

PR #438, opened 2019-07-10
Remove stream subscription task before canceling it, and also rely on…

broneill:
… same mapping when resync'ng. This doesn't fix all the race conditions, but it does permit resync to fix things.

**Pull Request checklist**

**Current behavior :**
Shards which report as being active can sometimes be stuck. This is because of a race condition in which the ingestion task is canceled but still remains in the mapping.

**New behavior :**
Remove the task from the mapping before it's canceled. The resync task checks the same mapping now, avoiding any races between the two maps.

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/IngestionActor.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/IngestionActor.scala
@@ -117,7 +117,7 @@ private[filodb] final class IngestionActor(dataset: Dataset,
 
     // Start with the full set of all shards being ingested, and remove shards from this set
     // which must continue being ingested.
-    val shardsToStop = HashSet() ++ streams.keySet
+    val shardsToStop = HashSet() ++ streamSubscriptions.keySet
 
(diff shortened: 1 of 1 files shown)


broneill:
The streams map is required for teardown to be called from removeAndReleaseResources.

velvia:
Oh true.  IN that case, I think we should really do is have the `IngestionStream` and the Future to be in the same hash map.    Or maybe the future can even be part of the stream state/object.

broneill:
I started by bundling them into the same map but realized that many more changes are required. I don't understand what the relationship between these objects is, as far as life cycle is concerned.

broneill:
What logging do you recommend? There's already quite a bit.

vishramachandran:
For starters we need indicators which clearly show
1. that stopIngestion was invoked (or not) for a shard and its completion status.
2. that the ingestion future completed, and that it was a success or failure.
3. that shard error/stopped status was sent to the singleton
4. why stopIngestion is being invoked - because of an actor death, or error, or some other reason.

PR #440, opened 2019-07-11
feat(core, *): PartitionSchema and DataSchemas: multi-schema part I

velvia:
NOTE: Strongly suggest going through the code in commit order, rather than browsing all code at once

**Pull Request checklist**

The `Dataset` class rules the world.  It blends ingestion stream and definition for partition keys and data schema, and is used virtually everywhere.

**New behavior :**

This is the start of multi-schema changes for FiloDB.  
It defines multiple new schema definition classes which are intended to eventually replace `Dataset`:
* `PartitionSchema` contains the schema for the partition key, and is to be shared across all datasets in a FiloDB cluster
* `DataSchema` defines the columns for a single data type.   Multiple DataSchemas will be supported at once, in the future.
* There are no more "Row key columns".  The first DataSchema column is the "row key", and it must be a `TimestampColumn` or `LongColumn`. 

The `Dataset` class implementation now points at a `Schema` which is just a `PartitionSchema` plus a `DataSchema`.  All usage of Dataset remains essentially the same as today.

PartitionSchemas and DataSchemas are now defined in the root/application server config.  Each source/dataset config points back at the schema it supports.  

**BREAKING CHANGES**

Noncompatible change to config files.

### DIFF
--- a/conf/timeseries-128shards-source.conf
+++ b/conf/timeseries-128shards-source.conf
@@ -1,20 +1,5 @@
     dataset = "prometheus"
-
-    definition {
-      partition-columns = ["tags:map"]
-      data-columns = ["timestamp:ts", "value:double:detectDrops=true"]
-      row-key-columns = [ "timestamp" ]
-      downsamplers = [ ]
-    }
-
-    options {
-      shardKeyColumns = [ "__name__", "_ns" ]
-      ignoreShardKeyColumnSuffixes = { "__name__" = ["_bucket", "_count", "_sum"] }
-      valueColumn = "value"
-      metricColumn = "__name__"
-      ignoreTagsOnPartitionKeyHash = [ "le" ]
-      copyTags = { }
-    }
+    schema = "prometheus"
 
     num-shards = 128
     min-num-nodes = 2
--- a/coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
@@ -328,7 +328,7 @@ class QueryEngine(dataset: Dataset,
                                       options: QueryOptions,
                                       lp: RawChunkMeta, spreadProvider : SpreadProvider): PlanResult = {
     // Translate column name to ID and validate here
-    val colName = if (lp.column.isEmpty) dataset.options.valueColumn else lp.column
+    val colName = if (lp.column.isEmpty) dataset.schema.data.valueColName else lp.column
     val colID = dataset.colIDs(colName).get.head
     val renamedFilters = renameMetricFilter(lp.filters)
     val metaExec = shardsFromFilters(renamedFilters, options, spreadProvider).map { shard =>
@@ -368,13 +368,13 @@ class QueryEngine(dataset: Dataset,
     * as those are automatically prepended.
     */
   private def getColumnIDs(dataset: Dataset, cols: Seq[String]): Seq[Types.ColumnId] = {
-    val realCols = if (cols.isEmpty) Seq(dataset.options.valueColumn) else cols
+    val realCols = if (cols.isEmpty) Seq(dataset.schema.data.valueColName) else cols
     val ids = dataset.colIDs(realCols: _*)
       .recover(missing => throw new BadQueryException(s"Undefined columns $missing"))
       .get
-    // avoid duplication if first ids are already row keys
-    if (ids.take(dataset.rowKeyIDs.length) == dataset.rowKeyIDs) { ids }
-    else { dataset.rowKeyIDs ++ ids }
+    // avoid duplication if first id is a timestamp, otherwise add it
+    if (ids.take(1) == Dataset.rowKeyIDs) { ids }
+    else { Dataset.rowKeyIDs ++ ids }
   }
 
   private def toChunkScanMethod(rangeSelector: RangeSelector): ChunkScanMethod = {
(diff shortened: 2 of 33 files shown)


core/src/main/scala/filodb.core/metadata/Schemas.scala
```diff
@@ -0,0 +1,237 @@
...
+ * The columns inside PartitionSchema are used for distribution and sharding, as well as filtering and searching
+ * for time series during querying.
+ * There should only be ONE PartitionSchema across the entire Database.
+ */
+final case class PartitionSchema(columns: Seq[Column],
+                                 predefinedKeys: Seq[String],
+                                 options: DatasetOptions) {
+  val binSchema = new RecordSchema(columns.map(c => ColumnInfo(c.name, c.columnType)), Some(0), predefinedKeys)
+
+  private val partKeyBuilder = new RecordBuilder(MemFactory.onHeapFactory, binSchema, 10240)
```
broneill: Magic value?

core/src/main/scala/filodb.core/metadata/Schemas.scala
```diff
@@ -0,0 +1,237 @@
...
+    else           Good(index)
+  }
+
+  /**
+   * Generates a unique 16-bit hash from the column names and types.  Sensitive to order.
+   */
+  def genHash(columns: Seq[Column]): Int = {
+    var hash = 7
+    for { col <- columns } {
+      // Use XXHash to get high quality hash for column name.  String.hashCode is _horrible_
```
broneill: The default string hashcode is actually quite good, but it's not very efficient, and it doesn't fill the upper bits when hashing short strings. Considering that you're tossing away the upper bits anyhow, it doesn't matter. When calling getBytes always specify a character encoding, or else you'll potentially get different results on different machines.

velvia: I actually tested it for our specific purposes, and found it to have a terrible distribution.  Not even close to the better algorithms.  The upper bits are actually the most important bits we use for other things, but in this case maybe it doesn't matter.  Anyhow, this will only be used at schema creation time, which is once on load, and good distribution is much more important, so I'd prefer better distribution over efficiency :)

core/src/main/scala/filodb.core/metadata/Schemas.scala
velvia: Ah ok, it was copy and pasted.  Thanks for catch

coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
```diff
@@ -368,13 +368,13 @@ class QueryEngine(dataset: Dataset,
...
+    val realCols = if (cols.isEmpty) Seq(dataset.schema.data.valueColName) else cols
     val ids = dataset.colIDs(realCols: _*)
       .recover(missing => throw new BadQueryException(s"Undefined columns $missing"))
       .get
-    // avoid duplication if first ids are already row keys
-    if (ids.take(dataset.rowKeyIDs.length) == dataset.rowKeyIDs) { ids }
-    else { dataset.rowKeyIDs ++ ids }
+    // avoid duplication if first id is a timestamp, otherwise add it
+    if (ids.take(1) == Dataset.rowKeyIDs) { ids }
+    else { Dataset.rowKeyIDs ++ ids }
```
vishramachandran: I am changing this method too in my PR. Looks like we need to merge.

core/src/main/scala/filodb.core/metadata/Dataset.scala
```diff
@@ -26,31 +25,30 @@ import filodb.memory.format.{BinaryVector, RowReader, TypedIterator, ZeroCopyUTF
...
-  val rowKeyColumns   = rowKeyIDs.map(dataColumns)
-  val rowKeyRouting   = rowKeyIDs.toArray
-
-  val ingestionSchema = RecordSchema2.ingestion(this)  // TODO: add predefined keys yo
-  val comparator      = new RecordComparator(ingestionSchema)
-  val partKeySchema   = comparator.partitionKeySchema
+final case class Dataset(name: String, schema: Schema) {
+  def options: DatasetOptions = schema.partition.options
+  def dataColumns: Seq[Column] = schema.data.columns
+  def partitionColumns: Seq[Column] = schema.partition.columns
```
vishramachandran: Why def for the three above?

conf/timeseries-128shards-source.conf
```diff
@@ -1,20 +1,5 @@
...
-
-    options {
-      shardKeyColumns = [ "__name__", "_ns" ]
-      ignoreShardKeyColumnSuffixes = { "__name__" = ["_bucket", "_count", "_sum"] }
-      valueColumn = "value"
-      metricColumn = "__name__"
-      ignoreTagsOnPartitionKeyHash = [ "le" ]
-      copyTags = { }
-    }
+    schema = "prometheus"
```
vishramachandran: Can you please remind what this will evolve into when there are multiple schemas ?

velvia: This line will be gone when we move to multi-schema, or will be a restricted set of schemas that each dataset is allowed to have.  I think you prefer restrictions (though that can be added later too)

core/src/main/scala/filodb.core/metadata/Dataset.scala
velvia: 🤷🏻‍♂️  I could change it to val.  def uses no memory, but that doesn't matter here.  They are quick references, maybe val is slightly faster?

PR #441, opened 2019-07-15
feat(various): Assert Thread Name of various operations

vishramachandran:
**Pull Request checklist**

Currently there is a lot of confusion and un-clarity and risk of bugs because what operation executed on which thread pool does not stand out so much so that it is almost impossible for a new developer to confidently change code without risking introducing new issues with concurrency.

**New behavior :**

This is an interim solution for introducing cleanliness, clarity and maintainability by clearly conveying to reader/developer which thread pool code is executed in.

Using this assertion, also fixed a bug where index recovery was being executed on akka actor thread instead of ingestion thread.

Also bringing clarity in threadpool by making implicitly passed scheduler arguments explicit.
Also renamed `ec` and `sched` threadpools to `ioPool` in many places. Former names are too ambiguous and unclear and can lead to bugs.

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/NodeConfiguration.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/NodeConfiguration.scala
@@ -1,0 +1,16 @@
+package filodb.coordinator
+
+import com.typesafe.config.Config
+
+import filodb.core.GlobalConfig
+
+
+/** Mixin used for nodes and tests. */
+trait NodeConfiguration {
+
+  /** The global Filo configuration. */
+  val systemConfig: Config = GlobalConfig.systemConfig
+
+}
+
+
--- a/coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
@@ -13,7 +13,7 @@ import scala.util.control.NonFatal
 
 import filodb.coordinator.queryengine2.QueryEngine
 import filodb.core._
-import filodb.core.memstore.{MemStore, TermInfo}
+import filodb.core.memstore.{FiloSchedulers, MemStore, TermInfo}
 import filodb.core.metadata.Dataset
 import filodb.core.query.ColumnFilter
 import filodb.query._
@@ -96,7 +96,7 @@ final class QueryActor(memStore: MemStore,
       .start()
     q.execute(memStore, dataset, queryConfig)(queryScheduler, queryConfig.askTimeout)
      .foreach { res =>
-       require(Thread.currentThread().getName.startsWith(QuerySchedName))
+       FiloSchedulers.assertThreadName(QuerySchedName)
        replyTo ! res
        res match {
          case QueryResult(_, _, vectors) => resultVectors.record(vectors.length)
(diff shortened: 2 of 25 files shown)


core/src/main/scala/filodb.core/memstore/FiloSchedulers.scala
```diff
@@ -0,0 +1,19 @@
...
+object FiloSchedulers {
+  val IngestSchedName = "ingestion-shard"
+  val FlushSchedName = "flush-sched"
+  val IOSchedName = "filodb.io"
+  val QuerySchedName = "query-sched"
+  val PopulateChunksSched = "populate-odp-chunks"
+
+  def assertThreadName(name: String): Unit = {
+
+    if (GlobalConfig.systemConfig.getBoolean("filodb.scheduler.enable-assertions")) {
```
broneill: Assign the config boolean to a object-level val, and then the check is compiled away at runtime. Do we expect to change the config setting dynamically?

coordinator/src/main/scala/filodb.coordinator/NodeConfiguration.scala
```diff
@@ -0,0 +1,16 @@
+package filodb.coordinator
+
+import com.typesafe.config.Config
+
+import filodb.core.GlobalConfig
+
+
+/** Mixin used for nodes and tests. */
+trait NodeConfiguration {
```
broneill: A separate source file for a top-level trait/class? Yes! More of this.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -437,41 +440,50 @@ class TimeSeriesShard(val dataset: Dataset,
...
+          new IndexData(tb, b.segmentId, RecordContainer(b.segment.array()))
+        }
       }
-    }
-    val fut = Observable.flatten(timeBuckets: _*)
-      .foreach(tb => extractTimeBucket(tb, partIdMap))(ingestSched)
-      .map(_ => completeIndexRecovery())
-    fut.onComplete(_ => tracer.finish())
-    fut
+      Observable.flatten(timeBuckets: _*)
```
broneill: Is the change here the thread-safety bug you mention?

coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
```diff
@@ -93,9 +94,9 @@ final class QueryActor(memStore: MemStore,
     val span = Kamon.buildSpan(s"execplan2-${q.getClass.getSimpleName}")
       .withTag("query-id", q.id)
       .start()
-    implicit val _ = queryConfig.askTimeout
-    q.execute(memStore, dataset, queryConfig)
+    q.execute(memStore, dataset, queryConfig)(queryScheduler, queryConfig.askTimeout)
      .foreach { res =>
+       require(Thread.currentThread().getName.startsWith(QuerySchedName))
```
broneill: Call FiloSchedulers?

core/src/main/scala/filodb.core/memstore/FiloSchedulers.scala
vishramachandran: Good idea. done

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
vishramachandran: Yes - thats right

coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
vishramachandran: thanks for catching. yes

PR #444, opened 2019-07-16
bug(coordinator) Stopping ingestion didn't actually run the handler.

broneill:
Stopping ingestion didn't actually run the handler. Define an explicit cancel task which is passed to the Observable, and complete the stop request via actor messages only.

**Pull Request checklist**

**Current behavior :** 
When ingestion is stopped, no handler is invoked, and no resources are cleaned up. This gives the appearance than ingestion is still active, when in fact it isn't. In some cases this leads to stuck shards.

**New behavior :**
An explicit cancel handler is used now, which actually gets invoked when the ingestion Observable is canceled. I verified that it's invoked by the test suite by deliberately breaking it, exiting the process.

Also, the remove and release of resources is invoked via an actor message, preventing any odd thread-safety or ordering issues.

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/IngestionActor.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/IngestionActor.scala
@@ -11,6 +11,7 @@ import scala.util.control.NonFatal
 import akka.actor.{ActorRef, Props}
 import akka.event.LoggingReceive
 import kamon.Kamon
+import monix.eval.Task
 import monix.execution.{CancelableFuture, Scheduler, UncaughtExceptionReporter}
 import monix.reactive.Observable
 import net.ceedubs.ficus.Ficus._
@@ -98,6 +99,7 @@ private[filodb] final class IngestionActor(dataset: Dataset,
     case GetStatus               => status(sender())
     case e: IngestRows           => ingest(e)
     case e: ShardIngestionState  => resync(e, sender())
+    case e: IngestionStopped     => ingestionStopped(e.ref, e.shard)
   }
 
   /**
@@ -226,12 +228,20 @@ private[filodb] final class IngestionActor(dataset: Dataset,
       logger.info(s"Starting normal/active ingestion for dataset=${dataset.ref} shard=$shard at offset $offset")
       statusActor ! IngestionStarted(dataset.ref, shard, nodeCoord)
 
+      // Define a cancel task to run when ingestion is stopped.
+      val onCancel = Task {
+        val stopped = IngestionStopped(dataset.ref, shard)
+        self ! stopped
+        statusActor ! stopped
+      }
+
       val shardIngestionEnd = memStore.ingestStream(dataset.ref,
         shard,
         stream,
         flushSched,
         flushStream(startingGroupNo),
-        diskTimeToLiveSeconds)
+        diskTimeToLiveSeconds,
+        onCancel)
       // On completion of the future, send IngestionStopped
       // except for noOpSource, which would stop right away, and is used for sending in tons of data
       // also: small chance for race condition here due to remove call in stop() method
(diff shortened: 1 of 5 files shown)


core/src/main/scala/filodb.core/memstore/TimeSeriesMemStore.scala
```diff
@@ -135,7 +145,9 @@ extends MemStore with StrictLogging {
                   }.collect { case Some(flushGroup) => flushGroup }
                   .mapAsync(numParallelFlushes) { f => shard.createFlushTask(f).executeOn(flushSched).asyncBoundary }
                            // asyncBoundary so subsequent computations in pipeline happen in default threadpool
-                  .foreach({ x => })(shard.ingestSched)
+                  .completedL
+                  .doOnCancel(cancelTask)
```
vishramachandran: We also need to add a doOnFinish handler with same cleanup tasks right ?

broneill: On normal completion or failure, the caller can continue doing that on the future object that's returned.

vishramachandran: I recommend passing single onEnd handler:

```
def onEnd(ex: Option[Throwable]): Task[Unit] = {
    ex match {
       case None => val stopped = IngestionStopped(dataset.ref, shard)
                                self ! stopped
                                statusActor ! stopped
       case Some(ex) => 
                                self ! IngestionStopped(dataset.ref, shard)
                                statusActor ! IngestionError(dataset.ref, shard)
}
```

and attach
```
.doOnCancel(Task { onEnd(None) } )
.doOnFinish(ex: Task { onEnd(ex) } )
```

Granted, there is a bit more cleanup to be done to the above code, but you get the idea ...

core/src/main/scala/filodb.core/memstore/TimeSeriesMemStore.scala
```diff
@@ -113,11 +115,19 @@ extends MemStore with StrictLogging {
...
+                     flushSched: Scheduler,
+                     diskTimeToLiveSeconds: Int): CancelableFuture[Unit] =
+    doIngestStream(dataset, shardNum, combinedStream, flushSched, diskTimeToLiveSeconds, Task(() => {}))
+
+  def doIngestStream(dataset: DatasetRef,
+                     shardNum: Int,
+                     combinedStream: Observable[DataOrCommand],
+                     flushSched: Scheduler,
+                     diskTimeToLiveSeconds: Int,
+                     cancelTask: Task[Unit]): CancelableFuture[Unit] = {
```
vishramachandran: Any reason why we could not retain one method with default arg value ?

broneill: I was getting bombarded with compiler errors with respect to method ambiguity and I just wanted to get the thing to work. I can try again.

coordinator/src/main/scala/filodb.coordinator/IngestionActor.scala
```diff
@@ -339,21 +349,14 @@ private[filodb] final class IngestionActor(dataset: Dataset,
...
-          logger.error(s"stopIngestion handler done. Ingestion failure for dataset=${dataset.ref} shard=$shard", f)
-          // release of resources on failure is already handled in the normalIngestion method
-      }(actorDispatcher)
-    }
-    shardIngestion.foreach(_.cancel())
+    streamSubscriptions.get(shard).foreach(_.cancel())
+  }
+
+  private def ingestionStopped(ref: DatasetRef, shard: Int): Unit = {
+    removeAndReleaseResources(ref, shard)
```
vishramachandran: This could be done on the endHandler itself. This way it would get scheduled on the ingestion thread.

broneill: This creates ordering races with the ingestion actor.

PR #445, opened 2019-07-18
feat(query): Ability to respond to gauge related PromQL queries from downsampled data

vishramachandran:
**Pull Request checklist**

Ability to respond to gauge queries from downsampled data.

* Use the right downsampled column based on the time range function used
* Modify the time range function based on whether we are querying from downsampled data. For example, count_over_time when applied on downsampled data is sum_over_time on the count column.

TODO:
* Currently presence of average downsample column is assumed. But we can do better by using just sum and count. 

Above todos can come in the next iteration (or I will look into this right away time permitting).

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
@@ -28,7 +28,6 @@ import filodb.query.exec._
   * ExecPlans as well as from the client to execute LogicalPlans.
   */
 class QueryEngine(dataset: Dataset,
-                  isDownsampledDataset: Boolean,
                   shardMapperFunc: => ShardMapper)
                    extends StrictLogging {
 
@@ -251,12 +250,9 @@ class QueryEngine(dataset: Dataset,
 
   private def materializePeriodicSeries(queryId: String,
                                         submitTime: Long,
-                                        options: QueryOptions,
-                                        lp: PeriodicSeries,
-                                        spreadProvider : SpreadProvider): PlanResult = {
-    val rawSeries = materializeRawSeries(queryId, submitTime, options,
-                                         lp.rawSeries.asInstanceOf[RawSeries],
-                                         None, spreadProvider)
+                                       options: QueryOptions,
+                                       lp: PeriodicSeries, spreadProvider : SpreadProvider): PlanResult = {
+    val rawSeries = walkLogicalPlanTree(lp.rawSeries, queryId, submitTime, options, spreadProvider)
     rawSeries.plans.foreach(_.addRangeVectorTransformer(PeriodicSamplesMapper(lp.start, lp.step, lp.end,
       None, None, Nil)))
     rawSeries
@@ -265,9 +261,7 @@ class QueryEngine(dataset: Dataset,
   private def materializeRawSeries(queryId: String,
                                    submitTime: Long,
                                    options: QueryOptions,
-                                   lp: RawSeries,
-                                   rangeFunction: Option[RangeFunctionId],
-                                   spreadProvider: SpreadProvider): PlanResult = {
+                                   lp: RawSeries, spreadProvider : SpreadProvider): PlanResult = {
     val colIDs = getColumnIDs(dataset, lp.columns)
     val renamedFilters = renameMetricFilter(lp.filters)
     val spreadChanges = spreadProvider.spreadFunc(renamedFilters)
@@ -278,7 +272,7 @@ class QueryEngine(dataset: Dataset,
     val execPlans = shardsFromFilters(renamedFilters, options, spreadProvider).map { shard =>
       val dispatcher = dispatcherForShard(shard)
       SelectRawPartitionsExec(queryId, submitTime, options.sampleLimit, dispatcher, dataset.ref, shard,
-        renamedFilters, toChunkScanMethod(lp.rangeSelector), isDownsampledDataset, colIDs)
+        renamedFilters, toChunkScanMethod(lp.rangeSelector), colIDs)
     }
     PlanResult(execPlans, needsStitch)
   }
@@ -377,12 +371,9 @@ class QueryEngine(dataset: Dataset,
     * as those are automatically prepended.
     */
   private def getColumnIDs(dataset: Dataset, cols: Seq[String]): Seq[Types.ColumnId] = {
-    val ids = dataset.colIDs(cols: _*)
+    dataset.colIDs(cols: _*)
       .recover(missing => throw new BadQueryException(s"Undefined columns $missing"))
       .get
-    // avoid duplication if first ids are already row keys
-    if (ids.take(dataset.rowKeyIDs.length) == dataset.rowKeyIDs) { ids }
-    else { dataset.rowKeyIDs ++ ids }
   }
 
   private def toChunkScanMethod(rangeSelector: RangeSelector): ChunkScanMethod = {
--- a/coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
@@ -28,6 +28,7 @@ import filodb.query.exec._
   * ExecPlans as well as from the client to execute LogicalPlans.
   */
 class QueryEngine(dataset: Dataset,
+                  isDownsampledDataset: Boolean,
                   shardMapperFunc: => ShardMapper)
                    extends StrictLogging {
 
@@ -131,7 +132,8 @@ class QueryEngine(dataset: Dataset,
   private def walkLogicalPlanTree(logicalPlan: LogicalPlan,
                                   queryId: String,
                                   submitTime: Long,
-                                  options: QueryOptions, spreadProvider: SpreadProvider): PlanResult = {
+                                  options: QueryOptions,
+                                  spreadProvider: SpreadProvider): PlanResult = {
 
     logicalPlan match {
       case lp: RawSeries                   => materializeRawSeries(queryId, submitTime, options, lp, spreadProvider)
@@ -252,7 +254,9 @@ class QueryEngine(dataset: Dataset,
                                         options: QueryOptions,
                                         lp: PeriodicSeries,
                                         spreadProvider : SpreadProvider): PlanResult = {
-    val rawSeries = walkLogicalPlanTree(lp.rawSeries, queryId, submitTime, options, spreadProvider)
+    val rawSeries = materializeRawSeries(queryId, submitTime, options,
+                                         lp.rawSeries.asInstanceOf[RawSeries],
+                                         None, spreadProvider)
     rawSeries.plans.foreach(_.addRangeVectorTransformer(PeriodicSamplesMapper(lp.start, lp.step, lp.end,
       None, None, Nil)))
     rawSeries
@@ -262,6 +266,7 @@ class QueryEngine(dataset: Dataset,
                                    submitTime: Long,
                                    options: QueryOptions,
                                    lp: RawSeries,
+                                   rangeFunction: Option[RangeFunctionId],
                                    spreadProvider: SpreadProvider): PlanResult = {
     val colIDs = getColumnIDs(dataset, lp.columns)
     val renamedFilters = renameMetricFilter(lp.filters)
@@ -273,7 +278,7 @@ class QueryEngine(dataset: Dataset,
     val execPlans = shardsFromFilters(renamedFilters, options, spreadProvider).map { shard =>
       val dispatcher = dispatcherForShard(shard)
       SelectRawPartitionsExec(queryId, submitTime, options.sampleLimit, dispatcher, dataset.ref, shard,
-        renamedFilters, toChunkScanMethod(lp.rangeSelector), colIDs)
+        renamedFilters, toChunkScanMethod(lp.rangeSelector), isDownsampledDataset, colIDs)
     }
     PlanResult(execPlans, needsStitch)
   }
@@ -329,7 +334,8 @@ class QueryEngine(dataset: Dataset,
   private def materializeRawChunkMeta(queryId: String,
                                       submitTime: Long,
                                       options: QueryOptions,
-                                      lp: RawChunkMeta, spreadProvider : SpreadProvider): PlanResult = {
+                                      lp: RawChunkMeta,
+                                      spreadProvider : SpreadProvider): PlanResult = {
     // Translate column name to ID and validate here
     val colName = if (lp.column.isEmpty) dataset.options.valueColumn else lp.column
     val colID = dataset.colIDs(colName).get.head
--- a/coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
@@ -249,8 +249,9 @@ class QueryEngine(dataset: Dataset,
 
   private def materializePeriodicSeries(queryId: String,
                                         submitTime: Long,
-                                       options: QueryOptions,
-                                       lp: PeriodicSeries, spreadProvider : SpreadProvider): PlanResult = {
+                                        options: QueryOptions,
+                                        lp: PeriodicSeries,
+                                        spreadProvider : SpreadProvider): PlanResult = {
     val rawSeries = walkLogicalPlanTree(lp.rawSeries, queryId, submitTime, options, spreadProvider)
     rawSeries.plans.foreach(_.addRangeVectorTransformer(PeriodicSamplesMapper(lp.start, lp.step, lp.end,
       None, None, Nil)))
@@ -260,7 +261,8 @@ class QueryEngine(dataset: Dataset,
   private def materializeRawSeries(queryId: String,
                                    submitTime: Long,
                                    options: QueryOptions,
-                                   lp: RawSeries, spreadProvider : SpreadProvider): PlanResult = {
+                                   lp: RawSeries,
+                                   spreadProvider: SpreadProvider): PlanResult = {
     val colIDs = getColumnIDs(dataset, lp.columns)
     val renamedFilters = renameMetricFilter(lp.filters)
     val spreadChanges = spreadProvider.spreadFunc(renamedFilters)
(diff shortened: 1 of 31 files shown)


query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
```diff
@@ -222,6 +222,80 @@ class SumAndMaxOverTimeFuncHD(maxColID: Int) extends ChunkedRangeFunction[Transi
...
+  }
+}
+
+/**
+  * Sums Histograms over time and also computes Max over time of a Max field.
+  */
+class AvgWithSumAndCountOverTimeFuncL extends ChunkedRangeFunction[TransientRow] {
+  private val sumFunc = new SumOverTimeChunkedFunctionL
+  private val countFunc = new CountOverTimeChunkedFunction
+  private val countColId: Int = 4 // FIXME hard coded for now
```
broneill: Comment why 4 is correct.

vishramachandran: this has been fixed

coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
```diff
@@ -368,13 +371,9 @@ class QueryEngine(dataset: Dataset,
     * as those are automatically prepended.
     */
   private def getColumnIDs(dataset: Dataset, cols: Seq[String]): Seq[Types.ColumnId] = {
-    val realCols = if (cols.isEmpty) Seq(dataset.options.valueColumn) else cols
```
velvia: Why are the extra lines being removed here?  I don't see how that has to do with the change.

core/src/main/scala/filodb.core/metadata/Dataset.scala
```diff
@@ -37,6 +37,7 @@ final case class Dataset(name: String,
                          dataColumns: Seq[Column],
                          rowKeyIDs: Seq[Int],
                          downsamplers: Seq[ChunkDownsampler],
+                         hasDownsampledData: Boolean = false,
```
velvia: To make the merge easier, can we
1) move hasDownsampledData to `DatasetOptions`?
2) Is it possible for dataColId to be moved somewhere else as well?  If it hasn't been here before it likely can be somewhere else.

query/src/main/scala/filodb/query/exec/SelectRawPartitionsExec.scala
```diff
@@ -18,8 +19,8 @@ object SelectRawPartitionsExec {
   def histMaxColumn(dataset: Dataset, colIDs: Seq[Types.ColumnId]): Option[Int] = {
     colIDs.find { id => dataset.dataColumns(id).columnType == HistogramColumn }
           .flatMap { histColID =>
-            dataset.dataColumns.find { c => c.name == "max" && c.columnType == DoubleColumn }
-          }.map(_.id)
+            dataset.dataColumns.find(_.name == "max").map(_.id)
```
velvia: Why is the criteria that max be a DoubleColumn removed?  That should remain there.

query/src/main/scala/filodb/query/exec/SelectRawPartitionsExec.scala
```diff
@@ -38,18 +39,39 @@ final case class SelectRawPartitionsExec(id: String,
                                          colIds: Seq[Types.ColumnId]) extends LeafExecPlan {
   import SelectRawPartitionsExec._
 
-  require(colIds.nonEmpty)
-
   protected[filodb] def schemaOfDoExecute(dataset: Dataset): ResultSchema = {
-    val numRowKeyCols = colIds.zip(dataset.rowKeyIDs).takeWhile { case (a, b) => a == b }.length
+    require(dataset.rowKeyIDs.forall(rk => !colIds.contains(rk)),
+      "User selected columns should not include timestamp (row-key); it will be auto-prepended")
```
velvia: Not sure we need to check for this, or what this really has to do with this PR.  Also this code won't work post-schema as we got rid of rowKeyIDs.  Please just leave it out to make merges easier.

query/src/main/scala/filodb/query/exec/SelectRawPartitionsExec.scala
```diff
@@ -38,18 +39,39 @@ final case class SelectRawPartitionsExec(id: String,
...
+
+    val selectedColIds = selectColIds(dataset)
+    val numRowKeyCols = selectedColIds.zip(dataset.rowKeyIDs).takeWhile { case (a, b) => a == b }.length
```
velvia: numRowKeyCols will be 1 going forward, always.  Can just hardcode to 1, again this will help reduce conflicts

query/src/main/scala/filodb/query/exec/SelectRawPartitionsExec.scala
```diff
@@ -38,18 +39,39 @@ final case class SelectRawPartitionsExec(id: String,
...
+        // query is selecting specific columns
+        colIds
+      } else if (!dataset.hasDownsampledData) {
+        // needs to select raw data
+        colIds ++ dataset.colIDs(dataset.options.valueColumn).get
+      } else {
+        // need to select column based on range function
+        val colNames = rangeVectorTransformers.find(_.isInstanceOf[PeriodicSamplesMapper]).map { p =>
+          RangeFunction.downsampleColsFromRangeFunction(dataset, p.asInstanceOf[PeriodicSamplesMapper].functionId)
+        }.getOrElse(Seq("avg"))
```
velvia: Please use the default value column from the dataset

query/src/main/scala/filodb/query/exec/SelectRawPartitionsExec.scala
```diff
@@ -38,18 +39,39 @@ final case class SelectRawPartitionsExec(id: String,
...
+      if (colIds.nonEmpty) {
+        // query is selecting specific columns
+        colIds
+      } else if (!dataset.hasDownsampledData) {
+        // needs to select raw data
+        colIds ++ dataset.colIDs(dataset.options.valueColumn).get
+      } else {
+        // need to select column based on range function
+        val colNames = rangeVectorTransformers.find(_.isInstanceOf[PeriodicSamplesMapper]).map { p =>
+          RangeFunction.downsampleColsFromRangeFunction(dataset, p.asInstanceOf[PeriodicSamplesMapper].functionId)
```
velvia: I'm thinking philosophically that instead of embedding the choice of columns here in the execution phase, it is better to select it automatically in the query planning phase.  However in the future one might not know it is downsampled data until the execution phase.  Just that it keeps everything cleaner at the lower level, because now we have multiple places where this logic is.  Also should not need to automatically add rowKeyIDs here, it might already have been added before.

query/src/main/scala/filodb/query/exec/SelectRawPartitionsExec.scala
```diff
@@ -58,10 +80,14 @@ final case class SelectRawPartitionsExec(id: String,
                           queryConfig: QueryConfig)
                          (implicit sched: Scheduler,
                           timeout: FiniteDuration): Observable[RangeVector] = {
-    require(colIds.indexOfSlice(dataset.rowKeyIDs) == 0)
+    require(dataset.rowKeyIDs.forall(rk => !colIds.contains(rk)),
```
velvia: Again prefer not to add this check here.  rowKeyIDs will be gone; and likely how to select columns will have to be redone in the near future anyways.

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
```diff
@@ -222,6 +222,78 @@ class SumAndMaxOverTimeFuncHD(maxColID: Int) extends ChunkedRangeFunction[Transi
   }
 }
 
+/**
+  * Sums Histograms over time and also computes Max over time of a Max field.
```
velvia: This is not for histograms.... please make relevant

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
```diff
@@ -222,6 +222,78 @@ class SumAndMaxOverTimeFuncHD(maxColID: Int) extends ChunkedRangeFunction[Transi
...
+    countFunc.reset()
+  }
+
+  final def apply(endTimestamp: Long, sampleToEmit: TransientRow): Unit = {
+    sampleToEmit.setValues(endTimestamp, sumFunc.sum / countFunc.sum)
+  }
+
+  import BinaryVector.BinaryVectorPtr
+
+  final def addChunks(tsVector: BinaryVectorPtr, tsReader: bv.LongVectorDataReader,
```
velvia: I thought we were just going to use average column?   If so why is this still needed?

query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
```diff
@@ -225,82 +226,128 @@ trait ChunkedLongRangeFunction extends TimeRangeFunction[TransientRow] {
...
-    case None                 => () => new LastSampleChunkedFunctionL
-    case Some(CountOverTime)  => () => new CountOverTimeChunkedFunction()
-    case Some(SumOverTime)    => () => new SumOverTimeChunkedFunctionL
-    case Some(AvgOverTime)    => () => new AvgOverTimeChunkedFunctionL
-    case Some(MinOverTime)    => () => new MinOverTimeChunkedFunctionL
-    case Some(MaxOverTime)    => () => new MaxOverTimeChunkedFunctionL
-    case Some(StdDevOverTime) => () => new StdDevOverTimeChunkedFunctionL
-    case Some(StdVarOverTime) => () => new StdVarOverTimeChunkedFunctionL
-    case _                    => iteratingFunction(func, funcParams)
+  def longChunkedFunction(dataset: Dataset,
```
velvia: Prefer to pass in `DatasetOptions` here.  In the future, this would be replaced by Schema, or DataSchema.

query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
```diff
@@ -225,82 +226,128 @@ trait ChunkedLongRangeFunction extends TimeRangeFunction[TransientRow] {
...
+                          func: Option[RangeFunctionId],
+                          funcParams: Seq[Any] = Nil): RangeFunctionGenerator = {
```
velvia: I prefer to not change the indentation and keep func match on the previous line.

query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
```diff
@@ -225,82 +226,128 @@ trait ChunkedLongRangeFunction extends TimeRangeFunction[TransientRow] {
...
-    case Some(Delta)    if config.has("faster-rate") => () => new ChunkedDeltaFunction
-    case Some(CountOverTime)  => () => new CountOverTimeChunkedFunctionD()
-    case Some(SumOverTime)    => () => new SumOverTimeChunkedFunctionD
-    case Some(AvgOverTime)    => () => new AvgOverTimeChunkedFunctionD
-    case Some(MinOverTime)    => () => new MinOverTimeChunkedFunctionD
-    case Some(MaxOverTime)    => () => new MaxOverTimeChunkedFunctionD
-    case Some(StdDevOverTime) => () => new StdDevOverTimeChunkedFunctionD
-    case Some(StdVarOverTime) => () => new StdVarOverTimeChunkedFunctionD
-    case _                    => iteratingFunction(func, funcParams)
+                            funcParams: Seq[Any] = Nil): RangeFunctionGenerator = {
```
velvia: Again keep the old indentation the way it is, why change it?

velvia: It makes the changes much harder to see

query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
```diff
@@ -225,82 +226,128 @@ trait ChunkedLongRangeFunction extends TimeRangeFunction[TransientRow] {
...
-    case None                 => () => new LastSampleChunkedFunctionH
-    case Some(SumOverTime) if maxCol.isDefined => () => new SumAndMaxOverTimeFuncHD(maxCol.get)
-    case Some(SumOverTime)    => () => new SumOverTimeChunkedFunctionH
-    case Some(Rate)           => () => new HistRateFunction
-    case Some(Increase)       => () => new HistIncreaseFunction
-    case _                    => ???
+  def histChunkedFunction(dataset: Dataset,
+                          func: Option[RangeFunctionId],
+                          funcParams: Seq[Any] = Nil): RangeFunctionGenerator = {
+    val maxCol = dataset.dataColumns.find(_.name == "max").map(_.id)
```
velvia: I do not think the old method of finding max and passing it in really needed to be changed.  The new way spreads the logic for finding the max column in more places and has less validation (does not check it is double, for example)

coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
vishramachandran: Essentially, I am making the query planner not decide what column needs to be chosen. This would be done at runtime based on what range function is used etc.

I want one place where column selection happens. Without this change, it would be in multiple places.

core/src/main/scala/filodb.core/metadata/Dataset.scala
vishramachandran: 1. Done
2. Hmm... To where? When your PR introduces data schema, this would have to be moved in there.

query/src/main/scala/filodb/query/exec/SelectRawPartitionsExec.scala
vishramachandran: Sure. This was unintentionally removed due to refactor.

query/src/main/scala/filodb/query/exec/SelectRawPartitionsExec.scala
vishramachandran: As mentioned in another comment, I want one place where column selection happens. The require makes the logic easy to understand.

query/src/main/scala/filodb/query/exec/SelectRawPartitionsExec.scala
vishramachandran: done

query/src/main/scala/filodb/query/exec/SelectRawPartitionsExec.scala
vishramachandran: Hence removed it everywhere else but here. 

Fully agree that it should happen in one place. I took the "here" approach as against query planning.
Hence column selection was removed from everywhere else and placed right here. 

That is the reason there is the assertion "require rowKeys to not be present".

It is debatable where column selection needs to happen: here during runtime, or at planning time. Let's defer that decision to the phase when we address queries for multi-schema. A lot more information needs to go into that decision making. For now, I have colocated all relevant code in one place. If it needs to be moved elsewhere, it can happen in a later PR.

query/src/main/scala/filodb/query/exec/SelectRawPartitionsExec.scala
vishramachandran: done

query/src/main/scala/filodb/query/exec/SelectRawPartitionsExec.scala
vishramachandran: Agree that this will change in the future. Simplification of column selection forced me to put this check here since it was quite ambiguous whether row keys should/shod-not be selected. I just fixed it and set the rules. I am open to changing this in a later PR.

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
vishramachandran: Thanks. Done

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
vishramachandran: AvgOverTime cannot use avg column since average of averages is not the same.

query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
vishramachandran: I moved the hasDownsampledFlag to DatasetOptions now, but I really don't like moving the options around. Dont want to introduce a new pattern.

query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
vishramachandran: Can we please keep the extra paren? Makes it easier to read and add code.

query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
vishramachandran: sure. Intellij.

query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
vishramachandran: OK. I'll look at reverting. I had to change it after adding explicit config for max column, and then reverted it later.

query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
vishramachandran: I also had to get past the line width limitation.

coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
vishramachandran: I moved all the column selection logic to one place, during query execution. Please see further explanation in other threads.

query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
vishramachandran: done

query/src/main/scala/filodb/query/exec/SelectRawPartitionsExec.scala
velvia: 👍

PR #451, opened 2019-07-29
feat(query, core): QueryResult need not always be serializable

vishramachandran:
**Pull Request checklist**

QueryResult uses SerializableRangeVectors only.

**New behavior :**

Since QueryResult can be a result of an ExecPlan that is executed within the current JVM, we should not be forced into converting results into a SerializableRangeVector.

Hence the interface changes

### DIFF
--- a/query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
+++ b/query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
@@ -5,7 +5,7 @@ import scala.concurrent.duration.FiniteDuration
 import monix.execution.Scheduler
 import monix.reactive.Observable
 
-import filodb.core.{query, DatasetRef}
+import filodb.core.DatasetRef
 import filodb.core.binaryrecord2.BinaryRecordRowReader
 import filodb.core.memstore.{MemStore, PartKeyRowReader}
 import filodb.core.metadata.Column.ColumnType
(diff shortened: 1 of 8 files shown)


query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
```diff
@@ -83,17 +81,12 @@ final case class LabelValuesDistConcatExec(id: String,
...
-      var metadataResult = scala.collection.mutable.Set.empty[Map[ZeroCopyUTF8String, ZeroCopyUTF8String]]
-      resp.foreach(rv => {
-        metadataResult ++= rv(0).rows.map(rowReader => {
-          val binaryRowReader = rowReader.asInstanceOf[BinaryRecordRowReader]
-          rv(0).schema.toStringPairs(binaryRowReader.recordBase, binaryRowReader.recordOffset)
-            .map(pair => pair._1.utf8 -> pair._2.utf8).toMap
-        })
-      })
+      val metadataResult = resp.map { rv =>
+        rv(0).key.labelValues.map(pair => pair._1 -> pair._2)
```
sherali42: @vishramachandran this may not work, as the key is a dummy key. Map of label name/values are returned as rows, schema has one column with type as Map.

vishramachandran: Ah - I think I get it now.

vishramachandran: Fixed.

vishramachandran:
@sherali42 please take another look

vishramachandran:
@velvia Not anymore. @TanviBhavsar 's change (see her PR) is introducing an InProcessDispatcher which for which we should not force the use of SerializableRangeVector.

PR #453, opened 2019-07-31
bug(core): Release searchers after each acquire

vishramachandran:
**Pull Request checklist**

After each `acquire` of searcher, we need to call `release` so that searcher releases the segments and they are deleted after merge. Prior to this change, release calls were not being made.

Currently, we are seeing a constant growth of disk space because old segments are never deleted. Expecting that this change will fix it.

Also renamed commit operation to refresh. Those two are technically different in Lucene. We are actually not doing commit at the bottom of the call stack.

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
+++ b/core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
@@ -158,8 +158,18 @@ class PartKeyLuceneIndex(dataset: Dataset,
   def partIdsEndedBefore(endedBefore: Long): EWAHCompressedBitmap = {
     val collector = new PartIdCollector()
     val deleteQuery = LongPoint.newRangeQuery(PartKeyLuceneIndex.END_TIME, 0, endedBefore)
-    searcherManager.acquire().search(deleteQuery, collector)
+
+    withNewSearcher(s => s.search(deleteQuery, collector))
     collector.result
+  }
+
+  private def withNewSearcher(func: IndexSearcher => Unit): Unit = {
+    val s = searcherManager.acquire()
+    try {
+      func(s)
+    } finally {
+      searcherManager.release(s)
+    }
   }
 
   /**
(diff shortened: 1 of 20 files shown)


core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -158,10 +158,20 @@ class PartKeyLuceneIndex(dataset: Dataset,
   def partIdsEndedBefore(endedBefore: Long): EWAHCompressedBitmap = {
     val collector = new PartIdCollector()
     val deleteQuery = LongPoint.newRangeQuery(PartKeyLuceneIndex.END_TIME, 0, endedBefore)
-    searcherManager.acquire().search(deleteQuery, collector)
+
+    withNewSearcher(s => s.search(deleteQuery, collector))
     collector.result
   }
 
+  private def withNewSearcher(func: IndexSearcher => Unit): Unit = {
```
velvia: Just one suggestion: you could make `withNewSearcher` return a value to make code slightly better:

    def withNewSearcher[B](func: IndexSearcher => B): B

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -202,46 +212,54 @@ class PartKeyLuceneIndex(dataset: Dataset,
...
-  def indexNames: scala.Iterator[String] = {
-    val searcher = searcherManager.acquire()
-    val indexReader = searcher.getIndexReader
-    val segments = indexReader.leaves()
-    segments.asScala.iterator.flatMap { segment =>
-      segment.reader().getFieldInfos.asScala.toIterator.map(_.name)
-    }.filterNot { n => ignoreIndexNames.contains(n) }
+  def indexNames(limit: Int): Seq[String] = {
+
+    var ret: Seq[String] = Nil
```
velvia: See suggestion above so that `withNewSearcher` can return a value.  Makes the var totally unnecessary

PR #456, opened 2019-08-08
bug(query): RecordBuilder should be part of PromQlExec

TanviBhavsar:
**Pull Request checklist**

**Current behavior :**
When there are multiple remote queries RecordBuilder throws IllegalArgumentException in checkFieldNo as the same builder is getting used in generating QueryResponse

**New behavior :**
Every PromQlExec has its own builder

### DIFF
--- a/query/src/main/scala/filodb/query/exec/PromQlExec.scala
+++ b/query/src/main/scala/filodb/query/exec/PromQlExec.scala
@@ -28,6 +28,9 @@ case class PromQlExec(id: String,
   protected def args: String = params.toString
   import PromQlExec._
 
+  val recSchema = SerializableRangeVector.toSchema(columns)
+  val builder = SerializableRangeVector.toBuilder(recSchema)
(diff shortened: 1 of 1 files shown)


query/src/main/scala/filodb/query/exec/PromQlExec.scala
```diff
@@ -28,6 +28,9 @@ case class PromQlExec(id: String,
   protected def args: String = params.toString
   import PromQlExec._
 
+  val recSchema = SerializableRangeVector.toSchema(columns)
```
vishramachandran: recSchema can still be inside the companion object. It is immutable across queries and can be reused.

TanviBhavsar: Moved it to PromQlExec object

TanviBhavsar:
> Great catch!

Thanks Evan!

PR #460, opened 2019-08-12
bug[memory]: Guard BlockManager data structures with the lock.

broneill:
bug[memory]: Guard BlockManager data structures with the lock. This should protect blocks from being reclaimed before they're fully switched to the "in use" state with a timestamp.

**Pull Request checklist**

**Current behavior :**
Observed MatchError exceptions when ODP'ng data older than the retention period. Possible cause is immediate reclamation of the blocks before they're moved to the "in use" state.

**New behavior :**


broneill:
I'll follow up with another PR that adds logging.

PR #468, opened 2019-08-15
bug(query) : replace AkkaHttpBackend with AsyncHttpClientFutureBackend

mishra-arun:
**Pull Request checklist**

**Current behavior :
AkkaHttpBackend internally creates a new actor system which binds to a random port. On compute, we will need to manage a new port in this case otherwise it can fail with `address already in use` error.

**New behavior :**
Replacing akka backend with async-client backend. Also adding a read timeout to http request.

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
@@ -3,7 +3,7 @@ package filodb.coordinator.queryengine2
 import java.util.UUID
 import java.util.concurrent.ThreadLocalRandom
 
-import scala.concurrent.duration.FiniteDuration
+import scala.concurrent.duration._
 
 import akka.actor.ActorRef
 import com.typesafe.config.{Config, ConfigFactory}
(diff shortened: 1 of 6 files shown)


coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
```diff
@@ -82,13 +82,13 @@ class QueryEngine(dataset: Dataset,
         case route: RemoteRoute =>
           val timeRange = route.timeRange.get
           val queryParams = tsdbQueryParams.asInstanceOf[PromQlQueryParams]
-          val endpoint = queryEngineConfig.isEmpty() match {
-            case false => queryEngineConfig.getString("routing.buddy.http.endpoint")
-            case _     => ""
-          }
+          val endpoint = if (queryEngineConfig.hasPath("routing.buddy.http.endpoint"))
```
velvia: More idiomatic Scala with less repetition:

    val endpoint = queryEngineConfig.as[Option[String]]("routing.buddy.http.endpoint").getOrElse("")
    val readTimeout = queryEngineConfig.as[Option[FiniteDuration]]("routing.buddy.http.timeout").getOrElse(60.seconds)

mishra-arun: fixed.

PR #469, opened 2019-08-15
feat(core): Redefine chunk id to incorporate ingestion time.

broneill:
**Pull Request checklist**

**Current behavior :**
ChunkID is defined as a 43 bit user time (millis from 1970) combined with a 21 bit nano time disambiguator.

**New behavior :**
ChunkID is now defined as a 42 bit user time combined with a 22 bit ingestion time (seconds from 1970) modulo 48 days.

**BREAKING CHANGES**
This change in the chunkID definition might create collisions, although it's unlikely.

### DIFF
--- a/core/src/main/scala/filodb.core/store/package.scala
+++ b/core/src/main/scala/filodb.core/store/package.scala
@@ -98,19 +98,17 @@ package object store {
    * no wraparound in the middle of the day (full binary encoding supports 48.545 days)
    *
    * @param startTime milliseconds since 1970
+   * @param ingestionTime seconds since 1970
    */
-  @inline final def chunkID(startTime: Long): Long =
-    chunkID(startTime, System.currentTimeMillis() / 1000)
-
-  @inline final def chunkID(startTime: Instant, ingestionTime: Instant): Long =
-    chunkID(startTime.toEpochMilli(), ingestionTime.getEpochSecond())
+  @inline final def chunkID(startTime: Long, ingestionTime: Long): Long =
+    (startTime << startTimeShift) | Math.floorMod(ingestionTime, (48 * 24 * 60 * 60))
 
   /**
    * @param startTime milliseconds since 1970
    * @param ingestionTime seconds since 1970
    */
-  @inline final def chunkID(startTime: Long, ingestionTime: Long): Long =
-    (startTime << startTimeShift) | Math.floorMod(ingestionTime, (48 * 24 * 60 * 60))
+  @inline final def chunkID(startTime: Instant, ingestionTime: Instant): Long =
+    chunkID(startTime.toEpochMilli(), ingestionTime.getEpochSecond())
 
   /**
    * Returns the start time portion of the chunk ID, as milliseconds from 1970.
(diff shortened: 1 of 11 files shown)


broneill:
Not going to really help without an absolute ingestion time.

core/src/main/scala/filodb.core/store/package.scala
```diff
@@ -87,16 +88,37 @@ package object store {
...
+  @inline final def chunkID(startTime: Long, ingestionTime: Long): Long =
+    (startTime << startTimeShift) | Math.floorMod(ingestionTime, (48 * 24 * 60 * 60))
 
-  @inline final def chunkID(startTime: Long, currentTime: Long): Long =
-    (startTime << msBitOffset) | (currentTime & lowerBitsMask)
+  /**
+   * @param startTime milliseconds since 1970
+   * @param ingestionTime seconds since 1970
+   */
+  @inline final def chunkID(startTime: Instant, ingestionTime: Instant): Long =
```
velvia: Why have a variation for `Instant`?   Can we get rid of this?

core/src/main/scala/filodb.core/store/package.scala
```diff
@@ -87,16 +88,37 @@ package object store {
...
+
+  /**
+   * Returns the start time portion of the chunk ID, as milliseconds from 1970.
+   */
+  @inline final def startTimeFromChunkID(chunkID: Long): Long = chunkID >>> startTimeShift
+
+  /**
+   * Returns the ingestion time portion of the chunk ID, as seconds from 1970, modulo 48 days.
+   */
+  @inline final def ingestionTimeFromChunkID(chunkID: Long): Long = chunkID & ingestionTimeMask
```
velvia: This is not the ingestion Time, it is the ingestionTimeMod48Days.... the method name is kinda misleading

jmh/src/main/scala/filodb.jmh/IntSumReadBenchmark.scala
```diff
@@ -31,7 +31,7 @@ class IntSumReadBenchmark {
   import IntSumReadBenchmark._
   val NumRows = 10000
 
-  val chunkSet = ChunkSet(dataset, partKey, rowIt.map(TupleRowReader).take(NumRows).toSeq, TestData.nativeMem)
+  val chunkSet = ChunkSet(dataset, partKey, 0, rowIt.map(TupleRowReader).take(NumRows).toSeq, TestData.nativeMem)
```
velvia: Suggestion: default the ingestionTime to 0 so that we can avoid changing the call in tests.  Will make merging easier for other PRs.

core/src/main/scala/filodb.core/store/package.scala
broneill: For testing.

jmh/src/main/scala/filodb.jmh/IntSumReadBenchmark.scala
broneill: A default of 0 for testing is dangerous. It makes it too easy to forget about it in code that really needs to specify it.

PR #474, opened 2019-08-21
misc(cassandra): Stop querying the chunk index table and use the regular table instead.

broneill:
**Pull Request checklist**

**Current behavior :**
A special chunk index table is maintained and queried for obtaining chunk info records without having to load full chunks.

**New behavior :**
The regular chunk table has the same primary key as the index table, and so it can be queried for chunk infos without needing to maintain a separate table. The downside is that it might increase the load on the Cassandra server, but it means less tables need to be maintained and repaired (in case of disaster). A future version will remove the chunk index table entirely.


vishramachandran:
@broneill Also, did you modify the write path to not write the info to two tables? 
Not needed anymore with this change.

broneill:
My concern is that we might want to switch back if there's an issue. I don't want to have to backfill the index table, so a future commit (a week perhaps) will remove everything.

PR #477, opened 2019-08-22
feat(cassandra, core): Drop the chunk index table and replace it with a table keyed by ingestion time.

broneill:
**Pull Request checklist**

**Current behavior :**
The chunk index table simply maps chunk id to chunk set info, and ingestion time isn't record.

**New behavior :**
A new ingestion table is keyed by ingestion time, and the chunk set info records also store the ingestion time.

**BREAKING CHANGES**
This completely breaks compatibility with existing Cassandra records. Chunks cannot be discovered anymore because the new table will start initially empty. In addition, the format of the chunk set info records has changed, and so a backfill isn't really possible without re-encoding the records to include a false ingestion time. The info records are also stored in the tschunks table, but they cannot be decoded for the same reason.


cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
```diff
@@ -259,17 +260,17 @@ trait CassandraChunkSource extends RawChunkSource with StrictLogging {
...
                         partMethod: PartitionScanMethod,
                         chunkMethod: ChunkScanMethod = AllChunkScan): Observable[RawPartData] = {
-    val indexTable = getOrCreateIndexTable(dataset.ref)
+    val ingestionTable = getOrCreateIngestionTable(dataset.ref)
     logger.debug(s"Scanning partitions for ${dataset.ref} with method $partMethod...")
-    val (filters, indexRecords) = partMethod match {
+    val (filters, infoRecords) = partMethod match {
       case SinglePartitionScan(partition, _) =>
-        (Nil, indexTable.getIndices(partition))
+        (Nil, ingestionTable.getInfos(partition))
```
vishramachandran: Can we include a couple of Kamon traces to measure the breakdown of how much time we spend in fetching infos vs chunks.

core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
```diff
@@ -129,18 +130,26 @@ object ChunkSetInfo extends StrictLogging {
...
+   * the chunkID be readily available at offset 0, which is one reason to have it. It's
+   * possible that ChunkMap could call a method instead, but storing the chunkID also offers
+   * flexibility in the future, in case the definition changes.
    */
   val OffsetChunkID = 0
-  val OffsetNumRows = 8
-  val OffsetStartTime = 12
-  val OffsetEndTime = 20
-  val OffsetVectors = 28
+  val OffsetIngestionTime = 8
```
vishramachandran: Why is IngestionTime needed in the ChunkInfo ?
Is it useful at all long term ?

cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
```diff
@@ -63,36 +63,36 @@ extends ColumnStore with CassandraChunkSource with StrictLogging {
     val chunkTable = getOrCreateChunkTable(dataset)
     val partIndexTable = getOrCreatePartitionIndexTable(dataset)
     clusterConnector.createKeyspace(chunkTable.keyspace)
-    val indexTable = getOrCreateIndexTable(dataset)
+    val ingestionTable = getOrCreateIngestionTable(dataset)
```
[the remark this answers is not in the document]
broneill: ok

cassandra/src/main/scala/filodb.cassandra/columnstore/IngestionTable.scala
```diff
@@ -0,0 +1,108 @@
...
+    val query = inPartReadCql.bind().setList(0, partitions.map(toBuffer).asJava, classOf[ByteBuffer])
+    val it = session.execute(query).asScala.toIterator.map(fromRow)
+    Observable.fromIterator(it).handleObservableErrors
+  }
+
+  val tokenQ = "TOKEN(partition)"
+
+  def scanInfos(tokens: Seq[(String, String)]): Observable[InfoRecord] = {
+    def cql(start: String, end: String): String =
+      s"SELECT * FROM $tableString WHERE $tokenQ >= $start AND $tokenQ < $end " +
```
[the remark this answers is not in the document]
broneill: ok

cassandra/src/main/scala/filodb.cassandra/columnstore/IngestionTable.scala
```diff
@@ -0,0 +1,108 @@
...
+
+  def getMultiInfos(partitions: Seq[Array[Byte]]): Observable[InfoRecord] = {
+    val query = inPartReadCql.bind().setList(0, partitions.map(toBuffer).asJava, classOf[ByteBuffer])
+    val it = session.execute(query).asScala.toIterator.map(fromRow)
+    Observable.fromIterator(it).handleObservableErrors
+  }
+
+  val tokenQ = "TOKEN(partition)"
+
+  def scanInfos(tokens: Seq[(String, String)]): Observable[InfoRecord] = {
```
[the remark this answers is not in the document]
broneill: This is a feature which can appear in a separate PR.

core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
broneill: Two reasons: 1. I cannot easily stash the ingestion time anywhere else and pass it along down to the Cassandra layer. The full ingestion time cannot be derived from the chunk ID.
2. It's always a good idea to have all the information necessary to reconstruct a secondary index from the primary table. Without this, this the full ingestion time wouldn't be stored in the main chunk table.

cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
```diff
@@ -136,14 +136,15 @@ extends ColumnStore with CassandraChunkSource with StrictLogging {
...
-      indexTable.writeIndices(partition, indices, sinkStats, diskTimeToLive)
+  private def writeIngestion(dataset: Dataset,
+                             partition: Array[Byte],
+                             chunkset: ChunkSet,
+                             diskTimeToLive: Int): Future[Response] = {
+    asyncSubtrace("write-ingestion", "ingestion") {
+      val ingestionTable = getOrCreateIngestionTable(dataset.ref)
+      val info = chunkset.info
+      val infos = Seq((info.ingestionTime, info.startTime, ChunkSetInfo.toBytes(info)))
+      ingestionTable.writeIngestion(partition, infos, sinkStats, diskTimeToLive)
```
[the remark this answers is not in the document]
broneill: Any repair based on ingestion time would need this, and the storage overhead isn't that much. A separate TTL just makes things more complicated and doesn't offer much benefit, in my opinion.

cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
broneill: I don't know how to measure this time with Cassandra, Futures, Observables, etc. Is there an example of this somewhere?

cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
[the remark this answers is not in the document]
vishramachandran: Ah yes - downsampled records would indeed need it for repair. I forgot about that. But would we ever repair data more than a week old? Would reverse index be needed for old data?

I thought we discussed about having a lower TTL for this. What changed since that time ?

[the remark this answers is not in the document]
broneill: For the time being, ODP queries fetch records from the ingestion index table first, and so those records must live at least as long as the records they refer to. Otherwise, old data cannot be queried at all.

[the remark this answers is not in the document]
vishramachandran: Agreed. I only ask that there be a separate parameter for the IngestionTimeRecord TTL. The values fed for both can be the same.

cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
vishramachandran: Search for  `val span = Kamon.buildSpan("write-chunkset").start()` in CassandraColumnStore.scala

cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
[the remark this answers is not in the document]
broneill: It's trivial to change later when the need arises. I cannot see why it makes any difference, and adding another parameter just makes things a little more complicated with no clear benefit. I'd have to modify the ChunkSink trait too. If more secondary indexes are added, does each one require a separate TTL? This is a leaky abstraction.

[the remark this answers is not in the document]
vishramachandran: The idea was just to save storage space in Cassandra. We discussed that we didn't need to keep this around for 6 months, possibly more when it increases. I was pushing for it since I would be needing it when I write  downsampled data, and would pass separate lower TTL value there. 

Anyway, I defer to you to make this call. 

That said, I see your point on the leaky abstractions. This config value would be part of the cassandra config  passed into the impl constructor, and not be included in the Store interface.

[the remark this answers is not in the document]
broneill: The discussion regarding the TTL was before I understood exactly how the index table was being used. Replacing the index table and changing the TTL can introduce new problems that don't exist today. Until we can safely say that the index has no purpose other than doing repairs, it's safest to keep the TTLs consistent among the two chunk tables.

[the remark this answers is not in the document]
broneill: Also, this change is disruptive enough that I don't think its a good idea to make more changes than is necessary. A later PR can address new features and optimizations.

cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
```diff
@@ -63,36 +63,36 @@ extends ColumnStore with CassandraChunkSource with StrictLogging {
     val chunkTable = getOrCreateChunkTable(dataset)
     val partIndexTable = getOrCreatePartitionIndexTable(dataset)
     clusterConnector.createKeyspace(chunkTable.keyspace)
-    val indexTable = getOrCreateIndexTable(dataset)
+    val ingestionTable = getOrCreateIngestionTimeIndexTable(dataset)
```
velvia: I think `indexTable` is still the better variable name here.  It is still an index.  `ingestionTable` makes it sound like it is the primary table for ingestion data, which is not true.

cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
```diff
@@ -107,13 +107,13 @@ extends ColumnStore with CassandraChunkSource with StrictLogging {
                val span = Kamon.buildSpan("write-chunkset").start()
                val partBytes = dataset.partKeySchema.asByteArray(chunkset.partition)
                val future =
-                 for { writeChunksResp  <- writeChunks(dataset.ref, partBytes, chunkset, diskTimeToLive)
-                       writeIndexResp   <- writeIndices(dataset, partBytes, chunkset, diskTimeToLive)
-                                           if writeChunksResp == Success
+                 for { writeChunksResp    <- writeChunks(dataset.ref, partBytes, chunkset, diskTimeToLive)
+                       writeIngestionResp <- writeIngestion(dataset, partBytes, chunkset, diskTimeToLive)
```
velvia: Again I think `writeIndices` is still more descriptive than `writeIngestion`

cassandra/src/main/scala/filodb.cassandra/columnstore/IngestionTimeIndexTable.scala
```diff
@@ -0,0 +1,106 @@
...
+import filodb.memory.format.UnsafeUtils
+
+case class InfoRecord(binPartition: ByteBuffer, data: ByteBuffer) {
+  def partBaseOffset: (Any, Long, Int) = UnsafeUtils.BOLfromBuffer(binPartition)
+}
+
+/**
+ * Mapping to chunk set info records using a full ingestion time, positioned in the cluster key
+ * before the chunk start time. This makes it possible to find chunks based on the time they
+ * were actually ingested, which is useful for performing cross-DC repairs. In addition,
```
velvia: or downsampling

cassandra/src/main/scala/filodb.cassandra/columnstore/IngestionTimeIndexTable.scala
```diff
@@ -0,0 +1,106 @@
...
+ * queries for all chunk set infos for a given partition are more efficient than using
+ * TimeSeriesChunksTable. This is because the chunks are likely to be fetched from Cassandra
+ * due to locality when using TimeSeriesChunksTable.
```
velvia: and from reading from a much smaller table

core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
```diff
@@ -129,18 +130,26 @@ object ChunkSetInfo extends StrictLogging {
...
-   * +28 long[] pointers to each vector
+   * +8  long   ingestion time (milliseconds from 1970)
+   * +16 int    # rows in chunkset
+   * +20 long   start timestamp
+   * +28 long   end timestamp
+   * +36 long[] pointers to each vector
    *
    * Note that block metadata has a 4-byte partition ID appended to the front also.
+   *
+   * Note 2: The chunkID is currently defined from the start timestamp and the ingestion time,
```
velvia: We don't really need to store the start timestamp explicitly.... as it can be extracted completely from the chunkID right?   Why not just leave out the start timestamp field and derive it form the chunkID, then we save 8 bytes?

core/src/main/scala/filodb.core/store/ChunkSink.scala
```diff
@@ -64,8 +64,8 @@ class ChunkSinkStats {
   private val chunkBytesHist     = Kamon.histogram("chunk-bytes-per-call")
   private val chunkLenHist       = Kamon.histogram("chunk-length")
 
-  private val numIndexWriteCalls = Kamon.counter("index-write-calls-num")
-  private val indexBytesHist     = Kamon.histogram("index-bytes-per-call")
+  private val numIngestionWriteCalls = Kamon.counter("ingestion-write-calls-num")
```
velvia: I also think these should still be `index` and not `ingestion`....  `ingestion-write-calls` is gonna be really confusing

core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
broneill: I don't like the idea of hard-coding the chunkID TOO much. If the start timestamp is derived from the chunkID, there's no going back.

velvia: I suppose..... though any change in the chunkID schema will be a breaking change anyways, it's not really any different from a chunkSetInfo schema change.

cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
broneill: I think this work belongs in a separate PR.

core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
broneill: Changed definition to closely match the current one. Less compatibility issues.

PR #478, opened 2019-08-23
feat(query): changes function support

TanviBhavsar:
**Pull Request checklist**

Support for prometheus changes function


vishramachandran:
@TanviBhavsar Did you try writing a `ChunkedRangeFunction` implementation for `changes` ? It would be faster.

TanviBhavsar:
> @TanviBhavsar Did you try writing a `ChunkedRangeFunction` implementation for `changes` ? It would be faster. You'll need to add additional methods (besides sum and count) to the vector classes.

@vishramachandran I have added chunked function for changes

memory/src/main/scala/filodb.memory/format/vectors/DeltaDeltaVector.scala
```diff
@@ -253,6 +257,20 @@ object DeltaDeltaConstDataReader extends LongVectorDataReader {
       out
     }
   }
+
+  def changes(vector: BinaryVectorPtr, start: Int, end: Int): Double = {
```
vishramachandran: Please double check with @velvia  that code duplication cannot be avoided.

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
```diff
@@ -133,6 +133,45 @@ class SumOverTimeFunction(var sum: Double = Double.NaN, var count: Int = 0) exte
   }
 }
 
+class ChangesFunction(var prev1: Double = Double.NaN, var changes: Double = Double.NaN) extends RangeFunction {
```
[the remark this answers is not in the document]
velvia: I agree with Vish, let's remove this one

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
```diff
@@ -550,4 +589,37 @@ class StdVarOverTimeChunkedFunctionL extends VarOverTimeChunkedFunctionL() {
...
+class ChangesChunkedFunctionL extends ChangesChunkedFunction with
+  ChunkedLongRangeFunction{
+  final def addTimeLongChunks(longVect: BinaryVector.BinaryVectorPtr,
+                              longReader: bv.LongVectorDataReader,
+                              startRowNum: Int,
+                              endRowNum: Int): Unit = {
+    if (changes.isNaN) {
+      changes = 0d
+    }
+    changes += longReader.sum(longVect, startRowNum, endRowNum)
```
velvia: Did you mean to call `changes` here instead of `sum`?  sum doesn't make sense right

scalastyle-config.xml
```diff
@@ -95,7 +95,7 @@
  </check>
  <check level="error" class="org.scalastyle.scalariform.NumberOfTypesChecker" enabled="true">
   <parameters>
-   <parameter name="maxTypes"><![CDATA[30]]></parameter>
+   <parameter name="maxTypes"><![CDATA[40]]></parameter>
```
velvia: It's probably better to add a scalastyle exception for one file than to globally increase this limit.

memory/src/main/scala/filodb.memory/format/vectors/LongBinaryVector.scala
```diff
@@ -227,6 +228,21 @@ object LongVectorDataReader64 extends LongVectorDataReader {
...
+  final def changes(vector: BinaryVectorPtr, start: Int, end: Int): Double = {
+    require(start >= 0 && end < length(vector), s"($start, $end) is out of bounds, length=${length(vector)}")
+    val prev = Double.NaN
+    var addr = vector + OffsetData + start * 8
+    val untilAddr = vector + OffsetData + end * 8 + 8   // one past the end
+    var changes = 0
+    while (addr < untilAddr) {
+      val nextDbl = UnsafeUtils.getLong(addr)
+      // There are many possible values of NaN.  Use a function to ignore them reliably.
+      if (!java.lang.Double.isNaN(nextDbl) && prev != nextDbl) changes += 1
```
velvia: It makes no sense to compare NaN for Long and DeltaDelta vectors, as NaN is not possible for integer values.

memory/src/main/scala/filodb.memory/format/vectors/LongBinaryVector.scala
```diff
@@ -138,6 +138,7 @@ trait LongVectorDataReader extends VectorDataReader {
    */
   def binarySearch(vector: BinaryVectorPtr, item: Long): Int
 
+  def changes(vector: BinaryVectorPtr, start: Int, end: Int): Double
```
velvia: I would change the signature of `changes` method to return Int.  There's really no reason to return Double here, and it potentially introduces other problems.

memory/src/main/scala/filodb.memory/format/vectors/DeltaDeltaVector.scala
```diff
@@ -253,6 +257,20 @@ object DeltaDeltaConstDataReader extends LongVectorDataReader {
...
+    require(start >= 0 && end < length(vector), s"($start, $end) is out of bounds, length=${length(vector)}")
+    var prev = Double.NaN
+    var changes = 0
+    var i = start
+    while (i <= end) {
+      val current = iterate(vector, i).next
```
velvia: There are several problems here:
1) It's not possible for a DeltaDelta to ever return NaN, as DeltaDeltas only ever return Long integer values.
2) What is happening when you call `Double.isNan` on `current`, which is a Long, is that Long is always converted to a Double, and isNaN will always return false.  So it's wasted.
3) Calling `iterate` returns an allocation every time, which is also a waste.

I would rewrite the changes methods in the LongBinaryVectors and elsewhere to not ever compare NaN as this is not possible, and to return Int counts always.

memory/src/main/scala/filodb.memory/format/vectors/DeltaDeltaVector.scala
```diff
@@ -253,6 +257,20 @@ object DeltaDeltaConstDataReader extends LongVectorDataReader {
       out
     }
   }
+
+  def changes(vector: BinaryVectorPtr, start: Int, end: Int): Double = {
+    require(start >= 0 && end < length(vector), s"($start, $end) is out of bounds, length=${length(vector)}")
+    var prev = Double.NaN
+    var changes = 0
```
velvia: So, this method belongs to a `ConstDeltaDeltaVector`.  This is a vector where everything is on a line slope already, so there's no point to iterate over each value.  Instead, there are two possible cases here:
a) slope == 0.  In this case, changes will always return 0, as the value is always constant - it never changes. (Or 1, I guess)
b) slope != 0.  In this case, every value is guaranteed to change - because the values are calculated.  Thus, changes should return the number of values in the range. 

The above would be much much faster and be optimal, no need to iterate on every value.

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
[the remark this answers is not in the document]
TanviBhavsar: Removed ChangesFunction

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
TanviBhavsar: Yes it should be changes. I was trying something and then forgot to change it

scalastyle-config.xml
TanviBhavsar: Added scalastyle:off and reverted the limit

memory/src/main/scala/filodb.memory/format/vectors/LongBinaryVector.scala
TanviBhavsar: Changed it to int

memory/src/main/scala/filodb.memory/format/vectors/LongBinaryVector.scala
TanviBhavsar: Removed NaN checks for LongBinaryVectors

memory/src/main/scala/filodb.memory/format/vectors/DeltaDeltaVector.scala
```diff
@@ -206,6 +206,20 @@ object DeltaDeltaDataReader extends LongVectorDataReader {
...
   }
+
+  def changes(vector: BinaryVectorPtr, start: Int, end: Int): Int = {
+    if (slope(vector) == 0)
+      return 0
+    require(start >= 0 && end < length(vector), s"($start, $end) is out of bounds, length=${length(vector)}")
+    var changes = 0
+    var i = start
+    while (i <= end) {
+      val current = iterate(vector, i).next
```
velvia: This is still a problem: calling `iterate()` with each element is really expensive

memory/src/main/scala/filodb.memory/format/vectors/DeltaDeltaVector.scala
```diff
@@ -253,6 +267,8 @@ object DeltaDeltaConstDataReader extends LongVectorDataReader {
       out
     }
   }
+
+  def changes(vector: BinaryVectorPtr, start: Int, end: Int): Int = 0
```
velvia: This is not always 0.  If the slope is 0, then it is 0; if it is nonzero, then the number of changes for `ConstDDV` is the number of elements (as it is guaranteed that every element changes at the same rate)

velvia: Feel free to make a test and test different scenarios

memory/src/main/scala/filodb.memory/format/vectors/DeltaDeltaVector.scala
TanviBhavsar: Made the changes as dicussed

memory/src/main/scala/filodb.memory/format/vectors/DeltaDeltaVector.scala
TanviBhavsar: Used length to get  changes when slope is not 0

memory/src/main/scala/filodb.memory/format/vectors/DeltaDeltaVector.scala
```diff
@@ -211,12 +211,16 @@ object DeltaDeltaDataReader extends LongVectorDataReader {
...
     require(start >= 0 && end < length(vector), s"($start, $end) is out of bounds, length=${length(vector)}")
+    val itr = iterate(vector, start)
+    var prev: Long = 0
     var changes = 0
-    var i = start
-    while (i <= end) {
-      val current = iterate(vector, i).next
-      changes += 1
-      i = i + 1
+    for {i <- start until end} {
```
velvia: Use optimized here, it will inline the function and make this a while loop underneath for better performance

memory/src/main/scala/filodb.memory/format/vectors/DeltaDeltaVector.scala
```diff
@@ -268,7 +272,12 @@ object DeltaDeltaConstDataReader extends LongVectorDataReader {
     }
   }
 
-  def changes(vector: BinaryVectorPtr, start: Int, end: Int): Int = 0
+ def changes(vector: BinaryVectorPtr, start: Int, end: Int): Int = {
+    if (slope(vector) == 0)
+      return 0
```
velvia: Style: use if/else so you don't have to do return

memory/src/main/scala/filodb.memory/format/vectors/DoubleVector.scala
```diff
@@ -247,7 +247,7 @@ object DoubleVectorDataReader64 extends DoubleVectorDataReader {
     count
   }
 
-  final def changes(vector: BinaryVectorPtr, start: Int, end: Int): Double = {
+  final def  changes(vector: BinaryVectorPtr, start: Int, end: Int): Double = {
```
velvia: Extra space

memory/src/main/scala/filodb.memory/format/vectors/DeltaDeltaVector.scala
```diff
@@ -206,6 +206,24 @@ object DeltaDeltaDataReader extends LongVectorDataReader {
     val innerIt = IntBinaryVector.simple(inner).iterate(inner, startElement)
     new DeltaDeltaIterator(innerIt, slope(vector), initValue(vector) + startElement * slope(vector).toLong)
   }
+
+  def changes(vector: BinaryVectorPtr, start: Int, end: Int): Int = {
+    if (slope(vector) == 0)
+      return 0
```
velvia: Same here use if/else

memory/src/main/scala/filodb.memory/format/vectors/DeltaDeltaVector.scala
TanviBhavsar: Added optimized

memory/src/main/scala/filodb.memory/format/vectors/DeltaDeltaVector.scala
TanviBhavsar: Incorporated all style feedback

PR #479, opened 2019-08-23
feat(util) : util to get label name for a label value from logical plan

TanviBhavsar:
**Pull Request checklist**

Function in LogicalPlanUtil to extract label name for a label value from Logical plan

### DIFF
--- a/query/src/main/scala/filodb/query/util/LogicalPlanUtil.scala
+++ b/query/src/main/scala/filodb/query/util/LogicalPlanUtil.scala
@@ -1,0 +1,56 @@
+package filodb.query.util
+
+import filodb.core.query
+import filodb.core.query.{ColumnFilter, Filter}
+import filodb.query._
+
+object LogicalPlanUtil {
+
+  private def getFilterValue(filter: Filter): Set[String] = {
+    filter match {
+      case f: query.Filter.Equals => Set(f.value.toString)
+      case f: query.Filter.And => getFilterValue(f.left).union(getFilterValue(f.right))
+      case f: query.Filter.In => f.values.map(_.toString)
+      case f: query.Filter.NotEquals => Set(f.value.toString)
+      case f: query.Filter.NotEqualsRegex => Set(f.value.toString)
+      case f: query.Filter.EqualsRegex => Set(f.value.toString)
+    }
+  }
+
+ private def getLabelFromFilters(filters: Seq[ColumnFilter], metricColumnName: String): Option[Set[String]] = {
+   val matchingFilters = filters.filter(_.column.equals(metricColumnName))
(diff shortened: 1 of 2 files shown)


query/src/main/scala/filodb/query/util/LogicalPlanUtil.scala
```diff
@@ -0,0 +1,56 @@
...
+
+ private def getLabelFromFilters(filters: Seq[ColumnFilter], metricColumnName: String): Option[Set[String]] = {
+   val matchingFilters = filters.filter(_.column.equals(metricColumnName))
+    if (matchingFilters.isEmpty)
+      None
+    else
+      Some(getFilterValue(matchingFilters.head.filter))
+  }
+
+  def getLabelNameFromLogicalPlan(logicalPlan: LogicalPlan, labelValue: String): Option[Set[String]] = {
```
tjackpaul: Method name is not correct, since it returns the values from filter and not the labelName

TanviBhavsar: Good catch! I have made the change

TanviBhavsar:
> Hi Tanvi,
> 
> So, the changes look fine for the most part, but there is a better design pattern you can use to reduce the copy and paste and having to list every possible kind of `LogicalPlan`. (You can apply this pattern too to `RoutingPlanner.scala` to reduce the amount of copy and paste and boilerplate code). The problem with this pattern is that every single place you are trying to look for something in the leaf node, such as both here and multiple places in `RoutingPlanner`, you need to keep repeating lists of LogicalPlans. It is easy to forget the lists and make mistakes.
> 
> Instead, use traits.
> 
> Create a `NonLeafLogicalPlan` like this:
> 
> ```scala
>     trait NonLeafLogicalPlan extends LogicalPlan {
>       def children: Seq[LogicalPlan]
>     }
> ```
> 
> Now, have every LogicalPlan that is not a leaf extend this trait and define this method for each of them.
> 
> Then, create a generic method that can walk through any LogicalPlan and find the nested children, until you get to the leaves. Something like a `findLeafLogicalPlans`.
> 
> All of the methods in `RoutingPlanner` as well as `getLabelValue` here can be based on that `findLeafLogicalPlans` or similar methods. This means all of the copy and paste and boilerplate code can now be gone, thanks to a generic way to traverse the tree which is not there before.
> 
> I'm asking for this now because we will want ways of traversing the tree for other things, such as optimizers, and it is generally really useful.
> 
> You can use the same technique for `query.Filter` as well.... define a generic valueStrings method for each filter, then you can just call that. Again, reduces the boilerplate a lot.
> 
> Thanks

I have added NonLeafLogicalPlan and made the required changes.

PR #483, opened 2019-08-29
feat(kafka, core): Ensure that ingestion time always appears monotonic.

broneill:
feat(kafka, core): Ensure that ingestion time always appears monotonic. Also report an ingestion lag metric.

**Pull Request checklist**

**Current behavior :**
Ingestion time can appear to go backwards, due to clock skew.

**New behavior :**
Ingestion time is selected as the maximum received from so far.

### DIFF
--- a/core/src/main/scala/filodb.core/binaryrecord2/RecordContainer.scala
+++ b/core/src/main/scala/filodb.core/binaryrecord2/RecordContainer.scala
@@ -47,12 +47,30 @@ final class RecordContainer(val base: Any, val offset: Long, maxLength: Int,
   }
 
   private[binaryrecord2] def updateTimestamp(): Unit = {
-    UnsafeUtils.setLong(base, offset + 8, System.currentTimeMillis())
+    setTimestamp(System.currentTimeMillis())
   }
 
   @inline final def version: Int = (UnsafeUtils.getInt(base, offset + 4) >> 24) & 0x00ff
   @inline final def isCurrentVersion: Boolean = version == Version
   @inline final def timestamp: Long = UnsafeUtils.getLong(base, offset + 8)
+  @inline private def setTimestamp(ts: Long): Unit = UnsafeUtils.setLong(base, offset + 8, ts)
+
+  /**
+   * Adjusts the timestamp to be monotonic by updating it to match the given timestamp, but
+   * only when it's higher. If the current timestamp is greater than or equal to the given
+   * timestamp, then the current timestamp is returned unmodified. The caller should remember
+   * the highest timestamp and pass it to the next container it receives.
+   *
+   * @return the current timestamp of this container
+   */
+  final def monotonicTimestamp(highestTimestamp: Long): Long = {
+    var ts = timestamp
+    if (highestTimestamp > ts) {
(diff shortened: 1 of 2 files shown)


core/src/main/scala/filodb.core/binaryrecord2/RecordContainer.scala
```diff
@@ -47,12 +47,30 @@ final class RecordContainer(val base: Any, val offset: Long, maxLength: Int,
...
+  /**
+   * Adjusts the timestamp to be monotonic by updating it to match the given timestamp, but
+   * only when it's higher. If the current timestamp is greater than or equal to the given
+   * timestamp, then the current timestamp is returned unmodified. The caller should remember
+   * the highest timestamp and pass it to the next container it receives.
+   *
+   * @return the current timestamp of this container
+   */
+  final def monotonicTimestamp(highestTimestamp: Long): Long = {
+    var ts = timestamp
```
velvia: No need to use a var here.  How about just:

    if (highestTimestamp > timestamp) setTimestamp(highestTimestamp)

kafka/src/main/scala/filodb/kafka/KafkaIngestionStream.scala
```diff
@@ -48,7 +54,20 @@ class KafkaIngestionStream(config: Config,
...
+
+      // We want to maintain the illusion that ingestion time never moves backwards. Because
+      // the record containers come from different sources, they might have clocks which are
+      // slightly skewed. Always choose the highest one, for simplicity and consistency.
+      highestTimestamp = rc.monotonicTimestamp(highestTimestamp)
+
+      // Note that if the lag is negative, report it as such. This makes it possible to observe
+      // cases where this server has a clock set to the past, or a source has a clock set to
+      // the future.
+      val lag = System.currentTimeMillis() - highestTimestamp
```
velvia: What is the latency of each `currentTimeMillis` call, just curious?

velvia: Also, we probably don't need to set the lag with every container.  Once a second is enough - consider moving this to the TimeSeriesShard.

Actually the more I think about it the more this logic should be in the TimeSeriesShard, not here in Kafka, as it is universal for all ingestion streams, not just Kafka.

broneill: Getting the current time is relatively cheap. It doesn't involve full system call overhead. Considering that the check is performed on the container, it's not called that much overall.

core/src/main/scala/filodb.core/binaryrecord2/RecordContainer.scala
broneill: I wanted to return a value from the function.

kafka/src/main/scala/filodb/kafka/KafkaIngestionStream.scala
broneill: Ideally, the time check should be global, and performing a per-shard adjustment means that the shards can be out sync with each other.

PR #489, opened 2019-09-06
fix(memory): Protect blocks from being marked as reclaimable multiple times.

broneill:
**Pull Request checklist**

**Current behavior :**
A background task can call BlockManager.markBucketedBlocksReclaimable, which can then steal blocks which are owned by BlockMemFactory instances. Because the BlockMemFactory isn't notified, it thinks it can still access the current block and mark it reclaimable again later. This makes it possible for the block to be shared, leading to corruption.

**New behavior :**
If there is a block owner assigned, it's consulted first. It can reject the attempt to mark its current block as reclaimable, if it's still being written to. If the attempt is accepted, then all blocks tracked by mem factory are marked as reclaimable, and all references to them are discarded.

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/DemandPagedChunkStore.scala
+++ b/core/src/main/scala/filodb.core/memstore/DemandPagedChunkStore.scala
@@ -73,21 +73,23 @@ extends RawToPartitionMaker with StrictLogging {
       tsShard.shardStats.numChunksPagedIn.increment(rawPartition.chunkSets.size)
       // One chunkset at a time, load them into offheap and populate the partition
       rawPartition.chunkSets.foreach { case RawChunkSet(infoBytes, rawVectors) =>
-        val memFactory = getMemFactory(timeBucketForChunkSet(infoBytes))
-        val chunkID = ChunkSetInfo.getChunkID(infoBytes)
+        if (!rawVectors.isEmpty) {
+          val memFactory = getMemFactory(timeBucketForChunkSet(infoBytes))
+          val chunkID = ChunkSetInfo.getChunkID(infoBytes)
 
-        memFactory.startMetaSpan()
-        val chunkPtrs = copyToOffHeap(rawVectors, memFactory)
-        val metaAddr = memFactory.endMetaSpan(writeMeta(_, tsPart.partID, infoBytes, chunkPtrs),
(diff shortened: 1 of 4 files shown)


memory/src/main/scala/filodb.memory/MemFactory.scala
```diff
@@ -212,18 +219,51 @@ class BlockMemFactory(blockStore: BlockManager,
...
+
+  private def requestBlock() = blockStore.requestBlock(bucketTime, optionSelf).get
 
   // tracks blocks that should share metadata
   private val metadataSpan: ListBuffer[Block] = ListBuffer[Block]()
 
+  // Last time this factory was used for allocation.
+  private var lastUsedNanos = now
+
+  private def now: Long = System.nanoTime()
```
velvia: How fast is nanoTime call?  
The other concern is that nanoTime can potentially wrap around, can't it?

broneill: It's plenty fast. Consider also that the block allocation is relatively infrequent, so not something to be concerned with. The nanotime value can wrap around, which is why modulo comparison (subtraction) is required.

core/src/main/scala/filodb.core/memstore/DemandPagedChunkStore.scala
```diff
@@ -73,21 +73,23 @@ extends RawToPartitionMaker with StrictLogging {
       tsShard.shardStats.numChunksPagedIn.increment(rawPartition.chunkSets.size)
       // One chunkset at a time, load them into offheap and populate the partition
       rawPartition.chunkSets.foreach { case RawChunkSet(infoBytes, rawVectors) =>
-        val memFactory = getMemFactory(timeBucketForChunkSet(infoBytes))
-        val chunkID = ChunkSetInfo.getChunkID(infoBytes)
+        if (!rawVectors.isEmpty) {
```
velvia: Please add a comment detailing the bug about metadata allocation error that might possibly happen that this fixes

broneill:
Also adding code to "Return the last metaAddr from endMetaSpan which actually has room" doesn't guarantee that the returned address is non-zero. The first might have been full as well. The newly added comment explains my motivation for just a simple fix.

PR #492, opened 2019-09-11
feat(downsampler): Spark job to read and downsample raw data from cassandra

vishramachandran:
**Pull Request checklist**

First draft of downsampling via spark job. Note that this is an initial commit for testing purposes only. Need to perform more testing E2E beyond local cluster, at larger scale.

TODO (will be done either in this, or later PR):
* Metrics
* Unit tests where possible
* Checkpointing and retries 

Note:
This is not a “framework” yet for performing more out of the cassandra data scans.
My goal is to get the functional features out first before adding any other capability.


downsampler/src/main/scala/filodb.downsampler/DownsamplerMain.scala
```diff
@@ -0,0 +1,112 @@
...
+import monix.reactive.Observable
+import org.apache.spark.sql.SparkSession
+
+import filodb.core.ErrorResponse
+import filodb.core.binaryrecord2.RecordSchema
+import filodb.core.memstore.{PagedReadablePartition, TimeSeriesPartition}
+import filodb.core.store.ChunkSet
+import filodb.memory.format.UnsafeUtils
+
+object DownsamplerMain extends App with StrictLogging {
```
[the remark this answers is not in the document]
vishramachandran: Yes - I believe this driver would run on the spark master node.

downsampler/src/main/scala/filodb.downsampler/PartitionDownsampler.scala
```diff
@@ -0,0 +1,106 @@
...
+    *
+    * @param nativeMemManager downsampled chunks are stored into offheap memory
+    *                         allocated by this memory manager
+    * @param bufferPool buffers for the downsampled partitions are procured from this pool
+    *
+    * @return a TimeSeriesPartition for each resolution to be downsampled.
+    *         NOTE THAT THE PARITIONS NEED TO BE FREED/SHUT DOWN ONCE CHUNKS ARE EXTRACTED FROM THEM
+    */
+  // scalastyle:off method.length parameter.number
+  def downsample(shardStats: TimeSeriesShardStats,
```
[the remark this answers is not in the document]
vishramachandran: Yes - lots of params indeed. This will be called a lot of times, just avoided the case class instance to save on the allocation. The number of params problem would stay either way.

memory/src/main/scala/filodb.memory/Block.scala
```diff
@@ -166,7 +166,7 @@ class Block(val address: Long, val capacity: Long, val reclaimListener: ReclaimL
       UnsafeUtils.setShort(UnsafeUtils.ZeroPointer, metaAddr, metaSize)
       metaAddr + 2
     } else {
-      0
+      throw new IllegalStateException(s"Could not allocate metadata of size $metaSize. Remaining=${remaining()}")
```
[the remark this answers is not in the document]
vishramachandran: Fixed

downsampler/src/main/scala/filodb.downsampler/DownsamplerMain.scala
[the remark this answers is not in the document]
vishramachandran: This is a standard spark application. 
Let me think of ways to unit test.

memory/src/main/scala/filodb.memory/Block.scala
```diff
@@ -159,15 +159,17 @@ class Block(val address: Long, val capacity: Long, val reclaimListener: ReclaimL
    * @return the Long address of the metadata space.  The metaSize is written to the location 2 bytes before this.
```
broneill: This comment is now wrong,

memory/src/main/scala/filodb.memory/Block.scala
```diff
@@ -157,7 +157,7 @@ class Block(val address: Long, val capacity: Long, val reclaimListener: ReclaimL
    * Allocates metaSize bytes for metadata storage.
    * @param metaSize the number of bytes to use for metadata storage.
    * @return the Long address of the metadata space.  The metaSize is written to the location 2 bytes before this.
-   *         If there is no capacity then 0 (null) is returned.
+   *         If there is no capacity then 0OutOfOffheapMemoryException is thrown.
```
broneill: Almost! There's a zero still!

downsampler/src/main/scala/filodb.downsampler/DownsamplerMain.scala
```diff
@@ -0,0 +1,108 @@
...
+    */
+  private def persistDownsampledChunks(downsampledChunksToPersist: MMap[FiniteDuration, Iterator[ChunkSet]]): Unit = {
+    // write all chunks to cassandra
+    val writeFut = downsampledChunksToPersist.map { case (res, chunks) =>
+      PerSparkExecutorState.cassandraColStore.write(downsampleDatasets(res).ref,
+        Observable.fromIterator(chunks), ttlByResolution(res))
+    }
+
+    writeFut.foreach { fut =>
+      val response = Await.result(fut, cassWriteTimeout)
```
velvia: Do we want writes to be serial?  There aren't many resolutions, it seems fine to write them all at once.

Also, since this is a bulk write, I think it's better to use the Spark Cassandra Connector to do it.

downsampler/src/main/scala/filodb.downsampler/FiloCassRelation.scala
```diff
@@ -0,0 +1,57 @@
...
+                                                    userStartTime, userEndTime, batchSize)
+                     .toIterator()
+                     .map { p => p.map(PartDataRow(_)) }
+      }
+  }
+}
+
+case class PartDataRow(partData: RawPartData) extends Row {
+  override def length: Int = 1
+  override def get(i: Int): Any = if (i ==0) partData else throw new IllegalArgumentException(s"invalid column $i")
```
velvia: space after `==`

downsampler/src/main/scala/filodb.downsampler/FiloCassRelation.scala
```diff
@@ -0,0 +1,57 @@
...
+  override def copy(): Row = PartDataRow(partData)
+}
+
+object PartDataRow {
+  val schema = StructType(
```
velvia: The schema needs to match the Row. 

Actually, I realized the problem here is that you should not be creating a 'Relation' at all.  Spark will have problems if the Row does not match the schema.  The reason why it works is that we are not using Spark SQL at all, so we aren't using relational style tables but just the RDD directly.  In that case the Relation is not needed at all, we should just get rid of it.

downsampler/src/main/scala/filodb.downsampler/PartitionDownsampler.scala
```diff
@@ -0,0 +1,106 @@
...
+        shardStats,
+        nativeMemManager,
+        1)
+      logger.trace(s"Creating new part=${part.hashCode}")
+      res -> part
+    }.toMap
+
+    val chunksets = partToDownsample.infos(AllChunkScan)
+    val timestampCol = 0
+    // TODO create a rowReader that will not box the vals below
```
velvia: I think this needs to be fixed before scale testing.  Using an array of Any will box everything and create huge GC churn.

It's fairly simple to use `RecordBuilder` to achieve this.  Just call reset() every time so that there is only one record ever written.

downsampler/src/main/scala/filodb.downsampler/PartitionDownsampler.scala
```diff
@@ -0,0 +1,106 @@
...
+      val tsReader = partToDownsample.chunkReader(timestampCol, vecPtr).asLongReader
+
+      // for each downsample resolution
+      resToPartition.foreach { case (resolution, part) =>
+        val resMillis = resolution.toMillis
+        var pStart = ((startTime - 1) / resMillis) * resMillis + 1 // inclusive startTime for downsample period
+        var pEnd = pStart + resMillis // end is inclusive
+        // for each downsample period
+        while (pStart <= endTime) {
+          if (pEnd >= userStartTime && pEnd <= userEndTime) {
```
velvia: Again, I don't see any reason why user time should even be here.

downsampler/src/main/scala/filodb.downsampler/PartitionDownsampler.scala
```diff
@@ -0,0 +1,106 @@
...
+            for {col <- downsamplers.indices optimized} {
+              val downsampler = downsamplers(col)
+              downsampler match {
+                case d: TimeChunkDownsampler =>
+                  downsampleRow(col) = d.downsampleChunk(partToDownsample, chunkset, startRowNum, endRowNum)
+                case d: DoubleChunkDownsampler =>
+                  downsampleRow(col) = d.downsampleChunk(partToDownsample, chunkset, startRowNum, endRowNum)
+                case h: HistChunkDownsampler =>
+                  downsampleRow(col) = h.downsampleChunk(partToDownsample, chunkset, startRowNum, endRowNum)
+                    .serialize()
```
velvia: Why does only the histogram downsampler have `serialize()`, what does this do?

downsampler/src/main/scala/filodb.downsampler/PartitionDownsampler.scala
```diff
@@ -0,0 +1,106 @@
...
+      val chunkset = chunksets.nextInfo
+      val startTime = chunkset.startTime
+      val endTime = chunkset.endTime
+      val vecPtr = chunkset.vectorPtr(timestampCol)
+      val tsReader = partToDownsample.chunkReader(timestampCol, vecPtr).asLongReader
+
+      // for each downsample resolution
+      resToPartition.foreach { case (resolution, part) =>
+        val resMillis = resolution.toMillis
+        var pStart = ((startTime - 1) / resMillis) * resMillis + 1 // inclusive startTime for downsample period
```
velvia: It seems the +1 makes it exclusive, but comment says inclusive

downsampler/src/main/scala/filodb.downsampler/PartitionDownsampler.scala
velvia: Oh never mind.  I'm not sure if this will work correctly, as the histogram ingestion is specifically tested and designed to work from BinaryRecords mostly.

downsampler/src/main/scala/filodb.downsampler/PerSparkExecutorState.scala
```diff
@@ -0,0 +1,73 @@
...
+                       "userTimeStart" -> userTimeStart.toString,
+                       "userTimeEnd" -> userTimeEnd.toString)
+
+  val schemas = Schemas.fromConfig(filodbConfig).get
+
+  val rawSchemas = rawSchemaNames.map { s => schemas.schemas(s)}
+  /**
+    * Downsample Schemas
+    */
+  val dsSchemas = rawSchemaNames.map { s => schemas.schemas(s).downsample.get}
```
velvia: This will fail, because not all schemas have downsamplers.

downsampler/src/main/scala/filodb.downsampler/PerSparkExecutorState.scala
```diff
@@ -0,0 +1,73 @@
...
+
+  /**
+    * Chunk Downsamplers by Raw Schema Id
+    */
+  val chunkDownsamplersByRawSchemaId = rawSchemas.map { s => s.schemaHash -> s.data.downsamplers }.toMap
```
velvia: I suggest using a `debox.Map` here, or some other map that does not box integers.  Otherwise every map lookup will involve boxing.

downsampler/src/main/scala/filodb.downsampler/PerSparkExecutorState.scala
```diff
@@ -0,0 +1,73 @@
...
+
+  /**
+    * Raw dataset from which we downsample data
+    */
+  val rawDataset = Dataset(rawDatasetName, rawSchemas.head)
```
velvia: We should avoid using Dataset if possible.  The first raw schema is probably not the only one we want to deal with.  Use of `Dataset` encourages the use of only one schema.

cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
```diff
@@ -148,6 +148,30 @@ extends ColumnStore with CassandraChunkSource with StrictLogging {
     }
   }
 
+  def getChunksByIngestionTimeRange(dataset: DatasetRef,
+                                    splits: Iterator[ScanSplit],
+                                    ingestionTimeStart: Long,
+                                    ingestionTimeEnd: Long,
+                                    userTimeStart: Long,
```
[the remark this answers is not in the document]
vishramachandran: This is a first pass solution until we resolve chunk alignment. I am going to have an offline conversation with you on this.

core/src/main/scala/filodb.core/memstore/PagedReadablePartition.scala
```diff
@@ -0,0 +1,102 @@
...
+                             override val partID: Int,
+                             pd: RawPartData,
+                             mem: NativeMemoryManager) extends ReadablePartition {
+
+  import PagedReadablePartition._
+  val chunkInfos = ArrayBuffer[ChunkSetInfo]()
+
+  var partitionKey: BinaryRegion.NativePointer = _
+
+  loadToOffHeap()
```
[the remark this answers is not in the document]
vishramachandran: This causes a leak in abstraction. It requires that the caller know that the load needs to be done, and requires two steps. The loadToOffHeap is just part of object construction and is abstracted away. I personally prefer this.  Not clear how the external caller invoking `loadToOffHeap` would make things immutable.

core/src/main/scala/filodb.core/memstore/PagedReadablePartition.scala
```diff
@@ -0,0 +1,102 @@
...
+  * are read.
+  */
+class PagedReadablePartition(override val schema: Schema,
+                             override val shard: Int,
+                             override val partID: Int,
+                             pd: RawPartData,
+                             mem: NativeMemoryManager) extends ReadablePartition {
+
+  import PagedReadablePartition._
+  val chunkInfos = ArrayBuffer[ChunkSetInfo]()
```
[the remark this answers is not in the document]
vishramachandran: 👍

core/src/main/scala/filodb.core/store/ReadablePartition.scala
```diff
@@ -42,6 +43,8 @@ trait ReadablePartition extends FiloPartition {
   // Returns string representation of partition key
   def stringPartition: String = schema.partKeySchema.stringify(partKeyBase, partKeyOffset)
 
+  def partitionKey: BinaryRegion.NativePointer
```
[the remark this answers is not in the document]
vishramachandran: `partitionKey` is defined in super class `ReadableParititon`.

This is one of the reasons invoking loadToOffHeap externally will not make this class immutable.

downsampler/src/main/scala/filodb.downsampler/DownsamplerSettings.scala
```diff
@@ -0,0 +1,44 @@
...
+  val batchSize = downsamplerConfig.getInt("num-partitions-per-cass-write")
+  val blockMemorySize = downsamplerConfig.getMemorySize("off-heap-block-memory-size").toBytes
+  val nativeMemManagerSize = downsamplerConfig.getMemorySize("off-heap-native-memory-size").toBytes
+  val cassWriteTimeout = 10.minutes
+
+  val now = downsamplerConfig.as[Option[Long]]("downsampler-execution-time").getOrElse(System.currentTimeMillis())
+  val chunkDuration = downsampleStoreConfig.flushInterval.toMillis
+  val userTimeEnd = (now / chunkDuration) * chunkDuration
+
+  val userTimeStart = userTimeEnd - chunkDuration
```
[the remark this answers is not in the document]
vishramachandran: Let us discuss offline and I can provide an explanation

downsampler/src/main/scala/filodb.downsampler/DownsamplerMain.scala
```diff
@@ -0,0 +1,108 @@
...
+    downsampleBatch(rawPartsBatch)
+  }
+  cassandraColStore.shutdown()
+  spark.sparkContext.stop()
+
+  // TODO migrate index entries
+
+
+  /**
+    * Downsample batch of raw partitions, and store downsampled chunks to cassandra
```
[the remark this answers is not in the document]
vishramachandran: Will do.

downsampler/src/main/scala/filodb.downsampler/DownsamplerMain.scala
```diff
@@ -0,0 +1,108 @@
...
+    downsampleResolutions.foreach { res =>
+      downsampledChunksToPersist(res) = Iterator.empty
+    }
+    val rawPartsToFree = ArrayBuffer[PagedReadablePartition]()
+    val downsampledPartsPartsToFree = ArrayBuffer[TimeSeriesPartition]()
+
+    rawPartsBatch.foreach { rawPart =>
+
+      val rawSchemaId = RecordSchema.schemaID(rawPart.partData.partitionKey, UnsafeUtils.arayOffset)
+      val rawPartSchema = schemas(rawSchemaId)
```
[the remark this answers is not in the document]
vishramachandran: `schemas` is within PerSparkExecutorState. It is constructed from config.

downsampler/src/main/scala/filodb.downsampler/DownsamplerMain.scala
```diff
@@ -0,0 +1,108 @@
...
+      val downsampleSchema = rawPartSchema.downsample.get
```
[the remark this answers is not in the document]
vishramachandran: 👍

downsampler/src/main/scala/filodb.downsampler/DownsamplerMain.scala
```diff
@@ -0,0 +1,108 @@
...
+      logger.debug(s"Downsampling partition ${rawPartSchema.partKeySchema.stringify(rawPart.partData.partitionKey)} ")
+
+      val rawReadablePart = new PagedReadablePartition(rawPartSchema, 0, 0,
+        rawPart.partData, memoryManager)
```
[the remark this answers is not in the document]
vishramachandran: The run should fail, and someone needs to step in and change the memory configuration or reduce size of the batch.
Any other suggestion ?

That said, the memory usage is determined by the size of the batch. Out of Off Heap  should happen only if there are way too many samples within a given time range (in many many partitions of the batch), or if the partKey is so long that it does not fit in memory.

If we have tuned this correctly and there isn't any BAD data, the chances of this happening are very slim. If it happens, on-call intervention is warranted.

downsampler/src/main/scala/filodb.downsampler/DownsamplerMain.scala
```diff
@@ -0,0 +1,108 @@
...
+      val downsampledParts = PartitionDownsampler.downsample(shardStats,
+        downsampleResolutions, downsampleSchema, chunkDownsamplersByRawSchemaId(rawSchemaId),
+        rawReadablePart, blockFactory, bufferPoolByRawSchemaId(rawSchemaId),
+        memoryManager, userTimeStart, // use userTime as ingestionTime for downsampled data
```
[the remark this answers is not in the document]
vishramachandran: Please refer to offline discussion, and documentation that I will add.

downsampler/src/main/scala/filodb.downsampler/DownsamplerMain.scala
vishramachandran: The parallelism is achieved by the cassandra splits.

I'll certainly look into the bulk-writer here. We can improve iteratively.

downsampler/src/main/scala/filodb.downsampler/FiloCassRelation.scala
vishramachandran: 👍

downsampler/src/main/scala/filodb.downsampler/PartitionDownsampler.scala
vishramachandran: Explained in offline discussion

downsampler/src/main/scala/filodb.downsampler/PartitionDownsampler.scala
vishramachandran: I just brought this over, as is, from the old downsampler. I believe you wrote it. Does it need a fix ?

downsampler/src/main/scala/filodb.downsampler/PerSparkExecutorState.scala
vishramachandran: `rawSchemaNames` (obtained by configuration) are the schemas that need to be downsampled by this downsampler.

downsampler/src/main/scala/filodb.downsampler/PerSparkExecutorState.scala
vishramachandran: 👍

downsampler/src/main/scala/filodb.downsampler/PerSparkExecutorState.scala
vishramachandran: 👍 Removed references to Dataset and used DatasetRef

core/src/main/scala/filodb.core/memstore/PagedReadablePartition.scala
```diff
@@ -0,0 +1,102 @@
...
+
+  override def hasChunks(method: ChunkScanMethod): Boolean = {
+    method == AllChunkScan && chunkInfos.isEmpty
+  }
+
+  override def hasChunksAt(id: ChunkID): Boolean = chunkInfos.exists(_.id == id)
+
+  override def earliestTime: Long = ???
+
+  override def partKeyBase: Array[Byte] = pd.partitionKey
```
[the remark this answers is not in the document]
vishramachandran: 👍

downsampler/src/main/scala/filodb.downsampler/PartitionDownsampler.scala
vishramachandran: This is old code as is.

vishramachandran: Added more docs.

core/src/main/scala/filodb.core/memstore/PagedReadablePartition.scala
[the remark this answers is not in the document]
velvia: You don't need to have the caller do it.  Just move the initialization code to an `apply` method in `PagedReadablePartition`.  Then the memory address becomes immutable, which is important for protecting against bugs like SEGV.

downsampler/src/main/scala/filodb.downsampler/DownsamplerMain.scala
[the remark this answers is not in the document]
velvia: One could cut off the batch early, clear the memory and start another batch.....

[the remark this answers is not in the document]
velvia: At least adda  TODO or something

cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
```diff
@@ -148,6 +148,34 @@ extends ColumnStore with CassandraChunkSource with StrictLogging {
...
+        indexTable.scanPartKeysByIngestionTime(split.tokens, ingestionTimeStart, ingestionTimeEnd)
+      case split => throw new UnsupportedOperationException(s"Unknown split type $split seen")
+    }
+
+    import filodb.core.Iterators._
+
+    val chunksTable = getOrCreateChunkTable(datasetRef)
+    partKeys.bufferTimedAndCounted(1.second, batchSize).map { parts =>
+      logger.debug(s"Querying cassandra for chunks from ${parts.size} partitions")
+      chunksTable.readRawPartitionRangeBB(parts, userTimeStart, userTimeEnd).toIterator().toSeq
```
velvia: This is going to serially read one partition at a time.  I know there is parallelism in Spark job, but I think this will be too slow in practice.   At least, let's add a TODO here and revisit.  I think you will actually need to group partitions to read in bulk, which is much faster.

core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
```diff
@@ -57,7 +57,7 @@ final case class InfoAppenders(info: ChunkSetInfo, appenders: TimeSeriesPartitio
  */
 class TimeSeriesPartition(val partID: Int,
                           val schema: Schema,
-                          partitionKey: BinaryRegion.NativePointer,
+                          override val partitionKey: BinaryRegion.NativePointer,
```
velvia: I still don't understand this part.  `partitionKey` was not a part of ReadablePartiton previously, why is it needed now?  It is duplicate, because that's exactly what `partKeyOffset` is.   We should not add an additional member to TSPartition, when it is redundant.  It adds extra memory usage for each TSPartition object.

core/src/main/scala/filodb.core/store/ReadablePartition.scala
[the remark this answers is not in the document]
velvia: Um, no... this PR added `partitionKey`, see above.  What I'm asserting is that this is not needed.

Instead do this:

    def partitionKey: NativePointer = partKeyOffset

PR #494, opened 2019-09-13
feat(core): Flush chunks on hourly boundaries.

broneill:
**Pull Request checklist**

**Current behavior :**
Chunks are flushed by a background task.

**New behavior :**
Chunks are flushed when ingestion time crosses an hourly boundary. The existing background task remains unchanged for now, but perhaps it can run less frequently.


core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -429,6 +440,19 @@ class TimeSeriesShard(val ref: DatasetRef,
             case e: Exception                   => logger.error(s"Unexpected ingestion err", e); disableAddPartitions()
           }
         } else {
+          val oldTimestamp = groupTimestamps(group)
+          val newTimestamp = Math.max(oldTimestamp, ingestionTime) // monotonic clock
```
velvia: This check should not happen within `onNext`, as the ingestion time does not change for each sample.  This makes the check much more expensive than it should be ... just check it after each call to ingest().

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -429,6 +440,19 @@ class TimeSeriesShard(val ref: DatasetRef,
...
+          groupTimestamps(group) = newTimestamp
+          if (newTimestamp > oldTimestamp && oldTimestamp != Long.MinValue
+              && (oldTimestamp / flushBoundaryMillis) != (newTimestamp / flushBoundaryMillis)) {
+            flushSched match {
+              case None => // none provided, so don't flush
+              case Some(sched) => {
+                // Flush out the group first such that the record is for a new hour.
+                createFlushTask(prepareFlushGroup(group)).executeOn(sched)
```
velvia: This logic does things on a different thread, and should not be here - it pollutes the IngestConsumer.

core/src/main/scala/filodb.core/memstore/TimeSeriesMemStore.scala
```diff
@@ -123,17 +122,12 @@ extends MemStore with StrictLogging {
                    cancelTask: Task[Unit]): CancelableFuture[Unit] = {
     val shard = getShardE(dataset, shardNum)
     combinedStream.map {
-                    case d: SomeData =>       shard.ingest(d)
+                    case d: SomeData =>       shard.ingest(d.records, d.offset, Some(flushSched))
```
velvia: I don't think ingest() should involve itself in triggering the flush at all. 

I think a much cleaner separation would be like this:
  1) Remove the `FlushCommand`... that won't be necessary any more
  2) Code could be like this:

     case d: SomeData => shard.ingest(d)
                                           if (shard.checkTimeToFlush()) {
                                              Some(shard.prepareFlushGroup(group))
                                           } else None

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -127,6 +127,11 @@ object TimeSeriesShard {
   val indexTimeBucketTtlPaddingSeconds = 24.hours.toSeconds.toInt
   val indexTimeBucketSegmentSize = 1024 * 1024 // 1MB
 
+  // Flush groups when ingestion time is observed to cross an hourly boundary. This simplifies
+  // disaster recovery -- chunks can be copied without concern that they may overlap in time.
+  // TODO make configurable if necessary
+  val flushBoundaryMillis = 1.hour.toMillis
```
velvia: See `StoreConfig`... there is a configuration already for the flush interval, we should just use that.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
broneill: It looks like the original task was scheduled on the same thread. I'll check again.

core/src/main/scala/filodb.core/memstore/TimeSeriesMemStore.scala
broneill: Yes. I left the FlushCommand in for now just to simplify the review.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
broneill: ok

core/src/main/scala/filodb.core/memstore/TimeSeriesMemStore.scala
broneill: The reason why ingest triggers the flush is that it cannot absorb any more data until the flush has been prepared. If some stateful check must be invoked on the shard after ingest has been called, it's too late.

velvia: This can be accomplished using the example I gave above.  It's not too late, as long as it happens before the next data is ingested.

broneill: Another reason why I decided to have the shard trigger the flush is that it does so as it's iterating over the record container. If TimeSeriesMemStore needed to trigger the flush, then the record container would need to be iterated over twice (or more).

velvia: We don't need to iterate over the record container to trigger the flush - there is only one ingestion time per recordContainer.  The trigger can be done before or after.

broneill: True dat.

broneill: ...however, there's the monotonic clock check, which is adjusted per group. A monotonic clock per record container requires switching to 1 flush group, which I'm hoping to avoid.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
broneill: Ingestion time can vary by group, and the monotonic adjustment shouldn't span groups.

core/src/main/scala/filodb.core/memstore/TimeSeriesMemStore.scala
velvia: > there's the monotonic clock check, which is adjusted per group

How does this work again?  The per group check?

broneill: The per group check is performed as sampled are extracted from the record container. which is heterogeneous.

broneill:
I've made changes such that the flushing is distributed across the hour (whatever flush interval is) by the number of flush groups. Note that the flush stream is completely gone now.

I modified the tests that were failing, and I removed one that's now invalid (it tested the removed flush stream code). The tests might still need a bit more work, but they do demonstrate that flushing by ingestion time is working as expected.

velvia:
@broneill  just to let you know - I haven't forgotten about this, just looked through all of the new changes.  I need to absorb the changes and the design a bit more.... specifically, why the change from initiating the flush tasks from within the ingest method itself from the original method.   I'm afraid that is still a very big change and would like to keep the original processing the way it is if possible.  But let me think about this some more.

core/src/main/scala/filodb.core/memstore/TimeSeriesMemStore.scala
```diff
@@ -101,49 +100,24 @@ extends MemStore with StrictLogging {
...
-                                                Some(FlushGroup(shard.shardNum, group, shard.latestOffset,
-                                                                flushTimeBucket))
-                    case a: Any => throw new IllegalStateException(s"Unexpected DataOrCommand $a")
-                  }.collect { case Some(flushGroup) => flushGroup }
-                  .mapAsync(numParallelFlushes) { f => shard.createFlushTask(f).executeOn(flushSched).asyncBoundary }
-                           // asyncBoundary so subsequent computations in pipeline happen in default threadpool
-                  .completedL
-                  .doOnCancel(cancelTask)
-                  .runAsync(shard.ingestSched)
+    stream.map { case d: SomeData => shard.ingest(d.records, d.offset, Some(flushSched)) }
```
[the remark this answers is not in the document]
broneill: I think this is vestigial and can be removed.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -462,6 +479,21 @@ class TimeSeriesShard(val ref: DatasetRef,
             case e: Exception                   => logger.error(s"Unexpected ingestion err", e); disableAddPartitions()
           }
         } else {
+          val oldTimestamp = groupTimestamps(group)
+          val newTimestamp = Math.max(oldTimestamp, ingestionTime) // monotonic clock
+          groupTimestamps(group) = newTimestamp
+          if (newTimestamp > oldTimestamp && oldTimestamp != Long.MinValue) {
+            val tsAdjust = flushOffsetMillis * group
+            if ((oldTimestamp - tsAdjust) / flushBoundaryMillis != (newTimestamp - tsAdjust) / flushBoundaryMillis) {
+              flushSched match {
```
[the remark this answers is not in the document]
broneill: Sounds right to me.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -462,6 +479,21 @@ class TimeSeriesShard(val ref: DatasetRef,
...
+                case None => // none provided, so don't flush
+                case Some(sched) => {
+                  // Flush out the group first such that the record is for a new hour.
+                  createFlushTask(prepareFlushGroup(group)).executeOn(sched)
```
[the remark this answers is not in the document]
broneill: How is this done? I'm not sure what the correct pattern to apply is.

core/src/main/scala/filodb.core/memstore/TimeSeriesMemStore.scala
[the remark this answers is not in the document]
broneill: Actually, it's still required. Some methods like TimeSeriesMemstore.ingest don't have access to a flushSched.

velvia:
Just chatted with @broneill offline, but adding it here for the record.
He's going to explore an approach where we can check the flush trigger outside of `IngestConsumer`, that way it won't need to be done on a per-sample basis, and allow us to keep the previous `mapAsync` approach of initiating the flush tasks, which keeps the ability to limit the number of simultaneous flush tasks going at the same time (for write load management).

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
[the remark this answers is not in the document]
velvia: Ah, I see that Vish called out the same issue here.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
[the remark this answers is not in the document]
broneill: Actually, I don't think these extra calculations can be omitted. I need to make a precise timestamp check before rounding down.

PR #495, opened 2019-09-16
feat(query):  date time functions support

TanviBhavsar:
Support for Prometheus date time functions:  hour, minute, month, year, days_in_month, day_of_month and day_of_week


broneill:
I think some of the boilerplate code and object construction can be reduced by using LocalDateTime directly: LocalDateTime.ofEpochSecond(value.toLong, 0, ZoneId.UTC).getYear(). 

Some objects might be allocated, some redundant calculations might be performed, but they'll get inllined, escape analysis will perform scalar replacement, and then the extra steps will be removed.

Note that the Joda-time library was better suited for raw field extraction without allocations, but internally it still performed redundant steps and it had magical caches. In the end, it wasn't any better than what the Java time library does today.

TanviBhavsar:
> Looks fine to me, but I would ask @broneill to have a look, he's more familiar with the JDK time functions.
> 
> Could we avoid creating an `Instant` with each function as that is an extra allocation? Better yet if there is a static way to calculate without allocating anything.
> 
> Did we implement time() function? That seems necessary to use these functions.

@velvia time() is not implemented yet. As discussed these functions can be used without time() also.

TanviBhavsar:
> I think some of the boilerplate code and object construction can be reduced by using LocalDateTime directly: LocalDateTime.ofEpochSecond(value.toLong, 0, ZoneId.UTC).getYear().
> 
> Some objects might be allocated, some redundant calculations might be performed, but they'll get inlined, escape analysis will perform scalar replacement, and then the extra steps will be removed.
> 
> Note that the Joda-time library was better suited for raw field extraction without allocations, but internally it still performed redundant steps and it had magical caches. In the end, it wasn't any better than what the Java time library does today.

Thanks @broneill! I have made changes to use LocalDateTime instead of Instant

PR #496, opened 2019-09-16
feat(core, query): Remove dataset from query APIs; add schema to leaf ExecPlans

velvia:
**Pull Request checklist**

All `ExecPlans` have `Dataset` as a parameter in most APIs

**New behavior :**

`ExecPlans` APIs have dataset removed, everywhere.  
Instead, the `ExecPlans` which need the schema, have the Schema injected into the constructor (basically only the leaf/raw ExecPlans).
As a bonus this opens up execution to the possibility of cross-dataset queries.


query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
```diff
@@ -286,17 +287,18 @@ object RangeFunction {
   /**
    * Returns a function to generate a ChunkedRangeFunction for Long columns
    */
-  def longChunkedFunction(dataset: Dataset,
+  def longChunkedFunction(schema: ResultSchema,
                           func: Option[RangeFunctionId],
                           funcParams: Seq[Any] = Nil): RangeFunctionGenerator = {
+    val isDownsampleGauge = (schema.columns == Schemas.dsGauge.dataInfos)
```
vishramachandran: As mentioned earlier, the resultSchema would just be the columns necessary for executing the downsample query. This would not necessarily be the same as full dsGuage schema.

query/src/main/scala/filodb/query/exec/SelectRawPartitionsExec.scala
```diff
@@ -32,64 +32,67 @@ final case class SelectRawPartitionsExec(id: String,
...
 
-  protected[filodb] def schemaOfDoExecute(dataset: Dataset): ResultSchema = {
+  def dataset: DatasetRef = datasetRef
+
+  protected[filodb] def schemaOfDoExecute(): ResultSchema = {
     require(!colIds.contains(0),
       "User selected columns should not include timestamp (row-key); it will be auto-prepended")
 
-    val selectedColIds = selectColIds(dataset)
+    val selectedColIds = selectColIds()
```
vishramachandran: Note that this will only return colIds required for serving the downsampled data. The inference of resultSchema == dsGuage later in the pipeline would be wrong.

query/src/test/scala/filodb/query/exec/WindowIteratorSpec.scala
```diff
@@ -405,21 +408,28 @@ class WindowIteratorSpec extends RawDataWindowingSpec {
...
                     useChunked = true).asChunkedD, queryConfig)
     chunkedIt.map(r => (r.getLong(0), r.getDouble(1))).filter(!_._2.isNaN).toList shouldEqual windowResults
   }
 
 
   it("should calculate query results from downsampled data") {
+    // schema stuff: this is important
+    val dsColNames = RangeFunction.downsampleColsFromRangeFunction(MetricsTestData.downsampleSchema,
+                                                                   Some(RangeFunctionId.AvgOverTime))
+    val dsColIDs = Seq(0) ++ MetricsTestData.downsampleSchema.colIDs(dsColNames: _*).get
```
vishramachandran: Not all columns would be selected.

query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
```diff
@@ -90,12 +89,11 @@ final case class PeriodicSamplesMapper(start: Long,
   }
 
   // Transform source double or long to double schema
-  override def schema(dataset: Dataset, source: ResultSchema): ResultSchema = {
-
-    if (dataset.options.hasDownsampledData) {
+  override def schema(source: ResultSchema): ResultSchema = {
+    // Special treatment for downsampled gauge schema - return regular timestamp/value
+    if (source.columns == Schemas.dsGauge.dataInfos) {
```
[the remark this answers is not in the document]
velvia: Your observation is indeed correct.
This is just a temporary thing.  It will be fixed in the next PR.
It's not possible to fix it without a major refactor, thus I'm delaying it to the next PR, which has the proper fix -- and I'm reluctant to deviate from the plan unless it's a really simple fix, but I can't think of anything simple.

query/src/main/scala/filodb/query/exec/SelectRawPartitionsExec.scala
velvia: Yes, this will be fixed in subsequent PR.

query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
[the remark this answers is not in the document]
velvia: Actually I think I can start on some of the intended changes from the plan now, then it won't be broken now.  Watch out for changes.

velvia:
@vishramachandran  I've fixed the downsampling queries by adjusting range functions to execute ahead of time.   Have a look!

query/src/main/scala/filodb/query/exec/SelectRawPartitionsExec.scala
```diff
@@ -22,6 +23,29 @@ object SelectRawPartitionsExec {
...
+  def replaceRangeFunction(transformers: Seq[RangeVectorTransformer],
+                           oldFunc: Option[RangeFunctionId],
+                           newFunc: Option[RangeFunctionId]): Seq[RangeVectorTransformer] =
+    transformers.map {
+      case p: PeriodicSamplesMapper if p.functionId == oldFunc => p.copy(functionId = newFunc)
+      case other: RangeVectorTransformer => other
+    }
+
+  // Optimize transformers, replacing range functions for downsampled gauge queries if needed
+  def optimizeForDownsample(schema: Schema, transformers: Seq[RangeVectorTransformer]): Seq[RangeVectorTransformer] = {
```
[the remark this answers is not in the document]
velvia: The dataset has many schemas nd should not have `hasDownsampledData`.  It may even mix up downsampled and normal schemas.

"generating wrong plan"... what is meant by this?  WE really have just two choices:
- generate a generic, schema-agnostic plan at planning time, then generating a more specific one later (current approach chosen)
OR - generate a schema-specific, final plan at planning time.

query/src/main/scala/filodb/query/PlanEnums.scala
```diff
@@ -62,6 +62,9 @@ object RangeFunctionId extends Enum[RangeFunctionId] {
 
   case object AvgOverTime extends RangeFunctionId("avg_over_time")
 
+  // Used only for ds-gauge schema
+  case object AvgWithSumAndCountOverTime extends RangeFunctionId("avg_with_sum_count_over_time")
```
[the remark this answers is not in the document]
vishramachandran: Downsampling details are not exposed to the user. For example, we may choose to downsample counter differently tomorrow. But that doesnt mean that user needs to invoke a different function to operate on the downsampled data. Hence, a new user-facing function is not warranted.

velvia:
@vishramachandran  

> Also, I didnt see a reason to change the earlier approach for serving the average. Why change something when it works? All that is needed is a more reliable way to figure out if the source schema has downsampled data or not.

1. The PSM has no access to the source schema, and I really don't think it should.  By the time PSM or any transformer is chosen, it should already have the right function that takes into account the source schema.  A downsampled avg_over_time function is fundamentally a different operation because the DS gauge schema is different, so at PSM time, the best isolation of abstraction and the cleanest approach is to choose a function specific to the different input schema, especially given it has two columns.
2.  "If the source schema has downsampled data or not" ... what you mean is, is the source schema a downsampled gauge.  Remember that this special treatment is specific for gauges.

The current approach is wrong because it leaks abstractions about the source schema into a transformers.  Transformers should only be concerned with transforming what data is coming into it, which is no longer the source schema.   The move to multi schema exposes the current design flaw and requires a cleaner solution.

vishramachandran:
It is clear that the choice of the RangeFunction requires knowledge of the schema. That we agree on. This is precisely why I proposed that we fuse PeriodicSamplesMapper and SelectRawPartitionsExec. The two don't live independently. For use cases where they need to be separate, we can create another leaf exec plan type. 

This solution addresses all concerns.

velvia:
> It is clear that the choice of the RangeFunction requires knowledge of the schema. That we agree on. This is precisely why I proposed that we fuse PeriodicSamplesMapper and SelectRawPartitionsExec. The two don't live independently. For use cases where they need to be separate, we can create another leaf exec plan type.

Let's chat in person.  I do not think this is the right direction.  Maintaining separation of concerns of transformers is a good goal, and the PSM and selecting raw data can and should be independent.
It is better to have schema-specific RangeFunctions as needed. The RangeFunction that is chosen can be independent of the user function name; that is, there should be an ExecPlan-level RangeFunction and a user-facing thing.

query/src/main/scala/filodb/query/PlanEnums.scala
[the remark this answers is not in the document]
velvia: We need a different `RangeFunction` for sure.  Right now, the ExecPlan uses the user-facing RangeFunctionId, but in fact the lower level needs a separate abstraction.

query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
```diff
@@ -226,14 +227,9 @@ trait ChunkedLongRangeFunction extends TimeRangeFunction[TransientRow] {
 object RangeFunction {
   type RangeFunctionGenerator = () => BaseRangeFunction
 
-
-  def downsampleColsFromRangeFunction(dataset: Dataset, f: Option[RangeFunctionId]): Seq[String] = {
+  def downsampleColsFromRangeFunction(schema: Schema, f: Option[RangeFunctionId]): Seq[String] = {
     f match {
       case None                   => Seq("avg")
-      case Some(Rate)             => Seq(dataset.schema.data.valueColName)
```
TanviBhavsar: Don't we have to support these functions now?

velvia: @TanviBhavsar  if you look a few lines down, I simply moved a whole bunch of them into the `case other =>` which does the same thing.

velvia:
@vishramachandran  have separated out `InternalRangeFunction` for ExecPlan execution, so no more publicly visible new function.  Have a look!

broneill:
The way the parser selects functions is by looking for named enums? There's no registration for this, and so any function becomes public? Instead, the registration of names to functions should be explicit, with a hashtable mapping. The enums themselves should have no names at all, and then we can add as many as we like without concern that implementation is leaking out.

This strategy is typical of parsers. The set of nodes that the represent the parse tree don't necessarily have to match the public language. Many intermediate transforms are made, such as adding "goto" nodes where none might exist in the public language.

velvia:
@broneill  maybe this is just splitting hairs, but technically the public name is a member of the Enum class/trait.  I actually think this is fairly explicit, as hash mappings often times it's easy to leave things out, but the `sealed trait` ensures that nothing is left out.  Anyways the last change was really to separate out an internal range function type, which is needed for better abstraction

broneill:
Because the PromQL names are embedded in the functions, these functions cannot be reused as easily for use in other query languages. Separation of name and function is more flexible, and it allows for internal transformations too. Totally out of scope at this point, however.

vishramachandran:
@velvia I personally don't prefer to use inheritance for solving this issue. It is tempting to use inheritance to solve such problems but it not always the best approach. The way I see it, we need a set of parallel constructs under the `exec` package and explicit mappings from Logical to Exec world in the Query Planner. That would be the cleanest and most readable. This is already being done for Logical and Exec Plans, and is needed for functions too. Please consider retaining the pattern.

PR #507, opened 2019-09-25
feat(core): Honor order in copyTags

tjackpaul:
**Pull Request checklist**

**Current behavior :**
Currently, we do not honor order in which tags are represented in DatasetOptions `copyTags`. Because of which gateway is not consistent w.r.t adding new tags.

**New behavior :**
Gateway will consistently add the new tag in the order in which it is added in `copytTags`

### DIFF
--- a/core/src/main/resources/filodb-defaults.conf
+++ b/core/src/main/resources/filodb-defaults.conf
@@ -29,6 +29,10 @@ filodb {
     predefined-keys = ["_ws_", "_ns_", "app", "__name__", "instance", "dc", "le", "job", "exporter"]
 
     options {
+      # Copy tags can be used to create a new labelPair from one of the existing labelPair in a TimeSeries.
+      # It honors the order in which the label keys are represented against a new key.
+      # Eg: For copyTags `"_ns_" = [ "exporter", "job" ]`, when an incoming TimeSeries already has both `exporter`
+      #     and `job`, then it will pickup only the value from `exporter` for `_ns_`.
       copyTags = {
         "_ns_" = [ "_ns", "exporter", "job" ]
       }
--- a/core/src/main/resources/filodb-defaults.conf
+++ b/core/src/main/resources/filodb-defaults.conf
@@ -30,8 +30,7 @@ filodb {
 
     options {
       copyTags = {
-        exporter = "_ns_"
-        job = "_ns_"
+        "_ns_" = [ "_ns", "exporter", "job" ]
       }
       ignoreShardKeyColumnSuffixes = { "__name__" = ["_bucket", "_count", "_sum"] }
       ignoreTagsOnPartitionKeyHash = ["le"]
--- a/gateway/src/main/scala/filodb/gateway/conversion/InputRecord.scala
+++ b/gateway/src/main/scala/filodb/gateway/conversion/InputRecord.scala
@@ -106,9 +106,7 @@ object PrometheusInputRecord {
     } else {
       val metric = metricTags.head._2
       val metricKey = metricTags.head._1
-      println(tags)
       val transformedTags = transformTags(tags.filterNot(_._1 == metricKey), promCounter.options)
-      println(transformedTags)
       (0 until tsProto.getSamplesCount).map { i =>
         val sample = tsProto.getSamples(i)
         PrometheusInputRecord(transformedTags, metric, sample.getTimestampMs, sample.getValue)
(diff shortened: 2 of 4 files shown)


core/src/main/resources/filodb-defaults.conf
```diff
@@ -30,8 +30,7 @@ filodb {
 
     options {
       copyTags = {
-        exporter = "_ns_"
-        job = "_ns_"
+        "_ns_" = [ "_ns", "exporter", "job" ]
```
velvia: Could we add a comment to explain what this means?   Thanks

gateway/src/main/scala/filodb/gateway/conversion/InputRecord.scala
```diff
@@ -47,7 +47,7 @@ trait InputRecord {
  * the partition key hash.
  * The tags should NOT include the metric name.
  */
-case class PrometheusInputRecord(tags: Map[String, String],
+case class PrometheusInputRecord(tags: collection.mutable.Map[String, String],
```
velvia: No, let's not make the input mutable, that leads to a whole class of potential bugs where value could change underneath as we are encoding.

gateway/src/main/scala/filodb/gateway/conversion/InputRecord.scala
```diff
@@ -96,17 +96,18 @@ object PrometheusInputRecord {
 
   // Create PrometheusInputRecords from a TimeSeries protobuf object
   def apply(tsProto: TimeSeries): Seq[PrometheusInputRecord] = {
-    val tags = (0 until tsProto.getLabelsCount).map { i =>
+    val tags = collection.mutable.HashMap[String, String]()
+    for ( i <- 0 until tsProto.getLabelsCount) {
```
velvia: This change is not necessary - it creates a map from the tags in the photo, which is exactly what the old code did.  However, the old code is shorter and creates an immutable map, which is preferable.

gateway/src/main/scala/filodb/gateway/conversion/InputRecord.scala
```diff
@@ -119,17 +120,14 @@ object PrometheusInputRecord {
...
-    val keys = tags.map(_._1).toSet
-    val extraTags = new collection.mutable.ArrayBuffer[(String, String)]()
-    tags.foreach { case (k, v) =>
-      if (options.copyTags contains k) {
-        val renamedKey = options.copyTags(k)
-        if (!(keys contains renamedKey))
-          extraTags += renamedKey -> v
+  def transformTags(tags: collection.mutable.HashMap[String, String],
+                    options: DatasetOptions): collection.mutable.Map[String, String] = {
+    for ((k, v) <- options.copyTags) {
```
velvia: Here is how you can write an immutable version of transformTags: filter and create the extra tags, then add them to the base collection.  Advantage: we keep the output immutable and free of potential mutation bugs.

    val extraTags = options.copyTags.collect {
      case (k, v) if !tags.contains(v) && tags.contains(k) => v -> tags(k)
    }
    tags ++ extraTags

gateway/src/main/scala/filodb/gateway/conversion/InputRecord.scala
tjackpaul: Please ignore the above change. It was part of an old commit where i was testing something.

gateway/src/main/scala/filodb/gateway/conversion/InputRecord.scala
tjackpaul: Please ignore the above change. It was part of an old commit where i was testing something.

gateway/src/main/scala/filodb/gateway/conversion/InputRecord.scala
tjackpaul: The above code is overriding until all the items in copyTags list is being iterated through. We do not want that to happen since we want to honor the order. Hence, i added the following check : 
```    for ((k, v) <- options.copyTags) {
      **if (!extraTags.contains(v)**
        && !tagsMap.contains(v)
        && tagsMap.contains(k)) {
        extraTags += v -> tagsMap(k)
      }
    }
```

core/src/main/resources/filodb-defaults.conf
tjackpaul: Added comments.

PR #510, opened 2019-09-26
bug(downsampler): Thread local memory constructs for multi-threaded spark executors

vishramachandran:
**Pull Request checklist**

When there are multiple executors per box, each thread needs its own memory constructs. Earlier, downsampler job was sharing memory constructs among all executors on a worker. 

Tested locally by spawning 3 node cass cluster causing multiple spark partitions and executors to be spun off.

### DIFF
--- a/spark-jobs/src/main/scala/filodb.downsampler/BatchDownsampler.scala
+++ b/spark-jobs/src/main/scala/filodb.downsampler/BatchDownsampler.scala
@@ -75,7 +75,9 @@ object BatchDownsampler extends StrictLogging with Instance {
     */
   private[downsampler] val rawDatasetRef = DatasetRef(settings.rawDatasetName)
 
-  private val maxMetaSize = dsSchemas.map(_.data.blockMetaSize).max
+  // FIXME * 2 exists to workaround an issue where we see underallocation for metaspan due to
+  // possible mis-calculation of max block meta size.
+  private val maxMetaSize = dsSchemas.map(_.data.blockMetaSize).max * 2
 
   /**
     * Datasets to which we write downsampled data. Keyed by Downsample resolution.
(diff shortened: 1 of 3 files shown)


broneill:
I have a general comment regarding the use of thread locals here. They're great for leaking vast amounts of memory because they don't get cleaned up right away. Do you have direct control over what thread instances are running? Will they stop when idle and be replaced by a new one later?

spark-jobs/src/main/scala/filodb.downsampler/BatchDownsampler.scala
```diff
@@ -124,19 +107,21 @@ object BatchDownsampler extends StrictLogging with Instance {
...
-                     downsampledChunksToPersist, userTimeStart, userTimeEnd)
+    try {
+      rawPartsBatch.foreach { rawPart =>
+        downsamplePart(rawPart, rawPartsToFree, downsampledPartsPartsToFree,
+          downsampledChunksToPersist, userTimeStart, userTimeEnd)
+      }
+      persistDownsampledChunks(downsampledChunksToPersist)
+    } finally {
+      // reclaim all blocks
+      offHeapMem.get().blockMemFactory.markUsedBlocksReclaimable()
```
broneill: Probably should move this to a close/free method on the offHeapMemObject.

vishramachandran: Right now, I don't think any abstraction is offered by the offHeapMemObject. It is just a place holder for thread locals. The user of the block mem is responsible for releasing.

vishramachandran:
Brian, as discussed in person, The executor jvm is shut down after the job is done, so release of memory after use is not a strong issue.

PR #515, opened 2019-09-29
bug(query): store last element of chunk in ChangesChunkedFunction

TanviBhavsar:
**Pull Request checklist**

 While calculating changes we have access to only elements of current chunk. Consecutive elements are compared to calculate changes. When window has more than one chunk, first element of current chunk should be compared with the last element of previous chunk

**New behavior :**

 Add prev to ChangesChunkedFunction which will store the last element of the previous chunk. Changes function now has ignorePrev to tell whether the first value of chunk should be compared with prev. The default value of previous is NaN. Changes is not incremented if prev is NaN. DoubleLongWrapDataReader calls  LongBinaryVector which converts double prev to long. So before calling changes ignorePrev is made false if prev is NaN

### DIFF
--- a/memory/src/main/scala/filodb.memory/format/vectors/DeltaDeltaVector.scala
+++ b/memory/src/main/scala/filodb.memory/format/vectors/DeltaDeltaVector.scala
@@ -208,28 +208,19 @@ object DeltaDeltaDataReader extends LongVectorDataReader {
   }
 
   def changes(vector: BinaryVectorPtr, start: Int, end: Int, prev: Long, ignorePrev: Boolean = false): (Long, Long) = {
-    println("Ind DeltaDeltaVector changes")
-    println("slope vector:" + slope(vector))
-    if (slope(vector) == 0)
-      println("slope is 0")
-//    if (slope(vector) == 0) {
-//      (0,0)
-//    } else {
       require(start >= 0 && end < length(vector), s"($start, $end) is out of bounds, length=${length(vector)}")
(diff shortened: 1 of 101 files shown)


memory/src/main/scala/filodb.memory/format/vectors/DeltaDeltaVector.scala
```diff
@@ -274,12 +270,21 @@ object DeltaDeltaConstDataReader extends LongVectorDataReader {
     }
   }
 
-  def changes(vector: BinaryVectorPtr, start: Int, end: Int): Int = {
+  def changes(vector: BinaryVectorPtr, start: Int, end: Int, prev: Long, ignorePrev: Boolean = false): (Long, Long) = {
+
+    require(start >= 0 && end < length(vector), s"($start, $end) is out of bounds, length=${length(vector)}")
+    val firstValue = iterate(vector, start).next
+    val lastValue = iterate(vector, end).next
```
velvia: You can call `apply` instead of `iterate`, this avoids the creation of two objects.  Also you don't need to call lastValue all the time.   So something like this:

    val firstValue = apply(vector, start)
    val changes = if (!ignorePrev && prev != firstValue) 1 else 0
    if (slope(vector) == 0) (0 + changes, apply(vector, end))

TanviBhavsar:
> Just a minor change to avoid allocations and not call `iterate`

Thanks Evan! I have changed iterate to apply

PR #527, opened 2019-10-14
bug(memory): Fix handling of blocks which are too full for more metadata.

broneill:
**Pull Request checklist**

**Current behavior :**
Sometimes a call to Block.allocMetadata fails with OutOfOffheapMemoryException, and this can also cause some unit tests to fail. It used to be that this method returned 0 instead of throwing an exception. Returning of zero was generally considered safe, except it was bad if all spanned blocks had no room for metadata.

**New behavior :**
Ensure that a non-zero address is returned from MemFactory.endMetaSpan, possibly by allocating another block. Don't throw an exception from Block.allocMetadata, since this is a regression.

### DIFF
--- a/memory/src/main/scala/filodb.memory/MemFactory.scala
+++ b/memory/src/main/scala/filodb.memory/MemFactory.scala
@@ -274,12 +274,6 @@ class BlockMemFactory(blockStore: BlockManager,
     * @return the Long native address of the last metadata block written
     */
   def endMetaSpan(metadataWriter: Long => Unit, metaSize: Short): Long = {
-    if (!metadataSpan.last.hasCapacity(metaSize + 2)) {
-      // Can be caused by the first block being nearly full, and allocateOffheap was never
-      // called.  Always make sure that a non-zero address is returned by this method.
-      doAllocateOffheap(0, metaSize)
-    }
-
     var metaAddr: Long = 0
     metadataSpan.foreach { blk =>
       // It is possible that the first block(s) did not have enough memory.  Don't write metadata to full blocks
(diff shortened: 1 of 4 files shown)


memory/src/main/scala/filodb.memory/MemFactory.scala
```diff
@@ -280,6 +280,19 @@ class BlockMemFactory(blockStore: BlockManager,
       metaAddr = blk.allocMetadata(metaSize)
       if (metaAddr != 0) metadataWriter(metaAddr)
     }
+
+    if (metaAddr == 0) {
```
velvia: I'm not sure we should do this.  The reason is, that metadata is supposed to be co-located with chunks in the same block, for the express purpose of properly being able to notify the TSPartition owners of their memory being recycled/reclaimed.  If it is located in a different block, then the ability to notify on free properly is compromised.  Granted, not having metadata allocated also compromises that.   I'm thinking, we need to actually address why the metadata could not be allocated in the first place.  It should not happen if the `allocMetadata` has already taken cared of capacity for the metadata, but maybe there are edge conditions we aren't thinking of.   Actually, what we really probably need is an atomic "allocate metadata and space for all chunks".

broneill: The edge case is quite simple. Start a metaspan and then end it. This used to cause a segv, but then an exception was thrown instead. It doesn't really fix the problem. I can perhaps change where the exception is thrown, but what is the contract supposed to be?

broneill: Ok, not that simple. The metaspan needs to be started and ended when there's no room left.

velvia: Hmmm.  So ideally what we want is:
1) If the endMetaspan was called with no previous chunks allocated in that span, go ahead and flip to a new block
2) however if chunks were already allocated in the current block, one needs to allocate metadata in the current block and not the next block

broneill: I think the problem lies in the fact that metadata size is both pinned to the BlockMemFactory (in the ctor) and the caller can override this. By eliminating the caller override, the start metaspan can check early and move to another block.

broneill:
@velvia I have a question regarding the meta size intent. Should it be fixed, passed into the ctor of MemFactory, or should it always be provided by the caller? Currently it does both, which is a problem.

PR #529, opened 2019-10-16
feat(core): Read-Only FiloDB capability for Querying Downsampled Data

vishramachandran:
**Pull Request checklist**

This PR adds new Memstore implementation to support serving downsampled data from Cassandra without caching it in off-heap memory. This read-only store does not perform ingestion.

`Memstore` BTW, should be renamed to just `TimeSeriesStore`.

PR also includes unit test that writes raw data to cassandra, runs downsampler batch job and reads downsampled data using PagedReadablePartition


spark-jobs/src/main/scala/filodb.downsampler/BatchDownsampler.scala
```diff
@@ -14,10 +14,11 @@ import filodb.cassandra.FiloSessionProvider
 import filodb.cassandra.columnstore.CassandraColumnStore
 import filodb.core.{DatasetRef, ErrorResponse, Instance}
 import filodb.core.binaryrecord2.RecordSchema
-import filodb.core.downsample.{ChunkDownsampler, DoubleChunkDownsampler, HistChunkDownsampler, TimeChunkDownsampler}
+import filodb.core.downsample._
```
broneill: I generally prefer to see this change going the other way -- fewer wildcards. I don't care too much, however.

core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesStore.scala
```diff
@@ -0,0 +1,175 @@
...
+  override def recoverStream(dataset: DatasetRef, shard: Int,
+                             stream: Observable[SomeData],
+                             startOffset: Long, endOffset: Long, checkpoints: Map[Int, Long],
+                             reportingInterval: Long): Observable[Long] = throw new UnsupportedOperationException()
+
+  override def numPartitions(dataset: DatasetRef, shard: Int): Int = throw new UnsupportedOperationException()
+
+  override def numRowsIngested(dataset: DatasetRef, shard: Int): Long = throw new UnsupportedOperationException()
+
+  override def latestOffset(dataset: DatasetRef, shard: Int): Long = throw new UnsupportedOperationException()
```
broneill: Isn't ??? the same as throwing this exception?

spark-jobs/src/main/scala/filodb.downsampler/BatchDownsampler.scala
vishramachandran: It was intellij auto imports at play. It introduces wildcard after a certain number of whitelisted imports.

core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesStore.scala
vishramachandran: Both throw exceptions, but different ones. `???` is technically not-implemented-error.

PR #534, opened 2019-10-24
feat(cass): Faster Cassandra reads for raw partition data

vishramachandran:
**Pull Request checklist**

There is no limit on user time length of a chunk. Because of this, it is not simple to issue one cassandra query and get only the relevant chunks for a given user time range in query. Currently, ODP logic currently load ALL chunkInfos for a partition into memory and then decide which chunk time ranges to page in. This involves two cassandra reads per partition and is expensive (or not even viable) especially when serving long term retained (downsampled) data which is almost always not going to be in memory.

**New behavior :**
This PR 
* limits the time-length of a chunk during ingestion
* issues just one CQL query to fetch the chunks. The maxChunkTime is subtracted from the startTime so bordering chunk that contain relevant data is included.
* Also refactors downsampler config so that maxChunkTime and related downsample config is available per dataset rather than in server config.

### DIFF
--- a/cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
+++ b/cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
@@ -264,7 +264,6 @@ case class CassandraTokenRangeSplit(tokens: Seq[(String, String)],
 }
 
 trait CassandraChunkSource extends RawChunkSource with StrictLogging {
-  import Iterators._
 
   def config: Config
   def filoSessionProvider: Option[FiloSessionProvider]
@@ -291,68 +290,27 @@ trait CassandraChunkSource extends RawChunkSource with StrictLogging {
   val partParallelism = 4
 
   def readRawPartitions(ref: DatasetRef,
+                        maxChunkTime: Long,
                         partMethod: PartitionScanMethod,
                         chunkMethod: ChunkScanMethod = AllChunkScan): Observable[RawPartData] = {
-    val indexTable = getOrCreateIngestionTimeIndexTable(ref)
-    logger.debug(s"Scanning partitions for $ref with method $partMethod...")
-    val (filters, infoRecords) = partMethod match {
-      case SinglePartitionScan(partition, _) =>
-        (Nil, indexTable.getInfos(partition))
-
-      case MultiPartitionScan(partitions, _) =>
-        (Nil, indexTable.getMultiInfos(partitions))
-
-      case FilteredPartitionScan(CassandraTokenRangeSplit(tokens, _), filters) =>
-        (filters, indexTable.scanInfos(tokens))
-
-      case other: PartitionScanMethod =>  ???
+    val partitions = partMethod match {
+      case MultiPartitionScan(p, _) => p
+      case SinglePartitionScan(p, _) => Seq(p)
+      case p => throw new UnsupportedOperationException(s"PartitionScan $p to be implemented later")
     }
-    // For now, partition filter func is disabled.  In the near future, this logic to scan through partitions
-    // and filter won't be necessary as we will have Lucene indexes to pull up specific partitions.
-    // val filterFunc = KeyFilter.makePartitionFilterFunc(dataset, filters)
-
-    partMethod match {
-      // Compute minimum start time from the infos, then do efficient range query MULTIGET
-      case MultiPartitionScan(partitions, _) =>
-        val chunkTable = getOrCreateChunkTable(ref)
-        Observable.fromTask(startTimeFromInfos(infoRecords, chunkMethod.startTime, chunkMethod.endTime))
-          .flatMap { case startTime =>
-            chunkTable.readRawPartitionRange(partitions, startTime, chunkMethod.endTime)
-          }
-      case other: PartitionScanMethod =>
-        // NOTE: we use hashCode as an approx means to identify ByteBuffers/partition keys which are identical
-        infoRecords.sortedGroupBy(_.binPartition.hashCode)
-                   .mapAsync(partParallelism) { case (_, binIndices) =>
-                     val infoBytes = binIndices.map(_.data.array)
-                     assembleRawPartData(ref, binIndices.head.binPartition.array, infoBytes, chunkMethod)
-                   }
-    }
-  }
-
-  // Returns the minimum start time from all the ChunkInfos that are within [startTime, endTime) query range
-  // Needed so we can do range queries efficiently by chunkID
-  private def startTimeFromInfos(infoRecords: Observable[InfoRecord],
-                                 startTime: Long,
-                                 endTime: Long): Task[Long] = {
-    infoRecords.foldLeftL(Long.MaxValue) { case (curStart, InfoRecord(_, data)) =>
-      val infoStartTime = ChunkSetInfo.getStartTime(data.array)
-      val infoEndTime = ChunkSetInfo.getEndTime(data.array)
-      if (infoStartTime <= endTime && infoEndTime >= startTime) {
-        Math.min(curStart, infoStartTime)
-      } else {
-        curStart
-      }
-    }
+    val chunkTable = getOrCreateChunkTable(ref)
+    val startTime = if (chunkMethod == AllChunkScan) Long.MinValue else chunkMethod.startTime - maxChunkTime
+    chunkTable.readRawPartitionRange(partitions, startTime, chunkMethod.endTime)
   }
 
   def getOrCreateChunkTable(dataset: DatasetRef): TimeSeriesChunksTable = {
-    chunkTableCache.getOrElseUpdate(dataset, { (dataset: DatasetRef) =>
+    chunkTableCache.getOrElseUpdate(dataset, { dataset: DatasetRef =>
       new TimeSeriesChunksTable(dataset, clusterConnector, ingestionConsistencyLevel)(readEc) })
   }
 
   def getOrCreateIngestionTimeIndexTable(dataset: DatasetRef): IngestionTimeIndexTable = {
     indexTableCache.getOrElseUpdate(dataset,
-                                    { (dataset: DatasetRef) =>
+                                    { dataset: DatasetRef =>
                                       new IngestionTimeIndexTable(dataset, clusterConnector)(readEc) })
   }
 
@@ -364,17 +322,4 @@ trait CassandraChunkSource extends RawChunkSource with StrictLogging {
 
   def reset(): Unit = {}
 
-  private def assembleRawPartData(ref: DatasetRef,
-                                  partKeyBytes: Array[Byte],
-                                  chunkInfos: Seq[Array[Byte]],
-                                  chunkMethod: ChunkScanMethod): Task[RawPartData] = {
-    val chunkTable = getOrCreateChunkTable(ref)
-
-    val filteredInfos = chunkInfos.filter { infoBytes =>
-      ChunkSetInfo.getStartTime(infoBytes) <= chunkMethod.endTime &&
-      ChunkSetInfo.getEndTime(infoBytes) >= chunkMethod.startTime
-    }
-
-    chunkTable.readRawPartitionData(partKeyBytes, filteredInfos)
-  }
 }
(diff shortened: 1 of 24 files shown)


broneill:
Is this limiting the size of the chunk based on user time? This interferes with any logic to flush chunks based on ingestion time only and might interfere with DR.

vishramachandran:
@broneill This change affects chunking only, and in theory should not affect when flush happens. Let's sync up in person.

cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
```diff
@@ -291,68 +292,33 @@ trait CassandraChunkSource extends RawChunkSource with StrictLogging {
...
-    infoRecords.foldLeftL(Long.MaxValue) { case (curStart, InfoRecord(_, data)) =>
-      val infoStartTime = ChunkSetInfo.getStartTime(data.array)
-      val infoEndTime = ChunkSetInfo.getEndTime(data.array)
-      if (infoStartTime <= endTime && infoEndTime >= startTime) {
-        Math.min(curStart, infoStartTime)
-      } else {
-        curStart
-      }
+      case FilteredPartitionScan(CassandraTokenRangeSplit(tokens, _), Nil)  =>
+        chunkTable.scanPartitionsBySplit(tokens)
```
velvia: Shouldn't we also pass in the start and end time?  C* is able to filter by partition key range as well, which would limit the scan time of this query significantly.

cassandra/src/main/scala/filodb.cassandra/columnstore/TimeSeriesChunksTable.scala
```diff
@@ -126,6 +126,22 @@ sealed class TimeSeriesChunksTable(val dataset: DatasetRef,
     Observable.fromFuture(futRawParts).flatMap { it: Iterator[RawPartData] => Observable.fromIterator(it) }
   }
 
+  def scanPartitionsBySplit(tokens: Seq[(String, String)]): Observable[RawPartData] = {
```
velvia: Add start and end times to allow for filtering here too

cassandra/src/main/scala/filodb.cassandra/columnstore/TimeSeriesChunksTable.scala
```diff
@@ -126,6 +126,22 @@ sealed class TimeSeriesChunksTable(val dataset: DatasetRef,
...
+      session.executeAsync(cql(start, end)).toIterator.handleErrors
+              .map { rowIt =>
+                rowIt.map { row => (row.getBytes(0), chunkSetFromRow(row, 1)) }
+                  .sortedGroupBy(_._1)
+                  .map { case (partKeyBuffer, chunkSetIt) =>
+                    RawPartData(partKeyBuffer.array, chunkSetIt.map(_._2).toBuffer)
+                  }
+              }
+    }
+    it.flatMap{ f => Observable.fromFuture(f).flatMap { it: Iterator[RawPartData] => Observable.fromIterator(it) } }
```
velvia: Use for comprehension so it's more readable

core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
```diff
@@ -127,31 +128,45 @@ extends ChunkMap(memFactory, initMapSize) with ReadablePartition {
...
-          // re-ingest every element, allocating new WriteBuffers
-          if (switchBuffers(blockHolder, encode = true)) { ingest(ingestionTime, row, blockHolder) }
-          else { _log.warn("EMPTY WRITEBUFFERS when switchBuffers called!  Likely a severe bug!!! " +
-                           s"Part=$stringPartition ts=$ts col=$col numRows=${currentInfo.numRows}") }
-          return
-        case other: AddResponse =>
+    if (ts - currentInfo.startTime > maxChunkTime) {
+      switchBuffersAndIngest(ingestionTime, ts, row, blockHolder, maxChunkTime)
+    } else {
+
```
velvia: No need for blank line here

core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
```diff
@@ -127,31 +128,45 @@ extends ChunkMap(memFactory, initMapSize) with ReadablePartition {
...
+                                     blockHolder: BlockMemFactory,
+                                     maxChunkTime: Long): Unit = {
+    // NOTE: a very bad infinite loop is possible if switching buffers fails (if the # rows is 0) but one of the
+    // vectors fills up.  This is possible if one vector fills up but the other one does not for some reason.
+    // So we do not call ingest again unless switcing buffers succeeds.
+    // re-ingest every element, allocating new WriteBuffers
+    if (switchBuffers(blockHolder, encode = true)) { ingest(ingestionTime, row, blockHolder, maxChunkTime) }
+    else { _log.warn("EMPTY WRITEBUFFERS when switchBuffers called!  Likely a severe bug!!! " +
+      s"Part=$stringPartition userTime=$userTime numRows=${currentInfo.numRows}") }
+
```
velvia: No need for extra line

core/src/main/scala/filodb.core/store/ChunkSource.scala
```diff
@@ -45,6 +45,7 @@ trait RawChunkSource {
    * Implemented by lower-level persistent ChunkSources to return "raw" partition data
    */
   def readRawPartitions(ref: DatasetRef,
+                        maxChunkTime: Long,
```
velvia: Please add doc for maxChunkTime

core/src/main/scala/filodb.core/store/IngestionConfig.scala
```diff
@@ -86,7 +88,10 @@ object StoreConfig {
   /** Pass in the config inside the store {}  */
   def apply(storeConfig: Config): StoreConfig = {
     val config = storeConfig.withFallback(defaults)
-    StoreConfig(config.as[FiniteDuration]("flush-interval"),
+    val flushInterval = config.as[FiniteDuration]("flush-interval")
+    val fallbackMaxChunkTime = (flushInterval.toMillis * 1.1).toLong.millis
```
velvia: 1.1 is good, but let's leave a comment why we use that (extra margin of time to ensure chunks aren't ended prematurely?  etc etc)

core/src/main/scala/filodb.core/downsample/DownsampleConfig.scala
```diff
@@ -1,11 +1,22 @@
...
+import scala.concurrent.duration.FiniteDuration
+
 import com.typesafe.config.{Config, ConfigFactory}
 import net.ceedubs.ficus.Ficus._
 
 final case class DownsampleConfig(config: Config) {
   val enabled = config.hasPath("enabled") && config.getBoolean("enabled")
-  val resolutions = if (enabled) config.as[Seq[Int]]("resolutions-ms") else Seq.empty
+
+  val resolutions = if (config.hasPath ("resolutions")) config.as[Seq[FiniteDuration]]("resolutions")
```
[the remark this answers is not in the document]
vishramachandran: I agree. This change crept in because maxChunkLength configuration is needed by downsampler too, and it was small enough to be done in same PR.

cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
```diff
@@ -153,12 +153,14 @@ extends ColumnStore with CassandraChunkSource with StrictLogging {
...
     * ** User/Ingestion End times are exclusive **
     */
+  // scalastyle:off parameter.number
   def getChunksByIngestionTimeRange(datasetRef: DatasetRef,
                                     splits: Iterator[ScanSplit],
                                     ingestionTimeStart: Long,
                                     ingestionTimeEnd: Long,
                                     userTimeStart: Long,
                                     userTimeEnd: Long,
+                                    maxChunkTime: Long,
```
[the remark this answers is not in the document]
vishramachandran: That would be a leak in abstraction to clients reading data from the chunk source. Everyone has to do the adjustment calculation.

cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
vishramachandran: We use FilteredPartitionScan sparingly and only in tests without supporting filters - this should be a different PR with different priority.

cassandra/src/main/scala/filodb.cassandra/columnstore/TimeSeriesChunksTable.scala
vishramachandran: We use FilteredPartitionScan sparingly and only in tests without supporting filters - this should be a different PR with different priority.

cassandra/src/main/scala/filodb.cassandra/columnstore/TimeSeriesChunksTable.scala
vishramachandran: it is a combination of futures and observables hence for-comprehension is not so feasible.

core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
```diff
@@ -127,31 +128,45 @@ extends ChunkMap(memFactory, initMapSize) with ReadablePartition {
...
-          // vectors fills up.  This is possible if one vector fills up but the other one does not for some reason.
-          // So we do not call ingest again unless switcing buffers succeeds.
-          // re-ingest every element, allocating new WriteBuffers
-          if (switchBuffers(blockHolder, encode = true)) { ingest(ingestionTime, row, blockHolder) }
-          else { _log.warn("EMPTY WRITEBUFFERS when switchBuffers called!  Likely a severe bug!!! " +
-                           s"Part=$stringPartition ts=$ts col=$col numRows=${currentInfo.numRows}") }
-          return
-        case other: AddResponse =>
+    if (ts - currentInfo.startTime > maxChunkTime) {
+      switchBuffersAndIngest(ingestionTime, ts, row, blockHolder, maxChunkTime)
```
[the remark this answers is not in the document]
vishramachandran: I'll add a comment. But I wonder if we need a stat. We dont have one for vector-too-small for instance. We are already publishing too many stats.

cassandra/src/main/scala/filodb.cassandra/columnstore/TimeSeriesChunksTable.scala
velvia: It should work.  Both results of flatMaps are observables

cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
```diff
@@ -291,6 +291,26 @@ trait CassandraChunkSource extends RawChunkSource with StrictLogging {
...
+    *
+    * 2. All chunks have samples with a range of userTime. During ingestion, we restrict the maximum
+    * range for the userTime. This restriction makes it possible to issue single CQL query to fetch
+    * all relevant chunks from cassandra. We do this by searching for searching in cassandra for chunkIds
+    * between `chunkID(queryStartTime - maxChunkTime)` and `chunkID(queryEndTime)`. The reason we need to
+    * subtract maxChunkTime from queryStartTime is for the range to include the first chunk which may have
+    * relevant data but may have a startTime outside the query range.
+    *
+    * @param ref dataset ref
+    * @param maxChunkTime maximum userTime allowed in a single chunk. This restriction makes it
```
velvia: Is it in mills?  Please add

cassandra/src/main/scala/filodb.cassandra/columnstore/TimeSeriesChunksTable.scala
```diff
@@ -139,7 +139,7 @@ sealed class TimeSeriesChunksTable(val dataset: DatasetRef,
                   }
               }
     }
-    it.flatMap{ f => Observable.fromFuture(f).flatMap { it: Iterator[RawPartData] => Observable.fromIterator(it) } }
+    res.flatMap{ f => Observable.fromFuture(f).flatMap { it: Iterator[RawPartData] => Observable.fromIterator(it) } }
```
velvia: Please try for comprehension, should be possible

PR #538, opened 2019-10-31
debug(query): Add debugging for binary join failures; don't require equal response size

velvia:
**Pull Request checklist**

BinaryJoins after multi schema query change often fail.

**New behavior :**

- Add debugging to joinKeys when cardinality is expected to be 1-1, but somehow is not.
- Disable check that responses equal LHS + RHS.  This is no longer true with multi schema as filters may return no QueryResult (unknown schema)

**BREAKING CHANGES**

Breaking changes may include:

**Other information**:

### DIFF
--- a/query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
+++ b/query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
@@ -56,7 +56,9 @@ final case class SetOperatorExec(id: String,
       case (QueryResult(_, _, result), i) => (result, i)
       case (QueryError(_, ex), _)         => throw ex
     }.toListL.map { resp =>
-      require(resp.size == lhs.size + rhs.size, "Did not get sufficient responses for LHS and RHS")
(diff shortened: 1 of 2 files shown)


query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
```diff
@@ -56,7 +56,9 @@ final case class SetOperatorExec(id: String,
       case (QueryResult(_, _, result), i) => (result, i)
       case (QueryError(_, ex), _)         => throw ex
     }.toListL.map { resp =>
-      require(resp.size == lhs.size + rhs.size, "Did not get sufficient responses for LHS and RHS")
+      // NOTE: We can't require this any more, as multischema queries may result in not a QueryResult if the
+      //       filter returns empty results.  The reason is that the schema will be undefined.
+      // require(resp.size == lhs.size + rhs.size, "Did not get sufficient responses for LHS and RHS")
```
vishramachandran: I still need to understand why/how this require becomes invalid.
This PR is now incomplete because subsequent lines in this algorithm depended on this assumption.

PR #539, opened 2019-10-31
feat(query, coordinator): scalar, vector and time function queries support

TanviBhavsar:
**Pull Request checklist**

- Support for scalar, vector and time functions which do not take any argument - time(), hour(), minute(), days_in_month(), day_of_month() and day_of_week()
- Support for Binary Operator queries between a metric and scalar function
- Binary operator between scalars support 
- Support for queries when instant function has a  scalar function as a  parameter


query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
```diff
@@ -618,13 +619,13 @@ class ChangesChunkedFunctionL extends ChangesChunkedFunction with
...
 
-abstract class QuantileOverTimeChunkedFunction(funcParams: Seq[Any],
+abstract class QuantileOverTimeChunkedFunction(funcParams: Seq[FuncArgs],
                                                var quantileResult: Double = Double.NaN,
                                                var values: Buffer[Double] = Buffer.empty[Double])
   extends ChunkedRangeFunction[TransientRow] {
   override final def reset(): Unit = { quantileResult = Double.NaN; values = Buffer.empty[Double] }
   final def apply(endTimestamp: Long, sampleToEmit: TransientRow): Unit = {
-    val q = funcParams.head.asInstanceOf[Number].doubleValue()
+    val q = funcParams.head.asInstanceOf[StaticFuncArgs].scalar
```
vishramachandran: Again require static arg. Dynamic arg not supported (yet)

query/src/main/scala/filodb/query/exec/rangefn/RangeInstantFunctions.scala
```diff
@@ -1,7 +1,8 @@
 package filodb.query.exec.rangefn
 
+import filodb.core.query.TransientRow
 import filodb.query.QueryConfig
-import filodb.query.exec.TransientRow
+//import filodb.query.exec.TransientRow
```
vishramachandran: remove

query/src/main/scala/filodb/query/exec/rangefn/InstantFunction.scala
```diff
@@ -350,29 +338,26 @@ case class DaysInMonthImpl(funcParams: Seq[Any]) extends EmptyParamsInstantFunct
...
+    require(scalarParam.length == 1, "Quantile (between 0 and 1) required for histogram quantile")
+    val q = scalarParam(0)
+    value.quantile(q)
+  }
 }
 
 /**
  * Histogram max quantile function for Histogram column + extra max (Double) column.
- * @param funcParams - a single value between 0 and 1, the quantile to calculate.
+ * //@param scalarParam - a single value between 0 and 1, the quantile to calculate.
```
vishramachandran: move

core/src/main/scala/filodb.core/query/RangeVector.scala
```diff
@@ -93,7 +96,92 @@ trait RangeVector {
...
+  override def numRowsInt: Int = timeValueMap.size
+}
 
+final case class RangeParams(start: Long, step: Long, end: Long)
+trait ScalarSingleValue extends ScalarVector {
+  def rangeParams: RangeParams
+  var numRowsInt : Int = 0
+
+  override def rows: Iterator[RowReader] = {
+    val rowList = new ListBuffer[TransientRow]()
```
velvia: This pattern uses up a whole bunch of memory, which is really not necessary.  It would be far more efficient to do something like the following:

    Iterator.from(rangeParams.start, rangeParams.step).takeWhile(_ <= rangeParams.end).map { i =>
      numRowsInt += 1
      new TransientRow(I*1000, getValue(I*1000))
    }

core/src/main/scala/filodb.core/query/RangeVector.scala
```diff
@@ -93,7 +96,92 @@ trait RangeVector {
...
+
+case class DayOfMonthScalar(rangeParams: RangeParams) extends ScalarSingleValue {
+  def value: Double = LocalDateTime.ofEpochSecond(rangeParams.start, 0, ZoneOffset.UTC).getDayOfMonth
+  override def getValue(time: Long): Double = value
+}
+
+case class DayOfWeekScalar(rangeParams: RangeParams) extends ScalarSingleValue {
+  def value: Double = {
+    val dayOfWeek = LocalDateTime.ofEpochSecond(rangeParams.start, 0, ZoneOffset.UTC).getDayOfWeek
+    if (dayOfWeek == 7) {
```
velvia: Can we collapse this if else into a single line?

core/src/main/scala/filodb.core/query/RangeVector.scala
```diff
@@ -127,17 +215,18 @@ final case class ChunkInfoRangeVector(key: RangeVectorKey,
...
-                                    val numRowsInt: Int,
-                                    containers: Seq[RecordContainer],
-                                    val schema: RecordSchema,
-                                    startRecordNo: Int) extends RangeVector with java.io.Serializable {
+final class SerializedRangeVector(val key: RangeVectorKey,
+                                  val numRowsInt: Int,
+                                  containers: Seq[RecordContainer],
+                                  val schema: RecordSchema,
+                                  startRecordNo: Int) extends RangeVector with SerializableRangeVector with
+  java.io.Serializable {
```
velvia: better to put all the extends and with's to a single line together

velvia: Also, java.io.Serializable should NOT be included.  We don't want to use Java serialization with this class, because it already serializes into a binary record.  Instead we just want Kryo serialization and to make sure Kryo serializer is used, in Serializer.scala.    If we need to include Serializable, then we are using the wrong path.  Please leave it out.

core/src/main/scala/filodb.core/query/TransientRow.scala
```diff
@@ -44,7 +44,7 @@ final class TransientRow(var timestamp: Long, var value: Double) extends Mutable
   def getInt(columnNo: Int): Int = throw new IllegalArgumentException()
   def getLong(columnNo: Int): Long = if (columnNo == 0) timestamp else throw new IllegalArgumentException()
   def getDouble(columnNo: Int): Double = if (columnNo == 1) value
-                                         else throw new IllegalArgumentException(s"Invalid col $columnNo")
+  else throw new IllegalArgumentException(s"Invalid col $columnNo")
```
velvia: Please fix the indentation, it looks horrible to be dedented

core/src/main/scala/filodb.core/query/TransientRow.scala
```diff
@@ -119,10 +119,10 @@ final class AvgAggTransientRow extends MutableRowReader {
   def getBoolean(columnNo: Int): Boolean = throw new IllegalArgumentException()
   def getInt(columnNo: Int): Int = throw new IllegalArgumentException()
   def getLong(columnNo: Int): Long = if (columnNo == 0) timestamp
-                                     else if (columnNo == 2) count
-                                     else throw new IllegalArgumentException()
+  else if (columnNo == 2) count
```
velvia: Please restore the indentation

core/src/main/scala/filodb.core/query/TransientRow.scala
```diff
@@ -119,10 +119,10 @@ final class AvgAggTransientRow extends MutableRowReader {
...
+  else throw new IllegalArgumentException()
   def getDouble(columnNo: Int): Double = if (columnNo == 1) avg
-                                         else throw new IllegalArgumentException()
+  else throw new IllegalArgumentException()
```
velvia: Same here

core/src/main/scala/filodb.core/query/TransientRow.scala
```diff
@@ -154,11 +154,11 @@ final class StdvarAggTransientRow extends MutableRowReader {
...
   def getInt(columnNo: Int): Int = throw new IllegalArgumentException()
   def getLong(columnNo: Int): Long = if (columnNo == 0) timestamp
-                                     else if (columnNo == 3) count
-                                     else throw new IllegalArgumentException()
+  else if (columnNo == 3) count
+  else throw new IllegalArgumentException()
   def getDouble(columnNo: Int): Double = if (columnNo == 1) stdVar
-                                         else if (columnNo == 2) avg
-                                         else throw new IllegalArgumentException()
+  else if (columnNo == 2) avg
```
velvia: Same here

prometheus/src/main/scala/filodb/prometheus/ast/Expressions.scala
```diff
@@ -28,21 +29,41 @@ trait Expressions extends Aggregates with Functions {
       vectorMatch.get.validate(operator, lhs, rhs)
     }
 
-
+    //scalastyle:off
```
velvia: Why is this needed?  Can we please add specific reason to escape scalastyle here, or just fix/refactor the code?

prometheus/src/main/scala/filodb/prometheus/ast/Functions.scala
```diff
@@ -12,64 +13,100 @@ trait Functions extends Base with Operators with Vectors {
...
-      } else if (filoFunctionIdOpt.isDefined) {
-        // No lookback needed as we are looking at chunk metadata only, not raw samples
-        val rangeSelector = timeParamToSelector(timeParams, 0)
-        val (filters, columns) = seriesParam match {
-          case i: InstantExpression => (i.columnFilters :+ i.nameFilter, i.columns)
-          case r: RangeExpression   => (r.columnFilters :+ r.nameFilter, r.columns)
+      if (vectorFn.isDefined) {
+        allParams.head match {
+          case num: ScalarExpression => VectorPlan(ScalarFixedDoublePlan(num.toScalar, RangeParams(timeParams.start, timeParams.step, timeParams.end)))
+          case function: Function => val nestedPlan = function.toPeriodicSeriesPlan(timeParams)
```
velvia: 1) align the `=>`
2) Let's have the val start on the next line at indented level

prometheus/src/main/scala/filodb/prometheus/ast/Functions.scala
```diff
@@ -12,64 +13,100 @@ trait Functions extends Base with Operators with Vectors {
...
+            }
+        }
+      } else if (allParams.isEmpty) {
+        ScalarTimeBasedPlan(scalarFunctionIdOpt.get, RangeParams(timeParams.start, timeParams.step, timeParams.end) )
+
+      }  else {
+        val seriesParam = allParams.filter(_.isInstanceOf[Series]).head.asInstanceOf[Series]
+
+       val stringParam = allParams.filter(!_.equals(seriesParam)).filter(_.isInstanceOf[InstantExpression]).
+         map(_.asInstanceOf[InstantExpression].realMetricName.replaceAll("^\"|\"$", ""))
```
velvia: Can we add some docs here?  No idea what all these transforms are for, and using regexes further makes it hard to understand.

prometheus/src/main/scala/filodb/prometheus/query/PrometheusModel.scala
```diff
@@ -84,16 +84,20 @@ object PrometheusModel {
                 (if (verbose) Map("_shards_" -> srv.key.sourceShards.mkString(","),
                                   "_partIds_" -> srv.key.partIds.mkString(","))
                 else Map.empty)
+    val samples = srv.rows.filter(!_.getDouble(1).isNaN).map { r =>
```
velvia: Why is this needed?
Also you need to check the output schema.  Not all the outputs have double as the output type, and this code would crash.

prometheus/src/main/scala/filodb/prometheus/query/PrometheusModel.scala
```diff
@@ -84,16 +84,20 @@ object PrometheusModel {
...
+      Sampl(r.getLong(0) / 1000, r.getDouble(1))
+    }.toSeq
+
+    val res = if (samples.isEmpty)
```
velvia: One line please.
Also seems this should have been a separate PR.

query/src/main/scala/filodb/query/LogicalPlan.scala
```diff
@@ -144,28 +144,58 @@ case class ScalarVectorBinaryOperation(operator: BinaryOperator,
...
+                             functionArgs: Seq[Any] = Nil) extends PeriodicSeriesPlan with NonLeafLogicalPlan {
+  override def children: Seq[LogicalPlan] = Seq(vectors)
+}
+
+//scalastyle:off
+trait FunctionArgsPlan extends LogicalPlan
+trait ScalarPlan extends LogicalPlan with PeriodicSeriesPlan with FunctionArgsPlan
+
+case class ScalarVaryingDoublePlan(vectors: PeriodicSeriesPlan,
+                      function: ScalarFunctionId,
```
velvia: Line up the params here too?

query/src/main/scala/filodb/query/LogicalPlan.scala
```diff
@@ -144,28 +144,58 @@ case class ScalarVectorBinaryOperation(operator: BinaryOperator,
...
+                      timeStepParams: RangeParams,
+                      functionArgs: Seq[FunctionArgsPlan] = Nil)  extends ScalarPlan with NonLeafLogicalPlan {
   override def children: Seq[LogicalPlan] = Seq(vectors)
 }
 
+
+case class ScalarTimeBasedPlan(function: ScalarFunctionId, rangeParams: RangeParams) extends ScalarPlan  {
```
velvia: You can remove the empty braces `{ }` they are not needed.  Same for others

query/src/main/scala/filodb/query/PlanEnums.scala
```diff
@@ -5,7 +5,7 @@ import enumeratum.{Enum, EnumEntry}
 //scalastyle:off
 sealed abstract class InstantFunctionId(override val entryName: String) extends EnumEntry
 
-object InstantFunctionId extends Enum[InstantFunctionId] {
+  object InstantFunctionId extends Enum[InstantFunctionId] {
```
velvia: Please undo indentation

query/src/main/scala/filodb/query/PlanEnums.scala
```diff
@@ -238,6 +238,41 @@ object SortFunctionId extends Enum[SortFunctionId] {
...
+
+}
+
+sealed abstract class VectorFunctionId(override val entryName: String) extends EnumEntry
+
+object VectorFunctionId extends Enum[VectorFunctionId] {
+  val values = findValues
+
+  case object Vector extends VectorFunctionId("vector")
+
```
velvia: No need for extra line

query/src/main/scala/filodb/query/exec/AggrOverRangeVectors.scala
```diff
@@ -57,6 +57,7 @@ final case class AggregateMapReduce(aggrOp: AggregationOperator,
                                     without: Seq[String],
                                     by: Seq[String]) extends RangeVectorTransformer {
   require(without == Nil || by == Nil, "Cannot specify both without and by clause")
+  override def funcParams: Seq[FuncArgs] = Nil
```
velvia: Shouldn't the default be Nil?  Also please remove the `override` as you are likely just implementing something.  override should be used only if you are truly overriding something

query/src/main/scala/filodb/query/exec/BinaryJoinExec.scala
```diff
@@ -63,6 +63,7 @@ final case class BinaryJoinExec(id: String,
   protected[exec] def compose(childResponses: Observable[(QueryResponse, Int)],
                               firstSchema: Task[ResultSchema],
                               queryConfig: QueryConfig): Observable[RangeVector] = {
+   // childResponses.
```
velvia: Remove empty comment?

query/src/main/scala/filodb/query/exec/ExecPlan.scala
```diff
@@ -39,7 +39,7 @@ final case class ExecResult(rvs: Observable[RangeVector], schema: Task[ResultSch
   * Convention is for all concrete subclasses of ExecPlan to
   * end with 'Exec' for easy identification
   */
-trait ExecPlan extends QueryCommand {
+trait ExecPlan extends QueryCommand  {
```
velvia: No need for extra space here

query/src/main/scala/filodb/query/exec/ExecPlan.scala
```diff
@@ -122,14 +122,20 @@ trait ExecPlan extends QueryCommand {
       } else {
         val finalRes = allTransformers.foldLeft((res.rvs, resSchema)) { (acc, transf) =>
           qLogger.debug(s"queryId: ${id} Setting up Transformer ${transf.getClass.getSimpleName} with ${transf.args}")
-          (transf.apply(acc._1, queryConfig, limit, acc._2), transf.schema(acc._2))
+          val paramRangeVector: Observable[ScalarVector] = if (transf.funcParams.isEmpty){
```
velvia: Space before {

query/src/main/scala/filodb/query/exec/ExecPlan.scala
```diff
@@ -249,6 +273,40 @@ abstract class LeafExecPlan extends ExecPlan {
   final def children: Seq[ExecPlan] = Nil
 }
 
+sealed trait FuncArgs{
```
velvia: Space before {

query/src/main/scala/filodb/query/exec/ExecPlan.scala
```diff
@@ -249,6 +273,40 @@ abstract class LeafExecPlan extends ExecPlan {
...
+  def getResult (implicit sched: Scheduler, timeout: FiniteDuration) : Observable[ScalarVector]
+}
+case class ExecPlanFuncArgs(execPlan: ExecPlan) extends FuncArgs {
```
velvia: final case class, to make sure it's not extendable.  All case classes need to have final

query/src/main/scala/filodb/query/exec/ExecPlan.scala
```diff
@@ -249,6 +273,40 @@ abstract class LeafExecPlan extends ExecPlan {
...
+  override def getResult(implicit sched: Scheduler, timeout: FiniteDuration): Observable[ScalarVector] = {
+    Observable.now(execPlan)
+      .mapAsync(Runtime.getRuntime.availableProcessors()) { plan =>
+        plan.dispatcher.dispatch(plan).onErrorHandle { case ex: Throwable =>
+          qLogger.error(s"queryId: ${execPlan.id} Execution failed for sub-query ${plan.printTree()}", ex)
+          QueryError(execPlan.id, ex)
+        }
+      }.map {
+      case (QueryResult(_, _, result) )=> result.head.asInstanceOf[ScalarVaryingDouble]
```
velvia: Need indentation here

query/src/main/scala/filodb/query/exec/ExecPlan.scala
```diff
@@ -249,6 +273,40 @@ abstract class LeafExecPlan extends ExecPlan {
   final def children: Seq[ExecPlan] = Nil
 }
 
+sealed trait FuncArgs{
+  def getResult (implicit sched: Scheduler, timeout: FiniteDuration) : Observable[ScalarVector]
+}
+case class ExecPlanFuncArgs(execPlan: ExecPlan) extends FuncArgs {
+  override def getResult(implicit sched: Scheduler, timeout: FiniteDuration): Observable[ScalarVector] = {
+    Observable.now(execPlan)
+      .mapAsync(Runtime.getRuntime.availableProcessors()) { plan =>
```
velvia: Instead of using `Observable.now(...).mapAsync` it is easier just to create the task from dispatch, then use `Observable.fromTask(...)`

query/src/main/scala/filodb/query/exec/ExecPlan.scala
```diff
@@ -249,6 +273,40 @@ abstract class LeafExecPlan extends ExecPlan {
...
+      }.map {
+      case (QueryResult(_, _, result) )=> result.head.asInstanceOf[ScalarVaryingDouble]
+      case (QueryError(_, ex)) => throw ex
+    }
+  }
+
+  override def toString: String = execPlan.printTree() + "\n"
+}
+
+case class StaticFuncArgs(scalar: Double, timeStepParams: RangeParams) extends FuncArgs {
```
velvia: final case class

query/src/main/scala/filodb/query/exec/MultiSchemaPartitionsExec.scala
```diff
@@ -86,8 +86,12 @@ final case class MultiSchemaPartitionsExec(id: String,
                                s"chunkMethod=$chunkMethod, filters=$filters, colName=$colName"
 
   // Print inner node's details for debugging
-  override def curNodeText(level: Int): String =
-    s"${super.curNodeText(level)}\n+Inner: ${Option(finalPlan).map(_.curNodeText(level)).getOrElse("None")}"
+  override def curNodeText(level: Int): String = {
+    val innerText= if (Option(finalPlan).isDefined) {
```
velvia: space before =

query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
```diff
@@ -38,11 +38,12 @@ final case class PeriodicSamplesMapper(start: Long,
   protected[exec] def args: String =
     s"start=$start, step=$step, end=$end, window=$window, functionId=$functionId, funcParams=$funcParams"
 
-
+  //scalastyle:off
```
velvia: Again let's not just use off without a specific reason to exclude it.  Try not to even turn off scalastyle, the reason it exists is to enforce good style, which this PR really needs

query/src/main/scala/filodb/query/exec/RangeVectorTransformer.scala
```diff
@@ -55,41 +60,59 @@ object RangeVectorTransformer {
   * range vectors
   */
 final case class InstantVectorFunctionMapper(function: InstantFunctionId,
-                                             funcParams: Seq[Any] = Nil) extends RangeVectorTransformer {
+                                             funcParams: Seq[FuncArgs] = Nil) extends RangeVectorTransformer {
   protected[exec] def args: String =
     s"function=$function, funcParams=$funcParams"
 
-  def apply(source: Observable[RangeVector],
+  def apply (source: Observable[RangeVector],
```
velvia: extra space here

query/src/main/scala/filodb/query/exec/ScalarTimeBasedExec.scala
```diff
@@ -0,0 +1,77 @@
...
+
+import filodb.core.DatasetRef
+import filodb.core.metadata.Column.ColumnType
+import filodb.core.query._
+import filodb.core.store.ChunkSource
+import filodb.query.{BadQueryException, QueryConfig, QueryResponse, QueryResult, ScalarFunctionId}
+import filodb.query.Query.qLogger
+import filodb.query.ScalarFunctionId.{DayOfMonth, DayOfWeek, DaysInMonth, Hour, Minute, Time, Year}
+
+case class ScalarTimeBasedExec(id: String,
```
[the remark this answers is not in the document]
TanviBhavsar: This is only for Time Scalars like DayOfMonth, Time, etc. Should I rename to TimeScalarGeneratorExec?

core/src/main/scala/filodb.core/query/RangeVector.scala
```diff
@@ -93,7 +96,92 @@ trait RangeVector {
...
+    map{ x=> new TransientRow(x._1, x._2)}.iterator
+  def getValue(time: Long): Double = timeValueMap(time)
+
+  override def numRowsInt: Int = timeValueMap.size
+}
 
+final case class RangeParams(start: Long, step: Long, end: Long)
+trait ScalarSingleValue extends ScalarVector {
+  def rangeParams: RangeParams
+  var numRowsInt : Int = 0
```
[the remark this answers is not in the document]
TanviBhavsar: ```numRowsInt```  is from SerializableRangeVector trait which should be implemented by ScalarVector. I increment numRowsInt when creating rows so kept it var
```for (i <- rangeParams.start to rangeParams.end by rangeParams.step) {
      rowList += new TransientRow(i*1000, getValue(i*1000))
      numRowsInt = numRowsInt + 1
    }```

File: query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala (hunk 30 above)
Comment (TanviBhavsar) [reply query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala]: I turned it off because the method is longer than 50 lines and scalastyle fails without it

File: core/src/main/scala/filodb.core/query/RangeVector.scala [hunk 34]
```diff
@@ -1,16 +1,19 @@
 package filodb.core.query
+//scalastyle:off
```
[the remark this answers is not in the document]
TanviBhavsar: Sorry I forgot to remove this

core/src/main/scala/filodb.core/query/RangeVector.scala
TanviBhavsar: I just renamed SerializableRangeVector to SerializedRangeVector and it had java.io.Serializable. We don't need it now? Just want to make sure that it won't break any existing code :)

coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
```diff
@@ -117,7 +118,8 @@ class QueryEngine(dsRef: DatasetRef,
       (tsdbQueryParams.isInstanceOf[PromQlQueryParams] &&
         !tsdbQueryParams.asInstanceOf[PromQlQueryParams].processFailure) || // This is a query that was part of
       // failure routing
-      !QueryRoutingPlanner.hasSingleTimeRange(rootLogicalPlan)) // Sub queries have different time ranges (unusual)
+      !QueryRoutingPlanner.hasSingleTimeRange(rootLogicalPlan) || // Sub queries have different time ranges (unusual)
+       rootLogicalPlan.isInstanceOf[ScalarTimeBasedPlan]) //No need to route since there is no metric
```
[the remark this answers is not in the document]
TanviBhavsar: Good suggestion! Made the change

query/src/main/scala/filodb/query/exec/AggrOverRangeVectors.scala
TanviBhavsar: Yes moved it to case class

query/src/main/scala/filodb/query/exec/ExecPlan.scala
```diff
@@ -249,6 +273,40 @@ abstract class LeafExecPlan extends ExecPlan {
...
 }
 
+sealed trait FuncArgs{
+  def getResult (implicit sched: Scheduler, timeout: FiniteDuration) : Observable[ScalarVector]
+}
+case class ExecPlanFuncArgs(execPlan: ExecPlan) extends FuncArgs {
+  override def getResult(implicit sched: Scheduler, timeout: FiniteDuration): Observable[ScalarVector] = {
+    Observable.now(execPlan)
+      .mapAsync(Runtime.getRuntime.availableProcessors()) { plan =>
+        plan.dispatcher.dispatch(plan).onErrorHandle { case ex: Throwable =>
```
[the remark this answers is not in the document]
TanviBhavsar: Yes, created Observable from dispatch result tasks

prometheus/src/main/scala/filodb/prometheus/ast/Functions.scala
TanviBhavsar: stringParam is for label_replace and label_join function which need strings as input.  Label replace/Join parameters are in quotes that should be removed. I first remove seriesParam, get the instantExpression and finally convert it into a normal string by removing character like quotes. Will add some documentation.

core/src/main/scala/filodb.core/query/RangeVector.scala
TanviBhavsar: Didn't know we can do this! Thanks!

query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
TanviBhavsar: Added a comment for this

query/src/main/scala/filodb/query/exec/RangeVectorTransformer.scala
```diff
@@ -55,41 +60,59 @@ object RangeVectorTransformer {
...
+        case cType: ColumnType =>
+          throw new UnsupportedOperationException(s"Column type $cType is not supported for instant functions")
+      }
+    }
+
+    if (funcParams.isEmpty) {
+      evaluate(source, Nil)
+    } else {
+      funcParams.head match {
+        case s: StaticFuncArgs   => evaluate(source, funcParams.map(x => x.asInstanceOf[StaticFuncArgs]).
```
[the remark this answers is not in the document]
TanviBhavsar: As discussed, we can't do it here

coordinator/src/main/scala/filodb.coordinator/queryengine2/QueryEngine.scala
```diff
@@ -258,8 +267,23 @@ class QueryEngine(dsRef: DatasetRef,
...
-    vectors
+    val funcArg = lp.scalar match {
+      case num: ScalarFixedDoublePlan  => Seq(StaticFuncArgs(num.scalar, num.timeStepParams))
+      case  s: ScalarVaryingDoublePlan => Seq(ExecPlanFuncArgs(generateLocalExecPlan(s, queryId, submitTime, options,
+        spreadProvider)))
+      case  t: ScalarTimeBasedPlan     => Seq(TimeFuncArgs(t.rangeParams))
+      case  _                          =>  throw new UnsupportedOperationException("Invalid logical plan")
+    }
+
+    if (vectors.plans.length > 1 && funcArg.isInstanceOf[ExecPlanFuncArgs]) {
```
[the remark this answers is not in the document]
TanviBhavsar: You are right. We don't need this check. I will remove. Thanks!

prometheus/src/main/scala/filodb/prometheus/query/PrometheusModel.scala
TanviBhavsar: As discussed I made the change because without it Result never has None.
So when there are no rows in result it returns rangevectorkeys without any rows.

TanviBhavsar:
Thanks @vishramachandran and @velvia! I have addressed your review comments

query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
```diff
@@ -38,12 +38,12 @@ final case class PeriodicSamplesMapper(start: Long,
   protected[exec] def args: String =
     s"start=$start, step=$step, end=$end, window=$window, functionId=$functionId"
 
-  //scalastyle:off
+ //scalastyle:off
```
velvia: Could this scalastyle be either made more specific or better yet removed?  Sorry don't remember if it was already there or not

query/src/main/scala/filodb/query/exec/RangeVectorTransformer.scala
```diff
@@ -64,53 +64,56 @@ final case class InstantVectorFunctionMapper(function: InstantFunctionId,
...
     } else {
       funcParams.head match {
         case s: StaticFuncArgs   => evaluate(source, funcParams.map(x => x.asInstanceOf[StaticFuncArgs]).
-          map(x => ScalarFixedDouble(x.timeStepParams, x.scalar)))
+          map(x => ScalarFixedDouble(x.timeStepParams, x.scalar)), queryConfig, limit, sourceSchema)
         case t: TimeFuncArgs     => evaluate(source, funcParams.map(x => x.asInstanceOf[TimeFuncArgs]).
-          map(x => TimeScalar(x.timeStepParams)))
-        case e: ExecPlanFuncArgs => paramResponse.map(param => evaluate(source, Seq(param))).flatten
+          map(x => TimeScalar(x.timeStepParams)), queryConfig, limit, sourceSchema)
+        case e: ExecPlanFuncArgs => paramResponse.map(param => evaluate(source, Seq(param), queryConfig,
```
velvia: Only use map(...) syntax if the closure is really small.  For this one refactor to

     paramResponse.map { param =>
        evalute(.....)
     }

coordinator/src/main/scala/filodb.coordinator/queryengine2/RoutingPlanner.scala
```diff
@@ -139,10 +139,19 @@ object QueryRoutingPlanner extends RoutingPlanner {
     * Check whether logical plan has a PeriodicSeriesPlan
     */
   def isPeriodicSeriesPlan(logicalPlan: LogicalPlan): Boolean = {
-    if (logicalPlan.isInstanceOf[RawSeriesPlan] || logicalPlan.isInstanceOf[MetadataQueryPlan]) {
+    if (!logicalPlan.isRoutable) {
       false
-    } else
-    true
+    } else if (logicalPlan.isInstanceOf[ScalarVectorBinaryOperation]) {
```
velvia: It occurs to me that this entire block would be much more readable as a match {.... }:

    logicalPlan match {
      case p: VectorPlan =>  isPeriodicSeriesPlan(p.scalars)
      case p: ApplyInstantFunction => isPeriodicSeriesPlan(p.vectors)
      ...
    }

See, no need for as/isInstanceOf

coordinator/src/test/scala/filodb.coordinator/queryengine2/QueryEngineSpec.scala
```diff
@@ -576,4 +575,6 @@ class QueryEngineSpec extends FunSpec with Matchers {
     child.params.step shouldEqual step
     child.params.processFailure shouldEqual(false)
   }
+
```
velvia: No need for extra lines

core/src/main/scala/filodb.core/query/RangeVector.scala
```diff
@@ -94,6 +96,123 @@ trait RangeVector {
...
+  */
+trait ScalarRangeVector extends SerializableRangeVector {
+  def key: RangeVectorKey = CustomRangeVectorKey(Map.empty)
+  def getValue(time: Long): Double
+}
+
+/**
+  * ScalarRangeVector which has time specific value
+  */
+case class ScalarVaryingDouble(private val timeValueMap: Map[Long, Double]) extends ScalarRangeVector {
```
velvia: Need final case class here

core/src/main/scala/filodb.core/query/RangeVector.scala
```diff
@@ -100,21 +101,21 @@ trait SerializableRangeVector extends RangeVector {
...
+trait ScalarRangeVector extends SerializableRangeVector {
   def key: RangeVectorKey = CustomRangeVectorKey(Map.empty)
   def getValue(time: Long): Double
 }
 
-case class ScalarVaryingDouble(private val timeValueMap: Map[Long, Double]) extends ScalarVector {
+case class ScalarVaryingDouble(private val timeValueMap: Map[Long, Double]) extends ScalarRangeVector {
   override def rows: Iterator[RowReader] = timeValueMap.toList.sortWith(_._1 < _._1).
-    map{ x=> new TransientRow(x._1, x._2)}.iterator
+                                            map{ x => new TransientRow(x._1, x._2)}.iterator
```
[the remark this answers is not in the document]
velvia: Also need space between map and {

core/src/main/scala/filodb.core/query/RangeVector.scala
```diff
@@ -94,6 +96,123 @@ trait RangeVector {
...
+}
+
+final case class RangeParams(start: Long, step: Long, end: Long)
+
+trait ScalarSingleValue extends ScalarRangeVector {
+  def rangeParams: RangeParams
+  var numRowsInt : Int = 0
+
+  override def rows: Iterator[RowReader] = {
+    Iterator.from(0, rangeParams.step.toInt).takeWhile(_ <= rangeParams.end - rangeParams.start).map { i =>
```
velvia: 👍

core/src/main/scala/filodb.core/query/RangeVector.scala
```diff
@@ -94,6 +96,123 @@ trait RangeVector {
...
+      val t = i + rangeParams.start
+      new TransientRow(t * 1000, getValue(t * 1000))
+    }
+  }
+}
+
+/**
+  * ScalarRangeVector which has one value for all time's
+  */
+case class ScalarFixedDouble(rangeParams: RangeParams, value: Double) extends ScalarSingleValue {
```
velvia: final case class for all these case classes

prometheus/src/main/scala/filodb/prometheus/ast/Functions.scala
```diff
@@ -12,64 +13,100 @@ trait Functions extends Base with Operators with Vectors {
...
+
       throw new IllegalArgumentException(s"Invalid function name [$name]")
     }
 
+    /**
+      *
+      * @return true when function is scalar or time
+      */
+    def isScalarFunction(): Boolean = {
+       name.equalsIgnoreCase("scalar") ||
```
velvia: 1) braces not really needed since it's just an expression
2) can we please make the body lines line up/indented at 2 spaces?

prometheus/src/main/scala/filodb/prometheus/ast/Functions.scala
velvia: ok thanks.  consider putting each filter/map on its own line - will be easier to debug stuff that way (and easier to read too)

query/src/main/scala/filodb/query/LogicalPlan.scala
```diff
@@ -144,28 +152,59 @@ case class ScalarVectorBinaryOperation(operator: BinaryOperator,
...
-                                      functionArgs: Seq[Any] = Nil) extends PeriodicSeriesPlan with NonLeafLogicalPlan {
+                             function: SortFunctionId) extends PeriodicSeriesPlan with NonLeafLogicalPlan {
   override def children: Seq[LogicalPlan] = Seq(vectors)
 }
 
+//scalastyle:off
+trait FunctionArgsPlan extends LogicalPlan
+trait ScalarPlan extends LogicalPlan with PeriodicSeriesPlan with FunctionArgsPlan
+
+case class ScalarVaryingDoublePlan(vectors: PeriodicSeriesPlan,
```
velvia: final case class for all these

query/src/main/scala/filodb/query/LogicalPlan.scala
```diff
@@ -144,28 +152,59 @@ case class ScalarVectorBinaryOperation(operator: BinaryOperator,
...
   * Apply Sort Function to a collection of RangeVectors
   */
 case class ApplySortFunction(vectors: PeriodicSeriesPlan,
-                                      function: SortFunctionId,
-                                      functionArgs: Seq[Any] = Nil) extends PeriodicSeriesPlan with NonLeafLogicalPlan {
+                             function: SortFunctionId) extends PeriodicSeriesPlan with NonLeafLogicalPlan {
   override def children: Seq[LogicalPlan] = Seq(vectors)
 }
 
+//scalastyle:off
```
velvia: Can we please not turn off scalastyle here?  Thx.... don't see any reason why we should need to

query/src/main/scala/filodb/query/exec/HistogramQuantileMapper.scala
```diff
@@ -20,12 +20,13 @@ object HistogramQuantileMapper {
   *
   * @param funcParams Needs one double quantile argument
   */
-case class HistogramQuantileMapper(funcParams: Seq[Any]) extends RangeVectorTransformer {
+case class HistogramQuantileMapper(funcParams: Seq[FuncArgs]) extends RangeVectorTransformer {
```
velvia: final case class

query/src/main/scala/filodb/query/exec/RangeVectorTransformer.scala
```diff
@@ -111,36 +137,44 @@ final case class InstantVectorFunctionMapper(function: InstantFunctionId,
...
                                           instantFunction: HistToDoubleIFunction,
+                                          scalar: Seq[ScalarRangeVector],
                                           result: TransientRow = new TransientRow()) extends Iterator[RowReader] {
   final def hasNext: Boolean = rows.hasNext
   final def next(): RowReader = {
     val next = rows.next()
-    val newValue = instantFunction(next.getHistogram(1))
-    result.setValues(next.getLong(0), newValue)
+    val timestamp = next.getLong(0)
+    val newValue = instantFunction(next.getHistogram(1), scalar.map(_.getValue(timestamp)))
```
velvia: This seems potentially really expensive - how big is the scalar?  map produces a new Seq for every sample....  can you please explain what is going on here?   Since most of the time there is no scalar, I wonder if we can make this a NOP most of the time.

query/src/main/scala/filodb/query/exec/RangeVectorTransformer.scala
```diff
@@ -111,36 +137,44 @@ final case class InstantVectorFunctionMapper(function: InstantFunctionId,
...
+                                           scalar: Seq[ScalarRangeVector],
                                            result: TransientRow = new TransientRow()) extends Iterator[RowReader] {
   final def hasNext: Boolean = rows.hasNext
   final def next(): RowReader = {
     val next = rows.next()
-    val newValue = instantFunction(next.getHistogram(1), next.getDouble(2))
-    result.setValues(next.getLong(0), newValue)
+    val timestamp = next.getLong(0)
+    val newValue = instantFunction(next.getHistogram(1),
+      next.getDouble(2), scalar.map(_.getValue(timestamp)))
```
velvia: Again same thing here ..... could we avoid the `scalar.map`?

query/src/main/scala/filodb/query/exec/RangeVectorTransformer.scala
```diff
@@ -240,4 +283,62 @@ final case class SortFunctionMapper(function: SortFunctionId) extends RangeVecto
...
+final case class VectorFunctionMapper() extends RangeVectorTransformer {
+  protected[exec] def args: String =
+    s"funcParams=$funcParams"
+
+  def apply(source: Observable[RangeVector],
+            queryConfig: QueryConfig,
+            limit: Int,
+            sourceSchema: ResultSchema,
+            paramResponse: Observable[ScalarRangeVector] = Observable.empty): Observable[RangeVector] = {
+    val resultRv = source.toListL.map { rvs =>
```
velvia: This can be simplified significantly.  Instead of using `toListL`, mapping over it, then converting back to an Observable, you can simply do `source.map { rv => new RangeVector{....} }`

Speaking of which, this is basically a NOP, as the new RangeVector just has the same keys and rows as the original.  So why is this transform even needed?

query/src/main/scala/filodb/query/exec/ScalarTimeBasedExec.scala
```diff
@@ -0,0 +1,80 @@
...
+    * Args to use for the ExecPlan for printTree purposes only.
+    * DO NOT change to a val. Increases heap usage.
+    */
+  override protected def args: String = s"params = $params, function = $function"
+
+  override def execute(source: ChunkSource,
+                       queryConfig: QueryConfig)
+                      (implicit sched: Scheduler,
+                       timeout: FiniteDuration): Task[QueryResponse] = {
+    val columns: Seq[ColumnInfo] = Seq(ColumnInfo("timestamp", ColumnType.LongColumn),
```
velvia: We should probably just have a fixed Object constant for this schema since it's used so often.  Instead of allocating a new one every time.

query/src/main/scala/filodb/query/exec/ScalarTimeBasedExec.scala
```diff
@@ -0,0 +1,80 @@
...
+      ColumnInfo("value", ColumnType.DoubleColumn))
+    val recSchema = SerializedRangeVector.toSchema(columns)
+    val resultSchema = ResultSchema(columns, 1)
+    val rangeVectors : Seq[RangeVector] = function match {
+      case Time        => Seq(TimeScalar (params))
```
velvia: Remove extra space

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
```diff
@@ -671,16 +676,17 @@ class QuantileOverTimeChunkedFunctionD(funcParams: Seq[Any]) extends QuantileOve
   }
 }
 
-class QuantileOverTimeChunkedFunctionL(funcParams: Seq[Any])
+class QuantileOverTimeChunkedFunctionL(funcParams: Seq[FuncArgs])
   extends QuantileOverTimeChunkedFunction(funcParams) with ChunkedLongRangeFunction {
   require(funcParams.size == 1, "quantile_over_time function needs a single quantile argument")
+  require(funcParams.head.isInstanceOf[StaticFuncArgs], "quantile parameter must be a number")
   require(funcParams.head.isInstanceOf[Number], "quantile parameter must be a number")
```
velvia: Wait, shouldn't this line be removed?  funcParams.head can't be both `StaticFuncArgs` and a `Number`

query/src/main/scala/filodb/query/exec/rangefn/InstantFunction.scala
```diff
@@ -48,7 +40,7 @@ trait HistToDoubleIFunction extends HistogramInstantFunction {
     * @param value Sample against which the function will be applied
     * @return Calculated value
     */
-  def apply(value: Histogram): Double
+  def apply(value: Histogram, scalarParam: Seq[Double] = Nil): Double
```
velvia: Why is the scalarParam needed here?

velvia: and how come it's always double instead of `FuncArg` etc?

query/src/main/scala/filodb/query/exec/rangefn/InstantFunction.scala
```diff
@@ -82,147 +74,137 @@ object InstantFunction {
...
-    * Validate the function before invoking the function.
-    */
-  require(funcParams.size == 1,
-    "Cannot use ClampMax without providing a upper limit of max.")
-  require(funcParams.head.isInstanceOf[Number],
-    "Cannot use ClampMax without providing a upper limit of max as a Number.")
-
-  override def apply(value: Double): Double =
-    scala.math.min(value, funcParams.head.asInstanceOf[Number].doubleValue())
+case class ClampMaxImpl() extends DoubleInstantFunction {
```
velvia: final case class... here and everywhere

query/src/main/scala/filodb/query/exec/rangefn/InstantFunction.scala
```diff
@@ -82,147 +74,137 @@ object InstantFunction {
...
   */
-case class Log10Impl(funcParams: Seq[Any]) extends EmptyParamsInstantFunction {
-  /**
-    * Validate the function before invoking the function.
-    */
-  require(funcParams.isEmpty, "No additional parameters required for Log10.")
-
-  override def apply(value: Double): Double = scala.math.log10(value)
+case class Log10Impl() extends EmptyParamsInstantFunction {
+
```
velvia: No need for extra line

core/src/main/scala/filodb.core/query/RangeVector.scala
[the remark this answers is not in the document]
TanviBhavsar: I have lined up. This is not the file I changed.  I think you are looking at an older version. Probably some issue with git UI.  Can you check this?
https://github.com/filodb/FiloDB/pull/539/files
Will add the space

query/src/main/scala/filodb/query/LogicalPlan.scala
TanviBhavsar: Without it, I get error ```Number of types declared in the file exceeds 30```
Adding a comment for this

query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
TanviBhavsar: This is also not the latest file. I have added ``` //scalastyle:off
 // Method is longer than 50 lines``

query/src/main/scala/filodb/query/exec/rangefn/InstantFunction.scala
TanviBhavsar: InstantFunction's need Double parameter. Instead of converting FuncArg to double value in every function I have done it in InstantFunctionMapper.  InstantFunctionMapper apply method converts FuncArgs to ScalarRangeVector based on type of FuncArgs. Then ```getValue(timestamp)``` of ScalarRangeVector is used to get the double value which is passed to the InstantFunction apply

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
TanviBhavsar: Ohh yes. I missed this. Thanks for letting me know! We don't use QuantileOverTimeChunkedFunctionL so the test cases were not failing

query/src/main/scala/filodb/query/exec/RangeVectorTransformer.scala
TanviBhavsar: Ohh yes, we don't need toListL. I will remove it.
Input to VectorFunctionMapper is a ScalarRangeVector. VectorFunctionMapper should generate a normal RangeVector

query/src/main/scala/filodb/query/exec/RangeVectorTransformer.scala
TanviBhavsar: Yes that is because the scalar can be an ExecPlan which will have a different value for every timestamp.

query/src/main/scala/filodb/query/exec/ExecPlan.scala
TanviBhavsar: This is also not the file I checked in yesterday. I have made it final.  Git UI is showing me```Outdated```

query/src/main/scala/filodb/query/LogicalPlan.scala
velvia: ok,  can we add a specific disabling just the number of types?  Thx

query/src/main/scala/filodb/query/exec/RangeVectorTransformer.scala
```diff
@@ -111,36 +138,44 @@ final case class InstantVectorFunctionMapper(function: InstantFunctionId,
...
+    val timestamp = next.getLong(0)
+    val newValue = instantFunction(nextVal, scalar.map(_.getValue(timestamp)))
+    result.setValues(timestamp, newValue)
     result
   }
 }
 
 private class H2DoubleInstantFuncIterator(rows: Iterator[RowReader],
                                           instantFunction: HistToDoubleIFunction,
+                                          scalar: Seq[ScalarRangeVector],
```
vishramachandran: Will we ever have more than one ScalarRangeVector when applying instant function ?
Cant we simplify the args ? This would apply to all the InstantFuncIterators

query/src/main/scala/filodb/query/exec/ScalarFixedDoubleExec.scala
```diff
@@ -0,0 +1,75 @@
...
+                                 submitTime: Long = System.currentTimeMillis()) extends LeafExecPlan {
+
+
+  /**
+    * Sub classes should override this method to provide a concrete
+    * implementation of the operation represented by this exec plan
+    * node
+    */
+  override def doExecute(source: ChunkSource, queryConfig: QueryConfig)
+                        (implicit sched: Scheduler, timeout: FiniteDuration): ExecResult = ???
```
vishramachandran: Use this detailed message instead of `???`
`throw new IllegalStateException("doExecute should not be called for ScalarFixedDoubleExec since it represents a readily available static value")`

query/src/main/scala/filodb/query/exec/ScalarFixedDoubleExec.scala
```diff
@@ -0,0 +1,75 @@
...
+    */
+  override protected def args: String = s"params = $params, value = $value"
+
+
+  override def execute(source: ChunkSource,
+                       queryConfig: QueryConfig)
+                      (implicit sched: Scheduler,
+                       timeout: FiniteDuration): Task[QueryResponse] = {
+    val columns: Seq[ColumnInfo] = Seq(ColumnInfo("timestamp", ColumnType.LongColumn),
+      ColumnInfo("value", ColumnType.DoubleColumn))
```
vishramachandran: Place this result schema in a re-usable static context instead of recreating for each invocation.

query/src/main/scala/filodb/query/exec/ScalarTimeBasedExec.scala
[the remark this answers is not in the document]
vishramachandran: Sure - that'd be ok too. I was aiming for a more generic name that that, but TimeScalarGeneratorExec is fine too.

query/src/main/scala/filodb/query/exec/ScalarTimeBasedExec.scala
```diff
@@ -0,0 +1,81 @@
...
+  val columns: Seq[ColumnInfo] = Seq(ColumnInfo("timestamp", ColumnType.LongColumn),
+    ColumnInfo("value", ColumnType.DoubleColumn))
+
+  /**
+    * Sub classes should override this method to provide a concrete
+    * implementation of the operation represented by this exec plan
+    * node
+    */
+  override def doExecute(source: ChunkSource, queryConfig: QueryConfig)
+                                  (implicit sched: Scheduler, timeout: FiniteDuration): ExecResult = ???
```
vishramachandran: better execption. See earlier comment.

query/src/main/scala/filodb/query/exec/rangefn/InstantFunction.scala
```diff
@@ -61,7 +53,7 @@ trait HDToDoubleIFunction extends HistogramInstantFunction {
     * @param value Sample against which the function will be applied
     * @return Calculated value
     */
-  def apply(h: Histogram, d: Double): Double
+  def apply(h: Histogram, d: Double, scalarParam: Seq[Double] = Nil): Double
```
vishramachandran: nitpick: Use plural `scalarParams` here as well as in implementations. I was wondering how there could be one param, but many values...

query/src/main/scala/filodb/query/exec/RangeVectorTransformer.scala
vishramachandran: never mind. It is a generic method that could apply for funcs with more than one arg. Some doc would have helped! It is very easy to forget!

prometheus/src/main/scala/filodb/prometheus/ast/Expressions.scala
```diff
@@ -27,25 +28,42 @@ trait Expressions extends Aggregates with Functions {
...
     override def toPeriodicSeriesPlan(timeParams: TimeRangeParams): PeriodicSeriesPlan = {
       if (lhs.isInstanceOf[ScalarExpression] && rhs.isInstanceOf[ScalarExpression]) {
         throw new UnsupportedOperationException("Binary operations on scalars is not supported yet")
       }
 
-      lhs match {
-        case expression: ScalarExpression if rhs.isInstanceOf[PeriodicSeries] =>
-          val scalar = expression.toScalar
-          val seriesPlan = rhs.asInstanceOf[PeriodicSeries].toPeriodicSeriesPlan(timeParams)
+      (lhs, rhs) match {
```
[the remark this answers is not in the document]
TanviBhavsar: Added default case

PR #542, opened 2019-11-01
feat(core): Flush chunks on hourly boundaries.

broneill:
**Pull Request checklist**

**Current behavior :**
Chunks are flushed by a separate flush command stream, driven by a background task.

**New behavior :**
Chunks are flushed when ingestion time crosses an hourly boundary.


coordinator/src/test/scala/filodb.coordinator/NodeCoordinatorActorSpec.scala
```diff
@@ -265,11 +265,10 @@ class NodeCoordinatorActorSpec extends ActorTest(NodeCoordinatorActorSpec.getNew
 
     it("should aggregate from multiple shards") {
       val ref = setupTimeSeries(2)
-      probe.expectMsgPF() { case CurrentShardSnapshot(ds, mapper) => }
       probe.send(coordinatorActor, IngestRows(ref, 0, records(dataset1, linearMultiSeries().take(30))))
-      probe.expectMsg(Ack(0L))
+      probe.fishForSpecificMessage() {case Ack(0L) => }
```
velvia: Nice.

core/src/main/scala/filodb.core/memstore/MemStore.scala
```diff
@@ -80,8 +79,6 @@ trait MemStore extends ChunkSource {
    * Sets up a shard of a dataset to continuously ingest new sets of records from a stream.
    * The records are immediately available for reads from that shard of the memstore.
    * Errors during ingestion are handled by the errHandler.
-   * Flushes to the ChunkSink are initiated by a potentially independent stream, the flushStream, which emits
-   *   flush events of a specific subgroup of a shard.
    * NOTE: does not check that existing streams are not already writing to this store.  That needs to be
```
velvia: Maybe add a comment about how flushes are done (since old doc had them)?

core/src/main/scala/filodb.core/memstore/TimeSeriesMemStore.scala
```diff
@@ -98,52 +97,36 @@ extends MemStore with StrictLogging {
...
-                                                                flushTimeBucket))
-                    case a: Any => throw new IllegalStateException(s"Unexpected DataOrCommand $a")
-                  }.collect { case Some(flushGroup) => flushGroup }
-                  .mapAsync(numParallelFlushes) { f => shard.createFlushTask(f).executeOn(flushSched).asyncBoundary }
-                           // asyncBoundary so subsequent computations in pipeline happen in default threadpool
-                  .completedL
-                  .doOnCancel(cancelTask)
-                  .runAsync(shard.ingestSched)
+    stream.flatMap {
+      case d: SomeData => {
```
velvia: Don't need the { around the case subblock here

core/src/main/scala/filodb.core/memstore/TimeSeriesMemStore.scala
```diff
@@ -98,52 +97,36 @@ extends MemStore with StrictLogging {
...
+        val tasks = shard.createFlushTasks(d.records)
+        shard.ingest(d)
```
velvia: Let's keep the old three line comment about how write buffers are switched

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -790,7 +809,36 @@ class TimeSeriesShard(val ref: DatasetRef,
     shardStats.purgedPartitions.increment(numDeleted)
   }
 
-  def createFlushTask(flushGroup: FlushGroup): Task[Response] = {
+  /**
+    * Creates zero or more flush tasks (up to the number of flush groups) based on examination
+    * of the record container's ingestion time. This should be called before ingesting the container.
+    */
```
velvia: Hmmm.  The createFlushTasks creates tasks, which are not actually executed until mapAsync.  We should add a comment that the tasks are designed not to be executed right away, instead they are executed after ingest method in a mapAsync.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -790,7 +809,36 @@ class TimeSeriesShard(val ref: DatasetRef,
...
+    val tasks = new ArrayBuffer[Task[Response]]()
+
+    var oldTimestamp = lastIngestionTime
+    val ingestionTime = Math.max(oldTimestamp, container.timestamp) // monotonic clock
+    var newTimestamp = ingestionTime
+
+    if (newTimestamp > oldTimestamp && oldTimestamp != Long.MinValue) {
+      var tsAdjust = 0L
+      for (group <- 0 until numGroups optimized) {
+        if (oldTimestamp / flushBoundaryMillis != newTimestamp / flushBoundaryMillis) {
```
velvia: I guess this is fine, though it would make a little more sense if the flushBoundaryMillis is adjusted per group instead (since it is the group's boundary that varies by group)

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -369,6 +371,21 @@ class TimeSeriesShard(val ref: DatasetRef,
...
 
+  /**
+    * Highest ingestion timestamp observed.
+    */
+  private var lastIngestionTime = Long.MinValue
+
+  // Flush groups when ingestion time is observed to cross a time boundary (typically an hour),
+  // plus a group-specific offset. This simplifies disaster recovery -- chunks can be copied
+  // without concern that they may overlap in time.
+  private val flushBoundaryMillis = storeConfig.flushInterval.toUnit(TimeUnit.MILLISECONDS).toLong
```
velvia: I *think* you can just do `flushInterval.toMillis` or `flushInterval.millis`

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
broneill: The size of the division is one hour (by config), and it must remain constant.

broneill:
I originally had a tsAdjust variable which was being added to each time and I was subtracting from tsAdjust. I accidentally left that variable in place, but now it's gone.

Consider these timestamps: 1:00PM, 2:00PM, 3:00PM. For group zero, the division check make sense. For group 1, it must flush at 1:01PM, 2:01PM, 3:01PM. If the actual times are :00, then nothing happens. If the actual times are :01, then the subtraction kicks in and it divides as 1:00, 2:00, 3:00 and the division check does the rest.

Note that the code would also work fine if the adjustment was added, but the applied offsets might be confusing. As long as the results are consistent, it's okay.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -95,6 +95,15 @@ class TimeSeriesShardStats(dataset: DatasetRef, shardNum: Int) {
...
   val evictedPartIdLookupMultiMatch = Kamon.counter("evicted-partId-lookup-multi-match").refine(tags)
+
+  /**
+    * Difference between the local clock and the received ingestion timestamps, in milliseconds.
+    * If this gauge is negative, then the received timestamps are ahead, and it will stay this
+    * way for a bit, due to the monotonic adjustment. When the gauge value is positive (which is
+    * expected), then the skew reflects the delay between the generation of the samples and
+    * receiving them, assuming that the clocks are in sync.
+    */
+  val ingestionClockSkew = Kamon.gauge("ingestion-clock-skew").refine(tags)
```
velvia: Could we call this `ingestion-clock-delay`?  Typically we think of delay as how far behind real time we are, which is pretty much what this measures.  I think of skew as more like difference in timestamps between different nodes/instances in the same cluster... or I guess between apps.

broneill: done

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -790,7 +817,42 @@ class TimeSeriesShard(val ref: DatasetRef,
...
+        }
+        oldTimestamp -= flushOffsetMillis
+        newTimestamp -= flushOffsetMillis
+      }
+    }
+
+    // Only update stuff if no exception was thrown.
+
+    if (ingestionTime != lastIngestionTime) {
+        lastIngestionTime = ingestionTime
```
vishramachandran: Can we update gauge much less frequently? No need to do for each for each kafka message. Instead once during flushing of group should be sufficient. Multiple updates of gauge within same reporting period is almost useless.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -790,7 +817,42 @@ class TimeSeriesShard(val ref: DatasetRef,
...
+    var newTimestamp = ingestionTime
+
+    if (newTimestamp > oldTimestamp && oldTimestamp != Long.MinValue) {
+      for (group <- 0 until numGroups optimized) {
+        if (oldTimestamp / flushBoundaryMillis != newTimestamp / flushBoundaryMillis) {
+          // Flush out the group before ingesting records for a new hour (by group offset).
+          tasks += createFlushTask(prepareFlushGroup(group))
+        }
+        oldTimestamp -= flushOffsetMillis
+        newTimestamp -= flushOffsetMillis
```
vishramachandran: I find it hard to understand the logic quickly. Conversion to something like this or more docs would help.

```
0.to(numGroups).filter(  <someCondition> ).map { group =>
   createFlushTask(prepareFlushGroup(group))
}
```

vishramachandran:
```for { g <- 0 to numGroups optimized 
        if   <some condition>
} yield {
     createFlushTask(prepareFlushGroup(group))
}```

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
broneill: When a gauge is updated, is a network message sent, or does it just update a number? I assumed that just a number is updated, which is cheap. Anything I do here is to reduce updating a number isn't likely to have much benefit.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
broneill: The subtraction is the confusing part. Changing the style of the code won't eliminate that aspect. The <someCondition> would need to depend on the stateful changing of the timestamps at each iteration.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
vishramachandran: I actually didnt ask for additional check per sample.
If we simply moved the setting of gauge into the createFlushTask method or nearby, it is sufficient.

broneill: The problem is that if I move to createFlushTask, I don't have the two timestamps to compare anymore.

PR #545, opened 2019-11-05
misc(query) : add metric column for Binary Join with Non Arithmetic operators

TanviBhavsar:
**Pull Request checklist**

**Current behavior :** 
Metric column is not part of Binary Join result. Prometheus has it when the operator is non-arithmetic.

**New behavior :**
Add metric column for non-arithmatic operator

### DIFF
--- a/query/src/main/scala/filodb/query/exec/BinaryJoinExec.scala
+++ b/query/src/main/scala/filodb/query/exec/BinaryJoinExec.scala
@@ -51,7 +51,7 @@ final case class BinaryJoinExec(id: String,
 
   val onLabels = on.map(Utf8Str(_)).toSet
   val ignoringLabels = ignoring.map(Utf8Str(_)).toSet
-  val ignoringLabelsForJoin = ignoring.map(Utf8Str(_)).toSet + metricColumn.utf8
+  val ignoringLabelsForJoin = ignoringLabels + metricColumn.utf8
   // if onLabels is non-empty, we are doing matching based on on-label, otherwise we are
   // doing matching based on ignoringLabels even if it is empty
   val onMatching = onLabels.nonEmpty
@@ -119,7 +119,6 @@ final case class BinaryJoinExec(id: String,
     // start from otherSideKey which could be many or one
     var result = otherSideKey.labelValues
     // drop metric name if math operator
-    // TODO use dataset's metricName column name here instead of hard-coding column
     if (binaryOp.isInstanceOf[MathOperator]) result = result - Utf8Str(metricColumn)
 
     if (cardinality == Cardinality.OneToOne) {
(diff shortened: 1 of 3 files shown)


query/src/main/scala/filodb/query/exec/BinaryJoinExec.scala
```diff
@@ -50,7 +50,8 @@ final case class BinaryJoinExec(id: String,
   require(!on.contains(metricColumn), "On cannot contain metric name")
 
   val onLabels = on.map(Utf8Str(_)).toSet
-  val ignoringLabels = ignoring.map(Utf8Str(_)).toSet + metricColumn.utf8
+  val ignoringLabels = ignoring.map(Utf8Str(_)).toSet
+  val ignoringLabelsForJoin = ignoring.map(Utf8Str(_)).toSet + metricColumn.utf8
```
velvia: Can the above be `ignoringLabelsForJoin = ignoringLabels + metricColumn.utf8` ?  Reads better and saves some computation

query/src/test/scala/filodb/query/exec/BinaryJoinExecSpec.scala
```diff
@@ -304,4 +304,30 @@ class BinaryJoinExecSpec extends FunSpec with Matchers with ScalaFutures {
...
+    // scalastyle:off
+    val lhs = QueryResult("someId", null, samplesLhsGrouping.map(rv => SerializableRangeVector(rv, schema)))
+    val rhs = QueryResult("someId", null, samplesRhsGrouping.map(rv => SerializableRangeVector(rv, schema)))
+    // scalastyle:on
+    // note below that order of lhs and rhs is reversed, but index is right. Join should take that into account
+    val result = execPlan.compose(Observable.fromIterable(Seq((rhs, 1), (lhs, 0))), tvSchemaTask, queryConfig)
+      .toListL.runAsync.futureValue
+
+    result.foreach { rv =>
+      rv.key.labelValues.contains("__name__".utf8) shouldEqual true
```
velvia: Would the `__name__` be metric if the schema is like that?   Is that OK?

query/src/main/scala/filodb/query/exec/BinaryJoinExec.scala
TanviBhavsar: Good suggestion! Made the change

query/src/test/scala/filodb/query/exec/BinaryJoinExecSpec.scala
TanviBhavsar: As discussed, changed the test case to have metric

TanviBhavsar:
Thanks @velvia! We remove metric column from result for Math Operator on line 123                          ``` if (binaryOp.isInstanceOf[MathOperator]) result = result - Utf8Str(metricColumn)```
This was already there. The problem was that ignoringLabels always had metricColumn and it got removed from the result for all operators

PR #551, opened 2019-11-15
feat(core): Simplify Cassandra Schema for Index Data Persistence

vishramachandran:
**Pull Request checklist**

Part Key data is stored in cassandra in a time bucketed manner. Part Keys that started/stopped ingesting in that time slice are added into the time bucket. Relevant keys from oldest time bucket that are still valid are rolled over to the latest bucket and the result is persisted at the end. Buckets have fixed retention.

The reason to use the above strategy is to avoid cassandra compaction issues.

**New behavior :**

This PR:
* Vastly simplifies partKey/start+endTime persistence in Cassandra with new simplified schema. We employ a table per shard for efficient download of entire table's contents using token range scans.
* Vastly simplifies partKey persistence logic in TimeSeriesShard. Earlier time bucket management and roll-over of keys is gone with simple flush of keys when necessary. 
* Simplification of index data recovery. No need to maintain partId map at the beginning because they are guaranteed to not repeat. Much less memory is required for bootstrap. Especially needed for long term data retention.

We have done some tests to confirm that a more simplified cassandra schema with partKey, start and end times columns is viable. The scale and churn can be handled by cassandra, and the download times are acceptable.

This change is absolutely necessary for the FiloDB cluster that will handle serving downsampled data. The old schema is operationally very expensive and not viable for long retention times. It is also essential for simple execution of chunk repair across DCs.

**BREAKING CHANGES**

Existing data needs to be wiped, or index data needs to be migrated from old cassandra schema to new cassandra schema.


cassandra/src/main/scala/filodb.cassandra/columnstore/PartitionKeysTable.scala
```diff
@@ -0,0 +1,69 @@
...
+
+  val createCql =
+    s"""CREATE TABLE IF NOT EXISTS $tableString (
+       |    partKey blob,
+       |    startTime bigint,
+       |    endTime bigint,
+       |    PRIMARY KEY (partKey)
+       |) WITH compression = {'sstable_compression': '$sstableCompression'}""".stripMargin
+
+  lazy val readCql = s"SELECT segmentid, segment " +
```
broneill: What is this for?

cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
```diff
@@ -59,42 +60,50 @@ extends ColumnStore with CassandraChunkSource with StrictLogging {
...
-          ixResp    <- indexTable.initialize()
-          pitResp   <- partIndexTable.initialize() } yield pitResp
+    for { ctResp    <- chunkTable.initialize() if ctResp == Success
+          ixResp    <- indexTable.initialize() if ixResp == Success
+          partKeyTablesResp <- partKeyTablesInit.runAsync if partKeyTablesResp.forall( _ == Success)
+        } yield Success
   }
 
-  def truncate(dataset: DatasetRef): Future[Response] = {
+  def truncate(dataset: DatasetRef, numShards: Int): Future[Response] = {
```
broneill: It looks like initialize and truncate really just need a common function with a callback passed to it.

core/src/main/scala/filodb.core/memstore/MemStore.scala
```diff
@@ -24,11 +25,9 @@ final case class ShardAlreadySetup(dataset: DatasetRef, shard: Int) extends
 sealed trait DataOrCommand
 // Typically one RecordContainer is a single Kafka message, a container with multiple BinaryRecords
 final case class SomeData(records: RecordContainer, offset: Long) extends DataOrCommand
-final case class IndexData(timeBucket: Int, segment: Int, records: RecordContainer) extends DataOrCommand
-final case class FlushIndexTimeBuckets(timeBucket: Int)
 
 final case class FlushGroup(shard: Int, groupNum: Int, flushWatermark: Long,
-                            flushTimeBuckets: Option[FlushIndexTimeBuckets])
+                            dirtyPartsToFlush: EWAHCompressedBitmap)
```
broneill: Given our recent observations with compressed bitmaps, why use one here?

cassandra/src/main/scala/filodb.cassandra/columnstore/PartitionKeysTable.scala
vishramachandran: good catch. Remnant from old code. Removable.

cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
vishramachandran: We sometimes use initialize to "create if not exists" in tests. I suggest we leave it as is since it is not used in day-to-day. For prod, we now have a script to initialize tables.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -415,6 +415,30 @@ class TimeSeriesShard(val ref: DatasetRef,
     }
   }
 
+  case class InMemPartitionIterator2(partIds: debox.Buffer[Int]) extends PartitionIterator {
```
broneill: Doc the intent of this class.

cassandra/src/main/scala/filodb.cassandra/columnstore/PartitionKeysTable.scala
```diff
@@ -0,0 +1,66 @@
...
+      connector.execStmtWithRetries(writePartitionCqlNoTtl.bind(
+        toBuffer(pk.partKey), pk.startTime: JLong, pk.endTime: JLong))
+    } else {
+      connector.execStmtWithRetries(writePartitionCql.bind(
+        toBuffer(pk.partKey), pk.startTime: JLong, pk.endTime: JLong, diskTimeToLive: JInt))
+    }
+  }
+
+  def scanPartKeys(tokens: Seq[(String, String)], shard: Int): Observable[PartKeyRecord] = {
+    def cql(start: String, end: String): String =
```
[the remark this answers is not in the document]
vishramachandran: Since usage is within the function, and the def is small enough, I'll leave as is.

PR #556, opened 2019-11-21
feat(prometheus): Use ::col for column selection in PromQL

velvia:
**Pull Request checklist**

To select the non-default column in a query, one uses a label: `__col__="min"` for example.
This is a magic filter which does not work like normal label filters.

**New behavior :**

This change switches the syntax to select a column to a `::col` suffix in either the metric name or the `__name__` label filter value.   
The thought is that this makes the transition easier for some histogram use cases, ie

`rate(my_histogram_count{....}[5m])` => `rate(my_histogram::count{...}[5m])`

**BREAKING CHANGES**

Existing queries that specify `__col__` filter would need to be rewritten.

### DIFF
--- a/README.md
+++ b/README.md
@@ -427,11 +427,11 @@ FiloDB can be queried using the [Prometheus Query Language](https://prometheus.i
 
 ### FiloDB PromQL Extensions
 
-Since FiloDB supports multiple schemas, with possibly more than one value column per schema, there needs to be a way to specify the target column to query.  This is done using the special `__col__` tag filter, like this request which pulls out the "min" column:
+Since FiloDB supports multiple schemas, with possibly more than one value column per schema, there needs to be a way to specify the target column to query.  This is done using a `::columnName` suffix in the metric name, like this request which pulls out the "min" column:
 
-    http_req_timer{_ws_="demo", _ns_="foo",__col__="min"}
+    http_req_timer::min{_ws_="demo", _ns_="foo"}
 
-By default if `__col__` is not specified then the `value-column` option of each data schema is used.
+By default if `::col` suffix is not specified then the `value-column` option of each data schema is used.
 
 Some special functions exist to aid debugging and for other purposes:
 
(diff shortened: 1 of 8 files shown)


prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
```diff
@@ -5,9 +5,6 @@ import filodb.core.query.ColumnFilter
 import filodb.query._
 
 object Vectors {
-  // The "tag" or key used to indicate the column to query to FiloDB.  A non-standard Prom extension.
-  val ColumnSelectorLabel = "__col__"
```
vishramachandran: This is a non-backward compatible change that needs to be called out during release. It will break any existing clients that use `__col__` notation in their PromQL. Needs to be addressed. (or called out)

prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
```diff
@@ -84,41 +81,58 @@ trait Vectors extends Scalars with TimeUnits with Base {
...
+      if (parts.size > 1) (parts(0), Some(parts(1))) else (metricName, None)
+    }
+
+    /**
+     * Converts LabelMatches to ColumnFilters.  Along the way:
+     * - extracts ::col column name expressions in metric names to columns
+     * - removes ::col in __name__ label matches as needed
+     */
+    protected def labelMatchesToFilters(labels: Seq[LabelMatch]) = {
+      var columns: List[String] = Nil
```
vishramachandran: Why is `columns` a list instead of an option ?

prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
```diff
@@ -130,31 +144,27 @@ trait Vectors extends Scalars with TimeUnits with Base {
     * It is possible to filter these time series further by
     * appending a set of labels to match in curly braces ({}).
     */
-
-  case class InstantExpression(override val metricName: Option[String],
-                               override val labelSelection: Seq[LabelMatch],
+  case class InstantExpression(val metricName: Option[String],
```
vishramachandran: I actually like having the `override` marker because it helps reader know that this member implements a super-class/trait def without additional UI help.

prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
```diff
@@ -84,41 +81,58 @@ trait Vectors extends Scalars with TimeUnits with Base {
...
 
-    protected def labelMatchesToFilters(labels: Seq[LabelMatch]) =
-      labels.map { labelMatch =>
-        val labelValue = labelMatch.value.replace("\\\\", "\\")
+    def realMetricName: String = mergeNameToLabels.find(_.label == PromMetricLabel).get.value
+
+    // Returns (trimmedMetricName, column) after stripping ::columnName
+    private def extractStripColumn(metricName: String): (String, Option[String]) = {
+      val parts = metricName.split("::", 2)
+      if (parts.size > 1) (parts(0), Some(parts(1))) else (metricName, None)
```
vishramachandran: Please add this require 

`require(parts(1).nonEmpty, "cannot use empty column name")`

README.md
```diff
@@ -450,11 +450,10 @@ Example of debugging chunk metadata using the CLI:
...
 
 * There is no need to append `_bucket` to the metric name.
-* However, you need to select the histogram column like `__col__="hist"`
-* To compute quantiles:  `histogram_quantile(0.7, sum_over_time(http_req_latency{app="foo",__col__="hist"}[5m]))`
-* To extract a bucket: `histogram_bucket(100.0, http_req_latency{app="foo",__col__="hist"})`
-* Sum over multiple Histogram time series:  `sum(sum_over_time(http_req_latency{app="foo",__col__="hist"}[5m]))` - you could then compute quantile over the sum.
-  - NOTE: Do NOT use `group by (le)` when summing `HistogramColumns`.  This is not appropriate as the "le" tag is not used.  FiloDB knows how to sum multiple histograms together correctly without grouping tricks.
+* To compute quantiles:  `histogram_quantile(0.7, sum(rate(http_req_latency{app="foo"}[5m])))`
+* To extract a bucket: `histogram_bucket(100.0, http_req_latency{app="foo"})`
```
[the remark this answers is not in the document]
velvia: That's not part of this PR  ;-)   We'll change that when we get to that

prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
velvia: I find it confusing.  I'd rather have override there only if it is truly overriding something.  I suppose am open on this though

prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
velvia: Yes, it has been called out in the PR notes.

prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
velvia: It could be an option I suppose.

velvia: Ah, now I remember.  It is a Seq to match the API of all the LogicalPlans which take Seq[String] columns.

velvia:
@vishramachandran addressed feedback

prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
vishramachandran: Can we really select more than one column? Is that even valid/allowed ? This is quite related to the scope of this PR and it would be good to clean this up.

velvia: Currently no.   I can change to an option

PR #557, opened 2019-11-22
feat(query): stddev support

xychen0510:
**Pull Request checklist**

Prometheus stddev function support

### DIFF
--- a/query/src/test/scala/filodb/query/exec/AggrOverRangeVectorsSpec.scala
+++ b/query/src/test/scala/filodb/query/exec/AggrOverRangeVectorsSpec.scala
@@ -130,11 +130,29 @@ class AggrOverRangeVectorsSpec extends RawDataWindowingSpec with ScalaFutures {
     compareIter(result8(0).rows.map(_.getDouble(1)), readyToAggr8.map { v =>
       stdvar(v.map(_.getDouble(1)))
     }.iterator)
+
+    // Stddev
+    val agg9 = RowAggregator(AggregationOperator.Stddev, Nil, tvSchema)
+    val resultObs9a = RangeVectorAggregator.mapReduce(agg9, false, Observable.fromIterable(samples), noGrouping)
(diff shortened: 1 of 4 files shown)


xychen0510:
> @xychen0510 looks fine, but I don't understand how the mean and count are used? The present method seems to just ignore them.

Good point! Similar to stdvar, count and mean are used for updating the new mean, which then be used to calculate the new standard deviation.

xychen0510:
> Looks great! It will be good to have a test case where samples have NaN

Nice suggestions! I added a new NaN test case in the new commit.

query/src/test/scala/filodb/query/exec/AggrOverRangeVectorsSpec.scala
```diff
@@ -130,13 +130,31 @@ class AggrOverRangeVectorsSpec extends RawDataWindowingSpec with ScalaFutures {
...
+
+    // Stddev
+    val agg9 = RowAggregator(AggregationOperator.Stddev, Nil, tvSchema)
+    val resultObs9a = RangeVectorAggregator.mapReduce(agg9, false, Observable.fromIterable(samples), noGrouping)
+    val resultObs9 = RangeVectorAggregator.mapReduce(agg9, true, resultObs9a, rv=>rv.key)
+    val result9 = resultObs9.toListL.runAsync.futureValue
+    result9.size shouldEqual 1
+    result9(0).key shouldEqual noKey
+
+    val readyToAggr9 = samples.toList.map(_.rows.toList).transpose
```
vishramachandran: can we use a debox.Buffer here instead of a List since samples will be boxed. Needed for stdvar and stddev methods as well.

xychen0510: Thanks for the suggestion! I think we should raise a specific PR for this since it covers tests for all aggregation methods.

vishramachandran: This is a spec... sorry - no need for any change. :) I assumed it is prod code.

PR #570, opened 2019-12-11
feat(core): Prometheus Counter and Histogram downsampling

vishramachandran:
**Pull Request checklist**

Capability to downsample Prometheus Counters and Histograms.

Simple strategy:
* Last value of counter or histogram used for every downsample period
* If there is a drop, then the last value before drop and the first value after drop are also included.
* Case of gauges and counters in same schema is handled using downsampler-period-marker which is an abstraction which identifies the row numbers at which downsampling should be done by looking at timestamps of rows or counter values. The period marker is configurable for schemas.

### DIFF
--- a/core/src/main/scala/filodb.core/downsample/DownsamplePeriodMarker.scala
+++ b/core/src/main/scala/filodb.core/downsample/DownsamplePeriodMarker.scala
@@ -1,0 +1,148 @@
+package filodb.core.downsample
+
+import debox.Buffer
+import enumeratum.{Enum, EnumEntry}
+import scalaxy.loops._
+
+import filodb.core.metadata.DataSchema
+import filodb.core.store.{ChunkSetInfoReader, ReadablePartition}
+import filodb.memory.format.PrimitiveVectorReader
+import filodb.memory.format.vectors.{CorrectingDoubleVectorReader, DoubleVectorDataReader}
+
+/**
+  * Enum of supported downsample period marker names
+  * @param entryName name of the marker
+  * @param periodMarkerClass its corresponding marker class used for instance construction
+  */
+sealed abstract class PeriodMarkerName(override val entryName: String, val periodMarkerClass: Class[_])
+  extends EnumEntry
+
+object PeriodMarkerName extends Enum[PeriodMarkerName] {
+  val values = findValues
+  case object Time extends PeriodMarkerName("time", classOf[TimeDownsamplePeriodMarker])
+  case object Counter extends PeriodMarkerName("counter", classOf[CounterDownsamplePeriodMarker])
+}
+
+/**
+  * Implementations of the trait identify row numbers that
+  * demark the boundaries of downsampling periods
+  */
+trait DownsamplePeriodMarker {
+
+  def name: PeriodMarkerName
+
+  /**
+    * Ids of Data Columns the marker works on.
+    * The column id values are fed in via downsampling configuration of the dataset
+    */
+  def inputColId: Int
+
+  /**
+    * Places row numbers for the given chunkset which marks the
+    * periods to downsample into the result buffer param
+    */
+  def getPeriods(part: ReadablePartition,
+                 chunkset: ChunkSetInfoReader,
+                 resMillis: Long,
+                 startRow: Int,
+                 endRow: Int): Buffer[Int]
+}
+
+/**
+  * Time based downsample period marker, which returns row numbers of the last sample for each downsample period
+  * @param inputColId requires the timestamp column 0
+  */
+class TimeDownsamplePeriodMarker(val inputColId: Int) extends DownsamplePeriodMarker {
+  require(inputColId == DataSchema.timestampColID)
+  override def getPeriods(part: ReadablePartition,
+                          chunkset: ChunkSetInfoReader,
+                          resMillis: Long,
+                          startRow: Int,
+                          endRow: Int): Buffer[Int] = {
+    val tsAcc = chunkset.vectorAccessor(DataSchema.timestampColID)
+    val tsPtr = chunkset.vectorAddress(DataSchema.timestampColID)
+    val tsReader = part.chunkReader(DataSchema.timestampColID, tsAcc, tsPtr).asLongReader
+
+    val startTime = tsReader.apply(tsAcc, tsPtr, startRow)
+    val endTime = tsReader.apply(tsAcc, tsPtr, endRow)
+
+    val result = debox.Buffer.empty[Int]
+    // A sample exactly for 5pm downsampled 5-minutely should fall in the period 4:55:00:001pm to 5:00:00:000pm.
+    // Hence subtract - 1 below from chunk startTime to find the first downsample period.
+    // + 1 is needed since the startTime is inclusive. We don't want pStart to be 4:55:00:000;
+    // instead we want 4:55:00:001
+    var pStart = ((startTime - 1) / resMillis) * resMillis + 1
+    var pEnd = pStart + resMillis // end is inclusive
+    // for each downsample period
+    while (pStart <= endTime) {
+      // fix the boundary row numbers for the downsample period by looking up the timestamp column
+      val endRowNum = Math.min(tsReader.ceilingIndex(tsAcc, tsPtr, pEnd), chunkset.numRows - 1)
+      result += endRowNum
+      pStart += resMillis
+      pEnd += resMillis
+    }
+    result
+  }
+  override def name: PeriodMarkerName = PeriodMarkerName.Time
+}
+
+/**
+  * Returns union of the following:
+  * (a) the results from TimeDownsamplePeriodMarker.
+  * (b) row number when counter drops
+  * (c) last row number before counter drops
+  * @param inputColId requires the counter column id
+  */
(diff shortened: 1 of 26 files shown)


core/src/main/scala/filodb.core/downsample/DownsamplePeriodMarker.scala
```diff
@@ -0,0 +1,148 @@
...
+  */
+trait DownsamplePeriodMarker {
+
+  def name: PeriodMarkerName
+
+  /**
+    * Ids of Data Columns the marker works on.
+    * The column id values are fed in via downsampling configuration of the dataset
+    */
+  def inputColId: Int
```
velvia: Comment says this reports Ids (plural) but definition only supports one Id.  Make this return Buffer[Int]?

core/src/main/scala/filodb.core/downsample/DownsamplePeriodMarker.scala
```diff
@@ -0,0 +1,148 @@
...
+
+  /**
+    * Places row numbers for the given chunkset which marks the
+    * periods to downsample into the result buffer param
+    */
+  def getPeriods(part: ReadablePartition,
```
velvia: Very Java-ish name.  How about just `periods`?

core/src/main/scala/filodb.core/downsample/DownsamplePeriodMarker.scala
```diff
@@ -0,0 +1,148 @@
...
+  }
+  override def name: PeriodMarkerName = PeriodMarkerName.Time
+}
+
+/**
+  * Returns union of the following:
+  * (a) the results from TimeDownsamplePeriodMarker.
+  * (b) row number when counter drops
+  * (c) last row number before counter drops
+  * @param inputColId requires the counter column id
```
velvia: Perhaps add a line or two explaining why we need to add the rows before/after drops

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -327,7 +327,7 @@ class TimeSeriesShard(val ref: DatasetRef,
     * TSP.ingesting is MUCH faster than bit.get(i) but we need the bitmap for faster operations
     * for all partitions of shard (like ingesting cardinality counting, rollover of time buckets etc).
     */
-  private[memstore] final val activelyIngesting = new EWAHCompressedBitmap
+  private[memstore] final val activelyIngesting = debox.Set.empty[Int]
```
velvia: How is this related to this PR?

core/src/main/scala/filodb.core/metadata/Schemas.scala
```diff
@@ -78,15 +79,18 @@ object DataSchema {
...
 
     for { dataColumns  <- Column.makeColumnsFromNameTypeList(dataColNameTypes)
           downsamplers <- validateDownsamplers(downsamplerNames, downsampleSchema)
+          periodMarker <- validatedDownsamplerPeriodMarker(dowsamplerPeriodMarker)
           valueColID   <- validateValueColumn(dataColumns, valueColumn)
           _            <- validateTimeSeries(dataColumns, Seq(0)) }
     yield {
-      DataSchema(name, dataColumns, downsamplers, Schemas.genHash(dataColumns), valueColID, downsampleSchema)
+      DataSchema(name, dataColumns, downsamplers, Schemas.genHash(dataColumns),
+        valueColID, downsampleSchema, periodMarker)
```
velvia: Please line up with name (first param)

core/src/main/scala/filodb.core/metadata/Schemas.scala
```diff
@@ -153,8 +158,13 @@ object PartitionSchema {
 
 /**
  * A Schema combines a PartitionSchema with a DataSchema, forming all the columns of a single ingestion record.
+ *
+ * The downsample member is a var because setting it at construction time can potentially result be impossible
+ * when downsample schema is same as raw schema
+ *
+ * Important Note: Serialization will be tricky since there may be loops in object graph. Avoid if possible.
```
velvia: Have to watch for this, I sort of remembered somewhere where Schema needed to be serialized.

core/src/main/scala/filodb.core/metadata/Schemas.scala
```diff
@@ -153,8 +158,13 @@ object PartitionSchema {
...
  */
-final case class Schema(partition: PartitionSchema, data: DataSchema, downsample: Option[Schema]) {
+final case class Schema(partition: PartitionSchema, data: DataSchema, var downsample: Option[Schema] = None) {
```
velvia: Can we please find a way to avoid making downsample a var?  Mutability + case classes is a bad mixture, will lead to initialization bugs and related headaches

memory/src/main/scala/filodb.memory/format/BinaryVector.scala
```diff
@@ -176,6 +176,11 @@ trait CounterVectorReader extends VectorDataReader {
    * @param meta CorrectionMeta with total running correction info.  lastValue is ignored
    */
   def updateCorrection(acc: MemoryReader, vector: BinaryVectorPtr, meta: CorrectionMeta): CorrectionMeta
+
+  /**
+    * Identifies row numbers at which histogram or counter has dropped
+    */
+  def dropPositions(acc2: MemoryReader, vector: BinaryVectorPtr): debox.Buffer[Int]
```
velvia: Can we have the resets function implementation use this method?

scripts/schema-truncate.sh
```diff
@@ -0,0 +1,27 @@
...
+
+cat << EOF
+TRUNCATE ${FILO_ADMIN_KEYSPACE}.checkpoints;
+TRUNCATE ${FILO_KEYSPACE}.${DATASET}_tschunks;
+TRUNCATE ${FILO_KEYSPACE}.${DATASET}_ingestion_time_index;
+EOF
+
+for SHARD in $(seq 0 $((NUM_SHARDS - 1))); do
+
+cat << EOF
```
velvia: Again, this kind of script is nowhere nearly as convenient as using filo-cli for a whole class of users.  I could not even use this script as is.
Please make this a separate PR, then we can do this properly.
In any case, please update README if desire is for users to use this script, and also update filo-cli instructions.  However, it should really be a separate PR.

core/src/main/scala/filodb.core/downsample/DownsamplePeriodMarker.scala
vishramachandran: Only one column in current implementation. I dont want to over-engineer. Easy to change when there is a real use case (multiple counters which can drop at different times for example).

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
vishramachandran: This was a small omission from previous PR that removed usage of EWAH Bitmap.

core/src/main/scala/filodb.core/metadata/Schemas.scala
vishramachandran: ok. I'll watch out during testing.

core/src/main/scala/filodb.core/metadata/Schemas.scala
vishramachandran: Please suggest options. This was the best way out. Immutability and cycles in object graphs is not possible unless I am missing something!

memory/src/main/scala/filodb.memory/format/BinaryVector.scala
vishramachandran: We should eventually use it, outside scope of this PR. Not a simple change because drops across chunks is not covered with this helper.

scripts/schema-truncate.sh
vishramachandran: I added it for convenience since quick capability was missing. We can add other approaches in a different PR.

core/src/main/scala/filodb.core/metadata/Schemas.scala
velvia: I don't think there is any problem with having a cycle in the case class, except for how you use it ie the `toString` method.

memory/src/main/scala/filodb.memory/format/MemoryReader.scala
```diff
@@ -163,7 +164,7 @@ class OnHeapByteBufferAccessor(buf: ByteBuffer) extends MemoryAccessor {
   }
 
   override def wrapInto(dBuf: DirectBuffer, addr: Long, length: Int): Unit = {
-    dBuf.wrap(buf, addr.toInt, length)
+    dBuf.wrap(buf, origBufPos + addr.toInt, length)
```
velvia: Good fix.

core/src/main/scala/filodb.core/metadata/Schemas.scala
vishramachandran: My point is you cant have a cycle with immutable case classes unless you use a var.

vishramachandran:
@velvia Please take another look. The truncate script is currently for dev environments. Since the number of tables have recently increased, the need for this script has become important for dev cycles.

PR #571, opened 2019-12-11
feat(gateway): Convert ILP lines to new gauge, counter, histogram schemas

velvia:
**Pull Request checklist**

In the gateway, Influx Line Protocol input was being converted to all "promCounter" schema, and histograms were being written using old Prom-style individual time series per bucket.

**New behavior :**

Influx Line Protocol inputs are now converted to gauge, counter, and histogram schemas.
summaries are not supported yet.

**BREAKING CHANGES**

Existing Prom-style histogram queries will need to be changed.

### DIFF
--- a/gateway/src/main/scala/filodb/gateway/conversion/InfluxProtocolParser.scala
+++ b/gateway/src/main/scala/filodb/gateway/conversion/InfluxProtocolParser.scala
@@ -8,6 +8,7 @@ import org.jboss.netty.buffer.ChannelBuffer
 import scalaxy.loops._
 
 import filodb.core.binaryrecord2.RecordBuilder
+import filodb.memory.format.UnsafeUtils
 
 trait KVVisitor {
   def apply(bytes: Array[Byte], keyIndex: Int, keyLen: Int, valueIndex: Int, valueLen: Int): Unit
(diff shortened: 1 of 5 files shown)


gateway/src/main/scala/filodb/gateway/conversion/InfluxProtocolParser.scala
```diff
@@ -173,6 +177,15 @@ object InfluxProtocolParser extends StrictLogging {
     }
   }
 
+  // Compares the FIRST key to some bytes
+  def firstKeyEquals(bytes: Array[Byte], offsets: Buffer[Int], compareTo: Array[Byte]): Boolean = {
+    require(offsets.nonEmpty)
+    val firstKeyOffset = keyOffset(offsets(0))
+    val firstKeyLen = valOffset(offsets(0)) - 1 - firstKeyOffset
+    UnsafeUtils.equate(bytes, UnsafeUtils.arayOffset + firstKeyOffset,
```
broneill: Can this be done without UnsafeUtils? Less unsafe is better.

velvia: I don't think there's an equivalent here that is as performant.....

broneill: The buffer has no feature to compare to an array?

velvia: The only buffer is the Netty one, but by this point it's no longer visible, thus would need a whole bunch of API changes.  I'll look into it separately.

PR #572, opened 2019-12-17
add function "last" for supporting last sample in a time window

sherali42:
**Pull Request checklist**

add `last` function for supporting last sample in a time window. This implementation ensures that timestamp of last sample is after start-time and before end-time of the window.

### DIFF
--- a/query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
+++ b/query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
@@ -77,7 +77,6 @@ trait RangeFunction extends BaseRangeFunction {
             window: Window,
             sampleToEmit: TransientRow,
             queryConfig: QueryConfig): Unit
-
 }
 
 /**
@@ -227,10 +226,10 @@ trait ChunkedLongRangeFunction extends TimeRangeFunction[TransientRow] {
 object RangeFunction {
   type RangeFunctionGenerator = () => BaseRangeFunction
 
-
+  // scalastyle:off cyclomatic.complexity
   def downsampleColsFromRangeFunction(dataset: Dataset, f: Option[RangeFunctionId]): Seq[String] = {
     f match {
-      case None | Some(Last)      => Seq("avg")
+      case None                   => Seq("avg")
       case Some(Rate)             => Seq(dataset.options.valueColumn)
       case Some(Irate)            => Seq(dataset.options.valueColumn)
       case Some(Increase)         => Seq(dataset.options.valueColumn)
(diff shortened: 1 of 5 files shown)


sherali42:
> Hey @sherali42 .... you don't need to pass in an extra parameter Window. There are already too many parameters to the `addChunks`. To check the number of samples in the window, you have to query the Timestamp vector reader and check the start and end row numbers. The window is not necessary.
> Also, the ChunkedWindowIterator does not use Windows at all, so I'm not sure how that is working.
> 
> I think this PR can be much, much much smaller. Just use the existing LastSampleRangeFunction which already does the right thing... we don't need to add another custom RangeFunction. It also already takes into account all the mechanics behind NaN and if there are no samples.

@velvia I have updated the code, could you please take a look now.

query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
```diff
@@ -77,6 +77,8 @@ trait RangeFunction extends BaseRangeFunction {
             window: Window,
             sampleToEmit: TransientRow,
             queryConfig: QueryConfig): Unit
+
```
velvia: Remove extra blank lines?

query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
```diff
@@ -394,14 +398,35 @@ object LastSampleFunction extends RangeFunction {
...
+  override def apply(startTimestamp: Long,
+            endTimestamp: Long,
+            window: Window,
+            sampleToEmit: TransientRow,
+            queryConfig: QueryConfig): Unit = {
+    if (window.size > 1)
+      throw new IllegalStateException(s"Window had more than 1 sample. Possible out of order samples. Window: $window")
+    // if no sample found in the window, set the value to NaN
+    if (window.size == 0 || startTimestamp > window.head.getLong(0)) {
```
vishramachandran: `startTimestamp > window.head.getLong(0)` is not needed. Always true.

query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
```diff
@@ -496,6 +521,30 @@ class LastSampleChunkedFunctionD extends LastSampleChunkedFuncDblVal() {
   }
 }
 
+// Return the last sample within the window
+class LastSampleInWindowChunkedFunctionD extends LastSampleChunkedFunctionD {
```
vishramachandran: The abstract class and implementation for double and histogram would be useful. Lets quickly add them and move on.

sherali42: i have added the stack to support double, long and histo types

query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
sherali42: looks like this condition is needed, as we add last sample before start of the window. This is where the logic is to include the last sample outside the window `SlidingWindowIterator.shouldAddCurToWindow`

PR #591, opened 2020-01-09
debug(core): Add very detailed debugString() BR method; add debugging to addPartition

velvia:
**Pull Request checklist**

**New behavior :**

Adds a very detailed `debugString` method to `RecordSchema` with offsets and details on every map key value. 
Adds debugging to addPartition before it creates a partKey.

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
+++ b/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
@@ -1107,6 +1107,7 @@ class TimeSeriesShard(val ref: DatasetRef,
     */
   private def addPartitionForIngestion(recordBase: Any, recordOff: Long, schema: Schema, group: Int) = {
     assertThreadName(IngestSchedName)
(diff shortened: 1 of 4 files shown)


core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -1107,6 +1107,7 @@ class TimeSeriesShard(val ref: DatasetRef,
     */
   private def addPartitionForIngestion(recordBase: Any, recordOff: Long, schema: Schema, group: Int) = {
     assertThreadName(IngestSchedName)
+    logger.debug(s"Adding part key XXX: ${schema.ingestionSchema.debugString(recordBase, recordOff)}")
```
broneill: Maybe a better search string than XXX?

velvia: I suppose we can leave out the XXX, or change it to "details"  :)

PR #593, opened 2020-01-10
bug(core): Fix/change predefined keys hashing algorithm for keys < 4 bytes

velvia:
**Pull Request checklist**

When `RecordBuilder` adds map key values to a BinaryRecord, it makes an assumption that keys are at least 4 bytes long in the `makeKeyKey` hashing code, reading an Int from a byte array.  If the keys are less than 4 bytes long this can result in undefined behavior.

**New behavior :**

The algorithm is changed to the following:
* If the key is <= 8 bytes long, the entire key is converted to a Long and used as a hash.
* If the key is > 8 bytes long, then XXHash 64-bit hash generator is used.

### DIFF
--- a/core/src/main/scala/filodb.core/binaryrecord2/RecordSchema.scala
+++ b/core/src/main/scala/filodb.core/binaryrecord2/RecordSchema.scala
@@ -4,6 +4,7 @@ import scala.collection.mutable.ArrayBuffer
 
 import org.agrona.DirectBuffer
 import org.agrona.concurrent.UnsafeBuffer
+import scalaxy.loops._
 
 import filodb.core.metadata.{Column, Schemas}
 import filodb.core.metadata.Column.ColumnType.{LongColumn, MapColumn, TimestampColumn}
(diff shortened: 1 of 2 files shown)


core/src/main/scala/filodb.core/binaryrecord2/RecordSchema.scala
```diff
@@ -419,22 +419,35 @@ object RecordSchema {
                                                        MapColumn -> 4,
                                                        HistogramColumn -> 4)
 
+  private val tlKeyBuf = new ThreadLocal[Array[Byte]]()
```
broneill: This is a rather odd way if initializing the thread local. There's already a method to do this -- no need for magic unsafe pointer.

core/src/main/scala/filodb.core/binaryrecord2/RecordSchema.scala
```diff
@@ -419,22 +419,35 @@ object RecordSchema {
...
+  private[binaryrecord2] def keyBuf: Array[Byte] = tlKeyBuf.get match {
+    case UnsafeUtils.ZeroPointer => val buf = new Array[Byte](8)
+                                    tlKeyBuf.set(buf)
+                                    buf
+    case b: Array[Byte]          => b
+  }
+
+  // Creates a Long by copying no more than 8 bytes from the bytes array, and zeroing out the rest
+  private def eightBytesToLong(bytes: Array[Byte], index: Int, len: Int): Long = {
+    val buf = keyBuf
```
broneill: I don't think this approach is any faster than a simpler approach (at least probably not that much). Read one byte at a time from the array and shift+xor it into a long. XXHash implementations already support shorter lengths, I thought.

velvia: That's true, let me replace this with the shift-XOR.  No state and probably just as fast.

XXHash does support shorter lengths, but the thought is that an exact match is probably just as fast and better than a hash.  Can discuss in person if you want.

core/src/main/scala/filodb.core/binaryrecord2/RecordSchema.scala
velvia: Did you mean ThreadLocal.withInitial(....)?   That's not easy to use from Scala

velvia:
@broneill pushed changes based on your feedback

PR #596, opened 2020-01-16
bug(query): Throw more precise error if unknown schema ID found

velvia:
**Pull Request checklist**

An NPE is thrown if a query happens to hit a partition/time series with a schema ID that is not known to FiloDB.  In theory this should not be possible, as both ingestion and the index/partkey recovery process skips partitions with unknown schema IDs. 

**New behavior :**

Unknown schemas will result in a new, specific `UnknownSchemaQueryErr`.  Upon seeing this error, operators need to truncate the data tables as this should only occur when an existing schema is changed, which is extremely rare.

### DIFF
--- a/query/src/main/scala/filodb/query/exec/MultiSchemaPartitionsExec.scala
+++ b/query/src/main/scala/filodb/query/exec/MultiSchemaPartitionsExec.scala
@@ -5,6 +5,7 @@ import scala.concurrent.duration.FiniteDuration
 import monix.execution.Scheduler
 
 import filodb.core.DatasetRef
+import filodb.core.metadata.Schemas
 import filodb.core.query.ColumnFilter
 import filodb.core.store._
 import filodb.query.Query.qLogger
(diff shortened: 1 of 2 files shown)


query/src/main/scala/filodb/query/exec/MultiSchemaPartitionsExec.scala
```diff
@@ -43,14 +44,15 @@ final case class MultiSchemaPartitionsExec(id: String,
...
-                           .getOrElse(schemas(
-                             lookupRes.firstSchemaId.getOrElse(schemas.schemas.values.head.schemaHash)))
 
     // If we cannot find a schema, or none is provided, we cannot move ahead with specific SRPE planning
+    // Also warn if unknown schema is found.  This is not supposed to happen since ingestion filters out unknown schemas
     schema.map { s => schemas.schemas(s) }
           .orElse(lookupRes.firstSchemaId.map(schemas.apply))
-          .map { sch =>
+          .filter { sch =>
+            if (sch == Schemas.UnknownSchema) qLogger.warn(s"Unknown schema ID ${lookupRes.firstSchemaId} skipped")
```
tjackpaul: Do we really need this as a warn? Won't it flood the logs for each query when the issue occurs?

velvia: Should be rare.  I can downgrade it to an info though.

velvia: I think we should leave it as WARN.   This only happens when we switch schemas and tables have not been truncated.  The WARNing should help alert users that that is what's going on.

Actually, let's throw a specific error so that operators can figure out what's going on easily.

velvia: ok fixed.

velvia:
@vishramachandran  @tjackpaul  @broneill  this is a super simple PR, please have a look

spark-jobs/src/main/scala/filodb.downsampler/BatchDownsampler.scala
```diff
@@ -118,7 +119,9 @@ object BatchDownsampler extends StrictLogging with Instance {
     try {
       rawPartsBatch.foreach { rawPart =>
         val rawSchemaId = RecordSchema.schemaID(rawPart.partitionKey, UnsafeUtils.arayOffset)
-        val pkPairs = schemas(rawSchemaId).partKeySchema.toStringPairs(rawPart.partitionKey, UnsafeUtils.arayOffset)
+        val schema = schemas(rawSchemaId)
+        if (schema == Schemas.UnknownSchema) throw UnknownSchemaQueryErr(rawSchemaId)
```
vishramachandran: If there is an unknown schema, we should log it and move on. Should not fail entire job with the exception.

velvia: I disagree.  There should never be unknown schema and it is something that needs operator intervention (namely clearing out old keys).  Logging and moving on does not fix the problem, which causes further problems down the line.   At the minimum this needs to be an ERROR in the logs, it really cannot just be ignored.  Also it's not restricted to downsampling, it was discovered during normal querying.

vishramachandran: I agree that operator intervention may be needed, but why should downsampling be stopped ? How would it benefit ? Would rerun produce a different result ? Lets discuss in person.

velvia:
@vishramachandran  comment addressed

PR #613, opened 2020-01-29
feat(query, http): hist_to_prom_vectors func, auto convert histogram HTTP output, ?histogramMap

velvia:
**Pull Request checklist**

Histogram query output is not represented properly in HTTP API output, currently a double value is returned which is meaningless.  There is also no way to convert histograms back to Prometheus-style vectors.

**New behavior :**

A new PromQL function, `hist_to_prom_vectors`, converts histograms to their equivalent Prometheus `_bucket` vectors - one vector per bucket, with `le` tag added.

HTTP query routes automatically use this function to convert histograms to Prometheus style vectors by default.  Remote read does the same thing as well.

A new query HTTP API param, `histogramMap`, if set to true, allows outputting histogram buckets as a JSON object, instead of converting to Prometheus bucket vectors.  This allows for programmatic processing of the entire histogram easily.

A bug is fixed in instant vector output, where `value` member is populated instead of `values`.


prometheus/src/main/scala/filodb/prometheus/query/PrometheusModel.scala
```diff
@@ -88,15 +106,42 @@ object PrometheusModel {
...
+        Result(tags,
+          // remove NaN in HTTP results
+          // Known Issue: Until we support NA in our vectors, we may not be able to return NaN as an end-of-time-series
+          // in HTTP raw query results.
+          if (samples.isEmpty) None else Some(samples),
+          None
+        )
+      case QueryResultType.InstantVector =>
+        Result(tags, None, samples.headOption)
+      case QueryResultType.Scalar => ???
```
vishramachandran: Forgot to replace ??? with impl ?

query/src/main/scala/filodb/query/exec/RangeVectorTransformer.scala
```diff
@@ -388,3 +391,96 @@ final case class AbsentFunctionMapper(columnFilter: Seq[ColumnFilter], rangePara
...
+  protected[exec] def args: String = s"HistToPromSeriesMapper(options=${sch.options})"
+
+  def funcParams: Seq[FuncArgs] = Nil
+  def apply(source: Observable[RangeVector],
+            queryConfig: QueryConfig,
+            limit: Int,
+            sourceSchema: ResultSchema,
+            paramResponse: Seq[Observable[ScalarRangeVector]]): Observable[RangeVector] = {
+    // Passthrough if RV is not histogram
+    if (valueColumnType(sourceSchema) != ColumnType.HistogramColumn) return source
```
vishramachandran: Shouldnt we let the user know to not apply the conversion if not needed ? I would prefer to error out.

query/src/main/scala/filodb/query/exec/PromQlExec.scala
```diff
@@ -66,7 +64,7 @@ case class PromQlExec(id: String,
         override def key: RangeVectorKey = CustomRangeVectorKey(r.metric.map (m => m._1.utf8 -> m._2.utf8))
 
         override def rows: Iterator[RowReader] = {
-          samples.iterator.map { v =>
+          samples.iterator.collect { case v: Sampl =>
```
vishramachandran: More thought needed here.

PromQLExec should invoke with histogramMap = true and then detect histogram types in output to facilitate cross-cluster query composition.

http/src/test/scala/filodb/http/PrometheusApiRouteSpec.scala
```diff
@@ -114,19 +122,68 @@ class PrometheusApiRouteSpec extends FunSpec with ScalatestRouteTest with AsyncT
...
+      resp.data.result.foreach { res =>
+        res.value shouldEqual None
+        res.values.get.length should be > (1)
+        res.metric.contains("le") shouldEqual true
+        res.metric("_metric_").endsWith("_bucket") shouldEqual true
+      }
+    }
+  }
+
+  it("should output native Histogram maps if histogramMap=true") {
```
vishramachandran: Can we add another unit test to ensure that the data points are returned in the correct format:

Today, each sample in json response is represented as 
```
{
"timestamp": 1581099990,
"value": 45.293414
}
```

Some change broke it.

It should have been
```
[
1581099990,
"45.293414"
]
```

prometheus/src/main/scala/filodb/prometheus/query/PrometheusModel.scala
```diff
@@ -2,12 +2,26 @@ package filodb.prometheus.query
...
 import remote.RemoteStorage._
 
-import filodb.core.query.{ColumnFilter, Filter, RangeVector}
-import filodb.query.{Data, ErrorResponse, ExplainPlanResponse, IntervalSelector, LogicalPlan, QueryResultType,
-  RawSeries, Result, Sampl, SuccessResponse}
-import filodb.query.exec.ExecPlan
+import filodb.core.metadata.Column.ColumnType
+import filodb.core.metadata.PartitionSchema
+import filodb.core.query.{ColumnFilter, ColumnInfo, Filter, RangeVector}
+import filodb.query.{QueryResult => FQR, _}
```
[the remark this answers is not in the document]
velvia: QueryResult is unfortunately used in Protobuf also.  Let me think of a better name.

prometheus/src/main/scala/filodb/prometheus/query/PrometheusModel.scala
velvia: There was never an implementation for Scalar, will add a TODO.

http/src/test/scala/filodb/http/PrometheusApiRouteSpec.scala
velvia: Yes, this PR should fix this, but it really depends on the encoders implicitly added in Circe, and it is really tricky to read in code and debug.

velvia: @vishramachandran hey I think it was due to an import removed by IntelliJ by mistake  ;-)
  // DO NOT REMOVE PromCirceSupport import below assuming it is unused - Intellij removes it in auto-imports :( .

query/src/main/scala/filodb/query/exec/RangeVectorTransformer.scala
velvia: Ok I removed it.  It's not needed anymore with refactor in http module.

query/src/main/scala/filodb/query/exec/PromQlExec.scala
velvia: I'm adding a TODO with notes.  Not easy as the current output has no provision for schema, and unless we add that creation of RangeVectors would be difficult.

velvia:
@vishramachandran  have another look, CR feedback should be addressed

PR #617, opened 2020-01-31
feat(query): LongTimeRangePlanner to stich raw/downsample query results

vishramachandran:
**Pull Request checklist**

LongTimeRangePlanner class is now implemented, and unit tested.


coordinator/src/main/scala/filodb.coordinator/queryplanner/LongTimeRangePlanner.scala
```diff
@@ -1,22 +1,50 @@
...
+                           downsampleClusterPlanner: QueryPlanner,
+                           rawDataRetentionMillis: Long,
+                           stitchDispatcher: => PlanDispatcher) extends QueryPlanner {
 
   def materialize(logicalPlan: LogicalPlan, qContext: QueryContext): ExecPlan = {
-    ???
-  }
+    logicalPlan match {
+      case p: PeriodicSeriesPlan =>
+        val lastRawTime = System.currentTimeMillis() - rawDataRetentionMillis
```
broneill: Using the wall clock here seems very very wrong. I would assume that the stitching would operate by finding the stitch point from the actual data, not by an externally provided "retention" timestamp.

vishramachandran: Good point - this first implementation solves the real-time ingested data. Moreover we downsample real-time data only - i.e. data where ingestion and user time is approx the same. 

We have to deal with use cases with historical data ingested separately. As much as I would like to solve this use case right away, it is a bigger discussion that deserves separate treatment and I don't want to include that in the scope of this PR.

coordinator/src/main/scala/filodb.coordinator/queryplanner/CompositePlanner.scala
```diff
@@ -17,17 +17,19 @@ import filodb.query.exec._
 class CompositePlanner(dsRef: DatasetRef,
                        schemas: Schemas,
                        shardMapperFunc: => ShardMapper,
-                       //                  downsampleMapperFunc: => ShardMapper,
+                       downsampleMapperFunc: => ShardMapper,
                        failureProvider: FailureProvider,
+                       earliestRawTimestamp: => Long,
```
velvia: How do we obtain this?

vishramachandran: Typically, `System.currentTimeMillis - rawDataRetentionMillis`

PR #630, opened 2020-02-12
feat(cassandra): Add method to copy chunks to a different C* column store.

broneill:
**Pull Request checklist**

This is the first step for supporting a spark job which can copy chunks (and the ingestion time index) to another C* cluster or keyspace. This will make it possible to perform disaster recovery and backfilling.

### DIFF
--- a/cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
+++ b/cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
@@ -1,9 +1,11 @@
 package filodb.cassandra.columnstore
 
 import java.net.InetSocketAddress
+import java.nio.ByteBuffer
 import java.util.concurrent.TimeUnit
 
-import scala.concurrent.{ExecutionContext, Future}
+import scala.collection.mutable.ArrayBuffer
+import scala.concurrent.{Await, ExecutionContext, Future}
 import scala.concurrent.duration._
 
 import com.datastax.driver.core.{ConsistencyLevel, Metadata, TokenRange}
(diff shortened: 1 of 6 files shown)


cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
```diff
@@ -215,6 +217,79 @@ extends ColumnStore with CassandraChunkSource with StrictLogging {
...
+      futures.clear()
+    }
+
+    var lastPartition: ByteBuffer = null
+
+    for (split <- splits) {
+      val tokens = split.asInstanceOf[CassandraTokenRangeSplit].tokens
+      val rows = sourceIndexTable.scanRowsByIngestionTime(tokens, ingestionTimeStart, ingestionTimeEnd)
+      for (row <- rows) {
+        val partition = row.getBytes(0) // partition
```
vishramachandran: I am not so much liking the leak in abstraction. CassColStore is needing to know about which columns are selected by CQL issued in index table class. In this case, it is ok to pay cost of case class instead of using "Row".

core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
```diff
@@ -185,23 +187,25 @@ object ChunkSetInfo extends StrictLogging {
...
+
+  def getChunkID(info: ByteBuffer): ChunkID = getChunkID(fromByteBuffer(info), 0)
+
+  def getIngestionTime(info: ByteBuffer): Long = getIngestionTime(fromByteBuffer(info), 0)
+
+  def getStartTime(info: ByteBuffer): Long = startTimeFromChunkID(getChunkID(info))
+
+  def getEndTime(info: ByteBuffer): Long = getEndTime(fromByteBuffer(info), 0)
+
+  def getNumRows(info: Array[Byte]): Int = getNumRows(fromArray(info), 0)
```
vishramachandran: We have to double check impact of object creation here. This was the reason I did not remove the original API in my on-heap vectors PR.

cassandra/src/main/scala/filodb.cassandra/columnstore/IngestionTimeIndexTable.scala
```diff
@@ -41,37 +36,40 @@ sealed class IngestionTimeIndexTable(val dataset: DatasetRef, val connector: Fil
...
+    * partition:      ByteBuffer
+    * ingestion_time: long
+    * start_time:     long
+    * info:           ByteBuffer
+    *
+    * The ChunkSetInfo object contains methods for extracting fields from the info column.
+    */
+  def scanRowsByIngestionTime(tokens: Seq[(String, String)],
+                              ingestionTimeStart: Long,
+                              ingestionTimeEnd: Long): Iterator[Row] = {
```
vishramachandran: I dont see the benefit of using Iterator here. The memory utilization for the entire method is not async or lazy. 

Instead, it may be simpler to materialize to a `List` which will enable simpler `groupBy` semantics to the consumer of this API.

cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
```diff
@@ -215,6 +217,79 @@ extends ColumnStore with CassandraChunkSource with StrictLogging {
...
+
+        if (!partition.equals(lastPartition)) {
```
vishramachandran: The idiomatic way of doing this would be to use takeWhile or groupBy which are not straightforward with Iterators. But I have pointed out elsewhere in this PR that using iterator here is not beneficial at all since we are synchronous and not lazy. We may as well use List.

cassandra/src/main/scala/filodb.cassandra/columnstore/IngestionTimeIndexTable.scala
```diff
@@ -41,37 +36,40 @@ sealed class IngestionTimeIndexTable(val dataset: DatasetRef, val connector: Fil
...
-    Observable.fromIterator(it).handleObservableErrors
-  }
-
-  def getMultiInfos(partitions: Seq[Array[Byte]]): Observable[InfoRecord] = {
-    val query = inPartReadCql.bind().setList(0, partitions.map(toBuffer).asJava, classOf[ByteBuffer])
-    val it = session.execute(query).asScala.toIterator.map(fromRow)
-    Observable.fromIterator(it).handleObservableErrors
+    * Test method which returns all rows for a partition.
+    */
+  def readAllRows(partKeyBytes: ByteBuffer): ResultSet = {
```
vishramachandran: Should explicitly call out APIs that are not async and not usable in real-time server.
If used, they can potentially block ingestion or query threads and slow down the server.

cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
broneill: Not much of leak really. All the code in this package is altogether a "unit" and nothing leaks out. If the table classes were in the same file as the CassandraColumnStore it would look like less of a "leak", but not really. I can add an extra wrapper around the row (no need to copy, since it's just a wrapper), but it would just map method names to numbers.

core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
broneill: The code here is highly inlinable. The wrapper objects would just turn into stack-allocated local variables.

cassandra/src/main/scala/filodb.cassandra/columnstore/IngestionTimeIndexTable.scala
broneill: Well, the number of rows received might be unbounded. An iterator makes more sense. Does C* driver read all rows from a query into memory up front, or does it do the right thing and stream as it goes?

cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
broneill: Again, need to understand how the driver works before switching to a List.

cassandra/src/main/scala/filodb.cassandra/columnstore/IngestionTimeIndexTable.scala
broneill: Good point.

cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
broneill: I should be able to declare these methods such that access is restricted just to CassandraColumnStore.

cassandra/src/main/scala/filodb.cassandra/columnstore/IngestionTimeIndexTable.scala
vishramachandran: Agreed. ResultSet is indeed lazy

broneill:
> Would also help if you added "sync" in method name so users dont overlook documentation.

Added a "NoAsync" suffix. Using "Sync" implies something much stronger.

PR #647, opened 2020-02-23
BUG(query) : handle invalid filodb-promql-functions syntax in parser

nikitag55:
BUG(query) : filodb-promql-functions like avg_over_time/holt_winters etc correctly gets executed when additional arbitrary string parameters are provided instead of throwing parser error like prom-promql functions.

After the input query functions are tokenized by the lexer, correctly validate the syntax and throw similar error like prom-promql in case of invalid syntax before the parsed query is being translated into their logical plan.

Above fix addresses below issues:

1.) Random additional strings passed along with mandatory parameters in the query functions now throw proper syntax error before a query enters the step of getting converted to their logical plan.

Example 1:

Correct syntax : `avg_over_time(some_metric[5m])`
Current Behavior: `avg_over_time(some_metric[5m], hello)` -> Gets correctly executed.
New Behavior : `avg_over_time(some_metric[5m], hello)` ->Expected only 1 arg, got 2

Example 2 :

Correct syntax : `holt_winters(some_metric[5m], 0.5, 0.5)`
Current Behavior:  `holt_winters(some_metric[5m], 0.5, 0.5, hello)` ->  Gets correctly executed.
New Behavior :  `holt_winters(some_metric[5m], 0.5, 0.5, hello)` -> Expected 3 args, got 4

2.) Earlier the parameters to the functions can be passed in any order UNLIKE prom-promql. Now, functions will accept parameters in correct order only.

Example 1 :

Correct syntax : Correct syntax :` holt_winters(some_metric[5m], 0.5, 0.5)`
Current Behavior:  `holt_winters(0.1, 0.1, some_metric[5m])`  ->  Gets correctly executed.
New Behavior :  `holt_winters(0.1, 0.1, some_metric[5m])`  -> Expected range-vector, got scalar

Example 2:

Correct syntax :  `quantile_over_time(1, some_metric[5m])`
Current Behavior:  `quantile_over_time(some_metric[5m], 1)` ->  Gets correctly executed.
New Behavior :  `quantile_over_time(some_metric[5m], 1)`  -> E xpected scalar, got range_vector


prometheus/src/main/scala/filodb/prometheus/ast/Functions.scala
```diff
@@ -21,6 +21,160 @@ trait Functions extends Base with Operators with Vectors {
...
+                throw new IllegalArgumentException(s"Invalid Trend factor. Expected: 0 < tf < 1, got: ${allParams(2).asInstanceOf[ScalarExpression].toScalar}")
+            }
+            else {
+              throw new IllegalArgumentException(s"${errExpectedScalar} ${RangeFunctionId.HoltWinters.entryName}, got ${allParams.head.getClass.getSimpleName}")
+            }
+
+          } else {
+            throw new IllegalArgumentException(s"${errExpected3Argument} ${RangeFunctionId.HoltWinters.entryName}, got ${allParams.size}")
+          }
+
```
whizkido: Move line 66 - 97 in its own function.

prometheus/src/main/scala/filodb/prometheus/ast/Functions.scala
```diff
@@ -21,6 +21,160 @@ trait Functions extends Base with Operators with Vectors {
...
+              throw new IllegalArgumentException(s"${errExpectedRangeVector} ${RangeFunctionId.PredictLinear.entryName}, got ${allParams.head.getClass.getSimpleName}")
+
+            // 2nd param needs to be a scalar
+            if (!allParams(1).isInstanceOf[ScalarExpression])
+              throw new IllegalArgumentException(s"${errExpectedScalar} ${RangeFunctionId.PredictLinear.entryName}, got ${allParams.head.getClass.getSimpleName}")
+
+          } else {
+            throw new IllegalArgumentException(s"${errExpected2Argument} ${RangeFunctionId.PredictLinear.entryName}, got ${allParams.size}")
+          }
+
```
whizkido: Move to function

prometheus/src/main/scala/filodb/prometheus/ast/Functions.scala
```diff
@@ -21,6 +21,160 @@ trait Functions extends Base with Operators with Vectors {
...
+              throw new IllegalArgumentException(s"${errExpectedScalar} ${RangeFunctionId.QuantileOverTime.entryName}, got ${allParams.head.getClass.getSimpleName}")
+
+            // 2nd param should be a range-vector
+            if (!allParams(1).isInstanceOf[RangeExpression])
+              throw new IllegalArgumentException(s"${errExpectedRangeVector} ${RangeFunctionId.QuantileOverTime.entryName}, got ${allParams.head.getClass.getSimpleName}")
+
+          } else {
+            throw new IllegalArgumentException(s"${errExpected2Argument} ${RangeFunctionId.QuantileOverTime.entryName}, got ${allParams.size}")
+          }
+
```
whizkido: Move to function

prometheus/src/main/scala/filodb/prometheus/ast/Functions.scala
nikitag55: done

prometheus/src/main/scala/filodb/prometheus/ast/Functions.scala
nikitag55: done

prometheus/src/main/scala/filodb/prometheus/ast/Functions.scala
```diff
@@ -21,6 +21,199 @@ trait Functions extends Base with Operators with Vectors {
       throw new IllegalArgumentException(s"Invalid function name [$name]")
     }
 
+    // Below code is for validating the syntax of promql functions belonging to RangeFunctionID.
+    // It takes care of validating syntax of the tokenized input query before creating the logical plan.
+    // In case of invalid params/invalid syntax, then throw exceptions with similar error-messages like promql.
+
+    // error messages
+    val errExpectedRangeVector : String = "Expected type range_vector in call to function "
```
velvia: `: String` is not needed, leave out

prometheus/src/main/scala/filodb/prometheus/ast/Functions.scala
```diff
@@ -21,6 +21,199 @@ trait Functions extends Base with Operators with Vectors {
...
+    val errExpected1Argument : String = "Expected 1 argument(s) in call to function "
+    val errExpected2Argument : String = "Expected 2 argument(s) in call to function "
+    val errExpected3Argument : String = "Expected 3 argument(s) in call to function "
+
+    /**
+     *
+     * Below function validates if the first argument to all the functions in RangeFunctionId which needs only one argument as range-expression
+     * is correct or not. If not, then throw Illegal Argument Exception with the proper error message.
+     */
+    def onlyRangeExprParamValidator(funcName : String) : Unit = {
```
velvia: Our style is to not have spaces in front of the colon, so please leave them out everywhere

prometheus/src/main/scala/filodb/prometheus/ast/Functions.scala
```diff
@@ -21,6 +21,199 @@ trait Functions extends Base with Operators with Vectors {
...
+      // needs to have 1 parameter
+      if (allParams.size == 1) {
+        // 1st Param should be a range-vector
+        if (!allParams.head.isInstanceOf[RangeExpression])
+          throw new IllegalArgumentException(s"${errExpectedRangeVector} ${funcName}, got ${allParams.head.getClass.getSimpleName}")
```
velvia: It's not necessary to use braces when the interpolated is just one variable.  For example: $funcName

prometheus/src/main/scala/filodb/prometheus/ast/Functions.scala
```diff
@@ -21,6 +21,199 @@ trait Functions extends Base with Operators with Vectors {
...
+
+        // 1st Param should be a range-vector
+        if (!allParams.head.isInstanceOf[RangeExpression])
+          throw new IllegalArgumentException(s"${errExpectedRangeVector} ${funcName}, got ${allParams.head.getClass.getSimpleName}")
+
+        // 2nd param needs to be a scalar
+        if (allParams(1).isInstanceOf[ScalarExpression]) {
+
+          // if it's a scalar, smoothing value needs to be between greater than 0 and less than 1
+          if(!(allParams(1).asInstanceOf[ScalarExpression].toScalar > 0 && allParams(1).asInstanceOf[ScalarExpression].toScalar < 1))
```
velvia: Code like this can be made more readable by factoring out `allParams(1).asInstanceOf[ScalarExpression].toScalar` to a val just above this.

vishramachandran:
@nikitag55 
Thanks for the nice effort to add the validation. Much appreciated!

I apologize for not chiming in on this topic earlier. I had to recall the validation patterns we were using and I just remembered them now. 

We do arg validation in the constructor of the ExecPlan which happens at query materialization time. For example see:

```
final case class HistogramQuantileMapper(funcParams: Seq[FuncArgs]) extends RangeVectorTransformer {

  import HistogramQuantileMapper._
  require(funcParams.size == 1, "histogram_quantile function needs a single quantile argument")
  require(funcParams.head.isInstanceOf[StaticFuncArgs], "Dynamic arg not supported yet")
```

However, this validation is missing in the code for some classes such as `InstantVectorFunctionMapper`. If we add them there it will be taken care of.

The above approach of adding require in the ExecPlans and RangeVectorTransformer case classes will fail the query execution at plan materialization time.

Adding them at the AST level has a couple of issues:

1. When we support new syntaxes or query languages, we have to repeat the validation there. Better to place the validation into the physical plan objects' constructors. Parsers can just be used for parsing. Arg and type validation is better be done at a place which knows more about the operation.

2. Not sure if your PR addresses more complex expressions where a func param is a complex expression in itself like `clamp_max(scalar(avg_over_time(foo{...}[5m])), bar{...})`

Let's discuss if you have further ideas or thoughts. Thanks.

nikitag55:
Thanks @vishramachandran @velvia . I will work on the comments/thoughts.

nikitag55:
> @nikitag55 Adding validation to query parsing is definitely a great idea and much needed! Thanks for the PR. Besides a few styling comments which can be incorporated, I think the main thing to chew on is about how to implement this so there is less repetition and more code reuse.
> 
> Do you see how right now there are various validators which must check everything, including the number of params and do validation on each one? There are many repeated elements, such as checking something is a range/instant expression, checking scalar ranges. If a new function is written, much code would need to be duplicated in its current form, and the table with function name to validator mappings must also be updated.
> 
> Here is my suggestion for how to make this well structured:
> 
> I would expand the RangeFunctionId trait to have a validationSpec param/method. The purpose of this param is a description of what kind of validation is needed, it should be good enough to fit most cases. Here is how it might look like:
> 
> ```scala
> sealed trait ParamSpec
> case object RangeVectorParam extends ParamSpec    // Param must be a range vector expression
> case object ScalarParamAny extends ParamSpec         // Param must be any Scalar expression
> final case class ScalarParamRange(min: Double, max: Double) extends ParamSpec   // Param must be a constant scalar in between min and max
> 
> sealed abstract class RangeFunctionId(val entryName: String, val paramSpecs: Seq[ParamSpec]) extends EnumEntry
> ```
> 
> Now, instead of a giant match with every function name, we can simply parse the RangeFunctionId's paramSpecs. It can check that the number of expected parameters is correct, and also check the type of each parameter, and even a range. This makes the specification of a function very clean and easy, and the checking code much less repetitive.
> 
> Would also seek @vishramachandran 's input on this as well.
> 
> Thanks again!

@velvia Hello Evan, Thanks for the review. Initially, I also thought to put it in `RangeFunctionId` trait. The reason why I didn't put up there is : I found that for other queries for example like simple Aggregates, this type of validation is being done in `Aggregates.scala`. For eg:    

```
 if (secondParamNeeded && allParams.size < 2)
      throw new IllegalArgumentException("2 parameters required for count_values, quantile, topk and bottomk")

    if (allParams.size == 2) {
      if (!secondParamNeeded) {
        throw new IllegalArgumentException("parameter is only required for count_values, quantile, topk and bottomk")
      }
      allParams.head match {
        case num: ScalarExpression =>
          parameter = Seq(num.toScalar)
        case s: InstantExpression =>
          parameter = Seq(s.metricName.get)
        case _ =>
          throw new IllegalArgumentException("First parameter to aggregate operator can be a string or number")
      }
    }
```

That's why I have my validations for functions in `Functions.scala`. I also totally agree to do this in `RangeFunctionId`. Let me know what you think still. I think the change is required for `Aggregates.scala` also.

velvia:
@nikitag55 yes you might need to change aggregates also.  I think that was done as a quick thing, but as we have more validation need to do it in a more structured place.  
That said, feel free to change Aggregates.scala in a separate PR if you'd like to limit the scope.

nikitag55:
Addressed all the requested changes.

nikitag55:
Awaiting final review @velvia @vishramachandran

prometheus/src/main/scala/filodb/prometheus/ast/Functions.scala
```diff
@@ -21,6 +21,70 @@ trait Functions extends Base with Operators with Vectors {
...
+              // If the obj is not Scalar Expression, then throw exception.
+              if (!paramObj.isInstanceOf[ScalarExpression])
+                throw new IllegalArgumentException(s"$errExpectedScalar $funcName, got ${allParams(index).getClass.getSimpleName}")
+
+              else {
+                val paramValue = paramObj.asInstanceOf[ScalarExpression].toScalar
+                if (!(paramValue > min && paramValue < max)) {
+                  // if index == 1 ; then it's a smoothing factor
+                  if (index == 1)
+                    throw new IllegalArgumentException(s"Invalid Smoothing factor. Expected: 0 < sf < 1, got: $paramValue")
```
velvia: This might not be true for other functions.  I would parameterize this;  put the expected error message in the `ScalarRangeParam` itself, so that it doesn't have to be in this Functions.scala.  Again, leave all the control to the functions.

prometheus/src/test/scala/filodb/prometheus/parse/ParserSpec.scala
```diff
@@ -186,7 +186,8 @@ class ParserSpec extends FunSpec with Matchers {
     parseSuccessfully("topk(5, some_metric)")
     parseSuccessfully("count_values(\"value\",some_metric)")
     parseSuccessfully("sum without(and, by, avg, count, alert, annotations)(some_metric)")
-    parseSuccessfully("sum_over_time(foo)")
+    // parseSuccessfully("sum_over_time(foo)")
```
velvia: You can remove the old case

prometheus/src/main/scala/filodb/prometheus/ast/Functions.scala
```diff
@@ -21,6 +21,70 @@ trait Functions extends Base with Operators with Vectors {
...
+    val functionId = RangeFunctionId.withNameLowercaseOnlyOption(name.toLowerCase)
+
+    if (functionId.nonEmpty) {
+      val funcName = functionId.get.entryName
+      // get the parameter spec of the function from RangeFunctionID
+      val paramSpec = functionId.get.paramSpec
+
+      // if the length of the args in param spec is NOT similar to the args in the i/p query, then the i/p query is INCORRECT
+      // throw invalid no. of args exception.
+      if(paramSpec.length != allParams.length)
```
velvia: Space before (

query/src/main/scala/filodb/query/PlanEnums.scala
```diff
@@ -53,50 +53,60 @@ object InstantFunctionId extends Enum[InstantFunctionId] {
   // TODO time, vector, scalar
 }
 
-sealed abstract class RangeFunctionId(override val entryName: String) extends EnumEntry
+sealed trait ParamSpec
+
+case object RangeVectorParam extends ParamSpec
```
velvia: I would suggest: add some docs, at least to the base ParamSpec trait, so we know these are used for validation.  Also, no need for blank lines between each case/variant.

prometheus/src/main/scala/filodb/prometheus/ast/Functions.scala
nikitag55: done

prometheus/src/main/scala/filodb/prometheus/ast/Functions.scala
nikitag55: Parameterized error messages. Error messages are now in RangeVectorParam, InstantVectorParam, ScalarParam and ScalarRangeParamto make everything consistent.

query/src/main/scala/filodb/query/PlanEnums.scala
nikitag55: > Also, no need for blank lines between each case/variant.

I added blank lines after every cases, because in the same file for other abstract class like InstantFunctionID and more, all the cases had blank lines. 

I have fixed this blank line issue in this whole file.

nikitag55:
@velvia Thanks for the review. I have addressed all the comments in the latest commit.

query/src/main/scala/filodb/query/PlanEnums.scala
```diff
@@ -9,104 +9,66 @@ object InstantFunctionId extends Enum[InstantFunctionId] {
...
-
-  case object SumOverTime extends RangeFunctionId("sum_over_time", Seq(RangeVectorParam))
-
-  case object Timestamp extends RangeFunctionId("timestamp", Seq(InstantVectorParam))
+  case object AvgOverTime extends RangeFunctionId("avg_over_time", Seq(RangeVectorParam()))
+  case object Changes extends RangeFunctionId("changes", Seq(RangeVectorParam()))
+  case object CountOverTime extends RangeFunctionId("count_over_time", Seq(RangeVectorParam()))
+  case object Delta extends RangeFunctionId("delta", Seq(RangeVectorParam()))
+  case object Deriv extends RangeFunctionId("deriv", Seq(RangeVectorParam()))
+  case object HoltWinters extends RangeFunctionId("holt_winters", Seq(RangeVectorParam(), ScalarRangeParam(0, 1, "Invalid Smoothing factor. Expected: 0 < sf < 1, got:"), ScalarRangeParam(0, 1, "Invalid Trend factor. Expected: 0 < tf < 1, got:")))
```
velvia: This line might be too long.... be sure to run ScalaStyle

vishramachandran:
@nikitag55 Can you resolve conflicts in the PR?

nikitag55:
@vishramachandran Done!

PR #659, opened 2020-02-27
BUG fix for (sum/avg/stddev)_over_time & zscore promql functions

nikitag55:
This commit addresses below mentioned issues:

1.) sum_over_time()

If the sequence contains all NaNs, current implementation is giving 0 as result. Instead, it should give NaN as result.

2.) avg_over_time()

If the sequence contains all NaNs, current implementation is giving 0 as result. Instead, it should give NaN as result.

3.) stddev_over_time()

On debugging above functions 1 & 2:

sum_over_time() function will calculate the sum of only non-NaN values and ignore all the NaN values which is CORRECT.

Similarly, avg_over_time() function is also computing the average as  sum of non-NaN values / count of non-NaN values in the sequence which is CORRECT

For stddev_over_time() computation, we need sum, squared sum, avg and count of the non-NaN values in the sequence.

Current code CORRECTLY calculates the  sum, squared sum by making use of non-NaN values.

But it has following two issues:

1.) count in this function is counting ALL the values irrespective of whether they are NaN or not NaN -> this will cause a problem when sequence is a combination of both NaN and not NaN values.

2.) while loop is used to iterate in the sequence, but the variable used to traverse the entire sequence which does work like i+=1, is done ONLY when value is not NaN -> this will cause a problem when sequence is a combination of both NaN and not NaN values because in order to traverse the ENTIRE sequence, we have to this  i+=1 always, irrespective of whether it is NaN or not NaN.

4.) z_score():

Because z_score() is based on avg_over_time() & stdev_over_time(), it is also giving WRONG results when sequence contains both NaN and not NaN values or only NaNs.

It works FINE when there are all not NaNs in the sequence.

Other improvements :

1.)  For unit testing, added more and more test-data sequences for sum/avg/stddev_over_time() & zscore which are combination of NaN & not NaN’s, empty sequence, or only NaN’s sequence.

2.) For unit testing, added zscore simple math function in the test file to see if the result given by promql zscore is correct or not.

3.) For unit testing, there are 3 functions in the test file : avg, squared sums & stdvar.
Limitation of these functions is : If the sequence contains combination of NaN and not NaN values, the computed result will be NaN only.
So, these functions are not good for the sequences which contains NaNs in it.

So in the test file, I have added 4 other functions for computing sum, squared sums, avg, stdvar, count ; which works well for the test-cases when sequences have NaNs wherein NaN is just ignored to compute any sort of result and have used the same to test sum_over_time(), avg_over_time(), stddev_over_time() with variety of test-cases.

4.) Earlier, for ZScoreChunkedFunction implementation an abstract class was also created. Removed it and made use of existing VarOverTimeChunkedFunctionD to achieve the result.

5.) Earlier, for Zscore implementation, LastSampleChunkedFunctionWithNanD was also added in RangeFunctions. Removed it and made use of existing VarOverTimeChunkedFunctionD to achieve the result.

**Pull Request checklist**

**New behavior :**

**BREAKING CHANGES**

Breaking changes may include:

**Other information**:

### DIFF
--- a/query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
+++ b/query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
@@ -155,11 +155,11 @@ class SumOverTimeChunkedFunctionD extends SumOverTimeChunkedFunction() with Chun
                                 doubleReader: bv.DoubleVectorDataReader,
                                 startRowNum: Int,
                                 endRowNum: Int): Unit = {
-    // NaN values are ignored by default in the sum method
-    if (sum.isNaN) {
-      sum = 0d
-    }
-    sum += doubleReader.sum(doubleVectAcc, doubleVect, startRowNum, endRowNum)
+
+    //Takes care of computing sum in all seq types : combination of NaN & not NaN & all numbers.
+    val chunkSum = doubleReader.sum(doubleVectAcc, doubleVect, startRowNum, endRowNum)
+    if(!JLDouble.isNaN(chunkSum) && JLDouble.isNaN(sum)) sum = 0d
+    sum += chunkSum
   }
 }
 
@@ -408,15 +408,15 @@ class AvgOverTimeFunction(var sum: Double = Double.NaN, var count: Int = 0) exte
   }
 }
 
-abstract class AvgOverTimeChunkedFunction(var sum: Double = Double.NaN, var count: Double = 0)
+abstract class AvgOverTimeChunkedFunction(var sum: Double = Double.NaN, var count: Int = 0)
   extends ChunkedRangeFunction[TransientRow] {
   override final def reset(): Unit = {
-    sum = Double.NaN;
-    count = 0d
+    sum = Double.NaN
+    count = 0
   }
 
   final def apply(endTimestamp: Long, sampleToEmit: TransientRow): Unit = {
-    sampleToEmit.setValues(endTimestamp, if (count > 0) sum/count else if (sum.isNaN()) sum else 0d)
+    sampleToEmit.setValues(endTimestamp, if (count > 0) sum / count else if (sum.isNaN()) sum else 0d)
   }
 }
 
@@ -426,10 +426,11 @@ class AvgOverTimeChunkedFunctionD extends AvgOverTimeChunkedFunction() with Chun
                                 doubleReader: bv.DoubleVectorDataReader,
                                 startRowNum: Int,
                                 endRowNum: Int): Unit = {
-    if (sum.isNaN) {
-      sum = 0d
-    }
-    sum += doubleReader.sum(doubleVectAcc, doubleVect, startRowNum, endRowNum)
+
+    // Takes care of computing sum in all seq types : combination of NaN & not NaN & all numbers.
+    val chunkSum = doubleReader.sum(doubleVectAcc, doubleVect, startRowNum, endRowNum)
+    if(!JLDouble.isNaN(chunkSum) && JLDouble.isNaN(sum)) sum = 0d
+    sum += chunkSum
     count += doubleReader.count(doubleVectAcc, doubleVect, startRowNum, endRowNum)
   }
 }
@@ -493,45 +494,64 @@ class StdVarOverTimeFunction(var sum: Double = 0d,
   }
 }
 
-abstract class VarOverTimeChunkedFunctionD(var sum: Double = 0d,
+abstract class VarOverTimeChunkedFunctionD(var sum: Double = Double.NaN,
                                            var count: Int = 0,
-                                           var squaredSum: Double = 0d) extends ChunkedDoubleRangeFunction {
-  override final def reset(): Unit = { sum = 0d; count = 0; squaredSum = 0d }
+                                           var squaredSum: Double = Double.NaN,
+                                           var lastSample: Double = Double.NaN) extends ChunkedDoubleRangeFunction {
+  override final def reset(): Unit = { sum = Double.NaN; count = 0; squaredSum = Double.NaN }
   final def addTimeDoubleChunks(doubleVectAcc: MemoryReader,
                                 doubleVect: BinaryVector.BinaryVectorPtr,
                                 doubleReader: bv.DoubleVectorDataReader,
                                 startRowNum: Int,
                                 endRowNum: Int): Unit = {
+    // Takes care of computing sum in all seq types : combination of NaN & not NaN & all numbers.
     val it = doubleReader.iterate(doubleVectAcc, doubleVect, startRowNum)
-    var _sum = 0d
-    var _sqSum = 0d
     var elemNo = startRowNum
+    var chunkSum = Double.NaN
+    var chunkSquaredSum = Double.NaN
+    var chunkCount = 0
     while (elemNo <= endRowNum) {
       val nextValue = it.next
       if (!JLDouble.isNaN(nextValue)) {
-        _sum += nextValue
-        _sqSum += nextValue * nextValue
-        elemNo += 1
+        if (chunkSum.isNaN()) chunkSum = 0d
+        if (chunkSquaredSum.isNaN()) chunkSquaredSum = 0d
+        if (elemNo == endRowNum) lastSample = nextValue
+        chunkSum += nextValue
+        chunkSquaredSum += nextValue * nextValue
+        chunkCount +=1
       }
+      elemNo += 1
     }
-    count += (endRowNum - startRowNum + 1)
-    sum += _sum
-    squaredSum += _sqSum
+    if(!JLDouble.isNaN(chunkSum) && JLDouble.isNaN(sum)) sum = 0d
+    sum += chunkSum
+    if(!JLDouble.isNaN(chunkSquaredSum) && JLDouble.isNaN(squaredSum)) squaredSum = 0d
+    squaredSum += chunkSquaredSum
+    count += chunkCount
   }
 }
 
 class StdDevOverTimeChunkedFunctionD extends VarOverTimeChunkedFunctionD() {
   final def apply(endTimestamp: Long, sampleToEmit: TransientRow): Unit = {
-    val avg = if (count > 0) sum/count else 0d
-    val stdDev = Math.sqrt(squaredSum/count - avg*avg)
+    var stdDev = Double.NaN
+    if (count > 0) {
+      val avg = sum / count
+      stdDev = Math.sqrt((squaredSum / count) - (avg * avg))
+    }
+    else if (sum.isNaN()) stdDev = sum
+    else stdDev = 0d
     sampleToEmit.setValues(endTimestamp, stdDev)
   }
 }
 
 class StdVarOverTimeChunkedFunctionD extends VarOverTimeChunkedFunctionD() {
   final def apply(endTimestamp: Long, sampleToEmit: TransientRow): Unit = {
-    val avg = if (count > 0) sum/count else 0d
-    val stdVar = squaredSum/count - avg*avg
+    var stdVar = Double.NaN
+    if (count > 0) {
+      val avg = sum / count
+      stdVar = (squaredSum / count) - (avg * avg)
+    }
+    else if (sum.isNaN()) stdVar = sum
+    else stdVar = 0d
     sampleToEmit.setValues(endTimestamp, stdVar)
   }
 }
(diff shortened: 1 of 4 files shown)


query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
```diff
@@ -155,11 +155,16 @@ class SumOverTimeChunkedFunctionD extends SumOverTimeChunkedFunction() with Chun
                                 doubleReader: bv.DoubleVectorDataReader,
                                 startRowNum: Int,
                                 endRowNum: Int): Unit = {
-    // NaN values are ignored by default in the sum method
-    if (sum.isNaN) {
-      sum = 0d
+    val it = doubleReader.iterate(doubleVectAcc, doubleVect, startRowNum)
```
TanviBhavsar: Removing doubleReader.sum and iterating will make it slower. Sum function is optimized based on type of data like for DeltaDeltaConstDataReader, DeltaDeltaIterator.
You can do something like:
```
val sumVector = doubleReader.sum(doubleVectAcc, doubleVect, startRowNum, endRowNum)
 if (!sumVector.isNaN && sum.isNaN)
   sum = 0d
 sum += sumVector
```

nikitag55: Correct. 

But above snippet will work correctly only if `sum` value inside `doubleReader.sum` is initialized with `Double.NaN` as `sum` both `doubleReader.sum` and `SumOverTimeChunkedFunctionD` is initialized with 0. Not sure, if it is set to 0d for some reason or not on both the places.

So, for PR thought to at least present the idea/bug with the changes done. I also think to make all the required changes in `doubleReader.sum` only.

TanviBhavsar: Yes as discussed initial sum should be NaN in all implementations of DoubleVectorDataReader.sum. We cannot make all changes in doubleReader.sum as we need to store sum of all chunks in SumOverTimeChunkedFunction. A window can have multiple chunks, so addTimeLongChunks can be called multiple times. We need to update SumOverTimeChunkedFunction sum for every chunk

velvia: As suggested, please move this fix into DoubleReader.sum itself -- the variant that reads from double vectors. (`DoubleVectorDataReader`)  This way if the vector is integral, we use the faster sum which doesn't need to test for NaN.

query/src/test/scala/filodb/query/exec/rangefn/AggrOverTimeFunctionsSpec.scala
```diff
@@ -146,20 +222,39 @@ class AggrOverTimeFunctionsSpec extends RawDataWindowingSpec {
   val numIterations = 10
 
   it("should correctly aggregate sum_over_time using both chunked and sliding iterators") {
-    val data = (1 to 240).map(_.toDouble)
```
velvia: I see that new data part was added to existing test case.  Can we please separate out the new data to a new test case?  Something about checking for NaN's.

query/src/test/scala/filodb/query/exec/rangefn/AggrOverTimeFunctionsSpec.scala
```diff
@@ -234,64 +329,101 @@ class AggrOverTimeFunctionsSpec extends RawDataWindowingSpec {
   }
 
   it("should aggregate count_over_time and avg_over_time using both chunked and sliding iterators") {
-    val data = (1 to 500).map(_.toDouble)
+    val test_data1 = Seq(
```
velvia: Here too

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
```diff
@@ -408,15 +407,15 @@ class AvgOverTimeFunction(var sum: Double = Double.NaN, var count: Int = 0) exte
...
   override final def reset(): Unit = {
-    sum = Double.NaN;
-    count = 0d
+    sum = Double.NaN
+    count = 0
   }
 
   final def apply(endTimestamp: Long, sampleToEmit: TransientRow): Unit = {
-    sampleToEmit.setValues(endTimestamp, if (count > 0) sum/count else if (sum.isNaN()) sum else 0d)
+    sampleToEmit.setValues(endTimestamp, sum / count)
```
velvia: I think it's still safer to have the divide by zero check here.  I see that in reset() the count is still 0, so there is still this possibility.

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
```diff
@@ -493,44 +492,44 @@ class StdVarOverTimeFunction(var sum: Double = 0d,
...
-    count += (endRowNum - startRowNum + 1)
-    sum += _sum
-    squaredSum += _sqSum
   }
 }
 
 class StdDevOverTimeChunkedFunctionD extends VarOverTimeChunkedFunctionD() {
   final def apply(endTimestamp: Long, sampleToEmit: TransientRow): Unit = {
-    val avg = if (count > 0) sum/count else 0d
+    val avg = sum/count
```
velvia: Here too, I think retaining divide by 0 is good idea

query/src/test/scala/filodb/query/exec/rangefn/AggrOverTimeFunctionsSpec.scala
```diff
@@ -173,8 +263,8 @@ class AggrOverTimeFunctionsSpec extends RawDataWindowingSpec {
       val chunkedIt = chunkedWindowItHist(data, rv, new SumOverTimeChunkedFunctionH(), windowSize, step)
       chunkedIt.zip(data.sliding(windowSize, step).map(_.drop(1))).foreach { case (aggRow, rawDataWindow) =>
         val aggHist = aggRow.getHistogram(1)
-        val sumRawHist = rawDataWindow.map(_(3).asInstanceOf[bv.MutableHistogram])
-                                      .foldLeft(emptyAggHist) { case (agg, h) => agg.add(h); agg }
+        val sumRawHist = rawDataWindow.map(_ (3).asInstanceOf[bv.MutableHistogram])
```
velvia: space?

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
nikitag55: yes, but Double.NaN/0 will always be double.NaN. Also, it's not possible for sum to be some value and count to be 0.

nikitag55: also, when count in reset() is 0, sum is Double.NaN. So, Double.NaN/0 = Double.NaN.

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
nikitag55: Same reason as above.

query/src/test/scala/filodb/query/exec/rangefn/AggrOverTimeFunctionsSpec.scala
nikitag55: oops. Fixed it.

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
nikitag55: Still you want the change?

query/src/test/scala/filodb/query/exec/rangefn/AggrOverTimeFunctionsSpec.scala
```diff
@@ -388,21 +479,21 @@ class AggrOverTimeFunctionsSpec extends RawDataWindowingSpec {
 
       data.sliding(windowSize, step).map(_.length - 2).toBuffer.zip(aggregated2).foreach {
         case (val1, val2) => if (val1 == -1) {
-                                val2.isNaN shouldEqual (true) // window does not have any element so changes will be NaN
-                              } else {
-                                val1 shouldEqual (val2)
-                              }
+          val2.isNaN shouldEqual (true) // window does not have any element so changes will be NaN
```
TanviBhavsar: Restore indentation

query/src/test/scala/filodb/query/exec/rangefn/AggrOverTimeFunctionsSpec.scala
```diff
@@ -516,32 +607,65 @@ class AggrOverTimeFunctionsSpec extends RawDataWindowingSpec {
...
 
-    var rv3 = timeValueRV(data3)
-    val chunkedIt3 = new ChunkedWindowIteratorD(rv3, 160000, 100000, 180000, 100000,
-      new ZScoreChunkedFunctionD(), queryConfig)
-    val aggregated3 = chunkedIt3.map(_.getDouble(1)).toBuffer
-    aggregated3(0).isNaN shouldBe true
+    (0 until numIterations).foreach { x =>
+      val windowSize = rand.nextInt(100) + 10
+      val step = rand.nextInt(50) + 5
+      info(s"  iteration $x  windowSize=$windowSize step=$step")
```
TanviBhavsar: Extra space

query/src/test/scala/filodb/query/exec/rangefn/AggrOverTimeFunctionsSpec.scala
```diff
@@ -194,11 +284,11 @@ class AggrOverTimeFunctionsSpec extends RawDataWindowingSpec {
       chunkedIt.zip(data.sliding(windowSize, step).map(_.drop(1))).foreach { case (aggRow, rawDataWindow) =>
         val aggHist = aggRow.getHistogram(1)
         val sumRawHist = rawDataWindow.map(_(4).asInstanceOf[bv.MutableHistogram])
-                                      .foldLeft(emptyAggHist) { case (agg, h) => agg.add(h); agg }
+          .foldLeft(emptyAggHist) { case (agg, h) => agg.add(h); agg }
         aggHist shouldEqual sumRawHist
 
         val maxMax = rawDataWindow.map(_(3).asInstanceOf[Double])
-                                  .foldLeft(0.0) { case (agg, m) => Math.max(agg, m) }
+          .foldLeft(0.0) { case (agg, m) => Math.max(agg, m) }
```
TanviBhavsar: Restore indentation

query/src/test/scala/filodb/query/exec/rangefn/AggrOverTimeFunctionsSpec.scala
nikitag55: done

query/src/test/scala/filodb/query/exec/rangefn/AggrOverTimeFunctionsSpec.scala
nikitag55: Actually this has come from develop code : https://github.com/filodb/FiloDB/blob/606a2f301a9265d893e76ac031ec506fe3f8f75a/query/src/test/scala/filodb/query/exec/rangefn/AggrOverTimeFunctionsSpec.scala#L378

I have fixed everywhere though.

query/src/test/scala/filodb/query/exec/rangefn/AggrOverTimeFunctionsSpec.scala
nikitag55: done

TanviBhavsar:
@nikitag55 You have changed changed sum += doubleReader.sum to sum = doubleReader.sum
A window can have multiple chunks. So sum, count, avg and stdDev would only have the last chunk result.  Refer my previous comment for details.

nikitag55:
> @nikitag55 You have changed changed sum += doubleReader.sum to sum = doubleReader.sum
> A window can have multiple chunks. So sum, count, avg and stdDev would only have the last chunk result. Refer my previous comment for details.

@TanviBhavsar Earlier I couldn't make the test cases such that window comprises of multiple chunks. I figured it out and observed how `addTimeDoubleChunks` will be called several times when there are several chunks in the window. Made changes as required to address such behavior. 


Please review.

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
vishramachandran: We have a TODO to represent no-value using a special NaN instead of regular NaN. We have to check if operations on that NaN result in the same NaN. Given this, it may be better to retain the explicit check.

nikitag55: Added the check @vishramachandran.

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
nikitag55: Added the check @velvia . Another issue I found is that in stddev, stdvar and zscore, that the count check is done just for computing avg and not for computing stddev & stdvar which is in the newline above the line you pointed here. The fix I added takes care of all now. Thanks for pointing out.

nikitag55:
Addressed all the changes. Please review @velvia @vishramachandran @TanviBhavsar

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
velvia: We can't rely on NaN.  It is possible, if it is an Integer encoded column, to end up with a sum of 0 and count of 0.    Anyways will have a look.

PR #663, opened 2020-03-04
feat(spark-jobs): Define a basic Spark job which copies chunks between Cassandra stores.

broneill:
**Pull Request checklist**

**Current behavior :** 
New feature.

### DIFF
--- a/spark-jobs/src/main/scala/filodb/repair/ChunkCopier.scala
+++ b/spark-jobs/src/main/scala/filodb/repair/ChunkCopier.scala
@@ -1,0 +1,135 @@
+package filodb.repair
+
+import java.io.File
+import java.time.Instant
+import java.time.format.DateTimeFormatter
+
+import com.typesafe.config.ConfigFactory
+import com.typesafe.scalalogging.StrictLogging
+import monix.execution.Scheduler
+import org.apache.spark.SparkConf
+import org.apache.spark.sql.SparkSession
+
+import filodb.cassandra.FiloSessionProvider
+import filodb.cassandra.columnstore.CassandraColumnStore
+import filodb.core.{DatasetRef, GlobalConfig}
+import filodb.core.store.ScanSplit
+
+/**
+  * Contains all the objects necessary for performing worker tasks. Is constructed from SparkConf
+  * and is cached.
+  */
+class ChunkCopier(conf: SparkConf) {
+  private def openConfig(str: String) = {
+    val sysConfig = GlobalConfig.systemConfig.getConfig("filodb")
+    ConfigFactory.parseFile(new File(conf.get(str))).getConfig("filodb").withFallback(sysConfig)
+  }
+
+  // Both "source" and "target" refer to file paths which define config files that have a
+  // top-level "filodb" section and a "cassandra" subsection.
+  val sourceConfig = openConfig("spark.filodb.chunkcopier.source.configFile")
+  val targetConfig = openConfig("spark.filodb.chunkcopier.target.configFile")
+
+  val sourceCassConfig = sourceConfig.getConfig("cassandra")
+  val targetCassConfig = targetConfig.getConfig("cassandra")
+
+  val sourceDatasetRef = DatasetRef.fromDotString(conf.get("spark.filodb.chunkcopier.source.dataset"))
+  val targetDatasetRef = DatasetRef.fromDotString(conf.get("spark.filodb.chunkcopier.target.dataset"))
+
+  // Examples: 2019-10-20T12:34:56Z  or  2019-10-20T12:34:56-08:00
+  private def parseDateTime(str: String) = Instant.from(DateTimeFormatter.ISO_OFFSET_DATE_TIME.parse(str))
+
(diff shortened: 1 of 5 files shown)


spark-jobs/src/main/scala/filodb/repair/ChunkCopier.scala
```diff
@@ -0,0 +1,134 @@
...
+  val targetConfig = openConfig("spark.filodb.chunkcopier.target.configFile")
+
+  val sourceCassConfig = sourceConfig.getConfig("cassandra")
+  val targetCassConfig = targetConfig.getConfig("cassandra")
+
+  val sourceDatasetRef = DatasetRef.fromDotString(conf.get("spark.filodb.chunkcopier.source.dataset"))
+  val targetDatasetRef = DatasetRef.fromDotString(conf.get("spark.filodb.chunkcopier.target.dataset"))
+
+  // Examples: 2019-10-20T12:34:56Z  or  2019-10-20T12:34:56-08:00
+  private def parseDateTime(str: String) = Instant.from(DateTimeFormatter.ISO_OFFSET_DATE_TIME.parse(str))
```
vishramachandran: Do you mind making this configuration consistent with DownsamplerMain? Perhaps the following line can also use the above approach ?

```
    val userTimeInPeriod: Long = spark.sparkContext.getConf.get("spark.filodb.downsampler.userTimeOverride",
      s"${System.currentTimeMillis() - settings.downsampleChunkDuration}").toLong
```

spark-jobs/src/main/scala/filodb/repair/ChunkCopier.scala
```diff
@@ -0,0 +1,134 @@
...
+
+  val ingestionTimeStart = parseDateTime(conf.get("spark.filodb.chunkcopier.ingestionTimeStart"))
+  val ingestionTimeEnd = parseDateTime(conf.get("spark.filodb.chunkcopier.ingestionTimeEnd"))
+
+  val diskTimeToLiveSeconds = conf.getTimeAsSeconds("spark.filodb.chunkcopier.diskTimeToLive")
+  val splitsPerNode = conf.getInt("spark.filodb.chunkcopier.splitsPerNode", 1)
+  val batchSize = conf.getInt("spark.filodb.chunkcopier.batchSize", 10000)
+
+  val readSched = Scheduler.io("cass-read-sched")
+  val writeSched = Scheduler.io("cass-write-sched")
```
vishramachandran: Reuse from DownsampleContext? Or rename that class as SparkJobContext ?

spark-jobs/src/main/scala/filodb/repair/ChunkCopier.scala
```diff
@@ -0,0 +1,134 @@
...
+      .getOrCreate()
+
+    val splits = copier.getScanSplits
+
+    logger.info(s"Cassandra split size: ${splits.size}. We will have this many spark partitions. " +
+      s"Tune splitsPerNode which was $copier.splitsPerNode if parallelism is low")
+
+    spark.sparkContext
+      .makeRDD(splits)
+      .foreachPartition(splitIter => ChunkCopier.lookup(conf).run(splitIter))
```
vishramachandran: Did you verify that `conf` is same for each worker task ? For example, no additional properties are added/removed/updated ?

broneill: Didn't check. Do you expect any to be added?

spark-jobs/src/main/scala/filodb/repair/ChunkCopier.scala
broneill: Which configuration consistent? You'd prefer that the timestamp be specified as seconds from 1970? The userTimeInPeriod line is capturing a duration, not an instant, and so it's different.

spark-jobs/src/main/scala/filodb/repair/ChunkCopier.scala
broneill: The SparkConf docs state that the config cannot change after running.

spark-jobs/src/main/scala/filodb/repair/ChunkCopier.scala
broneill: The design is a bit different here. I wanted to have the entire state encapsulated into one thing, and it's cached by a different key. I'm not sure which approach is better. I like the fact that the chunk copier code doesn't have to have transient fields all over the place, making it a bit simpler to understand and (hopefully) maintain.

spark-jobs/src/main/scala/filodb/repair/ChunkCopier.scala
vishramachandran: I would like userTimeInPeriod to be specified as a string like in your PR

broneill: But that's a duration, and it has a different syntax. Duration syntax is less standardized and many are unfamiliar with it.

broneill: Will fix in a separate PR.

PR #664, opened 2020-03-04
fix(query): fix for schema mismatch error in binaryOp with fixed scalar value

sherali42:
**Pull Request checklist**

**Current behavior :**
BinaryJoin queries with RangeVector and ScalarVector combination are failing with following error
> filodb.core.memstore.SchemaMismatch: Multiple schemas found, please filter. Expected schema..

**New behavior :**
ResultSchema comparison is made less restrictive to exclude `fixedVectorLen` and `coldIDs`


query/src/main/scala/filodb/query/LogicalPlan.scala
```diff
@@ -274,9 +274,9 @@ final case class ScalarVaryingDoublePlan(vectors: PeriodicSeriesPlan,
   */
 final case class ScalarTimeBasedPlan(function: ScalarFunctionId, rangeParams: RangeParams) extends ScalarPlan {
   override def isRoutable: Boolean = false
-  override def startMs: Long = rangeParams.start
-  override def stepMs: Long = rangeParams.step
-  override def endMs: Long = rangeParams.end
+  override def startMs: Long = rangeParams.start * 1000
+  override def stepMs: Long = rangeParams.step * 1000
+  override def endMs: Long = rangeParams.end * 1000
```
vishramachandran: Thanks for the fix! Can we rename rangeParams fields to `startSecs`, `endSecs` etc ?

query/src/main/scala/filodb/query/exec/ExecPlan.scala
```diff
@@ -406,7 +406,7 @@ abstract class NonLeafExecPlan extends ExecPlan {
       case QueryResult(_, schema, _) if rs == ResultSchema.empty =>
         schema     /// First schema, take as is
       case QueryResult(_, schema, _) =>
-        if (rs != schema) throw SchemaMismatch(rs.toString, schema.toString)
+        if (rs.hasSameColumnsAs(schema)) throw SchemaMismatch(rs.toString, schema.toString)
```
vishramachandran: This is tricky and needs more thought. 

Does this mean we can join counters and gauges ?

vishramachandran: Can we make this apply this fix only to joins ?
Doing it here, will apply for dist-concat as well...

sherali42: ack, fixed

query/src/main/scala/filodb/query/LogicalPlan.scala
sherali42: ack, fixed it

core/src/main/scala/filodb.core/query/ResultTypes.scala
```diff
@@ -52,9 +52,9 @@ final case class ResultSchema(columns: Seq[ColumnInfo], numRowKeyColumns: Int,
                               columns(1).colType == HistogramColumn && columns(2).colType == DoubleColumn
 
   def hasSameColumnsAs(other: ResultSchema): Boolean = {
-    // exclude fixedVectorLen
+    // exclude fixedVectorLen & coldIDs
```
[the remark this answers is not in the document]
sherali42: fixed

query/src/main/scala/filodb/query/exec/ExecPlan.scala
```diff
@@ -378,8 +378,7 @@ abstract class NonLeafExecPlan extends ExecPlan {
     // an empty schema.  Validate that the other schemas are the same.  Skip over empty schemas.
     var sch = ResultSchema.empty
     val processedTasks = childTasks.collect {
-      case (res @ QueryResult(_, schema, _), i) if schema != ResultSchema.empty =>
-        sch = reduceSchemas(sch, res)
+      case (res @ QueryResult(_, schema, _), i) if schema != ResultSchema.empty => sch = reduceSchemas(sch, res)
```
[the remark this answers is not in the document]
sherali42: reverted the change

PR #677, opened 2020-03-10
bug(query): AvgOverTime on downsample data fails

vishramachandran:
**Pull Request checklist**

Result schema for AvgOverTime on downsample data was wrong because of a
check that was not fixed during introduction of multi-schema queries.

Because of this we saw:

```
java.lang.IllegalArgumentException: Invalid col 2 
at filodb.core.query.TransientRow.getDouble(TransientRow.scala:47) 
at filodb.core.binaryrecord2.RecordSchema$$anonfun$11$$anonfun$apply$3.apply(RecordSchema.scala:77) 
at filodb.core.binaryrecord2.RecordSchema$$anonfun$11$$anonfun$apply$3.apply(RecordSchema.scala:77) 
at filodb.core.binaryrecord2.RecordBuilder.addFromReader(RecordBuilder.scala:287) 
at filodb.core.query.SerializedRangeVector$.apply(RangeVector.scala:316) 
at filodb.query.exec.ExecPlan$$anonfun$step2$1$2$$anonfun$7.apply(ExecPlan.scala:151) 
at filodb.query.exec.ExecPlan$$anonfun$step2$1$2$$anonfun$7.apply(ExecPlan.scala:141)```

### DIFF
--- a/query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
+++ b/query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
@@ -117,8 +117,12 @@ final case class PeriodicSamplesMapper(start: Long,
 
   // Transform source double or long to double schema
   override def schema(source: ResultSchema): ResultSchema = {
-    // Special treatment for downsampled gauge schema - return regular timestamp/value
-    if (functionId.contains(AvgWithSumAndCountOverTime)) {
+    // Special treatment for downsampled gauge schema.
+    // Source schema will be the selected columns. Result of this mapper should be regular timestamp/value
+    // since the avg will be calculated using sum and count
+    // FIXME dont like that this is hardcoded; but the check is needed.
+    if (functionId.contains(AvgWithSumAndCountOverTime) &&
+                        source.columns.map(_.name) == Seq("timestamp", "sum", "count")) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala [hunk 1]
```diff
@@ -118,9 +118,9 @@ final case class PeriodicSamplesMapper(start: Long,
   // Transform source double or long to double schema
   override def schema(source: ResultSchema): ResultSchema = {
     // Special treatment for downsampled gauge schema - return regular timestamp/value
-    if (source.columns == Schemas.dsGauge.dataInfos) {
+    if (functionId.contains(AvgWithSumAndCountOverTime)) {
```
velvia: Are you sure this is only needed for that function ID?  It seems to be all other downsampled gauge needs the transform too. Otherwise any other function on downsampled data just copies the original schema, which would contain multiple columns right?

vishramachandran: Source would be only selected columns (timestamp and one of the cols) which would be right. AvgWithSumAndCountOverTime is the only function that would have selected two columns.

PR #679, opened 2020-03-11
fix(query): Set Ops/binary joins fail when a range vector joined with timestamp

sherali42:
**Pull Request checklist**

- In `reduceSchemas`, ResultSchema comparison is made less restrictive for SetOps and Binary-Joins. fixedVectorLen & colIDs are excluded in the check.
- In order make it consistent, ResultSchema timestamp column type is changed from `LongColumn` to `TimestampColumn`.
- RangeParams fields are renamed to indicate time unit is Seconds

### DIFF
--- a/core/src/main/scala/filodb.core/query/ResultTypes.scala
+++ b/core/src/main/scala/filodb.core/query/ResultTypes.scala
@@ -52,9 +52,9 @@ final case class ResultSchema(columns: Seq[ColumnInfo], numRowKeyColumns: Int,
                               columns(1).colType == HistogramColumn && columns(2).colType == DoubleColumn
 
   def hasSameColumnsAs(other: ResultSchema): Boolean = {
-    // exclude fixedVectorLen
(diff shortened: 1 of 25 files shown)


core/src/main/scala/filodb.core/query/ResultTypes.scala
```diff
@@ -52,9 +52,9 @@ final case class ResultSchema(columns: Seq[ColumnInfo], numRowKeyColumns: Int,
                               columns(1).colType == HistogramColumn && columns(2).colType == DoubleColumn
 
   def hasSameColumnsAs(other: ResultSchema): Boolean = {
-    // exclude fixedVectorLen
+    // exclude fixedVectorLen & colIDs
     other.columns == columns && other.numRowKeyColumns == numRowKeyColumns &&
-      other.brSchemas == brSchemas && other.colIDs == colIDs
+      other.brSchemas == brSchemas
```
vishramachandran: @velvia can you confirm this ?

PR #680, opened 2020-03-12
bug(cassandra): Prepared stmts everywhere, fix stmt consistency

vishramachandran:
**Pull Request checklist**

We were not using prepared statements everywhere paying cost of inefficient queries.
Read consistency of ONE was not used everywhere.

**New behavior :**

* Organize CQL statements by preparing
* Set consistency of ONE for reads, and the configured writeConsistency for writes.

### DIFF
--- a/cassandra/src/main/scala/filodb.cassandra/columnstore/IngestionTimeIndexTable.scala
+++ b/cassandra/src/main/scala/filodb.cassandra/columnstore/IngestionTimeIndexTable.scala
@@ -19,7 +19,9 @@ import filodb.core.store.ChunkSinkStats
  * using TimeSeriesChunksTable. This is because the chunks are likely to be fetched from
  * Cassandra due to locality when using TimeSeriesChunksTable, and the chunks table is smaller.
  */
-sealed class IngestionTimeIndexTable(val dataset: DatasetRef, val connector: FiloCassandraConnector)
+sealed class IngestionTimeIndexTable(val dataset: DatasetRef,
+                                     val connector: FiloCassandraConnector,
+                                     writeConsistencyLevel: ConsistencyLevel)
                                     (implicit ec: ExecutionContext) extends BaseDatasetTable {
   import scala.collection.JavaConverters._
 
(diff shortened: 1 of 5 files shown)


cassandra/src/main/scala/filodb.cassandra/columnstore/IngestionTimeIndexTable.scala
```diff
@@ -63,36 +82,30 @@ sealed class IngestionTimeIndexTable(val dataset: DatasetRef, val connector: Fil
...
-      s"WHERE TOKEN(partition) >= $start AND TOKEN(partition) < $end " +
-      s"AND ingestion_time >= $ingestionTimeStart AND ingestion_time <= $ingestionTimeEnd " +
-      s"ALLOW FILTERING"
 
     tokens.iterator.flatMap { case (start, end) =>
-      session.execute(cql(start, end)).iterator.asScala
+      val stmt = scanCql1.bind(start.toLong: java.lang.Long,
+                               end.toLong: java.lang.Long,
+                               ingestionTimeStart: java.lang.Long,
+                               ingestionTimeEnd: java.lang.Long)
```
vishramachandran: @velvia  one confirmation I wanted from you is that for token scans, we were using String parameter from before for the Cassandra start/end tokens. I never understood why. Conversion of token to Long should not be an issue right ?

vishramachandran: In other words, tokens should be within limits of Long ?

vishramachandran: Confirmed that this works fine or Murmur3Partitioner tokens - which is why existing approach of converting to strings has been working.

velvia: Was this validated through tests?   Long _should_ be fine, though tokens can wrap also.  As long as we aren't doing manipulation of the tokens it's probably fine, but Datastax driver actually has Token objects.  I would think they would let you bind to the Token objects directly, that would be preferable to using longs

vishramachandran: I did try Token objects, but they are not Serializable. It breaks our spark jobs, and I dont want to open a can of worms there.

vishramachandran:
@velvia I tried string binding. It does not work.

PR #695, opened 2020-03-19
feat(spark-jobs): Chunk copier supports explicit deletion.

broneill:
**Pull Request checklist**

**Current behavior :**
Chunk copier supports 0 TTL, which is effectively like a delete.

**New behavior :**
Use 0 TTL to indicate that an explicit delete must be performed on the target, which can be the same as the source.

### DIFF
--- a/cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
+++ b/cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
@@ -219,7 +219,10 @@ extends ColumnStore with CassandraChunkSource with StrictLogging {
 
   /**
     * Copy a range of chunks to a target ColumnStore, for performing disaster recovery or
-    * backfills.
+    * backfills. This method can also be used to delete chunks, by specifying a negative ttl.
+    * If the target is the same as the source, then this effectively deletes from the source.
+    *
+    * @param diskTimeToLiveSeconds pass a negative value to delete chunks
     */
   // scalastyle:off null method.length
   def copyChunksByIngestionTimeRange(datasetRef: DatasetRef,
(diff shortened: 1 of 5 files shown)


cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
```diff
@@ -219,7 +219,10 @@ extends ColumnStore with CassandraChunkSource with StrictLogging {
...
   /**
     * Copy a range of chunks to a target ColumnStore, for performing disaster recovery or
-    * backfills.
+    * backfills. This method can also be used to delete chunks, by specifying a ttl of zero.
+    * If the target is the same as the source, then this effectively deletes from the source.
+    *
+    * @param diskTimeToLiveSeconds pass zero to delete chunks
     */
   // scalastyle:off null method.length
   def copyChunksByIngestionTimeRange(datasetRef: DatasetRef,
```
vishramachandran, line 228: Can we rename this method to account for the fact that it also does deletes?

cassandra/src/main/scala/filodb.cassandra/columnstore/IngestionTimeIndexTable.scala
```diff
@@ -154,4 +158,16 @@ sealed class IngestionTimeIndexTable(val dataset: DatasetRef,
       diskTimeToLiveSeconds: java.lang.Integer)
     )
   }
+
+  /**
+    * Deletes a single record, specified by a row from the scanRowsByIngestionTimeNoAsync method.
+    * Is used to delete records which are incorrect or inconsistent.
+    */
+  def deleteIndex(row: Row): Future[Response] = {
```
vishramachandran, line 166: Need more docs on what the expectation of Row is.

PR #713, opened 2020-04-08
Cherry-pick changes to support `Last` function from 0.8-maintenance branch 

snsingh:
Cherry-pick changes to support `Last` function from 0.8-maintenance branch

### DIFF
--- a/query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
+++ b/query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
@@ -255,6 +255,7 @@ object RangeFunction {
       case Some(MinOverTime)      => Seq("min")
       case Some(MaxOverTime)      => Seq("max")
       case other                  => Seq(schema.data.valueColName)
+      case Some(Last)             => Seq("avg")
     }
   }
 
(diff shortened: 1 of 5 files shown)


snsingh:
@sherali42 @tjackpaul PTAL at the changes cherry-picked from 08-maintenance branch and modify as needed. A lot has changed in the develop branch around the sample code change. I have made changes to the best of my knowledge. A thorough review of the PR is needed.

query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
```diff
@@ -436,11 +441,11 @@ extends ChunkedRangeFunction[R] {
 
     // update timestamp only if
     //   1) endRowNum >= 0 (timestamp within chunk)
-    //   2) timestamp is within stale window; AND
+    //   2) timestamp is within window; AND
     //   3) timestamp is greater than current timestamp (for multiple chunk scenarios)
     if (endRowNum >= 0) {
       val ts = tsReader(tsVectorAcc, tsVector, endRowNum)
-      if ((endTime - ts) <= queryConfig.staleSampleAfterMs && ts > timestamp)
```
TanviBhavsar, line 443: Why are we removing staleSampleAfterMs?

sherali42: @TanviBhavsar this is to make sure that lookback time falls within the `window`, right now even if the window is 3mins, lookback time can go upto 5mins because of `queryConfig.staleSampleAfterMs`.

vishramachandran: @sherali42 the above check is still needed. If window was given as 30mins, the last sample holds good only for 5 mins. That check needs to stay.

vishramachandran: ok Sher convinced me that since data worth only a 5 minute (stale window length) lookback window will be scanned, we should be fine removign the condition.

PR #714, opened 2020-04-09
feat(query, core): TimeSeries Metadata query can include start/end times

vishramachandran:
**Pull Request checklist**

Metadata queries include partition keys along with start/end times. Slightly more expensive query when reqested.

TODO
* There is a bug in printing the part key key-value pairs when tested from CLI
* Test from HTTP


vishramachandran:
There is now a fetchStartEndTimes param in the logical plan that defaults to false. The result schema has two additional columns.

PR #719, opened 2020-04-13
fix(query): Fix extraction of +Inf bucket from histogram RV

velvia:
**Pull Request checklist**

Extracting the `+Inf` bucket out of a histogram vector via `_bucket_="+Inf"` or `histogram_bucket(+Inf` was not working due to a bug in the `HistogramBucketImpl`.  The problem was that we were using code to compare the bucket value using double epsilon, ie (expected - actual) < (epsilon), but subtraction of infinity from infinity yields NaN in JVM.  

**New behavior :**

We add code to specialize the +Inf case so we wouldn't need to do an epsilon comparison.

### DIFF
--- a/query/src/main/scala/filodb/query/exec/rangefn/InstantFunction.scala
+++ b/query/src/main/scala/filodb/query/exec/rangefn/InstantFunction.scala
@@ -369,7 +369,8 @@ final case class HistogramBucketImpl() extends HistToDoubleIFunction {
     val bucket = scalarParams(0)
     if (bucket == Double.PositiveInfinity) {
       // Just get the top bucket if bucket scheme has +Inf at top, or return NaN
-      if (value.bucketTop(value.numBuckets - 1) == Double.PositiveInfinity) value.topBucketValue else Double.NaN
+      if (value.bucketTop(value.numBuckets - 1) == Double.PositiveInfinity) value.topBucketValue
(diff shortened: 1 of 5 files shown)


query/src/main/scala/filodb/query/exec/rangefn/InstantFunction.scala
```diff
@@ -367,9 +367,16 @@ final case class HistogramBucketImpl() extends HistToDoubleIFunction {
   final def apply(value: Histogram, scalarParams: Seq[Double]): Double = {
     require(scalarParams.length == 1, "Bucket/le required for histogram bucket")
     val bucket = scalarParams(0)
-    for { b <- 0 until value.numBuckets optimized } {
-      if (Math.abs(value.bucketTop(b) - bucket) <= 1E-10) return value.bucketValue(b)
+    if (bucket == Double.PositiveInfinity) {
+      // Just get the top bucket if bucket scheme has +Inf at top, or return NaN
+      if (value.bucketTop(value.numBuckets - 1) == Double.PositiveInfinity) value.topBucketValue else Double.NaN
```
vishramachandran, line 372: else throw Invalid data exception ? This condition should never happen right ? Lets not silently return NaN and suppress the error - it wont surface. User will never know about it. It'll be harder to debug quantile problems

velvia: Sure

PR #720, opened 2020-04-16
bug(core): RecordSchema.toStringPairs had bug with BR columns

vishramachandran:
**Pull Request checklist**

When string-pairing BR columns, different base/offset needs to be applied.

### DIFF
--- a/query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
+++ b/query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
@@ -101,7 +101,7 @@ final case class PartKeysExec(queryContext: QueryContext,
                               shard: Int,
                               partSchema: PartitionSchema,
                               filters: Seq[ColumnFilter],
-                              fetchStartEndTimes: Boolean,
+                              fetchFirstLastSampleTimes: Boolean,
                               start: Long,
                               end: Long) extends LeafExecPlan {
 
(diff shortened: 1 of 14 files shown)


query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
```diff
@@ -120,8 +120,8 @@ final case class PartKeysExec(queryContext: QueryContext,
     }
     Kamon.currentSpan().mark("creating-resultschema")
     val sch = new ResultSchema(Seq(ColumnInfo("TimeSeries", ColumnType.BinaryRecordColumn),
-      ColumnInfo("StartTime", ColumnType.LongColumn),
-      ColumnInfo("EndTime", ColumnType.LongColumn)), 3,
+      ColumnInfo("_startTime_", ColumnType.LongColumn),
```
tjackpaul, line 123: Will it be even clear if it is `_partStartTime_` ?

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
```diff
@@ -120,8 +120,8 @@ final case class PartKeysExec(queryContext: QueryContext,
...
+      ColumnInfo("_endTime_", ColumnType.LongColumn)), 3,
```
tjackpaul, line 124: Same goes here. Something like `_partEndTime_`, since we already have `_partIds_`

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
vishramachandran: Lets leave it short. I dont want the long name. It is for debugging anyway

vishramachandran: renamed actually

PR #726, opened 2020-04-20
fix(core): When removing a chunk, remove all lower ones too.

broneill:
**Pull Request checklist**

**Current behavior :**
When a block is reclaimed, chunks can be removed from partitions leaving "gaps" or "holes" in the time range. ODP isn't able to detect these holes, and so data remains missing until after a restart. 

**New behavior :**
Removing a chunk forces all lower chunks in the partition to be removed as well. ODP can detect the missing data properly and fill in the gap. Although more chunks are effectively reclaimed with this new logic, it's compatible with the original model of time-ordered reclamation. In such a model, gaps aren't expected anyhow.


vishramachandran:
@broneill Can we log and meter the floor elements removed?

broneill:
The shard stats should already be capturing the the number of elements removed. I updated that code in this PR to increment by the amount removed by the floor op.

PR #730, opened 2020-04-24
feat(query): add schema name to RangeVectorKey

tjackpaul:
**Pull Request checklist**

Add schema name to RangeVectorKey so that the same can be printed in results if needed.

### DIFF
--- a/prometheus/src/main/scala/filodb/prometheus/query/PrometheusModel.scala
+++ b/prometheus/src/main/scala/filodb/prometheus/query/PrometheusModel.scala
@@ -103,7 +103,7 @@ object PrometheusModel {
     */
   def toPromResult(srv: RangeVector, verbose: Boolean, typ: QueryResultType): Result = {
     val tags = srv.key.labelValues.map { case (k, v) => (k.toString, v.toString)} ++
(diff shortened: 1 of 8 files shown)


tjackpaul:
> Looks fine.... when would we need multiple schema names returned? Is that for aggregating schemas from multiple places in upper level aggregators? I didn't see the code for doing that.

We do not have it right now, since the aggregates across different schemas throw an error currently.  It might change in the future, hence added it as multiple schema names here.

prometheus/src/main/scala/filodb/prometheus/query/PrometheusModel.scala
```diff
@@ -149,6 +147,12 @@ object PrometheusModel {
     }
   }
 
+  def addVerboseLabelPairs(rvk: RangeVectorKey): Map[String, String] = {
```
vishramachandran, line 150: You are not adding verbose label pairs but creating them. Consider naming method as `makeVerboseLabels`

tjackpaul: take care of

PR #734, opened 2020-04-29
bug(core): Check if paged chunk exists in chunkMap before copying to block

vishramachandran:
**Pull Request checklist**

Before:
Paged chunk from Cassandra was copied to off-heap ODP block without checking if it exists in chunkMap. If it already existed, allocated memory is wasted. This with concurrent queries from dashboards caused too many chunks to be paged in and allocated but not added to chunkMap.

After:
We first do the contains check in the single threaded populateOdpChunks method and only then allocate in off-heap.

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/OnDemandPagingShard.scala
+++ b/core/src/main/scala/filodb.core/memstore/OnDemandPagingShard.scala
@@ -223,11 +223,10 @@ TimeSeriesShard(ref, schemas, storeConfig, shardNum, bufferMemoryManager, rawSto
         case req: TimeRangeChunkScan      =>  if (partition.numChunks > 0) {
                                                 val memStartTime = partition.earliestTime
                                                 if (req.startTime < memStartTime && partStartTime < memStartTime) {
-                                                  val toODP = TimeRangeChunkScan(req.startTime, memStartTime - 1)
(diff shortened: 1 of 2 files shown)


core/src/main/scala/filodb.core/memstore/OnDemandPagingShard.scala
```diff
@@ -223,11 +223,10 @@ TimeSeriesShard(ref, schemas, storeConfig, shardNum, bufferMemoryManager, rawSto
         case req: TimeRangeChunkScan      =>  if (partition.numChunks > 0) {
                                                 val memStartTime = partition.earliestTime
                                                 if (req.startTime < memStartTime && partStartTime < memStartTime) {
-                                                  val toODP = TimeRangeChunkScan(req.startTime, memStartTime - 1)
+                                                  val toODP = TimeRangeChunkScan(req.startTime, memStartTime)
```
velvia, line 226: Maybe add a comment about the boundary condition and explain what bug could result if we used -1?

vishramachandran: The CQL to fetch chunks uses `<` operator, and not `<=` operator on the endTime. Hence `-1` is not needed.

PR #748, opened 2020-05-12
feat(core): Query sizing based limits

vishramachandran:
**Pull Request checklist**

We only limit the number of time series a query touches. This is less restrictive than we want.

**New behavior :**

We estimate data size based on schema, sample size, time range and block queries that go beyond limit. The guidance is to reduce time range, or number of time series queried.

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
+++ b/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
@@ -1488,7 +1488,7 @@ class TimeSeriesShard(val ref: DatasetRef,
           val numSamplesPerChunk = storeConfig.flushInterval.toMillis / assumedResolution
           val numChunksPerTs = (end-st + storeConfig.flushInterval.toMillis - 1)/ storeConfig.flushInterval.toMillis
           val estDataSize = schemas.bytesPerSampleSwag(schId) * numMatches * numSamplesPerChunk * numChunksPerTs
-          require(estDataSize > storeConfig.maxDataPerShardQuery,
+          require(estDataSize < storeConfig.maxDataPerShardQuery,
               s"Estimate of $estDataSize bytes exceeds limit of " +
               s"${storeConfig.maxDataPerShardQuery} bytes queried per shard. Try one or more of these: " +
               s"(a) narrow your query filters to reduce to fewer than the current $numMatches matches " +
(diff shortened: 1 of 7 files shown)


core/src/main/scala/filodb.core/metadata/Schemas.scala
```diff
@@ -260,6 +261,23 @@ final case class Schemas(part: PartitionSchema,
...
+  /**
+    * This is purely a SWAG to be used for query size estimation. Do not rely for other use cases.
+    */
+  val bytesPerSampleSwag: Map[Int, Double] = {
+    schemas.values.map { s  =>
+      val est = s.data.columns.map(_.columnType).map {
+        case ColumnType.LongColumn => 1
+        case ColumnType.IntColumn => 1
+        case ColumnType.TimestampColumn => 0.5
+        case ColumnType.HistogramColumn => 10
```
TanviBhavsar, line 273: Why is ColumnType.HistogramColumn 10? Also why is ColumnType.TimestampColumn 0.5 when TimestampColumn is Long?

vishramachandran: The idea is to have an average bytes per sample. Timestamps are consecutive and with regular interval would compress well. Histograms take more space because of the buckets - this estimate is a swag and in reality it would depend on number of buckets. Long does not assume regular interval and hence is higher than timestamps. I would like @velvia to confirm.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -1474,8 +1470,39 @@ class TimeSeriesShard(val ref: DatasetRef,
...
+    *     precise size estimation is costly
+    * (b) it also assigns bytes per sample based on schema which is much of a swag. In reality, it would depend on
+    *     number of histogram buckets, samples per chunk etc.
+    * (c) Assumes reporting interval is 20s.
+    */
+  private def capDataScannedPerShardCheck(lookup: PartLookupResult) = {
+    lookup.firstSchemaId.foreach { schId =>
+      lookup.chunkMethod match {
+        case TimeRangeChunkScan(st, end) =>
+          val numMatches = lookup.partsInMemory.length + lookup.partIdsNotInMemory.length
```
velvia, line 1486: Could we move the code into a common place as it's repeated in both Shard classes?  Also, could we put the assumed resolution somewhere as a constant so we can easily change it if needed?

core/src/main/scala/filodb.core/metadata/Schemas.scala
whizkido: When we had done the testing in perf for histograms , the bytePerSample came to around 90 bytes for 64 bucket histograms.

PR #750, opened 2020-05-18
feat(core): ODP block allocation shouldn't force a reclamation.

broneill:
**Pull Request checklist**

**Current behavior :**
On Demand Paging can force block reclamation when it allocates new blocks, potentially reclaiming blocks that contain data in the requested query. This results in a "hole" in the query, although it generally is fixed the next time the query is issued.

**New behavior :**
Throw an exception instead of forcing reclamation, and now the headroom task must ensure that enough free space is available. A failed query is deemed better than broken results.


memory/src/main/scala/filodb.memory/BlockManager.scala
```diff
@@ -119,11 +119,14 @@ object PageAlignedBlockManager {
...
   * @param numPagesPerBlock       The number of pages a block spans
+  * @param onDemandTimeOrderedReclamation  When false, requestimg time ordered blocks fails if it
+                                           would force a reclamation
   */
 class PageAlignedBlockManager(val totalMemorySizeInBytes: Long,
                               val stats: MemoryStats,
                               reclaimer: ReclaimListener,
-                              numPagesPerBlock: Int)
+                              numPagesPerBlock: Int,
+                              onDemandTimeOrderedReclamation: Boolean = true)
```
vishramachandran, line 129: Can we leave out the default param = true? Cant tell if we have addressed all callers in the PR

broneill: Yes, and I'll add a test too.

memory/src/main/scala/filodb.memory/BlockManager.scala
```diff
@@ -178,7 +181,15 @@ class PageAlignedBlockManager(val totalMemorySizeInBytes: Long,
...
       stats.requestedBlocksMetric.increment(num)
 
-      if (freeBlocks.size < num) tryReclaim(num)
+      if (freeBlocks.size < num) {
+        if (bucketTime.isEmpty || onDemandTimeOrderedReclamation) {
+          tryReclaim(num)
+        } else {
+            val msg = s"Unable to allocate time ordered block(s) without forcing a reclamation: " +
+                      s"num_blocks=$num num_bytes=$memorySize freeBlocks=${freeBlocks.size}"
+            throw new IllegalStateException(msg)
```
vishramachandran, line 190: Is this an IllegalState ? This is the error user will see. It could mean overload of ODP queries (or ingestion) trying to use up all the headroom. I think this should be some kind of a "server overloaded" exception as opposed to illegal state.

broneill: Is there an existing class of exceptions for this sort of thing?

vishramachandran: I dont think so. We may have to create one.

broneill: What happens when a query exceeds any kind of limit? Nothing for that?

vishramachandran: We throw IllegalArgumentException for now.

broneill:
I'll just remove it.

PR #759, opened 2020-05-22
fix(core): Avoid reclamation when queries are running.

broneill:
**Pull Request checklist**

**Current behavior :** 
Ingestion can reclaim blocks needed by concurrent queries.

**New behavior :**
Ingestion waits (up to 16 seconds) for queries to finish.


vishramachandran:
@broneill Is it possible for the lock acquisition to be lazy only when reclaim needs to happen ?
Flush is going to happen constantly, and write lock will contend with queries. If write lock is attempted only when headroom is not available, it will happen much less often.

broneill:
Yes, it can be lazy. I need to think about the best approach.

PR #760, opened 2020-05-27
feat(core): Reduce block size to 400KB.

broneill:
**Pull Request checklist**

**Current behavior :** 
Block size is 4MB, which means that some memory is wasted due to chunkmap hole expansion.

**New behavior :**
Block size is 400KB, which increases the number of blocks, and so it should reduce the wastage.

Also replaced use of LinkedList with ArrayDeque reduce object allocation (old gen stuff) now that more Blocks exist.


broneill:
I can replace the linked lists too.

PR #761, opened 2020-05-27
perf(coord): Timeout query immediately if it stayed in mailbox for too long

vishramachandran:
**Pull Request checklist**

Using certain `Occurs` clauses enables scoring/ranking of search results which appears to be CPU intensive in profiler results. This PR uses Occurs.FILTER that explicitly disables scoring and weighting of results.

Also do timeout check before doing anything for query execution. This accounts for delays in delivery from actor mailbox.

TODO:
* JMH benchmark to measure effect. Right now it seems to be broken. Needs to be revived.
* More changes to possibly optimize other Lucene calls.

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
+++ b/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
@@ -1443,8 +1443,7 @@ class TimeSeriesShard(val ref: DatasetRef,
         val firstPartId = if (matches.isEmpty) None else Some(matches(0))
         val _schema = firstPartId.map(schemaIDFromPartID)
         val it1 = InMemPartitionIterator2(matches)
-        val partIdsToPage = debox.Buffer.empty[Int]
-        it1.filter(_.earliestTime > chunkMethod.startTime).foreach( partIdsToPage += _.partID)
+        val partIdsToPage = it1.filter(_.earliestTime > chunkMethod.startTime).map(_.partID)
         val partIdsNotInMem = it1.skippedPartIDs
         Kamon.currentSpan().tag(s"num-partitions-not-in-memory-$shardNum", partIdsNotInMem.length)
         val startTimes = if (partIdsToPage.nonEmpty) {
--- a/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
+++ b/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
@@ -1443,7 +1443,8 @@ class TimeSeriesShard(val ref: DatasetRef,
         val firstPartId = if (matches.isEmpty) None else Some(matches(0))
         val _schema = firstPartId.map(schemaIDFromPartID)
         val it1 = InMemPartitionIterator2(matches)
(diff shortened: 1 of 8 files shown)


query/src/main/scala/filodb/query/exec/MultiSchemaPartitionsExec.scala
```diff
@@ -99,6 +99,12 @@ final case class MultiSchemaPartitionsExec(queryContext: QueryContext,
   def doExecute(source: ChunkSource,
                 querySession: QuerySession)
                (implicit sched: Scheduler): ExecResult = {
+
+    // timeout can occur here if there is a build up in actor mailbox queue and delayed delivery
+    val queryTimeElapsed = System.currentTimeMillis() - queryContext.submitTime
+    if (queryTimeElapsed >= queryContext.queryTimeoutMillis)
+      throw QueryTimeoutException(queryTimeElapsed, this.getClass.getName)
```
velvia, line 106: I don't think MultiSchemaPartExec is the best place to do that.  Not all queries have MSPE in its query plan.  Instead, I want to run this in the QueryActor when it first gets the ExecPlan.  If it is done here, then futures have already dispatched in the thread pool and some resources have already been taken up.

vishramachandran: Good point. Addressed.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -1443,7 +1443,8 @@ class TimeSeriesShard(val ref: DatasetRef,
         val firstPartId = if (matches.isEmpty) None else Some(matches(0))
         val _schema = firstPartId.map(schemaIDFromPartID)
         val it1 = InMemPartitionIterator2(matches)
-        val partIdsToPage = it1.filter(_.earliestTime > chunkMethod.startTime).map(_.partID)
+        val partIdsToPage = debox.Buffer.empty[Int]
```
velvia, line 1446: NOTE: when the number of partitions is large, this may end up making much larger allocations than today.

run_benchmarks.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 sbt "jmh/jmh:run -rf json -i 15 -wi 10 -f3 -jvmArgsAppend -XX:MaxInlineLevel=20 \
  -jvmArgsAppend -Xmx4g -jvmArgsAppend -XX:MaxInlineSize=99 \
+ -prof jmh.extras.JFR:dir=/tmp/filo-jmh \
```
velvia, line 4: Is the idea here so we can always get JFR recording out?

vishramachandran: Correct

vishramachandran:
I am undoing most of this PR, retaining just the timeout checks. The JMH is not showing any improvement. It appears that Lucene's "score" methods actually do the job of search instead of scoring. Scoring is disabled already. I was able to debug through the steps and confirm it.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
vishramachandran: ack.

PR #770, opened 2020-06-04
bug(core): Release unwanted vectors eagerly when serving queries from downsampled data

vishramachandran:
**Pull Request checklist**

Downsample server pages all columns of schema since there is no way to fetch specific schema FiloDB columns from cassandra table. 

This contributes to on-heap usage since RawPartData holds on to the vectors until the query/scan is fully done.  This PR releases the unwanted vectors early.

Other fixes:
* Also fixing DownsampleMainSpec to use synthetic timestamps to bypass Query Planner restriction on data TTL.
* Fixed timeout check to return QueryError instead of failing actor

### DIFF
--- a/core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
+++ b/core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
@@ -294,6 +294,7 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
         downsampleTtls.last.toMillis -> downsampledDatasetRefs.last
       case TimeRangeChunkScan(startTime, _) =>
         var ttlIndex = downsampleTtls.indexWhere(t => startTime > System.currentTimeMillis() - t.toMillis)
+        // -1 return value means query startTime is before the earliest retention. Just pick the highest resolution
         if (ttlIndex == -1) ttlIndex = downsampleTtls.size - 1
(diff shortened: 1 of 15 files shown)


core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
```diff
@@ -292,19 +293,21 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
         // since it is the highest resolution/ttl
         downsampleTtls.last.toMillis -> downsampledDatasetRefs.last
       case TimeRangeChunkScan(startTime, _) =>
-        val ttlIndex = downsampleTtls.indexWhere(t => startTime > System.currentTimeMillis() - t.toMillis)
+        var ttlIndex = downsampleTtls.indexWhere(t => startTime > System.currentTimeMillis() - t.toMillis)
```
broneill, line 296: I don't understand this change.

vishramachandran: This picks the downsample resolution based on the earliest time of the query. If earliest time of query is older than any retention then -1 is returned. Then we have to resort to the highest resolution with highest TTL

broneill: Sounds good. Please add this comment to the code.

PR #787, opened 2020-06-12
feat(query): Instrument query scheduler to get metrics

vishramachandran:
**Pull Request checklist**

1. Removed additional task due to querySession.close and folded it into QueryActor inline.
2. Kamon instrumentation of query scheduler for metrics

Before:
```
QueryInMemoryBenchmark.someOverlapQueries      thrpt   42  1017.285 ± 46.556  ops/s
```

After: 
```
QueryInMemoryBenchmark.someOverlapQueries      thrpt   42  1129.631 ± 33.969  ops/s
```

Relevant new metrics added:
```
# Counter
executor_tasks_submitted_total{type="ThreadPoolExecutor",name="query-sched-prometheus"}
# Counter
executor_tasks_completed_total{type="ThreadPoolExecutor",name="query-sched-prometheus"}
# Histogram
executor_threads_active{type="ThreadPoolExecutor",name="query-sched-prometheus"} 
# Histogram
executor_queue_size_count{type="ThreadPoolExecutor",name="query-sched-prometheus"}
```

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
@@ -100,6 +100,19 @@ final class QueryActor(memStore: MemStore,
   private val resultVectors = Kamon.histogram("queryactor-result-num-rvs").withTags(TagSet.from(tags))
   private val queryErrors = Kamon.counter("queryactor-query-errors").withTags(TagSet.from(tags))
 
+  /**
+    * Instrumentation adds following metrics on the Query Scheduler
+    *
+    * # Counter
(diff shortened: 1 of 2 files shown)


vishramachandran:
I posted the before benchmark. The overhead isnt significant. It appears Akka dispatcher metrics are already enabled. Perhaps enabling the kamon-akka module just turns it on.

I would prefer not to add another knob esp because this is the query scheduler - we would need the metrics. Let's add it if we determine that the overhead is insignificant.

I dont think a README is needed. There are tons of metrics already. I'll add a note in the PR about the metric names.

PR #788, opened 2020-06-12
misc(query): add cardinality limits to binary-join, group-by queries

sherali42:
**Pull Request checklist**

- add cardinality limits on group-by and binary join queries
  - if the limits are breached throw `BadQueryException`
- limits are specified in the query context with default limits of 100k


query/src/main/scala/filodb/query/exec/AggrOverRangeVectors.scala
```diff
@@ -74,10 +75,12 @@ final case class AggregateMapReduce(aggrOp: AggregationOperator,
       sourceSchema.fixedVectorLen.filter(_ <= querySession.queryConfig.fastReduceMaxWindows).map { numWindows =>
         RangeVectorAggregator.fastReduce(aggregator, false, source, numWindows)
       }.getOrElse {
-        RangeVectorAggregator.mapReduce(aggregator, skipMapPhase = false, source, grouping)
+        RangeVectorAggregator.mapReduce(aggregator, skipMapPhase = false, source, grouping,
```
vishramachandran, line 78: Can you pls confirm that this code runs during the monix query pipeline setup stage, and not during scan stage?

query/src/main/scala/filodb/query/exec/AggrOverRangeVectors.scala
```diff
@@ -33,7 +33,8 @@ final case class ReduceAggregateExec(queryContext: QueryContext,
     val task = for { schema <- firstSchema }
                yield {
                  val aggregator = RowAggregator(aggrOp, aggrParams, schema)
-                 RangeVectorAggregator.mapReduce(aggregator, skipMapPhase = true, results, rv => rv.key)
+                 RangeVectorAggregator.mapReduce(aggregator, skipMapPhase = true, results, rv => rv.key,
+                   querySession.qContext.groupByCardLimit)
```
vishramachandran, line 37: Can you pls confirm that this code runs during the monix query pipeline setup stage, and not during scan stage?

query/src/main/scala/filodb/query/exec/AggrOverRangeVectors.scala
```diff
@@ -126,11 +129,17 @@ object RangeVectorAggregator extends StrictLogging {
...
+                grouping: RangeVector => RangeVectorKey,
+                cardinalityLimit: Int = Int.MaxValue): Observable[RangeVector] = {
     // reduce the range vectors using the foldLeft construct. This results in one aggregate per group.
     val task = source.toListL.map { rvs =>
       // now reduce each group and create one result range vector per group
       val groupedResult = mapReduceInternal(rvs, rowAgg, skipMapPhase, grouping)
+
+      // if group-by cardinality breaches the limit, throw exception
+      if (groupedResult.size > cardinalityLimit)
+        throw new BadQueryException(s"This query results in more than $cardinalityLimit group-by cardinality limit. " +
```
vishramachandran, line 141: Can you pls confirm that this code runs during the monix query pipeline setup stage, and not during scan stage?

query/src/main/scala/filodb/query/exec/BinaryJoinExec.scala
```diff
@@ -61,10 +61,15 @@ final case class BinaryJoinExec(queryContext: QueryContext,
...
 
+  //scalastyle:off method.length
   protected[exec] def compose(childResponses: Observable[(QueryResponse, Int)],
                               firstSchema: Task[ResultSchema],
                               querySession: QuerySession): Observable[RangeVector] = {
     val taskOfResults = childResponses.map {
+      case (QueryResult(_, _, result), _)
+        if (result.size  > queryContext.joinQueryCardLimit && cardinality == Cardinality.OneToOne) =>
+        throw new BadQueryException(s"This query results in more than ${queryContext.joinQueryCardLimit} " +
+          s"join cardinality. Try applying more filters.")
```
vishramachandran, line 72: My concerns with Binary join is not result cardinality - it is in fact the cardinality of the operands. They would cause so much high-card data to be passed over the network.

This check is conducted after the network IO happens, so it may be too late.

Can we push some checks down to the child plans to limit the number of RVs? And can we foresee the number of result vectors early on before scan stage ? 

I am not saying this is easy or possible, we need to evaluate.

vishramachandran:
@sherali42 Thanks for the changes, the group-by checks I think look good. Please double check they run before scan starts.

For Binary join, the new checks added are useful, but they don't avoid the costly network IO from leaf plans. Need that capability too!

query/src/main/scala/filodb/query/exec/AggrOverRangeVectors.scala
sherali42: @vishramachandran yes this is before scanning.

query/src/main/scala/filodb/query/exec/AggrOverRangeVectors.scala
sherali42: @vishramachandran yes this is before scanning.

core/src/main/resources/filodb-defaults.conf
```diff
@@ -150,6 +150,12 @@ filodb {
     # Maximum number of samples to return in a query
     sample-limit = 1000000
 
+    # Binary Join Cardinality limit
+    join-cardinality-limit = 10000
```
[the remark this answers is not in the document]
sherali42: bumped it uptown 25k

core/src/main/resources/filodb-defaults.conf
```diff
@@ -150,6 +150,12 @@ filodb {
...
+
+    # Group by Cardinality limit
+    group-by-cardinality-limit = 10000
```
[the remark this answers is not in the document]
sherali42: @vishramachandran reduced it to 1000

query/src/main/scala/filodb/query/exec/BinaryJoinExec.scala
sherali42: @vishramachandran yes you are right. i will work on it as a separate PR.

PR #795, opened 2020-06-19
feat(query, coordinator): support regex in shard keys

TanviBhavsar:
**Pull Request checklist**

Support regex in shard keys using ShardKeyRegexPlanner. Value for regex is retrieved using regexFieldMatcher. A LogicalPlan is created for every value of the regex. The LogicalPlan's are stitched using `ReduceAggregateExec` for aggregate functions and `DistConcatExec` for remaining queries


coordinator/src/main/scala/filodb.coordinator/queryplanner/ShardKeyRegexPlanner.scala
```diff
@@ -0,0 +1,55 @@
...
+  * @param nonRegexShardColumn shard column which is mandatory and cannot be a regex
+  * @param dataset dataset
+  * @param multiPartitionPlanner multiPartition query planner
+  * @param regexFieldMatcher used to get values for nonRegexShardColumn matching shard key column regex.
+  */
+
+class ShardKeyRegexPlanner(nonRegexShardColumn: String,
+                           dataset: Dataset,
+                           multiPartitionPlanner: QueryPlanner,
+                           regexFieldMatcher: ShardColumnValues => Seq[String]) extends QueryPlanner {
```
vishramachandran, line 22: Have you considered using the In operator of ColumnFilter instead of using Regex representation in the logical plan?
Also I am unclear on the design of `regexFieldMatcher` function.

query/src/main/scala/filodb/query/LogicalPlan.scala
```diff
@@ -335,6 +391,8 @@ case class ApplyAbsentFunction(vectors: PeriodicSeriesPlan,
   override def startMs: Long = vectors.startMs
   override def stepMs: Long = vectors.stepMs
   override def endMs: Long = vectors.endMs
+  override def updatePeriodicSeriesFilter(column: String, filter: Filter): PeriodicSeriesPlan = this.copy(vectors =
```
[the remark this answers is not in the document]
TanviBhavsar: If I add a LogicalPlan utility method like` copyWithUpdatedTimeRange` I will have to add all logical plans again. 
```
logicalPlan match {
      case lp: PeriodicSeries =>  lp.copy(rawSeries = rawSeries.copyRawSeriesFilter(column, filter)
     case lp: ApplyInstantFunction =>  lp.copy(vectors=vectors.copyRawSeriesFilter(column, filter))
      case lp: Aggregate  => lp.copy(vectors = vectors.copyRawSeriesFilter(column, filter))
...............
```

If a new LogicalPlan is created this method will fail if we don't add it in the switch case. We have had this issue in the past. Adding it in the interface will ensure that method is implemented for every new logical plan

coordinator/src/main/scala/filodb.coordinator/queryplanner/ShardKeyRegexPlanner.scala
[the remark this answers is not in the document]
TanviBhavsar: Just to make it modular. When we support binary join for shard key regex and multipartition the logic will increase and become more complicated. Also in the future we can use ShardKeyRegexPlanner with any new planner if needed or even limit to just SinglePartitionPlanner for testing. For now, it is probably fine to have it in multiPartitionPlanner. Let me know your thoughts.

TanviBhavsar: regexFieldMatcher will return all values for ShardColumnValues.regexColumn(_ns_) matching ShardColumnValues .regexValue for ShardColumnValues. nonRegexColumn(_ws_)

coordinator/src/main/scala/filodb.coordinator/queryplanner/ShardKeyRegexPlanner.scala
```diff
@@ -0,0 +1,65 @@
...
+   else throw new UnsupportedOperationException("Regex not supported for Binary Join")
+  }
+
+  private def materializeAggregate(aggregate: Aggregate, queryContext: QueryContext): ExecPlan = {
+    val execPlans = generateExec(aggregate, getNonMetricShardKeyFilters(aggregate).head, queryContext)
+    if (execPlans.size == 1) execPlans.head
+    else ReduceAggregateExec(queryContext, InProcessPlanDispatcher, execPlans, aggregate.operator, aggregate.params)
+  }
+
+  private def materializeNonAggregate(logicalPlan: LogicalPlan, queryContext: QueryContext): ExecPlan = {
```
vishramachandran, line 57: rename to `materializeOthers`

query/src/main/scala/filodb/query/LogicalPlan.scala
```diff
@@ -10,6 +10,13 @@ sealed trait LogicalPlan {
     * It is false for RawSeriesLikePlan, MetadataQueryPlan, RawChunkMeta, ScalarTimeBasedPlan and ScalarFixedDoublePlan
     */
   def isRoutable: Boolean = true
+  def replaceFilters(filters: Seq[ColumnFilter]): LogicalPlan = {
```
vishramachandran, line 13: Some scaladoc on the purpose and contract would help people writing more implementations

query/src/main/scala/filodb/query/LogicalPlan.scala
```diff
@@ -10,6 +10,13 @@ sealed trait LogicalPlan {
...
+    this match {
+      case p: PeriodicSeriesPlan => p.replacePeriodicSeriesFilters(filters)
+      case r: RawSeriesLikePlan  => r.replaceRawSeriesFilters(filters)
+      case _                     => throw new UnsupportedOperationException("Logical plan not supported for update")
```
vishramachandran, line 17: Remove the default case. Since this is a sealed class, it should catch at compile time. Otherwise it defeats the purpose of adding methods this way - we will only catch missing implementations at runtime.

TanviBhavsar:
> lgtm.. thank you @TanviBhavsar.. 👍
> 
> could you please help me recall why we need to convert LabelValue constraints to filter as part of this PR?

Thanks Jackson! I changed LabelValues filters to ColumnFilter to support Regex and other filters.  Before this change, it supported only Equals

tjackpaul:
>Thanks Jackson! I changed LabelValues filters to ColumnFilter to support Regex and other filters. Before this change, it supported only Equals

Ok great! So, with this change we will start supporting all filters against LabelValues metadata queries?

TanviBhavsar:
> > Thanks Jackson! I changed LabelValues filters to ColumnFilter to support Regex and other filters. Before this change, it supported only Equals
> 
> Ok great! So, with this change we will start supporting all filters against LabelValues metadata queries?

Yes

TanviBhavsar:
> Some minor comments. LGTM otherwise. Thanks for the iterations!

Thanks @vishramachandran ! I have addressed your review comments.

PR #796, opened 2020-06-23
fix(memory): Halt the process when locks aren't released properly.

broneill:
**Pull Request checklist**

**Current behavior :**
When locks aren't released properly, an error is logged, but this state is never recoverable. The process must be restarted.

**New behavior :**
Restart the process after the lock error state has been logged the first time.

### DIFF
--- a/memory/src/main/scala/filodb.memory/data/ChunkMap.scala
+++ b/memory/src/main/scala/filodb.memory/data/ChunkMap.scala
@@ -130,8 +130,14 @@ object ChunkMap extends StrictLogging {
     if (numLocksReleased > 0) {
       logger.error(s"Number of locks was non-zero: $numLocksReleased. " +
         s"This is indicative of a possible lock acquisition/release bug.")
+      haltAndCatchFire
     }
     execPlanTracker.put(t, execPlan)
+  }
+
+  def haltAndCatchFire(): Unit = {
+    logger.error(s"Shutting down process since it may be in an unstable/corrupt state.")
+    Runtime.getRuntime.halt(1)
   }
 }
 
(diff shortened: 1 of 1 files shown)


memory/src/main/scala/filodb.memory/data/ChunkMap.scala
```diff
@@ -130,9 +130,15 @@ object ChunkMap extends StrictLogging {
...
       logger.error(s"Number of locks was non-zero: $numLocksReleased. " +
         s"This is indicative of a possible lock acquisition/release bug.")
+      haltAndCatchFire
     }
     execPlanTracker.put(t, execPlan)
   }
+
+  def haltAndCatchFire(): Unit = {
+    logger.error(s"Shutting down process since it may be in an unstable/corrupt state.")
+    Runtime.getRuntime.halt(1)
```
vishramachandran, line 140: Can we use a unique exit code ?

broneill: I suppose. What would it reveal that the log message didn't?

vishramachandran:
Can we dump as much debug information before halting ?

broneill:
I'm not sure what can be added. We already try to track queries. Perhaps you should take over and add debug stuff that you feel is important?

memory/src/main/scala/filodb.memory/data/ChunkMap.scala
vishramachandran: Exit code is usually displayed in compute and  quicker to identify

vishramachandran:
ack. Please push this in. Thx.

PR #799, opened 2020-06-25
bug(core): Populate ODPed chunks in reverse order

vishramachandran:
**Pull Request checklist**

OPDed chunks are populated oldest first. This can potentially cause another query to assume that chunks are already there in memory because it checks the oldest chunk time in the chunk-map.

**New behavior :**

Populate chunks in newest first order. This way concurrent queries will take the ODP path. Not efficient since ODP will be done twice but, acceptable for now. At least it will yield right query results.


vishramachandran:
@velvia Chunks from RawPartData are already on heap. I believe the implementation of Seq decides how reverse iteration is done. Depending on the underlying collection, it may be materialized in reverse order if neeed.

PR #803, opened 2020-06-26
fix(core): Fix various bugs that left locks stuck.

broneill:
**Pull Request checklist**

**Current behavior :** 
Some unit tests fail because they leave locks stuck, and this also has been observed in the running service itself.

**New behavior :**
Unit tests perform more validation to help narrow down root cause, and a bunch of fixes have been made.


query/src/main/scala/filodb/query/exec/RangeVectorTransformer.scala
```diff
@@ -142,12 +142,29 @@ final case class InstantVectorFunctionMapper(function: InstantFunctionId,
   }
 }
 
+private abstract class FuncIterator(rows: Iterator[RowReader])  extends Iterator[RowReader] {
```
vishramachandran, line 145: Have you evaluated impact if you use this abstract class in the RangeVector trait ?

broneill: no

broneill: What issues do you expect?

vishramachandran: I think it is worth a shot. There may be much good out of this. Right now developers can forget to inherit from this interface.

broneill: I'm just going for the simplest changes for now, avoiding any possible destabilization.

broneill:
I created a more generic wrapper class that can be reused, but the "lack of proper failure handling" is so pervasive, there's no quick fix and it's way out of scope for this PR.

PR #804, opened 2020-06-29
feat(query): Introducing RangeVectorCursor

vishramachandran:
**Pull Request checklist**

Today `RangeVector`s expose `rows` as `Iterator[RowReader]` which cannot be closed.
Because of this, there isnt a cascading way to close the ChunkInfoIterator, thereby causing room for developer
error where it is not closed in some instances.

This PR changes `RangeVector` to use `RangeVectorCursor` which is expected to cascade the close 
call to all wrapped iterators, finally culminating with close of the `ChunkInfoIterator`(s), thereby releasing 
the locks reliably. This PR will force developers during compile time to think about closing their underlying cursors.


core/src/main/scala/filodb.core/query/RangeVectorCursor.scala
```diff
@@ -0,0 +1,42 @@
...
+  * that force developer to care for "closing" the cursor before
+  * completing the query
+  */
+trait RangeVectorCursor extends Iterator[RowReader] { self =>
+  /**
+    * This method mut release all resources (example locks) acquired
+    * for the purpose of executing this query
+    */
+  def close(): Unit
+  def map2(f: RowReader => RowReader): RangeVectorCursor = new RangeVectorCursor {
```
broneill, line 22: map2? Is this a common naming convention?

broneill:
By convention, the closable cursor should also implement java.io.Closeable or java.lang.AutoCloseable or whatever Scala-specific thing that's equivalent.

core/src/main/scala/filodb.core/query/RangeVectorCursor.scala
vishramachandran: There is already a `map` in `Iterator` but it returns an `Iterator`. Since I cannot use the same signature with different return value, I used `map2`. I'll rename to `mapRow`

query/src/main/scala/filodb/query/exec/AggrOverRangeVectors.scala
```diff
@@ -132,7 +134,7 @@ object RangeVectorAggregator extends StrictLogging {
       // now reduce each group and create one result range vector per group
       val groupedResult = mapReduceInternal(rvs, rowAgg, skipMapPhase, grouping)
       groupedResult.map { case (rvk, aggHolder) =>
-        val rowIterator = aggHolder.map(_.toRowReader)
+        val rowIterator = new CustomCloseCursor(aggHolder.map(_.toRowReader))(rvs.foreach(_.rows().close()))
```
sherali42, line 137: seems like we are invoking close redundantly for each group?

vishramachandran: Good Point. I'll try to address.

vishramachandran: Addressed. Thanks for catching!

PR #807, opened 2020-06-30
fix(query): Add lock validation to some unit tests and apply fixes.

broneill:
**Pull Request checklist**

**Current behavior :**
Some of the unit tests fail when lock release validation is enabled.

**New behavior :**
Fix the issues that weren't releasing the locks and re-enable validation.


query/src/test/scala/filodb/query/exec/rangefn/InstantFunctionSpec.scala
```diff
@@ -335,7 +335,12 @@ class InstantFunctionSpec extends RawDataWindowingSpec with ScalaFutures {
             else if (val1.isNaN) val2.isNaN shouldEqual true
             else val1 shouldEqual val2 +- 0.0001
         }
+        // The zip method appears to short circuit and doesn't always fully drain all the
+        // iterators. This can leave locks stuck. Calling size has the side effect of fully
+        // draining the iterator, which will in turn ensure that locks are released.
+        ex.size
+        res.size
```
vishramachandran, line 342: Can we close cursor instead of draining it ? Better yet, we should replace the NoCloseCursor in the test with custom close call.

query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
```diff
@@ -182,27 +182,36 @@ extends RangeVectorCursor with StrictLogging {
...
+        }
+      }
+      rangeFunction.apply(wit.curWindowStart, wit.curWindowEnd, sampleToEmit)
+    } catch {
+      case e: Throwable => {
+        wit.unlock()
+        wit.close()
+        throw e
       }
     }
```
vishramachandran, line 214: This is more of a bandaid IMO. The real fix should be to catch the exception upstream. Doesnt the "finally" catch the exception and close the cursor ?

query/src/test/scala/filodb/query/exec/rangefn/InstantFunctionSpec.scala
broneill: I have no idea where the cursor is coming from. The tests that fail aren't creating cursors directly as far as I can tell. I'll have to take heap dump and reverse engineer the code.

PR #816, opened 2020-07-09
fix(memory): Blocks shouldn't be reclaimable until after metadata is …

broneill:
…filled in. Also, minor thread safety improvement (currently not an issue).

**Pull Request checklist**

**Current behavior :** 
A potential race exists when allocating over multiple blocks. The blocks can be reclaimed before metadata has been written. The thread which is filling in the metadata can run concurrently with any thread performing reclamation.

**New behavior :**
Don't make full blocks reclaimable until after the metadata has been filled in.

### DIFF
--- a/memory/src/main/scala/filodb.memory/MemFactory.scala
+++ b/memory/src/main/scala/filodb.memory/MemFactory.scala
@@ -295,7 +295,7 @@ class BlockMemFactory(blockStore: BlockManager,
       if (blk != metadataSpan.last) {
         if (markFullBlocksAsReclaimable) {
           blk.markReclaimable()
-        } else {
+        } else synchronized {
           fullBlocks += blk
         }
       }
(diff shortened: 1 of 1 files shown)


broneill:
I don't see any issues with the new statements. Anything that you can see as an issue?

memory/src/main/scala/filodb.memory/MemFactory.scala
```diff
@@ -318,10 +325,12 @@ class BlockMemFactory(blockStore: BlockManager,
       }
     } else {
       val newBlock = requestBlock()
-      if (markFullBlocksAsReclaimable) {
-        block.markReclaimable()
-      } else {
-        fullBlocks += block
+      if (!metadataSpanActive) {
```
velvia, line 328: My main Q is, here, `metadataSpanActive` will always be true because we always allocate chunks in the middle of an span ie between startMetadata and endMetadata.   Thus the code in there will effectively not be called.... while the foreach loop in endMeta seems to mark things as reclaimable, I'm wondering if we are missing any cases by effectively not calling the above.  Also if we aren't really going to use that code might as well comment it out.

broneill: There's already a check below for if (metadataSpanActive), so I assumed both cases are possible.

broneill: It is possible to allocate memory directly without a span.

PR #821, opened 2020-07-16
feat(query): Step Multiple Notation for PromQL Lookback

vishramachandran:
**Pull Request checklist**

* Support for queries where lookback can be mentioned as a multiple of step.
* When such a notation is used, step query param becomes mandatory
* When this notation is used with counter based range functions, publish interval padding is done automatically so we don't miss counter increases between lookback windows. Publish interval is extracted from `_step_` tag of the time series.

Example
* `sum_over_time(foo{...}[5i])` with step = 10s will be equivalent to `sum_over_time(foo{...}[50s])` since lookback = 5 * 10s
* `sum_over_time(foo{...}[1i])` with step = 10s will be equivalent to `sum_over_time(foo{...}[10s])` since lookback = 1 * 10s
* `rate(foo{...}[1i])` with step = 60s and publish interval of 10s will be equivalent to `rate(foo{...}[70s])` since lookback = 1 * 60s + 10s. Publish interval is padded when both counter functions and step-factor notation is used.

For Future:
* We should explore adding publishInterval as a first class column in the partKey schema as opposed to a tag. It can be extracted out in a more performant manner.


query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
```diff
@@ -73,29 +77,49 @@ final case class PeriodicSamplesMapper(start: Long,
       case c: ChunkedRangeFunction[_] if valColType == ColumnType.HistogramColumn =>
         source.map { rv =>
           val histRow = if (hasMaxCol) new TransientHistMaxRow() else new TransientHistRow()
+          val rdrv = rv.asInstanceOf[RawDataRangeVector]
+          val pubInt =
+            if (functionId.exists(_.onCumulCounter) && stepFactorNotationUsed) rdrv.publishInterval.getOrElse(0L)
```
velvia: I think, rather than adding another parameter to the PSM which already has a lot, we should encapsulate the extra padding by abstracting this as a new `WindowParam` AST/trait, which can be NoWindow, SimpleWindow, WindowWithStepPadding etc.

ddl.cql
```diff
@@ -0,0 +1,121 @@
+
+CREATE KEYSPACE IF NOT EXISTS filodb_admin WITH replication = {'class': 'SimpleStrategy', 'replication_factor': '1'};
+
+CREATE KEYSPACE IF NOT EXISTS filodb WITH replication = {'class': 'SimpleStrategy', 'replication_factor': '1'};
+
+CREATE KEYSPACE IF NOT EXISTS filodb_downsample WITH replication = {'class': 'SimpleStrategy', 'replication_factor': '1'};
+
+CREATE TABLE IF NOT EXISTS filodb_admin.checkpoints (
```
broneill, line 8: Why is this change in this PR? Seems out of place.

core/src/main/scala/filodb.core/memstore/PublishIntervalFinder.scala
```diff
@@ -0,0 +1,33 @@
...
+
+  import UTF8Str._
+  val stepTag = "_step_".utf8
+
+  override def findPublishIntervalMs(pkSchemaId: Int, pkBase: Any, pkOffset: Long): Option[Long] = {
+    var result: Option[Long] = None
+    if (pkSchemaId == Schemas.global.part.hash) {
+      val consumer = new MapItemConsumer {
+        override def consume(keyBase: Any, keyOffset: Long, valueBase: Any, valueOffset: Long, index: Int): Unit = {
+          val keyUtf8 = new UTF8Str(keyBase, keyOffset + 1, UTF8StringShort.numBytes(keyBase, keyOffset))
```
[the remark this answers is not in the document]
vishramachandran: 👍 I'll get in touch with you on this.

core/src/test/scala/filodb.core/TestData.scala
```diff
@@ -473,7 +473,7 @@ object MetricsTestData {
   val timeseriesSchemas = Schemas(timeseriesSchema)
 
   val timeseriesDatasetWithMetric = Dataset.make("timeseries",
-    Seq("tags:map"),
+    Seq("_metric_:string", "tags:map"),
```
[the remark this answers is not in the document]
vishramachandran: I thought this was a bug in tests. The val name is `timeseriesDatasetWithMetric` but it didnt have the metric column in it. The tests work after this change.

query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
vishramachandran: I was aware of this, but chose to accept the additional param. It is hard for the planner to make this call so early. The padding depends on
* whether counter based function is used
* whether publish interval is even available
* step factor notation was used

Some of this information is available only at runtime.

vishramachandran:
@velvia `_step_` is not mandatory in collection. If publish interval is not available, the lookback padding will be skipped, and the queries will behave as if hard coded lookback was done.

I will update the docs.

The `step` query param is needed in instant queries ONLY if step multiple notation is used. Otherwise there is no impact to existing queries in any way. 

> Having hard coded reliance on _step_ tag 

The publish interval detection is a trait. We can add other ways to detect the value in the future. So it is not "hard coded" as such.

> It also breaks scenarios where data is not published regularly and where there are no intervals.

It does not. If `_step_` tag is not provided in the partKey, the lookback extension is omitted. Queries will not error out or be rejected.

> Relying on _step_ seems really fragile to me. 

I do not disagree. I have called out that in future, we have to devise ways to collect the publish interval in a more reliable fashion. It should be added as a first class field in the partition schema. But it remains as a fact that there will be use cases where publish interval is not provided - one is Prom Remote write. I have already accounted for that possibility in this PR.

If you are thinking about how to respond to these queries without adding the publish interval, yes - sure. We have met multiple times to discuss this, but no other elegant + reliable way comes to my mind. This is an inherent byproduct of cumulative counters and how PromQL evaluates counters. This solution was arrived at in the interest of time.

Alternative solutions considered:
* Reverse iterate samples, but how far behind should we go ? How would longer (say 5m) publish intervals be treated?
* Any other?

I am very open to improving this iteratively.

README.md
```diff
@@ -494,6 +494,45 @@ One major difference FiloDB has from the Prometheus data model is that FiloDB su
...
+
+**Problem 1:** PromQL has an important gotcha on range value. Hard coded lookback
+can yield to queries with unintended effects. Lets consider the expression `rate(foo[5m])`. The result
+of this query would have completely different meanings with different step values. Since step parameter
+typically is not controlled by user (as in Grafana), it can have silent negative effects. For example, if
+lookback becomes smaller than step because longer time range was chosen, then query silently drops samples.
+
+To alleviate this, we introduce a new range notation where lookback can be specified as a multiple of step.
+The notation `[1i]` would cause lookback to be 1 times step. Notation `[2i]` would make lookback as 2 times
+step and so on. This feature is in alignment with [MetricsQL](https://github.com/VictoriaMetrics/VictoriaMetrics/wiki/MetricsQL) FiloDB
```
broneill, line 511: Omit FiloDB from this sentence.

README.md
```diff
@@ -494,6 +494,45 @@ One major difference FiloDB has from the Prometheus data model is that FiloDB su
...
+
+The value provided within square brackets in the Range Vector Selector expression of the PromQL expression
+is commonly known as "range" or "lookback". It indicates how far behind each instant
+the query engine will lookback to normalize the value for any given instant.
+
+**Problem 1:** PromQL has an important gotcha on range value. Hard coded lookback
+can yield to queries with unintended effects. Lets consider the expression `rate(foo[5m])`. The result
+of this query would have completely different meanings with different step values. Since step parameter
+typically is not controlled by user (as in Grafana), it can have silent negative effects. For example, if
+lookback becomes smaller than step because longer time range was chosen, then query silently drops samples.
```
velvia, line 507: Samples are only dropped if the Lookback is below the publish interval, it has nothing to do with step.  Please correct....  or clarify what you mean here.

README.md
```diff
@@ -494,6 +494,45 @@ One major difference FiloDB has from the Prometheus data model is that FiloDB su
...
+This notation will help users who do not know what range value to use in their queries. The `[1i]` should be a good
+default for all queries. 2 or more times step can be used when smoothing is needed.
+
+**Problem 2:** There is yet another gotcha that we solve. If step == lookback, lookback windows in consecutive instants
+become adjacent and non-overlapping. In the case of cumulative counters, where rate or increase is used, this can
+still yield to wrong results. The increase or decrease between the adjacent windows would be omitted from the query. 
+
+The solution that we have adopted to solve this is to extend the lookback window by the publish interval of the
+time series. The padding is done if the publish interval is supplied. The publish interval is supplied as an
+internal `_step_` tag within the time series tags. The intent here is to make it better over time. by making it
```
velvia, line 523: Could we rename `_step_` to something like `_publish_` or something?  Also please add `_step_` or whatever to the list of `predefined-keys` so we can save on wire space.

README.md
vishramachandran: samples are indeed dropped if lookback is less than publish interval. I can explain offline. This has already been documented elsewhere. It is a long explanation, and needs to have a separate context and effort. We need to document how the query engine works in regards to lookback and step.

README.md
vishramachandran: It is unfortunate that `_step_` was chosen - I'll make it configurable.

README.md
vishramachandran: updated docs

README.md
```diff
@@ -504,7 +504,8 @@ the query engine will lookback to normalize the value for any given instant.
 can yield to queries with unintended effects. Lets consider the expression `rate(foo[5m])`. The result
 of this query would have completely different meanings with different step values. Since step parameter
 typically is not controlled by user (as in Grafana), it can have silent negative effects. For example, if
-lookback becomes smaller than step because longer time range was chosen, then query silently drops samples.
+user expands the time range and hardcoded lookback becomes smaller than step, the same query will no
+longer do what the user orginally intends to do because samples wont be included in the query.
```
velvia, line 508: Suggestion: replace
     the same query will no
 longer do what the user orginally intends to do because samples wont be included in the query

with

   the query lookback may not include enough samples for an effective query

PR #824, opened 2020-07-21
misc(memory): Remove unnecessary objects.

broneill:
**Pull Request checklist**

**Current behavior :**
Lots of AtomicReferences, AtomicBooleans, and Loggers found in a heap dump. Most were from the Block class.

**New behavior :**
Removed all of these useless objects. Note that block ownership feature was never truly used. The call to "own" a block and to check ownership was always performed in the same thread, in the BlockMemFactory.allocateOffheap method.

### DIFF
--- a/memory/src/main/scala/filodb.memory/Block.scala
+++ b/memory/src/main/scala/filodb.memory/Block.scala
@@ -1,39 +1,8 @@
 package filodb.memory
 
 import java.nio.ByteBuffer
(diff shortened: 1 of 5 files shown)


broneill:
We've gone down a path in which blocks are guarded using a latching mechanism, which effectively tracks thread ownership. It's not currently done at the block level, but it's at the shard level for now.

PR #825, opened 2020-07-22
bug(query): handle multiple schemas in remote exec

TanviBhavsar:
**Pull Request checklist**

**Current behavior :**
 Remote Exec uses only untyped schema as query response does not have schema

**New behavior :**
- Pass histogramMap = true in URL params and create histogram schema when DataSampl is of type HistSampl
- Return empty QueryResult when remote query response is empty 
- Ignore column names in reduce schema to allow stitching for gauge schema as it has DoubleColumn name as `value ` instead of `number` for untyped schema
- RemoteExec creates histogram with CustomBucket, but local histogram can have different bucket type like GeometricBucket. Allow addition of different types of buckets if bucketTops are equal.
- Don't remove empty histogram for remote query as it is needed for stitching with local results


query/src/main/scala/filodb/query/exec/PromQlRemoteExec.scala
```diff
@@ -89,16 +92,24 @@ case class PromQlRemoteExec(queryEndpoint: String,
                             params: PromQlQueryParams,
                             numberColumnRequired: Boolean = false) extends RemoteExec {
 
-  private val columns = Seq(ColumnInfo("timestamp", ColumnType.TimestampColumn),
+  override def args: String = s"${params.toString}, queryEndpoint=$queryEndpoint, " +
+    s"requestTimeoutMs=$requestTimeoutMs, limit=${queryContext.sampleLimit}, " +
+    s"numberColumnRequired=${numberColumnRequired}"
+  private val defaultColumns = Seq(ColumnInfo("timestamp", ColumnType.TimestampColumn),
     ColumnInfo(if (numberColumnRequired) "number" else "value", ColumnType.DoubleColumn))
```
tjackpaul, line 99:
```
ColumnInfo(if (numberColumnRequired) "number" else "value", ColumnType.DoubleColumn))
```

Can be replaced by following since we ignore the column name while comparing schemas:

```
ColumnInfo("value", ColumnType.DoubleColumn))
```

query/src/main/scala/filodb/query/exec/PromQlRemoteExec.scala
```diff
@@ -89,16 +92,24 @@ case class PromQlRemoteExec(queryEndpoint: String,
...
-  private val recSchema = SerializedRangeVector.toSchema(columns)
-  private val resultSchema = ResultSchema(columns, 1)
+  private val defaultRecSchema = SerializedRangeVector.toSchema(defaultColumns)
+  private val defaultResultSchema = ResultSchema(defaultColumns, 1)
+
+  val histColumns = Seq(ColumnInfo("timestamp", ColumnType.TimestampColumn),
+    ColumnInfo("h", ColumnType.HistogramColumn))
+  val histRecSchema = SerializedRangeVector.toSchema(histColumns)
+  val histResultSchema = ResultSchema(histColumns, 1)
```
tjackpaul, line 106: Rest of the variables are `private`. So does it make sense to keep `histColumns`, `histRecSchema`, `histResultSchema` as private too?

core/src/main/scala/filodb.core/query/ResultTypes.scala
```diff
@@ -57,6 +57,10 @@ final case class ResultSchema(columns: Seq[ColumnInfo], numRowKeyColumns: Int,
     other.columns == columns && other.numRowKeyColumns == numRowKeyColumns &&
       other.brSchemas == brSchemas
   }
+  def hasSameColumnTypes(other: ResultSchema): Boolean = {
```
velvia, line 60: empty line before?

memory/src/main/scala/filodb.memory/format/vectors/Histogram.scala
```diff
@@ -220,8 +220,9 @@ final case class MutableHistogram(buckets: HistogramBuckets, values: Array[Doubl
    * If the scheme is different, then an approximation is used so that the resulting histogram has
    * an approximate sum of the individual distributions, with the original scheme.  Modifies itself.
    */
-  final def addNoCorrection(other: HistogramWithBuckets): Unit =
-    if (buckets == other.buckets) {
+  final def addNoCorrection(other: HistogramWithBuckets): Unit = {
+    // Allow addition when type of bucket is different
+    if (buckets.allBucketTops.sameElements(other.buckets.allBucketTops)) {
```
velvia, line 225: Sorry but this is problematic.   
1) The bucket tops are not the only thing that matters.  For certain schemes such as DDS, other factors might be important like other parameters, and so this equivalency test is not right.
2) It breaks the abstraction given to buckets to compare themselves.  Comparing all the elements for geometric buckets might not perform as well.

What I suggest is instead to add a method `similarForMath` to each bucket which returns true for this purpose, if the buckets are similar enough for additions.   In that method, it can then switch on different types of `other` and do what is optimal and correct.  This would let subtypes decide when they can be added.

query/src/main/scala/filodb/query/exec/PromQlRemoteExec.scala
```diff
@@ -136,11 +167,42 @@ case class PromQlRemoteExec(queryEndpoint: String,
...
+
+        override def key: RangeVectorKey = CustomRangeVectorKey(r.metric.map(m => m._1.utf8 -> m._2.utf8))
+
+        override def rows(): RangeVectorCursor = {
+          import NoCloseCursor._
+
+          samples.iterator.collect { case v: HistSampl =>
+            row.setLong(0, v.timestamp * 1000)
+            val sortedBucketsWithValues = v.buckets.toArray.map(x => (x._1.toDouble, x._2)).sortBy(_._1)
+            val hist = MutableHistogram(CustomBuckets(sortedBucketsWithValues.map(_._1)),
```
velvia, line 192: You should sync with @HimaVarsha94 .... about her work to implement DDSketches, to make sure that using MutableHistograms would be correct. 

The only other reservation about this approach is that we basically need to keep all of Data in memory, which perhaps can be expensive.  However I don't have any better suggestions for now -- maybe we could serialize the incoming data into something cheaper, but that's more complicated.

TanviBhavsar:
Thanks @tjackpaul  and @velvia! I have addressed your review comments

PR #828, opened 2020-07-24
perf(coord): Query Scheduler switch to ForkJoinPool

vishramachandran:
**Pull Request checklist**

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
@@ -1,8 +1,7 @@
 package filodb.coordinator
 
 import java.lang.Thread.UncaughtExceptionHandler
-import java.util.concurrent.{ScheduledThreadPoolExecutor, ThreadFactory}
-import java.util.concurrent.atomic.AtomicLong
+import java.util.concurrent.{ForkJoinPool, ForkJoinWorkerThread}
 
 import scala.util.control.NonFatal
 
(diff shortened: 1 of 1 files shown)


coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
```diff
@@ -117,22 +113,20 @@ final class QueryActor(memStore: MemStore,
...
-              logger.error("Uncaught Exception in Query Scheduler", e)
-          })
+        thread.setUncaughtExceptionHandler(exceptionHandler)
+        thread.setName(s"$schedName-${thread.getPoolIndex}")
         thread
       }
     }
-    // TODO retaining old fixed size pool for now - later change to fork join pool.
-    val executor = new ScheduledThreadPoolExecutor(numSchedThreads, thFactory)
+    val executor = new ForkJoinPool( numSchedThreads, threadFactory, exceptionHandler, true)
```
broneill, line 129: Is async mode correct? Does monix detect the use of FJP and rely on it for joining tasks?

vishramachandran: Javadoc.
     * @param asyncMode if true, establishes local first-in-first-out
     * scheduling mode for forked tasks that are never joined. This
     * mode may be more appropriate than default locally stack-based
     * mode in applications in which worker threads only process
     * event-style asynchronous tasks.  For default value, use {@code
     * false}.

Monix does fall into above category. Moreoever, monix schedulers keeps async = true.

PR #830, opened 2020-07-24
feat(coordinator): Query concurrency limiter + single thread per query

velvia:
**Pull Request checklist**

Currently, when ExecPlans come into the QueryActor mailbox, they are processed immediately, and Tasks for actual query plan execution are then added onto a Scheduler.   This results in some issues when a flood of ExecPlans comes to a QueryActor:
1) As the scheduler is used as a queue, and Tasks and associated state use up significant memory, this results in `OutOfMemoryError`s and significant GC pressure, leading to host restarts or much longer latencies
2) A very large number of concurrent tasks from different queries leads to high overhead and thread switching / Monix task management overhead costs, which can be seen in profiler output.

**New behavior :**

This PR introduces a new class, the `QueryScheduler`.  

The `QueryScheduler` allows only n concurrent queries to be executed at the same time; all ExecPlans belonging to the same queryID may be then converted into Tasks and executed; only n concurrent queryIDs are allowed at once.   Incoming ExecPlans which do not belong in one of the currently executing queryIDs are put into a queue, and only when a query finishes executing is a new query allowed into the task pool.  Limiting concurrent queries is designed to address 1) the OOM and memory issues primarily.  It also has other benefits including tracking of and potential cancellation of running queries.

At the same time, the `QueryScheduler` will assign a single-threaded scheduler to each running query (rather all ExecPlans belonging to the same queryID).  This is designed to reduce the task and thread switching overhead, and in fact does appear to boost performance significantly, approximately 2x over the original design.

JMH performance measurements, QueryInMemoryBenchmark.someOverlapQueries:

Original QueryActor with fixed thread pool:
```
[info] Benchmark                                   Mode  Cnt     Score    Error  Units
[info] QueryInMemoryBenchmark.someOverlapQueries  thrpt   20  1217.110 ± 31.830  ops/s
```

Original QueryActor with ForkJoinPool:
```
[info] Benchmark                                   Mode  Cnt     Score    Error  Units
[info] QueryInMemoryBenchmark.someOverlapQueries  thrpt   20  1622.374 ± 91.850  ops/s
```

QueryScheduler with single-threaded pool per query (this PR):
```
[info] Benchmark                                   Mode  Cnt     Score    Error  Units
[info] QueryInMemoryBenchmark.someOverlapQueries  thrpt   20  2458.496 ± 44.633  ops/s
```

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
@@ -24,7 +24,8 @@ import filodb.query.exec.ExecPlan
 
 
 // A class to manage a queue and a set of currently executing queries, allowing no more than concurrentQueries
-// at a time.  When the query and execplans finishes execution, then they are removed.
+// at a time.  When the query and execplans finishes execution, then they are removed.  This has multiple benefits
+// including limiting memory usage and tracking running queries.
 // concurrentQueries defines the number of current queries where all execPlans belonging to the same queryID
 // is counted as a single query.
 // A queue keeps track of queries for which there is no space.  It is not expected for there to be more than
--- a/coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
@@ -44,11 +44,11 @@ class QueryQueue(concurrentQueries: Int) {
 
   // Given a plan and replyTo, executes the plan if there is room or it is a currently executing query.
   // If there isn't room, add it to the queue.
-  def execOrWait(plan: ExecPlan, replyTo: ActorRef, newPlanFunc: (ExecPlan, ActorRef) => Future[Unit])
-                (implicit sched: Scheduler): Unit = {
+  def execOrWait(plan: ExecPlan, replyTo: ActorRef, newPlanFunc: (ExecPlan, ActorRef, Scheduler) => Future[Unit])
+                (sched: Scheduler): Unit = {
     val id = plan.queryContext.queryId
     if ((currentIDs contains id) || currentIDs.size < concurrentQueries) {
-      execute(plan, replyTo, newPlanFunc)
+      execute(plan, replyTo, newPlanFunc)(sched)
     } else {
       queue.enqueue((plan, replyTo))
     }
@@ -57,11 +57,11 @@ class QueryQueue(concurrentQueries: Int) {
   def queueSize: Int = queue.length
 
   // Pop next item(s?) off top of queue so long as there is room to execute another query.
-  def tryPop(newPlanFunc: (ExecPlan, ActorRef) => Future[Unit])
-            (implicit sched: Scheduler): Unit = synchronized {
+  def tryPop(newPlanFunc: (ExecPlan, ActorRef, Scheduler) => Future[Unit])
+            (sched: Scheduler): Unit = synchronized {
     if (currentIDs.size < concurrentQueries && queue.nonEmpty) {
       val (nextPlan, replyTo) = queue.dequeue
-      execute(nextPlan, replyTo, newPlanFunc)
+      execute(nextPlan, replyTo, newPlanFunc)(sched)
     }
   }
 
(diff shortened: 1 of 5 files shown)


coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
```diff
@@ -25,6 +22,88 @@ import filodb.core.store.CorruptVectorException
...
+                  }
+                }
+                if (newspace) {
+                  tryPop(newPlanFunc)
+                }
+    }(sched)
+  }
+
+  // Currently active query IDs
+  def currentQueryIDs: Set[String] = currentIDs.keySet.toSet
```
broneill, line 104: Not used anywhere?

coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
```diff
@@ -25,6 +22,88 @@ import filodb.core.store.CorruptVectorException
...
+  val currentIDs = new mHashMap[String, Int]().withDefaultValue(0)
+  val queue = new mQueue[(ExecPlan, ActorRef)]
+  val idToSched = new mHashMap[String, Scheduler]()
+  val schedulers = new mQueue[Scheduler]()
+
+  // Given a plan and replyTo, executes the plan if there is room or it is a currently executing query.
+  // If there isn't room, add it to the queue.
+  def execOrWait(plan: ExecPlan,
+                 replyTo: ActorRef,
+                 newPlanFunc: (ExecPlan, ActorRef, Scheduler) => Future[Unit]): Unit = {
```
broneill, line 48: Synchronization is required here.

coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
```diff
@@ -25,6 +22,88 @@ import filodb.core.store.CorruptVectorException
...
+    planFut.onComplete {
+      case _ => synchronized {
+                  currentIDs(queryId) -= 1
+                  if (currentIDs(queryId) == 0) {
+                    currentIDs.remove(queryId)
+                    schedulers.enqueue(idToSched.remove(queryId).get)
+                    newspace = true
+                  }
+                }
+                if (newspace) {
```
broneill, line 97: No need to use newspace var. Can keep the tryPop in the sync block because tryPop is synchronized anyhow.

coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
```diff
@@ -25,6 +22,88 @@ import filodb.core.store.CorruptVectorException
...
+// A queue keeps track of queries for which there is no space.  It is not expected for there to be more than
+// one query in the queue belonging to the same queryID, this is because only high-level ExecPlans spawn more
+// low-level ones, and spawned ones arrive in the Actor mailbox and will go striaght to executing if the parent
+// plan is already executing.
+// A single threaded scheduler is created for each query (all ExecPlans with the same query ID).  This reduces
+// thread contention significantly and reduces task switching overhead, boosting query throughput especially
+// when the tasks are small.
+class QueryScheduler(concurrentQueries: Int) {
+  val currentIDs = new mHashMap[String, Int]().withDefaultValue(0)
+  val queue = new mQueue[(ExecPlan, ActorRef)]
```
vishramachandran, line 40: Need to instrument queue so metrics are available in Kamon

vishramachandran:
@velvia This PR should be merged to integration and double fixed to develop so we arent forced to bring in additional changes from develop. Needed to compare apples to apples.

coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
```diff
@@ -25,6 +22,88 @@ import filodb.core.store.CorruptVectorException
...
+      case _ => synchronized {
+                  currentIDs(queryId) -= 1
+                  if (currentIDs(queryId) == 0) {
+                    currentIDs.remove(queryId)
+                    schedulers.enqueue(idToSched.remove(queryId).get)
+                    newspace = true
+                  }
+                }
+                if (newspace) {
+                  tryPop(newPlanFunc)
```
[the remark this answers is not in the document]
velvia: > So attempt to place query in queue. But this doesnt happen yet. Instead context switches.

what do you mean context switches?  The call to execOrWait() (which is now synchronized) would complete and the queue would be added before any context switches.
tryPop() also is synchronized, so I don't think this scenario is possible.
Also, the tryPop has been moved back inside the synchronized block above.

coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
velvia: It will be used, eventually  :)

velvia:
@vishramachandran  @broneill   PR feedback addressed, including cap on max queue size and Kamon metrics

coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
broneill: It's not thread safe, so either remove it or add docs.

velvia: It's a read, but might not be consistent due to multi threading, is that what you mean?   I'll add docs

broneill: If reading while another thread is writing, the behavior is undefined. You might get an exception, or missing entry, or temporary duplication, etc.

PR #839, opened 2020-07-31
fix(core): move status from recovery to active if there is no data to read from ingestion source

sherali42:
**Pull Request checklist**

Currently if there are checkpoints for a dataset and there is no data at InputSource, recovery process waits indefinitely for data to arrive, and in turn Shard status gets stuck at `StatusRecovery(0)`.
This PR is to 
 - timeout when there is no data to read `.timeoutOnSlowUpstreamTo(timeout, Observable.now(endOffset))`
 - return endOffset to complete recovery
 - mark the status as active
 - enable normal ingestion

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/TimeSeriesMemStore.scala
+++ b/core/src/main/scala/filodb.core/memstore/TimeSeriesMemStore.scala
@@ -161,8 +161,8 @@ extends MemStore with StrictLogging {
       }
         // Nothing to read from source, blocking indefinitely. In such cases, 60sec timeout causes it
         // to return endOffset, making recovery complete and to start normal ingestion.
-        .timeoutOnSlowUpstreamTo(timeout, Observable.now(endOffset))
-        .collect {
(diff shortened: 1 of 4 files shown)


core/src/main/scala/filodb.core/memstore/TimeSeriesMemStore.scala
```diff
@@ -158,7 +158,11 @@ extends MemStore with StrictLogging {
           startOffsetValidated = true
         }
         shard.ingest(r)
-      }.collect {
+      }
+        // Nothing to read from source, blocking indefinitely. In such cases, 60sec timeout causes it
+        // to return endOffset, making recovery complete and to start normal ingestion.
+        .timeoutOnSlowUpstreamTo(timeout, Observable.now(endOffset))
```
velvia, line 164: Could we dedent two spaces on these two lines so that they line up with the previous closing brace, and so that `.collect {` appears one level up from the case?

sherali42: @velvia fixed the indentation

PR #840, opened 2020-07-31
fix(memory): Reset the meta position when reclaiming blocks.

broneill:
**Pull Request checklist**


memory/src/main/scala/filodb.memory/Block.scala
```diff
@@ -161,6 +161,7 @@ class Block(val address: Long, val capacity: Long, val reclaimListener: ReclaimL
       reclaimListener.onReclaim(metaPointer + 2, metaSize)
       metaPointer += (2 + metaSize)
     }
+    _metaPosition = capacity.toInt
```
velvia, line 164: I would put this in the free() method, which also resets the position.

broneill: This is actually the right place, in my opinion. Once the reclaim listener has been called, then the meta position is effectively reset.

PR #849, opened 2020-08-07
bug(query): use localPlanner when shard key filters does not have regex

TanviBhavsar:
**Pull Request checklist**

**Current behavior :** 

Shard key regex logic is applied even for queries that do not have shard key regex. For queries like `histogram_quantile(0.2, sum(http_requests))` PromQlRemoteExec is created which has InstantVectorFunctionMapper.
PromQlRemoteExec just gets results by making HTTP call for ` sum(http_requests)` to the remote endpoint. It does not apply rangevectorTransformers

**New behavior :**

When shard key filters do not have regex simply return localPlanner.materialize without executing shard key regex logic. This way PromQLRemoteExec is created for the entire query `histogram_quantile(0.2, sum(http_requests))`

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/queryplanner/ShardKeyRegexPlanner.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryplanner/ShardKeyRegexPlanner.scala
@@ -2,7 +2,6 @@ package filodb.coordinator.queryplanner
 
 import filodb.core.metadata.{Dataset, Schemas}
 import filodb.core.query.{ColumnFilter, PromQlQueryParams, QueryContext}
-import filodb.core.query.Filter.{EqualsRegex, NotEqualsRegex}
 import filodb.query._
 import filodb.query.exec._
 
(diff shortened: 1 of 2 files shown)


coordinator/src/main/scala/filodb.coordinator/queryplanner/ShardKeyRegexPlanner.scala
```diff
@@ -63,6 +64,12 @@ class ShardKeyRegexPlanner(dataset: Dataset,
     LogicalPlan.getRawSeriesFilters(logicalPlan)
       .map { s => s.filter(f => dataset.options.nonMetricShardColumns.contains(f.column))}
 
+  /**
+    * Returns true when shard key filters do not have regex
+    */
+  private def withoutShardKeyRegex(logicalPlan: LogicalPlan) = getNonMetricShardKeyFilters(logicalPlan).
+    forall(_.forall(f => !f.filter.isInstanceOf[EqualsRegex] && !f.filter.isInstanceOf[NotEqualsRegex]))
```
tjackpaul, line 71: It would be better if you can just check for `Equals` instead of doing other way. Just a suggestion.

PR #867, opened 2020-08-21
bug(query): binary join with AND has LHS range vectors when corresponding RHS range vector is empty   

TanviBhavsar:
**Pull Request checklist**

**Current behavior :** 
 Binary join with AND adds all LHS range vectors which have matching keys in RHS. When RHS has empty range vectors, keys are present with all values as NaN. AND  operator picks value from LHS  which can be non NaN. So final result has range vectors from LHS even if RHS is empty

**New behavior :**
Don't add Range vector if matching RHS range vector is empty


TanviBhavsar:
> Nice work root causing the bug !

Thanks Vish! I have addressed your review comment

PR #869, opened 2020-08-25
fix(prometheus): Parse labels without stack overflow.

broneill:
**Pull Request checklist**

**Current behavior :**
Parsing long quoted identifiers causes a stack overflow error due to the use of a regex.

**New behavior :**
Removed the regex and use a custom parser instead. This also means that no special escape character cleanup is required later.


prometheus/src/main/scala/filodb/prometheus/parse/Parser.scala
```diff
@@ -15,8 +19,52 @@ trait BaseParser extends Expressions with JavaTokenParsers with RegexParsers wit
...
+          return Failure("quote character expected", in)
+        }
+
+        val bob = new StringBuilder()
+
+        while (offset < source.length) {
+          var c = source.charAt(offset); offset += 1
+
+          if (c == quote) {
+            return Success(Identifier(bob.toString()), in.drop(offset - in.offset))
```
vishramachandran, line 46: How is a string like `"dfsd"df` treated ?

prometheus/src/main/scala/filodb/prometheus/parse/Parser.scala
```diff
@@ -15,8 +19,52 @@ trait BaseParser extends Expressions with JavaTokenParsers with RegexParsers wit
     "[a-zA-Z_:][a-zA-Z0-9_:\\-\\.]*".r ^^ { str => Identifier(str) }
   }
 
-  protected lazy val labelValueIdentifier: PackratParser[Identifier] =
-    "([\"'])(?:\\\\\\1|.)*?\\1".r ^^ { str =>  Identifier(str.substring(1, str.size-1)) } //remove quotes
+  protected lazy val labelValueIdentifier: PackratParser[Identifier] = {
```
vishramachandran, line 22: Please add comment on why you had to write a custom parser here for posterity.

whizkido: Yes , please .

broneill: Because it's easier to read than a regex with 6 backslashes in a row? ;-)

prometheus/src/main/scala/filodb/prometheus/parse/Parser.scala
broneill: It should stop the identifier parsing at the quote, and the trailing df is handled by the next parsing step. It will likely fail.

PR #872, opened 2020-08-26
bug(query): AND returns LHS row values if corresponding RHS row value is NaN

TanviBhavsar:
**Pull Request checklist**

**Current behavior :**
 AND returns LHS row values if corresponding RHS row value is not present which is represented by NaN

**New behavior :**
Add LHS row to result only when corresponding RHS row is not NaN


vishramachandran (changes requested):
As discussed, please address the consumption of rhs range vector rows due to call of `isEmpty` method call.

vishramachandran (approved):
Please explain in comments why you are asserting for SRV.

PR #878, opened 2020-09-04
bug(query): lucene matches timeseries for notEquals filter even if the column is not present

TanviBhavsar:
**Pull Request checklist**

**Current behavior :** 
Filters like `container != ""` returns timeseries which do not have container tag 

**New behavior :**
Change lucene query to only have timeseries which have the tag present


tjackpaul (approved):

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -451,9 +451,9 @@ class PartKeyLuceneIndex(ref: DatasetRef,
         new TermQuery(term)
       case NotEquals(value) =>
         val term = new Term(column, value.toString)
-        val allDocs = new MatchAllDocsQuery
         val booleanQuery = new BooleanQuery.Builder
-        booleanQuery.add(allDocs, Occur.FILTER)
+        val termAll = new Term(column, ".*")
+        booleanQuery.add(new RegexpQuery(termAll, RegExp.NONE), Occur.FILTER)
```
vishramachandran, line 456: Is there a different way to express this, than use a RegExp query ? Typically regexp queries are costly.

What happens if you remove that Query from the filter ?

TanviBhavsar: If I remove the RegexpQuery it does not return result even when `!=` condition is satisfied. Like if input has `"host" -> "host1"` it does not return any results for query having `ColumnFilter("host", Filter.NotEquals("host2".utf8))`or `ColumnFilter("host", Filter.NotEquals(""))`

TanviBhavsar: Also added a test case. It fails if I remove the RegexpQuery from the BooleanQuery

vishramachandran (approved):

PR #880, opened 2020-09-09
bug(core): Revert _step_ predefined key since we already use it; fix binary joins

vishramachandran:
**Pull Request checklist**

`_step_` tag was earlier introduced as predefined tag. As a result new partKeys are created which aren't equivalent to old ones. In addition, binary joins have issues since the one-side in OneToOne or OneToMany joins now have duplicate join keys.

**New behavior :**

* Rename `_step_` tag with unused `_pi_` (stands for publish interval). Test case included to ensure name update works fine.
* Include tags like `_step_` and `_pi_` in the `on` specifier of joins

### DIFF
--- a/core/src/main/resources/filodb-defaults.conf
+++ b/core/src/main/resources/filodb-defaults.conf
@@ -25,9 +25,12 @@ filodb {
     columns = ["_metric_:string", "tags:map"]
 
     # Predefined keys allow space to be saved for over the wire tags with the given keys
-    # WARN: It is suggested to only ADD to predefined keys.
-    #       Changing/Renaming existing keys will cause incorrect tags to be reported and probably require a clean wipe.
-    predefined-keys = ["_ws_", "_ns_", "app", "__name__", "instance", "dc", "le", "job", "exporter", "_step_"]
+    # WARN: 1. It is suggested to only ADD to predefined keys that are not used already,
+    #          which means only new tags with underscore prefixed and suffixed.
+    #       2. Changing/Renaming existing keys will cause incorrect tags to be reported and probably require a clean wipe.
+    #       3. Even when adding new keys, be careful. If that tag is already being used, there is no
+    #          equality of partKeys between old and new time series.
+    predefined-keys = ["_ws_", "_ns_", "app", "__name__", "instance", "dc", "le", "job", "exporter", "_pi_"]
 
     options {
       # Copy tags can be used to create a new labelPair from one of the existing labelPair in a TimeSeries.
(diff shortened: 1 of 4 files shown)


query/src/main/scala/filodb/query/exec/BinaryJoinExec.scala
```diff
@@ -51,11 +51,11 @@ final case class BinaryJoinExec(queryContext: QueryContext,
   require(!on.contains(metricColumn), "On cannot contain metric name")
 
   val onLabels = on.map(Utf8Str(_)).toSet
+  val withExtraOnLabels = onLabels ++ Seq("_pi_", "_step_") // needed since these are internal tags
```
tjackpaul, line 54: Adding only new tag ‘pi’ is not sufficient? 

Also, do we have to keep this forever? Or for a short term would suffice?

tjackpaul (approved):
lgtm.. thanks for the quick fix.. 

had a question.. added the same in comment

core/src/main/resources/filodb-defaults.conf
```diff
@@ -25,9 +25,12 @@ filodb {
...
     # Predefined keys allow space to be saved for over the wire tags with the given keys
-    # WARN: It is suggested to only ADD to predefined keys.
-    #       Changing/Renaming existing keys will cause incorrect tags to be reported and probably require a clean wipe.
-    predefined-keys = ["_ws_", "_ns_", "app", "__name__", "instance", "dc", "le", "job", "exporter", "_step_"]
+    # WARN: 1. It is suggested to only ADD to predefined keys that are not used already,
+    #          which means only new tags with underscore prefixed and suffixed.
+    #       2. Changing/Renaming existing keys will cause incorrect tags to be reported and probably require a clean wipe.
+    #       3. Even when adding new keys, be careful. If that tag is already being used, there is no
+    #          equality of partKeys between old and new time series.
+    predefined-keys = ["_ws_", "_ns_", "app", "__name__", "instance", "dc", "le", "job", "exporter", "_pi_"]
```
broneill, line 33: Comments for what these special keys mean? I cannot find any references to `_pi_`

query/src/main/scala/filodb/query/exec/BinaryJoinExec.scala
vishramachandran: Technically yes. But this issue will also occur when customers change the publish interval themselves. Joins will fail and they have to provide this internal tag explicitly. Instead, we could do it automatically for them.

core/src/main/resources/filodb-defaults.conf
vishramachandran: `_pi_` stands for Publish interval. It is intentionally a new unused tag name so the new time series can be differentiated from the ones that already have `_step_` tag.

vishramachandran: Will add docs.

velvia:
I strongly recommend that predefined keys NOT be removed from the list in defaults.   Change the gateway instead to not use it, but removing something creates the possibility that there could exist two possible meanings in persisted part keys, which could lead to corruption.   While you could use deployment controls etc to guard against this, as a matter of principle, the predefined-keys list should be treated as append-only.

TanviBhavsar (approved):

vishramachandran:
@velvia Thanks for the heads-up. We were aware of the consequences while proceeding with this. It was a necessary change.

PR #894, opened 2020-09-17
measure(query): Measure time elapsed for different points in execplan

vishramachandran:
**Pull Request checklist**

More metrics to measure time elapsed since start of query execution.

Very useful to identify at which point query execution is stalling.

### DIFF
--- a/query/src/main/scala/filodb/query/exec/BinaryJoinExec.scala
+++ b/query/src/main/scala/filodb/query/exec/BinaryJoinExec.scala
@@ -2,6 +2,7 @@ package filodb.query.exec
 
 import scala.collection.mutable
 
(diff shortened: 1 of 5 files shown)


broneill (commented):
Will milliseconds provide enough precision? Will you need to keep narrowing down further? This might end up looking like manual instrumentation, which adds a lot of overhead and complexity. Can you run a test in which you run with an instrumenting profiler instead?

tjackpaul (approved):

vishramachandran:
@broneill as discussed in person, the intent is to narrow down the issues seen during load testing. It is not reproducible locally. If we find that these metrics aren't useful longer term, we can remove them later after the debug effort.

PR #903, opened 2020-09-23
bug(query): BinaryJoinExec is created for queries having nested scalar expressions

TanviBhavsar:
**Pull Request checklist**

**Current behavior :** 
BinaryJoinExec is created for queries having nested scalar expressions like `foo + 10/2`

**New behavior :**
Create ScalarVectorBinaryOperation for such queries


vishramachandran (approved):
Thanks. Lets add the query to functional tests.

prometheus/src/main/scala/filodb/prometheus/ast/Expressions.scala
```diff
@@ -95,26 +96,35 @@ trait Expressions extends Aggregates with Functions {
...
-
-            BinaryJoin(seriesPlanLhs, operator.getPlanOperator, cardinality, seriesPlanRhs,
-              onLabels.getOrElse(Nil), ignoringLabels.getOrElse(Nil),
-              vectorMatch.flatMap(_.grouping).map(_.labels).getOrElse(Nil))
-
+            //10/2 + foo
+            if (hasScalarResult(lh)) {
+              val scalar = lh.toSeriesPlan(timeParams).asInstanceOf[ScalarPlan]
+              val seriesPlan = rh.toSeriesPlan(timeParams)
+              ScalarVectorBinaryOperation(operator.getPlanOperator, scalar, seriesPlan, scalarIsLhs = true)
```
vishramachandran, line 103: if both LHS and RHS are scalar, it come here. Wouldn't it be wrong ? It should be a scalar/vector operation

TanviBhavsar: If both LHS and RHS are scalars it will go to the if block above(starting from line 45) and it will be converted to ScalarBinaryOperation

vishramachandran: Ack. Please ensure we have a test case to verify that.

TanviBhavsar: Yes we have 4 test cases in ScalarQueriesSpec

PR #911, opened 2020-09-30
fix(core,memory): Allow blocks to be reclaimed when ODP fails.

broneill:
**Pull Request checklist**

**Current behavior :** 
When ODP fails to allocate, it leaves the metadata span open, and the blocks it contains aren't marked as reclaimable. The next time a span is started, the existing blocks within the span are lost.

**New behavior :**
When ODP fails, an explicit discard method is called to mark the blocks as reclaimable.

### DIFF
--- a/memory/src/main/scala/filodb.memory/MemFactory.scala
+++ b/memory/src/main/scala/filodb.memory/MemFactory.scala
@@ -319,7 +319,11 @@ class BlockMemFactory(blockStore: BlockManager,
   def discardMetaSpan(): Unit = {
     if (metadataSpanActive) {
       metadataSpan.foreach { blk =>
-        blk.markReclaimable()
+        if (markFullBlocksAsReclaimable) {
+          blk.markReclaimable()
+        } else synchronized {
+          fullBlocks += blk
+        }
       }
       metadataSpan.clear()
       metadataSpanActive = false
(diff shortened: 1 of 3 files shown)


memory/src/main/scala/filodb.memory/MemFactory.scala
```diff
@@ -311,6 +311,21 @@ class BlockMemFactory(blockStore: BlockManager,
...
 
+  /**
+    * Discards the current metadata span, to be called when a failure occurs during allocation.
+    * The blocks are then marked as reclaimable, and they are fully reclaimed later by the
+    * headroom task.
+    */
+  def discardMetaSpan(): Unit = {
+    if (metadataSpanActive) {
+      metadataSpan.foreach { blk =>
+        blk.markReclaimable()
```
vishramachandran, line 322: Is this correct ? Should it be marked as reclaimable ?
`markFullBlocksAsReclaimable` needs to be checked?

broneill: I had that check originally, but it seemed unnecessary.

broneill: When discard is called, it's due to an exception. Keeping blocks around that might have nothing in it doesn't make sense. In practice, it won't make a difference because the mark full mode is only used with ODP, and discard isn't needed for ingestion.

broneill: I'll flip it back, at least for consistency.

vishramachandran (approved):
Thanks for the iterations!

PR #912, opened 2020-10-01
fix(core): Shutdown process if headroom task is failing.

broneill:
**Pull Request checklist**

**Current behavior :**
If memory cannot be reclaimed (perhaps a memory leak), the process must be manually restarted.

**New behavior :**
Exit and restart automatically after 5 minutes.

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
+++ b/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
@@ -1500,7 +1500,21 @@ class TimeSeriesShard(val ref: DatasetRef,
 
   private def startHeadroomTask(sched: Scheduler): Unit = {
     sched.scheduleWithFixedDelay(1, 1, TimeUnit.MINUTES, new Runnable {
-      def run() = blockStore.ensureHeadroom(storeConfig.ensureHeadroomPercent)
+      var numFailures = 0
+
+      def run() = {
+        val numFree = blockStore.ensureHeadroom(storeConfig.ensureHeadroomPercent)
+        if (numFree > 0) {
+          numFailures = 0
+        } else {
+          numFailures += 1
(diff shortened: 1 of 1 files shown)


core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -1500,7 +1500,21 @@ class TimeSeriesShard(val ref: DatasetRef,
...
+      def run() = {
+        val numFree = blockStore.ensureHeadroom(storeConfig.ensureHeadroomPercent)
+        if (numFree > 0) {
+          numFailures = 0
+        } else {
+          numFailures += 1
+          if (numFailures >= 5) {
+            logger.error(s"Headroom task was unable to free memory for $numFailures consecutive attempts. " +
+              s"Shuting down process. shard=$shardNum")
+            Runtime.getRuntime.halt(1)
```
vishramachandran, line 1514: Can we use a unique exit code, so we can identify cause from compute infra logs ?
Would also be nice to Prefix ALL shutdown log statements with a common term like "SHUTTING DOWN" so it can be searched in logs

vishramachandran (approved):

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
vishramachandran: Can we dump additional debug information here?
For example latch state, time ordered vs raw blocks used etc.

broneill: We'd need to define an error code registry, which seems out of scope. I'd rather not rely on capturing such codes anyhow, since it depends on environment specifics. For now, I'll call into the existing haltAndCatchFire method.

sherali42 (approved):
👍

PR #914, opened 2020-10-05
fix(memory): Forcibly reclaim time-ordered-blocks when running out of…

broneill:
… memory.

**Pull Request checklist**

**Current behavior :**
Can run out of memory even though time-ordered blocks still exist. This can be caused by unknown failures which leave the block in an unreclaimable state.

**New behavior :**
Forcibly reclaim time-ordered blocks as a last resort.


vishramachandran (approved):
This is a temporary fix. We need to understand why the blocks are not being marked as reclaimable when it is supposed to be by design. There is a remnant bug.

PR #915, opened 2020-10-06
bug(core): Mark ingestion blocks reclaimable on cassandra failure

vishramachandran:
**Pull Request checklist**

* Blocks are not marked reclaimable if there is a cassandra error, but they are released back to pool where they will never be marked as reclaimable.
* When we run out of memory in ingestion pipeline, we continue without cleanup

**New behavior :**

* Mark blocks as reclaimable even though the chunks are not completely flushed. 
* Shutdown process if we run out of ingestion memory.

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
+++ b/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
@@ -961,8 +961,11 @@ class TimeSeriesShard(val ref: DatasetRef,
     result.onComplete { resp =>
       assertThreadName(IngestSchedName)
       try {
-        // Mark used blocks as reclaimable even on failure. Even if cassandra write fails or other errors occur,
-        // we cannot leave blocks as not reclaimable and also release the factory back into pool.
+        // COMMENTARY ON BUG FIX DONE: Mark used blocks as reclaimable even on failure. Even if cassandra write fails
+        // or other errors occur, we cannot leave blocks as not reclaimable and also release the factory back into pool.
+        // Earlier, we were not calling this with the hope that next use of the blockMemFactory will mark them
+        // as reclaimable. But the factory could be used for a different flush group. Not the same one. It can
+        // succeed, and the wrong blocks can be marked as reclaimable.
         // Can try out tracking unreclaimed blockMemFactories without releasing, but it needs to be separate PR.
         blockHolder.markUsedBlocksReclaimable()
         blockFactoryPool.release(blockHolder)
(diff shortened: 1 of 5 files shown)


memory/src/main/scala/filodb.memory/data/Shutdown.scala
```diff
@@ -0,0 +1,15 @@
+package filodb.memory.data
+
+import com.typesafe.scalalogging.StrictLogging
+
+object Shutdown extends StrictLogging {
+
+  def haltAndCatchFire(e: Exception, unitTest: Boolean = false): Unit = {
+    if (unitTest) {
+      throw new Error(e)
```
broneill, line 9: Why wrap it with Error? Just throw it back out.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -962,6 +961,10 @@ class TimeSeriesShard(val ref: DatasetRef,
     result.onComplete { resp =>
       assertThreadName(IngestSchedName)
       try {
+        // Mark used blocks as reclaimable even on failure. Even if cassandra write fails or other errors occur,
+        // we cannot leave blocks as not reclaimable and also release the factory back into pool.
+        // Can try out tracking unreclaimed blockMemFactories without releasing, but it needs to be separate PR.
+        blockHolder.markUsedBlocksReclaimable()
```
broneill, line 967: What if the used blocks aren't full yet? They might be reclaimed concurrently while filling up more.

memory/src/main/scala/filodb.memory/data/Shutdown.scala
vishramachandran: I just carried over old code. You are right though. I will change.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
vishramachandran: This block holder is used only by this flush task - no one else because it is checked out of the pool.
So no one else should be filling these blocks.
Once the task is complete, we accept unused portions as defragmentation cost.

vishramachandran: The absence of markUsedBlocksReclaimable  on flush failure will result in bug. The released factory will be reused for a different flush task, and these blocks will be marked as reclaimable anyway if that task succeeds (even if the chunks are not flushed). It is simpler to mark them as reclaimable rightaway anyway.

Yes - there is a risk of us reclaiming the chunks before flush. But this change doesnt make it worse.

broneill (approved):

PR #918, opened 2020-10-07
bug(memory): Fix mark reclaimable bugs; Remove Time-ordered-blocks; ADD UNIT TEST!

vishramachandran:
**Pull Request checklist**

* There is a leak in memory caused due to blocks not being marked as reclaimable
* Not reclaiming ODP blocks quickly leads to more holes
* Mark as Reclaimable bug and Hole expansion together leads to ODP amplification.

**New behavior :**

* Fix bug where blocks are not marked as reclaimable when full
* Remove time ordered blocks. Replace ODP block mem factories per time bucket with just one block-mem-factory.
* Reclaim logic will reclaim ALL ODP blocks first (except currently used block) before reclaiming any ingestion block. Concurrency between ODP queries and reclaims are protected with reclaimlock - this way paged data is not reclaimed before query completes.
* The intention here is to keep ODPed data until memory saturates. Once that happens, ODPed data stays in memory momentarily until query completes.
* Hole expansion logic continues to help by discarding chunk map pointers when newer paged data is reclaimed before older paged data (this can occur due to query order).

**Added much needed unit test for BlockMemFactory - validates logic of marking as reclaimable**


broneill (approved):

vishramachandran:
BTW, The implications of this PR is that time ordered blocks are expected be reclaimed very quickly after allocation once headroom reclaim kicks in. This implies that once off-heap memory is saturated, ODPed data will not stay in memory for long. There are implications for clusters holding long term data in memory and rely more on ODP. This PR will not help those use cases.

Furthermore, I want to clarify that further simplification can be done for this clear predictable behavior where ODP blocks are reclaimed before Ingestion blocks. There is actually no need for time ordered blocks. The original intended time ordered blocks for time ordered reclamation has a lot of issues. If we decide to simplify code further for improved maintainability, it has to be a different PR.

vishramachandran:
@whizkido @broneill I have removed time ordered blocks completely and simplified to a single deque of ODP blocks.

broneill (approved):

memory/src/main/scala/filodb.memory/BlockManager.scala
```diff
@@ -221,8 +203,8 @@ class PageAlignedBlockManager(val totalMemorySizeInBytes: Long,
         val allocated = new Array[Block](num)
         (0 until num).foreach { i =>
           val block = freeBlocks.remove()
-          if (bmf.nonEmpty) block.setOwner(bmf.get)
-          use(block, bucketTime)
+          if (ownerBmf.nonEmpty) block.setOwner(ownerBmf.get)
+          use(block, odp)
```
sherali42, line 207: this block will be executed only if `odp = false`, otherwise `MemoryRequestException` gets thrown, right? and `use` method is not being used anywhere else.

vishramachandran: No - The foreach block is executed `num` number of times. It essentially takes the blocks from "free blocks list" and places them in the "used block list". It will be run for both ODP and ingestion.

whizkido (approved):
I approve this pr. 

We need to do the heap analysis for this so as to confirm the behavior . Lets do this in dev once merged.

PR #926, opened 2020-10-21
bug(query): predence is not applied based on brackets

TanviBhavsar:
**Pull Request checklist**

**Current behavior :**
For query like `(metric1 + metric2) * (metric3 + metric4)`  multiplication is executed before addition

**New behavior :**
Add PrecedenceExpression to allow brackets to override operator precedence


prometheus/src/main/scala/filodb/prometheus/ast/Expressions.scala
```diff
@@ -41,10 +43,22 @@ trait Expressions extends Aggregates with Functions {
...
+       case p: PrecedenceExpression  => p.expression
+       case _                        => lhs
+
+     }
+
+      val rhsWithPrecedence = rhs match {
+        case p: PrecedenceExpression  => p.expression
+        case _                        => rhs
+
+      }
```
vishramachandran, line 56: Why don't you fold this logic into hasScalarResult ?

vishramachandran (approved):
Approved with suggestion

prometheus/src/main/scala/filodb/prometheus/ast/Expressions.scala
TanviBhavsar: As discussed we need it for subsequent case statements

PR #955, opened 2020-11-10
bug(query): "next on empty iterator" when using sum over topk result

TanviBhavsar:
**Pull Request checklist**

**Current behavior :** 
If we execute topK query with k as 1 from t1 to t5 we just get max value for every timestamp from all range vectors. So if for a times series only t1 has top value, t2, t3 etc won't be present in final result. When sum is applied on such a query `next on empty iterator` error is thrown as some timestamps are missing.

**New behavior :**
Add NaN for timestamps which do not have any value in topk result

### DIFF
--- a/query/src/main/scala/filodb/query/exec/aggregator/TopBottomKRowAggregator.scala
+++ b/query/src/main/scala/filodb/query/exec/aggregator/TopBottomKRowAggregator.scala
@@ -1,6 +1,7 @@
 package filodb.query.exec.aggregator
 
-import scala.collection.mutable
+import scala.collection.{ mutable}
+import scala.collection.mutable.ListBuffer
 
 import filodb.core.binaryrecord2.RecordBuilder
 import filodb.core.memstore.FiloSchedulers
@@ -24,6 +25,10 @@ class TopBottomKRowAggregator(k: Int, bottomK: Boolean) extends RowAggregator {
   private val rvkStringCache = mutable.HashMap[RangeVectorKey, ZeroCopyUTF8String]()
 
   case class RVKeyAndValue(rvk: ZeroCopyUTF8String, value: Double)
+  val colSchema = Seq(ColumnInfo("timestamp", ColumnType.TimestampColumn),
+    ColumnInfo("value", ColumnType.DoubleColumn))
+  val recSchema = SerializedRangeVector.toSchema(colSchema)
+
   class TopKHolder(var timestamp: Long = 0L) extends AggregateHolder {
     val valueOrdering = Ordering.by[RVKeyAndValue, Double](kr => kr.value)
     implicit val ordering = if (bottomK) valueOrdering else valueOrdering.reverse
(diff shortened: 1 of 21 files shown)


query/src/main/scala/filodb/query/exec/aggregator/TopBottomKRowAggregator.scala
```diff
@@ -81,27 +86,46 @@ class TopBottomKRowAggregator(k: Int, bottomK: Boolean) extends RowAggregator {
     acc
   }
 
-  def present(aggRangeVector: RangeVector, limit: Int): Seq[RangeVector] = {
-    val colSchema = Seq(ColumnInfo("timestamp", ColumnType.TimestampColumn),
-      ColumnInfo("value", ColumnType.DoubleColumn))
-    val recSchema = SerializedRangeVector.toSchema(colSchema)
+  private def addRecordToBuilder(builder: RecordBuilder, value: Double, timeStampMs: Long): Unit = {
```
vishramachandran, line 89: nit: timestamp before value

query/src/main/scala/filodb/query/exec/aggregator/TopBottomKRowAggregator.scala
```diff
@@ -81,27 +86,46 @@ class TopBottomKRowAggregator(k: Int, bottomK: Boolean) extends RowAggregator {
...
-            builder.startNewRecord(recSchema)
-            builder.addLong(row.getLong(0))
-            builder.addDouble(row.getDouble(i + 1))
-            builder.endRecord()
+            rvkSeen += rvk
+            val builder = resRvs.getOrElseUpdate(rvk, createBuilder(rangeParams, t))
+            addRecordToBuilder(builder, row.getDouble(i + 1), row.getLong(0))
           }
-          i += 2
+           i += 2
```
vishramachandran, line 125: nit: indent

query/src/main/scala/filodb/query/exec/aggregator/TopBottomKRowAggregator.scala
```diff
@@ -81,27 +86,46 @@ class TopBottomKRowAggregator(k: Int, bottomK: Boolean) extends RowAggregator {
...
           if (row.filoUTF8String(i) != CustomRangeVectorKey.emptyAsZcUtf8) {
             val rvk = CustomRangeVectorKey.fromZcUtf8(row.filoUTF8String(i))
-            val builder = resRvs.getOrElseUpdate(rvk, SerializedRangeVector.newBuilder())
-            builder.startNewRecord(recSchema)
-            builder.addLong(row.getLong(0))
-            builder.addDouble(row.getDouble(i + 1))
-            builder.endRecord()
+            rvkSeen += rvk
+            val builder = resRvs.getOrElseUpdate(rvk, createBuilder(rangeParams, t))
+            addRecordToBuilder(builder, row.getDouble(i + 1), row.getLong(0))
```
vishramachandran, line 123: `row.getLong(0)` is same as `t*1000` right ? Can just be easier to read ?

query/src/test/scala/filodb/query/exec/AggrOverRangeVectorsSpec.scala
```diff
@@ -518,7 +525,31 @@ class AggrOverRangeVectorsSpec extends RawDataWindowingSpec with ScalaFutures {
     result(0).rows.map(_.getDouble(1)).toList shouldEqual counts
   }
 
-  it("should sum and compute max of histogram & max RVs") {
+  it ("should add NaN in topK") {
+
+    val samples: Array[RangeVector] = Array(
+      toRv(Seq((1000L,5.1), (2000L, 5.6d), (3000L, 4.0d))),
+      toRv(Seq((1000L,5.0), (2000L, 5.7d), (3000L, 4.4d)), ignoreKey2)
+    )
```
vishramachandran, line 533: Can you make sure this test case includes NaN in the beginning, NaN in the middle and NaN at the end of presented range vector?

vishramachandran (approved):
Approved with minor comments

TanviBhavsar:
> Approved with minor comments

Thanks Vish! I have addressed your comments

PR #962, opened 2020-11-13
feat(core): add addititional condition to create chunks at flush interval boundary, only when enabled

sherali42:
**Pull Request checklist**

switch buffers and create an encoded chunk when current sample's timestamp crosses flush boundary.  e.g. for a flush-interval of 1hour, if new sample falls in different hour than last sample, then switch buffers and create chunk. This helps in aligning chunks across Active/Active HA clusters and facilitates chunk migration between the clusters during disaster recovery.
Note: Enabling this feature might result into creation of smaller suboptimal chunks.

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
+++ b/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
@@ -1184,7 +1184,8 @@ class TimeSeriesShard(val ref: DatasetRef,
         val tsp = part.asInstanceOf[TimeSeriesPartition]
         brRowReader.schema = schema.ingestionSchema
         brRowReader.recordOffset = recordOff
-        tsp.ingest(ingestionTime, brRowReader, overflowBlockFactory, maxChunkTime)
+        tsp.ingest(ingestionTime, brRowReader, overflowBlockFactory,
+          storeConfig.createChunkAtFlushBoundary, Option(flushBoundaryMillis), maxChunkTime)
         // Below is coded to work concurrently with logic in updateIndexWithEndTime
         // where we try to de-activate an active time series
         if (!tsp.ingesting) {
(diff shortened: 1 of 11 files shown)


core/src/main/scala/filodb.core/store/IngestionConfig.scala
```diff
@@ -42,6 +43,7 @@ final case class StoreConfig(flushInterval: FiniteDuration,
   import collection.JavaConverters._
   def toConfig: Config =
     ConfigFactory.parseMap(Map("flush-interval" -> (flushInterval.toSeconds + "s"),
+                               "create-chunk-at-flush-boundary" -> createChunkAtFlushBoundary,
```
vishramachandran, line 46: Rename to `time-aligned-chunks-enabled`

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -1184,7 +1184,8 @@ class TimeSeriesShard(val ref: DatasetRef,
         val tsp = part.asInstanceOf[TimeSeriesPartition]
         brRowReader.schema = schema.ingestionSchema
         brRowReader.recordOffset = recordOff
-        tsp.ingest(ingestionTime, brRowReader, overflowBlockFactory, maxChunkTime)
+        tsp.ingest(ingestionTime, brRowReader, overflowBlockFactory,
+          storeConfig.createChunkAtFlushBoundary, Option(flushBoundaryMillis), maxChunkTime)
```
vishramachandran, line 1188: Dont create a new `Option` object for each sample. Will result in too many objects. Create a val and reuse.

vishramachandran (changes requested):
One important change needed, and I can approve. Thanks!

core/src/main/scala/filodb.core/store/IngestionConfig.scala
sherali42: fixed the name

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
sherali42: update the code, please take a look.

vishramachandran (approved):

PR #988, opened 2021-01-06
bug(cass): Retry with exponential backoff when scanning

vishramachandran:
**Pull Request checklist**

Token scan from ingestion index table during downsample use paged iterator, and this sometimes times out causing entire task to repeat causing more writes, more compaction on cassandra, and more read timeouts.

Putting in place retry with exponential backoff.

### DIFF
--- a/cassandra/src/main/scala/filodb.cassandra/Util.scala
+++ b/cassandra/src/main/scala/filodb.cassandra/Util.scala
@@ -3,9 +3,11 @@ package filodb.cassandra
 import java.nio.ByteBuffer
 
 import scala.concurrent.{Future, Promise}
+import scala.util.Random
 
 import com.datastax.driver.core._
-import com.datastax.driver.core.exceptions.DriverException
+import com.datastax.driver.core.exceptions.{DriverException, ReadTimeoutException}
+import com.typesafe.scalalogging.StrictLogging
 import monix.reactive.Observable
 
 import filodb.core._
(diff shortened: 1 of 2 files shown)


cassandra/src/main/scala/filodb.cassandra/Util.scala
```diff
@@ -84,4 +86,35 @@ object Util {
...
+  //scalastyle:off null
+  override def next(): Row = {
+    var retries = 0
+    val maxRetries = 5 // Hardcode maxRetries for now. If needed make it a config later.
+    var nxt: Row = null
+    while (retries < maxRetries && nxt == null) {
+      try {
+        nxt = inner.next()
+      } catch {
+        case e: ReadTimeoutException =>
```
zznate, line 108: We still have to think about UnavailableException in this context. This is a bit more tricky because we are dealing directly with the semantics of consistency and hiding behavior here may be a surprise to the user. Definitely should be bubbled up into a big, giant comment and in docs regardless.

I think for now, dealing with just the RTE is the easiest fix as that will avoid the common case of retrying when it just takes a little longer to churn through. 

For reference, there is a good explanation on the different types of exceptions on the Java-Driver's documentation about retry logic: https://github.com/datastax/java-driver/tree/3.x/manual/retries#onreadtimeout

cassandra/src/main/scala/filodb.cassandra/Util.scala
```diff
@@ -84,4 +86,35 @@ object Util {
...
+          retries += 1
+          if (retries == 5) throw e
+          val jitter = Random.nextInt(3000)
+          val sleepTime = Math.pow(2, retries + 1).toLong * 1000 + jitter
```
zznate, line 112: Yeah, this is fine otherwise. Nice and simple with a ceiling.

zznate (commented):
Suggestions/ideas on handing other exception types.

sherali42 (approved):
👍

PR #949, opened 2020-11-06
feat(repair): spark job to migrate partitionKeys in case of disaster recovery

sharath-raghavan:
**Modifications**
- Added a Spark job that can be triggered on demand to copy partitionKey records.  
- Time interval can be configured as specified below. Spark job would copy all the records falling in this interval.
```
conf.set("spark.filodb.partitionkeys.copier.ingestionTimeStart", "2020-10-13T00:00:00Z")
conf.set("spark.filodb.partitionkeys.copier.ingestionTimeEnd", "2020-10-13T05:00:00Z")
```
- This job would be useful in cases of disaster recovery or back fills.

**Pull Request checklist**

- [ ] The commit(s) message(s) follows the contribution [guidelines](CONTRIBUTING.md) ?
- [ ] Tests for the changes have been added (for bug fixes / features) ?
- [ ] Docs have been added / updated (for bug fixes / features) ?

**Current behavior :** 
- NA

**New behavior :**
- data repair feature is new. This job would be useful in cases of disaster recovery or back fills.

**BREAKING CHANGES**
- NA

**Other information**:

PR #993, opened 2021-01-19
Index copier tweaks

aandis:
### Modifications

Addresses review comments from #949.

* https://github.com/filodb/FiloDB/pull/949#discussion_r520932534 - use the existing config for getting `splitsPerNode` value.
* https://github.com/filodb/FiloDB/pull/949#discussion_r520937150 - write data using `CassandraColumnStore.writePartKeys` to set part key ttl based on end time.
* https://github.com/filodb/FiloDB/pull/949#discussion_r520942378 - write data to `PartitionKeysByUpdateTimeTable` along with writing to part keys table.

**Pull Request checklist**


cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
```diff
@@ -248,16 +253,19 @@ extends ColumnStore with CassandraChunkSource with StrictLogging {
...
+          case Some(targetPkr) => pkRecordWithHash(compareAndGet(partKeyRecord, targetPkr))
+          case None => pkRecordWithHash(partKeyRecord)
         }
-      }
+      )
+
+      val updateHour = System.currentTimeMillis() / 1000 / 60 / 60
+      Await.result(
+        writePartKeys(targetDatasetRef, shard, Observable.fromIterable(partKeys), diskTimeToLiveSeconds, updateHour),
+        Duration.Inf // TODO - Use cassandra write timeout.
```
aandis, line 267: This needs to be updated.

spark-jobs/src/main/scala/filodb/repair/PartitionKeysCopier.scala
```diff
@@ -54,15 +54,22 @@ class PartitionKeysCopier(conf: SparkConf) {
...
+  private val targetDatasetConfig = datasetConfig(targetConfig)
   private val sourceDatasetRef = DatasetRef.fromDotString(sourceDataset)
   private val targetDatasetRef = DatasetRef.fromDotString(conf.get("spark.filodb.partitionkeys.copier.target.dataset"))
   private val sourceSession = FiloSessionProvider.openSession(sourceCassConfig)
   private val targetSession = FiloSessionProvider.openSession(targetCassConfig)
 
-  private val numOfShards: Int = getShardNum
+  val schemas = Schemas.fromConfig(sourceConfig).get
+  private[repair] def partKeyHashFn = (partKey: PartKeyRecord) =>
+    Option(schemas.part.binSchema.partitionHash(partKey, UnsafeUtils.arayOffset))
```
sherali42, line 66: this is incorrect, it should be
`Option(schemas.part.binSchema.partitionHash(partKey.partKey, UnsafeUtils.arayOffset))`

core/src/main/scala/filodb.core/store/ChunkSink.scala
```diff
@@ -13,7 +13,18 @@ import monix.reactive.Observable
 
 import filodb.core._
 
-case class PartKeyRecord(partKey: Array[Byte], startTime: Long, endTime: Long, hash: Option[Int])
+case class PartKeyRecord(partKey: Array[Byte], startTime: Long, endTime: Long, hash: Option[Int]) {
+  override def equals(obj: Any): Boolean = {
```
sherali42, line 17: why do we need to override these?

sherali42 (changes requested):

core/src/main/scala/filodb.core/store/ChunkSink.scala
aandis: We needed to override equals in `PartKeyRecord` because - 
1. When you do `scanPartKeysByStartTime` and `scanPartKeysByEndTime`, you get duplicates.
1. To get rid of duplicates we put them in a Set.
1. The default equals method that comes with the PartKeyRecord caseClass does not work because dsPartKey inside it is an array which is compared by reference. So the comparison fails even for two identical part key records

spark-jobs/src/main/scala/filodb/repair/PartitionKeysCopier.scala
aandis: fixed in 9491bcd

aandis: fixed in 9491bcd

cassandra/src/main/scala/filodb.cassandra/columnstore/CassandraColumnStore.scala
```diff
@@ -237,43 +237,47 @@ extends ColumnStore with CassandraChunkSource with StrictLogging {
...
+    for (split <- splits; shard <- 0 until numOfShards) {
+      val tokens = split.asInstanceOf[CassandraTokenRangeSplit].tokens
+      val srcPartKeysTable = getOrCreatePartitionKeysTable(datasetRef, shard)
+      val targetPartKeysTable = target.getOrCreatePartitionKeysTable(targetDatasetRef, shard)
+      // CQL does not support OR operator. So we need to query separately to get the timeSeries partitionKeys
+      // which were born or died during the data loss period (aka repair window).
+      val rowsByStartTime = srcPartKeysTable.scanRowsByStartTimeRangeNoAsync(tokens, repairStartTime, repairEndTime)
+      val rowsByEndTime = srcPartKeysTable.scanRowsByEndTimeRangeNoAsync(tokens, repairStartTime, repairEndTime)
+      // add to a Set to eliminate duplicate entries.
+      val records = rowsByStartTime.++(rowsByEndTime).map(PartitionKeysTable.rowToPartKeyRecord)
```
vishramachandran, line 278: Try your best to use iterator instead of bringing everything into memory here.
There are times when contents of a single split may blow out the heap allocation. These things tend to happen at wrong times.

It may be ok to pay the cost of writing same part key twice into cassandra - that may be an acceptable trade off for the few number of such PKs that both start and end in the repair period. 

This will eliminate the need to implement equals on `PartitionKeyRecord` as well.

vishramachandran (approved):
Thanks for following up on Sharath's PR. Conditionally approved - we need to remove the memory requirement.

sherali42 (approved):

PR #1001, opened 2021-01-29
bug(query): Stitch and uniqify LHS and RHS before binary join

vishramachandran:
**Pull Request checklist**

Range Vectors need to be stitched and made distinct before joins because duplicates can exist due to spread changes. Duplicates cause errors in join matching.

### DIFF
--- a/query/src/test/scala/filodb/query/exec/BinaryJoinGroupingSpec.scala
+++ b/query/src/test/scala/filodb/query/exec/BinaryJoinGroupingSpec.scala
@@ -211,14 +211,7 @@ class BinaryJoinGroupingSpec extends AnyFunSpec with Matchers with ScalaFutures 
     val result = execPlan.compose(Observable.fromIterable(Seq((rhs, 1), (lhs, 0))), tvSchemaTask, querySession)
       .toListL.runAsync.futureValue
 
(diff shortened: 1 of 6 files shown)


query/src/test/scala/filodb/query/exec/BinaryJoinGroupingSpec.scala
```diff
@@ -308,12 +309,12 @@ class BinaryJoinGroupingSpec extends AnyFunSpec with Matchers with ScalaFutures
       )
     )
     result.size shouldEqual 4
-    result.map(_.key.labelValues) sameElements(expectedLabels) shouldEqual true
+    result.map(_.key.labelValues).toSet shouldEqual expectedLabels.toSet
 
-    result(0).rows.map(_.getDouble(1)).toList shouldEqual List(0.75)
```
TanviBhavsar, line 313: Rows ordering has changed. Won't expectedLabels also change?

TanviBhavsar (approved):
Looks good to me. Just a minor comment

query/src/test/scala/filodb/query/exec/BinaryJoinGroupingSpec.scala
vishramachandran: RV ordering has changed. RV rows should remain same. ExpectedLabels should remain same as well.
These changes only account for RV order change

PR #1013, opened 2021-02-04
bug(query): stitch vectors in LHS and RHS during binary join with set operator

TanviBhavsar:
**Pull Request checklist**

Optimally stitching same RVs from multiple shards while performing Binary Join for LAND, OR and LUNLESS

### DIFF
--- a/query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
+++ b/query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
@@ -1,13 +1,11 @@
 package filodb.query.exec
 
 import scala.collection.mutable
-import scala.collection.mutable.ListBuffer
-
+import scala.collection.mutable.ArrayBuffer
 import kamon.Kamon
 import kamon.metric.MeasurementUnit
 import monix.eval.Task
 import monix.reactive.Observable
-
 import filodb.core.memstore.SchemaMismatch
 import filodb.core.query._
 import filodb.memory.format.{RowReader, ZeroCopyUTF8String => Utf8Str}
@@ -15,6 +13,7 @@ import filodb.memory.format.ZeroCopyUTF8String._
 import filodb.query._
 import filodb.query.BinaryOperator.{LAND, LOR, LUnless}
 
+case class ResultVal(resultRv: RangeVector, stitchedOtherSide: RangeVector)
 /**
   * Set operator between results of lhs and rhs plan.
   *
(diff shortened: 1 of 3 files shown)


query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
```diff
@@ -133,39 +155,72 @@ final case class SetOperatorExec(queryContext: QueryContext,
...
       } else if (jk.isEmpty) {
         // "up AND ON (dummy) vector(1)" should be equivalent to up as there's no dummy label
-        result += lhs
+        val arrayBuffer = result.getOrElse(jk, ArrayBuffer())
+        arrayBuffer.append(lhs)
+        result.put(jk, arrayBuffer)
       }
     }
-    result.toList
+    result.values.flatten
```
vishramachandran, line 169: See if you can improve allocation by returning iterator of Range Vectors. Eventually, you can also create Observable from Iterator. This will reduce O(n) allocation.

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
```diff
@@ -97,25 +97,47 @@ final case class SetOperatorExec(queryContext: QueryContext,
...
     rhsRvs.foreach { rv =>
       val jk = joinKeys(rv.key)
       // Don't add range vector if it is empty
-      if (jk.nonEmpty && !isEmpty(rv, rhsSchema))
-        rhsMap.put(jk, rv)
+      if (jk.nonEmpty && !isEmpty(rv, rhsSchema)) {
+        if (rhsMap.contains(jk)) {
+          val resVal = rhsMap(jk)
+          if (resVal.key.labelValues == rv.key.labelValues) {
+            rhsMap.put(jk, StitchRvsExec.stitch(rv, resVal) )
```
vishramachandran, line 116: Lets document why we are stitching.. "When spread changes, we need to account for multiple Range Vectors with same key coming from different shards. Each of these range vectors would contain data for different time ranges". I missed adding this piece of information in the BinaryJoinExec as well clearly. Do you mind adding there too ?

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
```diff
@@ -133,39 +155,72 @@ final case class SetOperatorExec(queryContext: QueryContext,
...
+        arrayBuffer.append(lhs)
+        result.put(jk, arrayBuffer)
       }
     }
-    result.toList
+    result.values.flatten
   }
+  // scalastyle:on method.length
 
   private def setOpOr(lhsRvs: List[RangeVector]
```
vishramachandran, line 173: This is outside the scope of this PR, but good to add while we are at it. Do you mind adding English description of what each of AND, OR and UNLESS must do along with nuances? This helps because we keep forgetting and it takes some reverse engineering effort to recall - docs will make it easier. Thanks much!

vishramachandran (approved):
Approved with comments. Thank you very much!

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
TanviBhavsar: Added description

vishramachandran (approved):
Thanks for the changes!

PR #1021, opened 2021-02-10
bug(core): Overflow BlockMemFactory is not reclaimed correctly after flush

vishramachandran:
**Pull Request checklist**

Overflow BlockMemFactory in TimeSeriesShard today depends on ODP blocks and hence is not correctly reclaimed after flush of chunks placed in it.

This PR removes overflow BMF in the shard and makes the BlockMemFactory pool track the flush group for which the
BMF is used. Subsequent use in overflow chunks will retrieve same BMF, until the BMF is checked out for flush in the FlushEvent for the group. At that point, it is removed from the map and future calls for overflow BMF will yield a new BMF.
Once the BMF checked out for flush is released after flush to Cassandra, the BMF is put back into free pool.

### DIFF
--- a/memory/src/main/scala/filodb.memory/BlockMemFactoryPool.scala
+++ b/memory/src/main/scala/filodb.memory/BlockMemFactoryPool.scala
@@ -1,5 +1,7 @@
 package filodb.memory
 
+import scala.collection.mutable
+
 import com.typesafe.scalalogging.StrictLogging
 
 /**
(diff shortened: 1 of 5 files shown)


memory/src/main/scala/filodb.memory/BlockMemFactoryPool.scala
```diff
@@ -15,32 +17,54 @@ class BlockMemFactoryPool(blockStore: BlockManager,
                           metadataAllocSize: Int,
                           baseTags: Map[String, String]) extends StrictLogging {
   private val factoryPool = new collection.mutable.Queue[BlockMemFactory]()
+  private val checkedOut = new mutable.HashMap[Int, BlockMemFactory]()
 
   def poolSize: Int = factoryPool.length
```
broneill, line 22: Should make this synchronized too while you're at it.

memory/src/main/scala/filodb.memory/BlockMemFactoryPool.scala
```diff
@@ -15,32 +17,54 @@ class BlockMemFactoryPool(blockStore: BlockManager,
...
-    fact.tags = baseTags ++ moreTags
-    fact
   }
 
+  /**
+   * Checks out a BlockMemFactory for flush. Subsequent call for fetch for this
+   * flush group will return new BlockMemFactory
+   */
+  def checkoutForFlush(flushGroup: Int): BlockMemFactory = synchronized {
+    checkedOut.remove(flushGroup).getOrElse(fetchForOverflow(flushGroup))
```
broneill, line 51: What if remove returns null?

memory/src/main/scala/filodb.memory/BlockMemFactoryPool.scala
```diff
@@ -15,32 +17,54 @@ class BlockMemFactoryPool(blockStore: BlockManager,
...
     }
-    fact.tags = baseTags ++ moreTags
-    fact
   }
 
+  /**
+   * Checks out a BlockMemFactory for flush. Subsequent call for fetch for this
+   * flush group will return new BlockMemFactory
+   */
+  def checkoutForFlush(flushGroup: Int): BlockMemFactory = synchronized {
```
broneill, line 50: Perhaps rename to "removeForFlush"?

broneill: Not critical.

memory/src/main/scala/filodb.memory/BlockMemFactoryPool.scala
vishramachandran: remove does not return null. It returns a None which will cause new BMF to be returned

memory/src/main/scala/filodb.memory/BlockMemFactoryPool.scala
vishramachandran: will rename per in-person discussion

memory/src/main/scala/filodb.memory/BlockMemFactoryPool.scala
vishramachandran: 👍

broneill (approved):

PR #1027, opened 2021-02-16
misc(core): fix character encoding to use UTF-8 instead of platform default

sherali42:
**Pull Request checklist**

In few places Strings use platform default character encoding. This change is to use `UTF-8` encoding.

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/ratelimit/RocksDbCardinalityStore.scala
+++ b/core/src/main/scala/filodb.core/memstore/ratelimit/RocksDbCardinalityStore.scala
@@ -1,6 +1,7 @@
 package filodb.core.memstore.ratelimit
 
 import java.io.File
+import java.nio.charset.StandardCharsets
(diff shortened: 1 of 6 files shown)


broneill (approved):

core/src/main/scala/filodb.core/memstore/ratelimit/RocksDbCardinalityStore.scala
```diff
@@ -228,7 +229,7 @@ class RocksDbCardinalityStore(ref: DatasetRef, shard: Int) extends CardinalitySt
 
     breakable {
       while (it.isValid()) {
-        val key = new String(it.key())
+        val key = new String(it.key(), StandardCharsets.UTF_8)
```
vishramachandran, line 232: encoding also needs to use UTF8

vishramachandran (changes requested):

PR #1053, opened 2021-03-12
feat(query): absent over time function support

TanviBhavsar:
**Pull Request checklist**

Supoort for `absent_over_time()`


coordinator/src/main/scala/filodb.coordinator/queryplanner/PlannerMaterializer.scala
```diff
@@ -126,20 +125,27 @@ trait  PlannerMaterializer {
...
-    def materializeAbsentFunction(qContext: QueryContext,
+   def addAbsentFunctionMapper(vectors: PlanResult,
+                               columnFilters: Seq[ColumnFilter],
+                               rangeParams: RangeParams,
+                               queryContext: QueryContext): PlanResult = {
+    if (vectors.plans.length > 1) {
+      val targetActor = pickDispatcher(vectors.plans)
+      val topPlan = LocalPartitionDistConcatExec(queryContext, targetActor, vectors.plans)
+      topPlan.addRangeVectorTransformer(AbsentFunctionMapper(columnFilters, rangeParams,
+        dsOptions.metricColumn))
```
vishramachandran, line 136: Should not need to concat. You can add the RVT to each child plan. It will get concatenated if needed eventually.

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
```diff
@@ -284,7 +284,10 @@ class SingleClusterPlanner(dsRef: DatasetRef,
       series.plans.foreach(_.addRangeVectorTransformer(PeriodicSamplesMapper(newStartMs, lp.stepMs,
         lp.endMs, window, Some(execRangeFn), qContext, lp.stepMultipleNotationUsed,
         paramsExec, lp.offsetMs, rawSource)))
-      series
+      if (execRangeFn == InternalRangeFunction.AbsentOverTime)
+        addAbsentFunctionMapper(series, lp.columnFilters,
+        RangeParams(lp.startMs / 1000, lp.stepMs / 1000, lp.endMs / 1000), qContext)
+      else series
```
vishramachandran, line 290: There is a leak in abstraction. There is some hidden magic (translation of absentOverTime to last sample) that is happening during execution, that this planner depends on. It is hard for reader concerned with planning to understand whats going on without digging very deep into implementation. 

Instead, we should plan with LastSample and apply the mapper right here so that the "magic" is visible in one place. 

Please also add a short description on the secret sauce of range function + existing mapper to achieve absent over time.

prometheus/src/main/scala/filodb/prometheus/ast/Functions.scala
```diff
@@ -180,12 +180,13 @@ trait Functions extends Base with Operators with Vectors {
...
         } else {
           val rangeExpression = seriesParam.asInstanceOf[RangeExpression]
+
           PeriodicSeriesWithWindowing(
             rangeExpression.toSeriesPlan(timeParams, isRoot = false),
             timeParams.start * 1000 , timeParams.step * 1000, timeParams.end * 1000,
             rangeExpression.window.millis(timeParams.step * 1000),
             rangeFunctionId, rangeExpression.window.timeUnit == IntervalMultiple,
-            otherParams, rangeExpression.offset.map(_.millis(timeParams.step * 1000)))
+            otherParams, rangeExpression.offset.map(_.millis(timeParams.step * 1000)), rangeExpression.columnFilters)
```
vishramachandran, line 189: I didnt get this change. Why does AbsentFunctionMapper need columnFilters ?

vishramachandran (changes requested):
Looks good. Needs some reorganizing. Thanks!

coordinator/src/main/scala/filodb.coordinator/queryplanner/PlannerMaterializer.scala
TanviBhavsar: If we apply AbsentFunctionMapper before concatenation of child responses results can be incorrect. Let's say some_metric has child results some_metric{instance="inst-0"} and some_metric{instance="inst-1"} and we execute some_metric{instance="inst-1"}. If at t1 inst=0 is present but inst-1 is not present,  AbsentFunctionMapper will return some_metric{instance="inst-1"}. When we concatenate final output will return  some_metric{instance="inst-1"} which is incorrect. Since some_metric{instance="inst-0"} is present absent(some_metric) should not return anything

prometheus/src/main/scala/filodb/prometheus/ast/Functions.scala
TanviBhavsar: absent_over_time creates range vector with column filters in original query.

coordinator/src/main/scala/filodb.coordinator/queryplanner/PlannerMaterializer.scala
vishramachandran: Are we doing the concatenation for the sake of label names ? We have to evaluate other means.
Premature concatenation is a performance issue. For example, if sum is done over these time series, all time series will be moved to one node and then aggregated instead of map-reduce. Lets talk on this one.

broneill:
The Antlr parser will need changes too.

broneill:
Actually, the Antlr parser won't need changes since you're only adding a new function. It should work just fine.

vishramachandran (approved):

vishramachandran:
Please add some docs in the planner to clarify that absent related functions, although arent documented as aggregation, perform aggregation implicitly. It will be confusing and non-obvious if not clarified

PR #1057, opened 2021-03-16
feat(core): Support for enabling/disabling Partial Query Results

vishramachandran:
**Pull Request checklist**

This PR enables input `partialResultsOk` flag. Output result now contains a `mayBePartial` flag accompanying a reason.

* Today we continue with query even if index / kafka recovery is not complete yielding partial results. If partialResults are enabled, we continue query processing. Otherwise, query is aborted with error.
* AskTimeoutException from one or more shards  errors entire query today. If `partialResultsOk` input flag is enabled, then AskTimeoutException is converted to EmptyResult and query processing continues.


query/src/main/scala/filodb/query/exec/SelectChunkInfosExec.scala
```diff
@@ -42,6 +43,13 @@ final case class SelectChunkInfosExec(queryContext: QueryContext,
   def doExecute(source: ChunkSource,
                 querySession: QuerySession)
                (implicit sched: Scheduler): ExecResult = {
+    if (!source.isReadyForQuery(dataset, shard)) {
+      if (!queryContext.plannerParams.partialResultsOk) {
+        throw new ServiceUnavailableException(s"Unable to answer query since shard $shard is still bootstrapping")
+      }
+      querySession.resultCouldBePartial = true
+      querySession.partialResultsReason = Some(s"Shard $shard still bootstrapping")
```
tjackpaul, line 51: Should this be list to avoid overriding when multiple shards are still bootstrapping ?

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
```diff
@@ -112,7 +111,13 @@ final case class LabelValuesExec(queryContext: QueryContext,
   def doExecute(source: ChunkSource,
                 querySession: QuerySession)
                (implicit sched: Scheduler): ExecResult = {
-    val parentSpan = Kamon.currentSpan()
+    if (!source.isReadyForQuery(dataset, shard)) {
+      if (!queryContext.plannerParams.partialResultsOk) {
+        throw new ServiceUnavailableException(s"Unable to answer query since shard $shard is still bootstrapping")
+      }
+      querySession.resultCouldBePartial = true
+      querySession.partialResultsReason = Some(s"Shard $shard still bootstrapping")
```
tjackpaul, line 119: Should this be list to avoid overriding when multiple shards are still bootstrapping ?

core/src/main/scala/filodb.core/query/QueryContext.scala
```diff
@@ -31,7 +31,8 @@ case class PlannerParams(applicationId: String = "filodb",
                         splitSizeMs: Long = 1.day.toMillis,
                         skipAggregatePresent: Boolean = false,
                         processFailure: Boolean = true,
-                        processMultiPartition: Boolean = false)
+                        processMultiPartition: Boolean = false,
+                        partialResultsOk: Boolean = false)
```
tjackpaul, line 35: Rename this to `blockPartialResults` or `allowPartialResults`

query/src/main/scala/filodb/query/ResultTypes.scala
```diff
@@ -49,7 +51,9 @@ object QueryResultType extends Enum[QueryResultType] {
 
 final case class QueryResult(id: String,
                              resultSchema: ResultSchema,
-                             result: Seq[RangeVector]) extends QueryResponse {
+                             result: Seq[RangeVector],
+                             mayBePartial: Boolean = false,
+                             partialResultReason: Option[String] = None) extends QueryResponse {
```
tjackpaul, line 56: It should be a seq/list to keep adding multiple messages from each shard?

query/src/main/scala/filodb/query/ResultTypes.scala
```diff
@@ -49,7 +51,9 @@ object QueryResultType extends Enum[QueryResultType] {
 
 final case class QueryResult(id: String,
                              resultSchema: ResultSchema,
-                             result: Seq[RangeVector]) extends QueryResponse {
+                             result: Seq[RangeVector],
+                             mayBePartial: Boolean = false,
```
tjackpaul, line 55: `isPartial`

core/src/main/scala/filodb.core/query/QueryContext.scala
```diff
@@ -83,7 +84,9 @@ object QueryContext {
   */
 case class QuerySession(qContext: QueryContext,
                         queryConfig: QueryConfig,
-                        var lock: Option[Lock] = None) {
+                        var lock: Option[Lock] = None,
+                        var resultCouldBePartial: Boolean = false,
```
tjackpaul, line 88: `isPartial`

core/src/main/scala/filodb.core/query/QueryContext.scala
```diff
@@ -83,7 +84,9 @@ object QueryContext {
...
+                        var partialResultsReason: Option[String] = None) {
```
tjackpaul, line 89: It should be a seq/list to keep adding multiple messages from each shard?

tjackpaul (commented):
thanks for the quick PR @vishramachandran .. left few minor comments.. also , does it handle passing the partialResultReason across multiple clusters for the various stitching use-cases ?

query/src/main/scala/filodb/query/exec/SelectChunkInfosExec.scala
vishramachandran: I think one reason is good enough. Intent is to keep it simple. Any specific case where all reasons must be listed?

query/src/main/scala/filodb/query/ResultTypes.scala
vishramachandran: It is intentionally called `mayBePartial` and not `isPartial`. We can never be sure because the missed results could be empty.

tjackpaul: > We can never be sure because the missed results could be empty.

That means `mayBePartial` is always true with what you say? In this case if we are specifically targeting not answering queries or returning partial results when shards are not active, should the flag be specific about that instead of `mayBePartial` ?

query/src/main/scala/filodb/query/exec/SelectChunkInfosExec.scala
tjackpaul: When more than 1 shards are not active, why can not we return both the shard numbers in reason. It is against QueryResult object and not per timeseries, so it is already simple in that perspective. Sorry i did not get what do you mean by `keep it simple` here.

tjackpaul: Ok. As discussed, we can have a generic message here.

query/src/main/scala/filodb/query/ResultTypes.scala
tjackpaul: Ok. As discussed, it may be partial if the timeseries partitions required for the query is already loaded.

core/src/main/scala/filodb.core/query/QueryContext.scala
tjackpaul: Not needed, in favor of generic message.

query/src/main/scala/filodb/query/ResultTypes.scala
tjackpaul: Not needed, in favor of generic message.

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
tjackpaul: Not needed, in favor of generic message.

tjackpaul (approved):

PR #1072, opened 2021-03-30
bug(query): precedenceExpression is not removed for child expressions

TanviBhavsar:
**Pull Request checklist**

- [x ] Tests for the changes have been added (for bug fixes / features) ?

**Current behavior :** 
`Second parameter to aggregate operator sum should be a vector, is instead PrecedenceExpression(InstantExpression(Some(some_metric),List(),None))`  is thrown for queries like sum((some_metric))

**New behavior :**
Extract inner expression for all such queries


prometheus/src/main/scala/filodb/prometheus/parse/Parser.scala
```diff
@@ -386,8 +386,9 @@ trait Expression extends Aggregates with Selector with Numeric with Join {
     }
 
   lazy val expression: PackratParser[Expression] =
-    binaryExpression | subqueryExpression | aggregateExpression2 | aggregateExpression1 |
-      function | unaryExpression | vector | numericalExpression | simpleSeries | precedenceExpression
+    binaryExpression | aggregateExpression2 | aggregateExpression1 |
+      function | unaryExpression | vector | numericalExpression | simpleSeries | precedenceExpression |
+      "(" ~> expression <~ ")"
```
vishramachandran, line 391: `subqueryExpression` got removed

TanviBhavsar: Ohh yes, missed it while rebasing. Thanks for letting me know!

prometheus/src/main/scala/filodb/prometheus/parse/Parser.scala
```diff
@@ -387,7 +387,8 @@ trait Expression extends Aggregates with Selector with Numeric with Join {
 
   lazy val expression: PackratParser[Expression] =
     binaryExpression | subqueryExpression | aggregateExpression2 | aggregateExpression1 |
-      function | unaryExpression | vector | numericalExpression | simpleSeries | precedenceExpression
+      function | unaryExpression | vector | numericalExpression | simpleSeries | precedenceExpression |
+      "(" ~> expression <~ ")"
```
broneill, line 391: What does this magic syntax mean?

TanviBhavsar: It is for queries like `(sum(some_metric))`

broneill: I mean, the magic "~>" and "<~" symbols. What do they mean?

TanviBhavsar: According to parsing.combinator, `p ~> q succeeds if `p` succeeds and `q` succeeds on the input  left over by `p`. `q` gets executed only if p is true and it returns result of q. Similarly `p <~ q` returns the result of p on success

https://github.com/scala/scala-parser-combinators/blob/2e43707531264ef30d0a623c5189939ff54841ca/shared/src/main/scala/scala/util/parsing/combinator/Parsers.scala#L307

broneill: 👍

broneill (approved):

sherali42 (approved):

PR #1074, opened 2021-04-01
bug(query):  remove BinaryExpression validation from constructor and assign precedence to LHS 

TanviBhavsar:
**Pull Request checklist**

**Current behavior :** 
`count((some_metric / 300) >= 1 ) or vector(0)`

- Since precedence is assigned after parsing is done above  query throws `comparisons between scalars must use BOOL modifier` when BinaryExpression is created
- In the above query, LHS is not assigned precedence

**New behavior:**

-  BinaryExpression validation is done after object is created
- In a BinaryExpression, precedence is assigned to LHS also to handle nested Binary Expression's as above


prometheus/src/main/scala/filodb/prometheus/parse/Parser.scala
```diff
@@ -485,7 +485,7 @@ object Parser extends Expression {
       case f: Function             => f.copy(allParams = f.allParams.map(assignPrecedence(_)))
       case a: AggregateExpression  => a.copy(params = a.params.map(assignPrecedence(_)), altFunctionParams = a.
                                      altFunctionParams.map(assignPrecedence(_)))
-      case b: BinaryExpression     => assignPrecedence(b.lhs, b.operator, b.vectorMatch, b.rhs)
+      case b: BinaryExpression     => assignPrecedence(assignPrecedence(b.lhs), b.operator, b.vectorMatch, b.rhs)
```
vishramachandran, line 488: No need to assign precedence to RHS ?

vishramachandran (approved):
approved with comment.

prometheus/src/main/scala/filodb/prometheus/parse/Parser.scala
TanviBhavsar: It is assigned in `assignPrecedence` function

PR #1081, opened 2021-04-09
perf(query): NaN Optimization in SerializedRangeVector

vishramachandran:
**Pull Request checklist**

**Current behavior :** 

When doing binary joins on high churn time series over longer time ranges, we send a lot of time series over the wire. Since these time series are ephemeral, we end up writing too many NaNs over the wire on the SerializedRangeVector. This increases network usage as well as heap usage on both ends.

**New behavior :**

We compress the SerializedRangeVector by including start/step/end in the range vector and omitting the NaNs entirely. While reading the values out at the receiver, we uncompress by inserting NaNs at appropriate steps where there is no value.

Work in Progress:
* Unit tests
* Local E2E tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ jdk:
   - openjdk11
 
 script:
-  - sbt ++$TRAVIS_SCALA_VERSION -Dakka.test.timefactor=3 -XX:MaxMetaspaceSize=512m -sbt-launch-repo https://repo1.maven.org/maven2 test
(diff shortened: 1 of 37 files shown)


broneill:
Is the approach to compress the data as the vector is being built, or at the point it's written? On the receiving side, is the data fully decompressed before processed, or is the data incrementally decompressed and thrown away as needed?

vishramachandran:
It is a trivial and cheap compression. NaN rows are thrown away. Thats it.
At decompression time, NaNs are reintroduced in the row iterator at appropriate points.

TanviBhavsar (approved):

PR #1083, opened 2021-04-13
perf(core): Parallel bootstrap into lucene index for downsample cluster

vishramachandran:
**Pull Request checklist**

Parallel additions into lucene index for downsample cluster

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/IndexBootstrapper.scala
+++ b/core/src/main/scala/filodb.core/memstore/IndexBootstrapper.scala
@@ -25,13 +25,12 @@ class IndexBootstrapper(colStore: ColumnStore) {
     * @param assignPartId the function to invoke to get the partitionId to be used to populate the index record
     * @return number of updated records
     */
-  def bootstrapIndex(index: PartKeyLuceneIndex,
-                     shardNum: Int,
-                     ref: DatasetRef)
-                     (assignPartId: PartKeyRecord => Int): Task[Long] = {
+  def bootstrapIndexRaw(index: PartKeyLuceneIndex,
+                        shardNum: Int,
+                        ref: DatasetRef)
+                       (assignPartId: PartKeyRecord => Int): Task[Long] = {
 
-    val recoverIndexLatency = Kamon.gauge("shard-recover-index-latency",
-      MeasurementUnit.time.milliseconds)
+    val recoverIndexLatency = Kamon.gauge("shard-recover-index-latency", MeasurementUnit.time.milliseconds)
       .withTag("dataset", ref.dataset)
       .withTag("shard", shardNum)
     val start = System.currentTimeMillis()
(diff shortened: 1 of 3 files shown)


core/src/main/scala/filodb.core/memstore/IndexBootstrapper.scala
```diff
@@ -25,13 +25,12 @@ class IndexBootstrapper(colStore: ColumnStore) {
     * @param assignPartId the function to invoke to get the partitionId to be used to populate the index record
     * @return number of updated records
     */
-  def bootstrapIndex(index: PartKeyLuceneIndex,
-                     shardNum: Int,
-                     ref: DatasetRef)
-                     (assignPartId: PartKeyRecord => Int): Task[Long] = {
+  def bootstrapIndexRaw(index: PartKeyLuceneIndex,
```
tjackpaul, line 28: Do we really have to rename this? `raw` does not seem to be a correct keyword here.

core/src/main/scala/filodb.core/memstore/IndexBootstrapper.scala
```diff
@@ -48,6 +47,36 @@ class IndexBootstrapper(colStore: ColumnStore) {
...
+                     shardNum: Int,
+                     ref: DatasetRef)
+                    (assignPartId: PartKeyRecord => Int): Task[Long] = {
+
+    val recoverIndexLatency = Kamon.gauge("shard-recover-index-latency", MeasurementUnit.time.milliseconds)
+      .withTag("dataset", ref.dataset)
+      .withTag("shard", shardNum)
+    val start = System.currentTimeMillis()
+    colStore.scanPartKeys(ref, shardNum)
+      .mapAsync(Runtime.getRuntime.availableProcessors()) { pk =>
```
tjackpaul, line 66: Should it be a config based instead which will help us in testing with different value?

tjackpaul (approved):
lgtm.. thanks for the quick pr @vishramachandran.. left a minor comment.. if we can change this to a config based, we can continue using single method itself with default parallelism as 1...

core/src/main/scala/filodb.core/memstore/IndexBootstrapper.scala
vishramachandran: I didn't want config overload. Better to cross that bridge later if really needed. This default seems ok.

core/src/main/scala/filodb.core/memstore/IndexBootstrapper.scala
vishramachandran: Raw was used since it is used for raw data cluster. The name change was intended to indicate the purpose of each method.

tjackpaul: Since raw was not the only other usecase after DS, i asked that question. By default a cluster can be used for anything: for example; to store rules, to store aggregated data, for raw data.

PR #1086, opened 2021-04-14
bug(query):  absent/absent_over_time return empty result for nonexistent metric

TanviBhavsar:
**Pull Request checklist**

**Current behavior :** 
AbsentFunctionMapper does not get executed for metrics that are not present 

**New behavior :**
- Remove AgggregatePresent from sum aggregator which is added for absent and absent_over_time. AbsentFunctionMapper is not executed if the schema is empty and the ExecPlan has rangeVectorTransformers that can't handle empty schema
- Return empty result in ReduceAggregateExec when input schema is empty

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/queryplanner/PlannerMaterializer.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryplanner/PlannerMaterializer.scala
@@ -183,7 +183,8 @@ trait  PlannerMaterializer {
     val aggregate = Aggregate(AggregationOperator.Sum, lp, Nil, Seq("job"))
     // Add sum to aggregate all child responses
     // If all children have NaN value, sum will yield NaN and AbsentFunctionMapper will yield 1
-    val aggregatePlanResult = PlanResult(Seq(addAggregator(aggregate, qContext, vectors, Seq.empty)))
+    val aggregatePlanResult = PlanResult(Seq(addAggregator(aggregate, qContext.copy(plannerParams =
+      qContext.plannerParams.copy(skipAggregatePresent = true)), vectors, Seq.empty))) // No need for present for sum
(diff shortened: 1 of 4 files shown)


coordinator/src/main/scala/filodb.coordinator/queryplanner/PlannerMaterializer.scala
```diff
@@ -183,7 +183,8 @@ trait  PlannerMaterializer {
     val aggregate = Aggregate(AggregationOperator.Sum, lp, Nil, Seq("job"))
     // Add sum to aggregate all child responses
     // If all children have NaN value, sum will yield NaN and AbsentFunctionMapper will yield 1
-    val aggregatePlanResult = PlanResult(Seq(addAggregator(aggregate, qContext, vectors, Seq.empty)))
+    val aggregatePlanResult = PlanResult(Seq(addAggregator(aggregate, qContext.copy(plannerParams =
+      qContext.plannerParams.copy(skipAggregatePresent = true)), vectors, Seq.empty))) // No need for present for sum
```
vishramachandran, line 187: Now that we have this, should the canHandleEmptyResult be removed from RangeVectorTransformer ?

vishramachandran (approved):
Approved with suggestion.

coordinator/src/main/scala/filodb.coordinator/queryplanner/PlannerMaterializer.scala
TanviBhavsar: We still need `canHandleEmptyResult` for AbsentFunctionMapper as the top level exec plan (ReduceAggrgateExec) will have AbsentFunctionMapper and it will not get executed for empty schema if canHandleEmptyResult is false

PR #1090, opened 2021-04-16
fix(core): map partkeys to labelvalues eagerly to avoid duplicate results

sherali42:
**Pull Request checklist**

- Currently LabelValueResultIterator lazily maps partIds and returns matching label/value pairs. This is causing duplicate label/value pairs in the leaf results and also prematurely breaching metadata query limits. The changes in this PR makes sure that the mapping/deduping is done when the first row is retrieved from the iterator.


core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
```diff
@@ -313,37 +314,41 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
...
       }
-      foundValue
+      rows.toSeq
     }
 
+    override def hasNext: Boolean = iterIndx < rows.size
+
     override def next(): Map[ZeroCopyUTF8String, ZeroCopyUTF8String] = {
-      numResultsReturned += 1
+      val currVal = rows(iterIndx)
```
vishramachandran, line 350: index access with hash set will not be efficient. Better to use simple iterator.

core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
```diff
@@ -313,37 +314,41 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
...
+    var iterIndx = 0
+    private lazy val rows = labelValues
 
-    override def hasNext: Boolean = {
-      var foundValue = false
-      while(partIndex < partIds.length && numResultsReturned < limit && !foundValue) {
-        val partId = partIds(partIndex)
+    def labelValues: Seq[Map[ZeroCopyUTF8String, ZeroCopyUTF8String]] = {
+      var partLoopIndx = 0
+      val rows = new mutable.HashSet[Map[ZeroCopyUTF8String, ZeroCopyUTF8String]]()
```
vishramachandran, line 327: Can you explain why we are using Set of Map here ?
Wont simple arraylist of values for each label work better here?

vishramachandran (changes requested):

core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
```diff
@@ -313,37 +314,41 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
...
 
         //retrieve PartKey either from In-memory map or from PartKeyIndex
         val nextPart = partKeyFromPartId(partId)
 
         // FIXME This is non-performant and temporary fix for fetching label values based on filter criteria.
         // Other strategies needs to be evaluated for making this performant - create facets for predefined fields or
         // have a centralized service/store for serving metadata
-        currVal = schemas.part.binSchema.toStringPairs(nextPart, UnsafeUtils.arayOffset)
+        val currVal = schemas.part.binSchema.toStringPairs(nextPart, UnsafeUtils.arayOffset)
           .filter(labelNames contains _._1).map(pair => {
```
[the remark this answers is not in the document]
sherali42: 👍 updated to use `colValues`

core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
sherali42: since each row is a collection of key/values, we need Map here

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -704,39 +705,38 @@ class TimeSeriesShard(val ref: DatasetRef,
...
         // have a centralized service/store for serving metadata
-        currVal = schemas.part.binSchema.toStringPairs(nextPart.base, nextPart.offset)
-          .filter(labelNames contains _._1).map(pair => {
-          pair._1.utf8 -> pair._2.utf8
-        }).toMap
-        foundValue = currVal.nonEmpty
-        partIndex += 1
+
+        val currVal = schemas.part.binSchema.colValues(nextPart.base, nextPart.offset, labelNames).
+          zipWithIndex.filter(_._1 != null).map{case(value, ind) => labelNames(ind).utf8 -> value.utf8}.toMap
```
vishramachandran, line 729: why would `_1` be null ?

sherali42: @vishramachandran we initialize an array with nulls inside `colValues`. If the requested labels are not there in the partkey, it will have nulls for those labels.
`val res = collection.mutable.ArrayBuffer.fill[String](cols.size)(UnsafeUtils.ZeroPointer.asInstanceOf[String])`.
So we need this check.

vishramachandran (approved):

PR #1091, opened 2021-04-16
bug(reset): increment reset for counter restart

TanviBhavsar:
**Pull Request checklist**

**Current behavior :** 
Reset value does not increase for samples like 10, NaN

**New behavior :**
Increment reset value for counter restart

### DIFF
--- a/query/src/main/scala/filodb/query/exec/rangefn/RangeInstantFunctions.scala
+++ b/query/src/main/scala/filodb/query/exec/rangefn/RangeInstantFunctions.scala
@@ -122,13 +122,14 @@ class ResetsFunction extends RangeFunction {
 
   def addedToWindow(row: TransientRow, window: Window): Unit = {
     val size = window.size
-    if (row.value.isNaN && !resets.isNaN) resets += 1
+    if (row.value.isNaN && !resets.isNaN && size > 1 && !window(size - 2).value.isNaN) resets += 1
     if (resets.isNaN && size > 0) resets = 0
     if (size > 1 && window(size - 2).value > row.value) resets += 1
   }
 
   def removedFromWindow(row: TransientRow, window: Window): Unit = {
-    if (window.size > 0 && row.value.isNaN) resets -= 1
+    val size = window.size
+    if (size > 1 && row.value.isNaN && !window(size - 2).value.isNaN) resets -= 1
     else if (window.size > 0 && row.value > window.head.value) resets -= 1
     else if (window.size == 0) resets = Double.NaN
   }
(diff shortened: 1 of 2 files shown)


query/src/main/scala/filodb/query/exec/rangefn/RangeInstantFunctions.scala
```diff
@@ -122,17 +122,15 @@ class ResetsFunction extends RangeFunction {
 
   def addedToWindow(row: TransientRow, window: Window): Unit = {
     val size = window.size
+    if (row.value.isNaN && !resets.isNaN) resets += 1
```
vishramachandran, line 125: (10, NaN, NaN, 5) would be counted as two because of this. Best is to simply treat NaN as 0 when comparing against previous or next value.

sherali42: true, two adjacent NaN's should not increment `resets`

query/src/main/scala/filodb/query/exec/rangefn/RangeInstantFunctions.scala
```diff
@@ -122,17 +122,15 @@ class ResetsFunction extends RangeFunction {
...
-    }
+    if (size > 1 && window(size - 2).value > row.value) resets += 1
   }
 
   def removedFromWindow(row: TransientRow, window: Window): Unit = {
-    if (window.size > 0 && row.value > window.head.value)
-      resets -= 1
-    else if (window.size == 0)
-      resets = Double.NaN
+    if (window.size > 0 && row.value.isNaN) resets -= 1
```
sherali42, line 131: same logic as `addedToWindow ` applicable here. two adjacent NaN's should not decrement `resets`

query/src/main/scala/filodb/query/exec/rangefn/RangeInstantFunctions.scala
TanviBhavsar: Made the change and added a test case. Thanks for letting me know!

query/src/main/scala/filodb/query/exec/rangefn/RangeInstantFunctions.scala
```diff
@@ -122,17 +122,20 @@ class ResetsFunction extends RangeFunction {
...
   def removedFromWindow(row: TransientRow, window: Window): Unit = {
-    if (window.size > 0 && row.value > window.head.value)
-      resets -= 1
-    else if (window.size == 0)
-      resets = Double.NaN
+    val currentValue = if (row.value.isNaN) 0 else row.value
+    if (window.size > 0) {
+      val prevValue = if (window.head.value.isNaN) 0 else window.head.value
+    if (currentValue > prevValue) resets -= 1
+    } else if (window.size == 0) resets = Double.NaN
```
vishramachandran, line 138: fix indentation. `else if` not needed. `else` is good enough

vishramachandran (approved):

sherali42 (approved):
👍

PR #1095, opened 2021-04-22
bug(core): Fix Partition Eviction Issues

vishramachandran:
**Pull Request checklist**

Currently, we use lucene index to identify partIds eligible for eviction. This is not proving to be reliable. Getting top-k partitions with oldest endTime from lucene is complex and is complex possibly has bugs. Even if fixed, it yields to complex eviction logic. There are cases when ODP may cause the partition to never be evicted at all because of existing watermark logic.

Hence:
* Explicitly maintain a queue of partIds eligible for eviction. When partition stops ingesting or is ODPed, the partId is added to this list
* Eviction happens very quickly from this queue. Lucene queries will not be an issue. ODPed partitions get evicted first to be consistent with block eviction.

Another issue is we still have ODP cannibalism where ODP queries can silently evict other TSPs required by the same ODP query. This is prevented by disallowing inline eviction for ODP queries. Instead, we evict regularly in the headroom task to create room for ODPed partitions.

### DIFF
--- a/memory/src/main/scala/filodb.memory/BlockManager.scala
+++ b/memory/src/main/scala/filodb.memory/BlockManager.scala
@@ -254,16 +254,6 @@ class PageAlignedBlockManager(val totalMemorySizeInBytes: Long,
   }
 
   /**
-   * Calculate lock timeout based on headroom space available.
-   * Lower the space available, longer the timeout.
-   */
-  def getHeadroomLockTimeout(ensurePercent: Double): Int = {
-    // Ramp up the timeout as the current headroom shrinks. Max timeout per attempt is a little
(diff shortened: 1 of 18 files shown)


memory/src/main/scala/filodb.memory/BlockManager.scala
```diff
@@ -225,15 +227,15 @@ class PageAlignedBlockManager(val totalMemorySizeInBytes: Long,
     try {
       val start = System.nanoTime()
       // Give up after waiting (in total) a little over 16 seconds.
```
broneill, line 229: nit: comment is now wrong

broneill (approved):

PR #1124, opened 2021-05-27
fix(standalone): shutdown server on initialization errors

aandis:
Fixes the following -

1. Server not shutting down on initialization errors.
2. NPE during shutdown due to http server not being initialised.

**Pull Request checklist**

### DIFF
--- a/standalone/src/main/scala/filodb.standalone/FiloServer.scala
+++ b/standalone/src/main/scala/filodb.standalone/FiloServer.scala
@@ -1,7 +1,6 @@
 package filodb.standalone
 
 import scala.concurrent.duration._
-import scala.util.control.NonFatal
 
 import akka.actor.ActorRef
 import akka.cluster.Cluster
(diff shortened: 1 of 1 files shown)


standalone/src/main/scala/filodb.standalone/FiloServer.scala
```diff
@@ -90,10 +91,6 @@ class FiloServer(watcher: Option[ActorRef]) extends FilodbClusterNode {
 
 object FiloServer extends StrictLogging {
   def main(args: Array[String]): Unit = {
-    try {
-      new FiloServer().start()
-    } catch { case e: Exception =>
-      logger.error("Could not start FiloDB server", e)
-    }
+    new FiloServer().start()
```
vishramachandran, line 94: We need a force shutdown here. There is a possibility of daemon threads holding up the process.

aandis: The force shutdown is already present at https://github.com/filodb/FiloDB/pull/1124/files#diff-c50749986da7e22c6df371808ceb7bfed314508b866b9a0db2d9bcd4727165e1R72-R75

Do we still need to add it here?

vishramachandran: Ah got it. Do you mind moving the try-catch here to the uppermost call - that way any future additions here will also benefit from it. Thanks.

vishramachandran (approved):
Approved with suggestion

PR #1127, opened 2021-05-29
feat(query): don't error histogram aggregation with heterogenous bucket scheme

TanviBhavsar:
**Pull Request checklist**

**Current behavior :** 
Exception is thrown for the entire query when histogram schema is not same even for a single instant 

**New behavior :**
Return NaN for timestamps where schema does not match and return aggregation result for other timestamps

### DIFF
--- a/memory/src/main/scala/filodb.memory/format/vectors/Histogram.scala
+++ b/memory/src/main/scala/filodb.memory/format/vectors/Histogram.scala
@@ -216,11 +216,10 @@ final case class MutableHistogram(buckets: HistogramBuckets, values: Array[Doubl
 
   /**
    * Adds the values from another Histogram.
-   * If the other histogram has the same bucket scheme, then the values are just added per bucket.
-   * If the scheme is different, then an approximation is used so that the resulting histogram has
-   * an approximate sum of the individual distributions, with the original scheme.  Modifies itself.
+   * If the other histogram has the same bucket scheme, then the values are just added per bucket and true is returned.
+   * If the scheme is different, it assigns NaN to values and false is returned.
    */
-  final def addNoCorrection(other: HistogramWithBuckets): Unit = {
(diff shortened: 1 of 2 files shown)


memory/src/main/scala/filodb.memory/format/vectors/Histogram.scala
```diff
@@ -216,20 +216,23 @@ final case class MutableHistogram(buckets: HistogramBuckets, values: Array[Doubl
...
   /**
    * Adds the values from another Histogram.
-   * If the other histogram has the same bucket scheme, then the values are just added per bucket.
-   * If the scheme is different, then an approximation is used so that the resulting histogram has
-   * an approximate sum of the individual distributions, with the original scheme.  Modifies itself.
+   * If the other histogram has the same bucket scheme, then the values are just added per bucket and true is returned.
+   * If the scheme is different, it assigns NaN to values and false is returned.
    */
-  final def addNoCorrection(other: HistogramWithBuckets): Unit = {
+  final def addNoCorrection(other: HistogramWithBuckets): Boolean = {
```
vishramachandran, line 222: Add docs for return value - when is it true/false

memory/src/main/scala/filodb.memory/format/vectors/Histogram.scala
```diff
@@ -248,8 +251,8 @@ final case class MutableHistogram(buckets: HistogramBuckets, values: Array[Doubl
    * Adds the values from another Histogram, making a monotonic correction to ensure correctness
    */
   final def add(other: HistogramWithBuckets): Unit = {
-    addNoCorrection(other)
-    makeMonotonic()
+    // Call makeMonotonic only when buckets have same schema
```
vishramachandran, line 254: document that it is NaN otherwise and monotonic correction isnt needed

vishramachandran (approved):

PR #1130, opened 2021-06-01
debug(query): add log for empty outputrange

TanviBhavsar:


vishramachandran (approved):
Add a more expressive message. "Empty RvRangeFound".. and print the plan and its children, (like in explain plan)

PR #1131, opened 2021-06-02
debug(query): add debug logs

TanviBhavsar:


query/src/main/scala/filodb/query/exec/AggrOverRangeVectors.scala
```diff
@@ -179,6 +179,7 @@ object RangeVectorAggregator extends StrictLogging {
           s"Try applying more filters")
       groupedResult.map { case (rvk, aggHolder) =>
         val rowIterator = new CustomCloseCursor(aggHolder.map(_.toRowReader))(aggHolder.close())
+        logger.debug(s"Creating IteratorBackedRangeVector with period: ${period}")
```
vishramachandran, line 182: log only if empty to avoid too much logging

query/src/main/scala/filodb/query/exec/ExecPlan.scala
```diff
@@ -166,8 +166,9 @@ trait ExecPlan extends QueryCommand {
               val execPlanString = queryWithPlanName(queryContext)
               val srv = SerializedRangeVector(rv, builder, recSchema, execPlanString)
               if (rv.outputRange.isEmpty)
-                qLogger.debug(s"Empty rangevector found. Rv class is:  ${rv.getClass.getSimpleName}, " +
-                  s"execPlan is: $execPlanString, execPlan children ${this.children}")
+                qLogger.debug(s"Empty outputRange found. Rv class is:  ${rv.getClass.getSimpleName}, " +
+                  s"execPlan is: $execPlanString, execPlan children ${this.children}, " +
+                  s"execPlan rangeVectorTransformers ${this.rangeVectorTransformers}")
```
vishramachandran, line 171: log the explain plan string, not the case class toString

vishramachandran (changes requested):

vishramachandran (approved):
Not sure if you got the suggestion of using `printTree` to log exec plan. It prints RVTs too, and is more easier to read than case class toString. Approving, use your discretion to change.

PR #1135, opened 2021-06-07
feat(repair): Use inline dataset config for repair jobs

aandis:
**Current behavior :** Repair jobs rely on `dataset-config` which needs to be parsed separately.

**New behavior :** Repair jobs will rely on `inline-dataset-config` which is part of source config. No separate file need to be parsed.

**Pull Request checklist**


vishramachandran (approved):
Can you add a comment in filo-defaults that external dataset file is deprecated ?

PR #1162, opened 2021-07-10
Scala 2.12 upgrade

aandis:
This is a followup to #870 with a rebase on current `develop`.

[x] Changes verified locally.
[x] Code compiles successfully.
[x] All tests are passing.
[x] Filodb server comes up.
[x] Can see dummy data with `TestTimeseriesProducer`. 

**Pull Request checklist**

**NOTICEABLE CHANGES**

c781d78 is the only change on top of #870 that needs to be reviewed.


prometheus/src/main/scala/filodb/prometheus/parse/LegacyParser.scala
```diff
@@ -356,6 +356,7 @@ trait ExpressionParser extends AggregatesParser with SelectorParser with Numeric
 
     "(" ~ expression ~ ")" ^^ {
       case "(" ~ ep ~ ")" => PrecedenceExpression(ep)
+      case _ => ???
```
vishramachandran, line 359: what was missing here ? Can we call explicitly ?

vishramachandran (commented):
Could you please address the one comment I added? Rest looks good.

Can we also run one JMH benchmark to confirm that things look good there.
We also need to bring up two servers locally, ingest and query synthetic data generated by GatewayServer test data producer

prometheus/src/main/scala/filodb/prometheus/parse/LegacyParser.scala
aandis: This was complaining that `match may not be exhaustive`. So I added this line to catch all other cases compiler things we're missing. Since the block was already working with this setup in the old version, this shouldn't be an issue.

vishramachandran (approved):
@aandis @szymonm Thanks for the excellent work on this PR! Much appreciated.

vishramachandran:
@aandis Please rebase, ensure that PRB runs successfully.

vishramachandran:
@aandis Please do not squash and merge, Instead retain individual commits.

szymonm:
Ha ha, I already lost hope it will land.

🤝

vishramachandran:
@szymonm We had to time this PR since we were in the process of stabilizing the mainline. This PR will need some rigorous integration testing and we had to make room for it. Hence the delay. Thanks for the patience! 🙏

aandis:
Closing and reopening to trigger ci.

PR #1169, opened 2021-07-21
perf(core) Lucene field reuse

vishramachandran:
**Pull Request checklist**

Reuse lucene fields to reduce allocation during Index bootstrapping.
For dynamic fields, using a weak hash map to cache the fields.

Also fix test compile issue that existed.

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
+++ b/core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
@@ -166,7 +166,7 @@ class PartKeyLuceneIndex(ref: DatasetRef,
                                                _.toString)
       val value = new String(valueBase.asInstanceOf[Array[Byte]],
                                unsafeOffsetToBytesRefOffset(valueOffset + 2), // add 2 to move past numBytes
-                               UTF8StringMedium.numBytes(valueBase, valueOffset))
+                               UTF8StringMedium.numBytes(valueBase, valueOffset), StandardCharsets.UTF_8)
       addIndexedField(key, value)
     }
   }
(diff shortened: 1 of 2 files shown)


core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -142,8 +182,8 @@ class PartKeyLuceneIndex(ref: DatasetRef,
         def fromPartKey(base: Any, offset: Long, partIndex: Int): Unit = {
           val strOffset = schema.binSchema.blobOffset(base, offset, pos)
           val numBytes = schema.binSchema.blobNumBytes(base, offset, pos)
-          val value = new BytesRef(base.asInstanceOf[Array[Byte]], strOffset.toInt - UnsafeUtils.arayOffset, numBytes)
-          addIndexEntry(colName.toString, value, partIndex)
+          val value = new String(base.asInstanceOf[Array[Byte]], strOffset.toInt - UnsafeUtils.arayOffset, numBytes)
```
sherali42, line 185: we would need `Charset` here?

sherali42 (changes requested):
we should add charset to `new String`

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
vishramachandran: good point

vishramachandran: added

vishramachandran:
Ran all unit tests locally

sherali42 (approved):
👍

PR #1176, opened 2021-08-02
bug(parser):  fix Antlr parsing errors

TanviBhavsar:
**Pull Request checklist**

- [ ] Docs have been added / updated (for bug fixes / features) 

Fix Antlr errors for following queries:

- Binary expression without any space between operator and operand: `1-1` 
- Metric name starting with ___: `rate(___http_requests_total{job="api-server"}[10m])`
- Scalar Binary Expression with multiple operator's without any space between operators:  `(http_requests_total{job="api-server"}==1)+4`


prometheus/src/main/java/filodb/prometheus/antlr/PromQL.g4
```diff
@@ -116,7 +116,7 @@ literal: NUMBER | STRING;
 
 // Number format is non-standard.
 // It doesn't support hex, NaN, Inf, and it doesn't require a digit after the decimal point.
-NUMBER: [-+] ? (
+NUMBER: (
```
vishramachandran, line 119: Why was this removed? We need to support numbers like `+1` and `-5`

vishramachandran: Maybe good to add a test case for it.

prometheus/src/main/scala/filodb/prometheus/parse/Parser.scala
```diff
@@ -48,8 +48,8 @@ object Parser extends StrictLogging {
   }
 
   // TODO: Once fully switched to AntlrParser, get rid of the special precedence methods.
-  private def parseQueryWithPrecedence(query: String): Expression = {
-    mode match {
+  private def parseQueryWithPrecedence(query: String, modeParam: Mode = mode): Expression = {
+    modeParam match {
```
vishramachandran, line 52: When does `mode`, the member come into play now with this change? Almost never right since it is always called with non-default ?

prometheus/src/main/java/filodb/prometheus/antlr/PromQL.g4
TanviBhavsar: Expressions like `+1`  get parsed to to unaryOperation using `unaryOp vectorExpression`. I have added test cases for this. I removed it because parsing of expressions like '1+1' was failing as the operator was getting matched to NUMBER because of [-+]

vishramachandran (approved):

PR #1191, opened 2021-08-16
bug(core): Make shutdown on eviction lock timeout less aggressive

vishramachandran:
**Pull Request checklist**

Currently we shutdown too prematurely even if memory is slightly lower than threshold, and we didn't get the lock 5 eviction iterations. This is too aggressive, considering that timeout calculated (due to just below threshold) will be very small.

New algorithm:
1. Don't shutdown if not able to reclaim for multiple attempts as done today. Rely on ingestion failure to trigger forced shutdown.
2. If really close to threshold, increase timeout much more steeply than being done today. Given that query timeout is 60s, the maximum timeout should be at least that.
3. Also track queries that are holding the shared eviction locks for debugging purposes. This involves some small cost in memory and latency.


sherali42 (approved):
👍

whizkido (approved):
lgtm , we should check the impact to queries for a 3 day window post this change

PR #1195, opened 2021-08-19
bug(function) Fixing last for subqueries and adding last_over_time

kvpetrov:
**Pull Request checklist**

**New behavior :**

**BREAKING CHANGES**

Breaking changes may include:

**Other information**:

### DIFF
--- a/query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
+++ b/query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
@@ -376,14 +376,8 @@ class SlidingWindowIterator(raw: RangeVectorCursor,
     * @param curWindowStart start time of the current window
     */
   private def shouldAddCurToWindow(curWindowStart: Long, cur: TransientRow): Boolean = {
-    // One of the following three conditions need to hold true:
-    val insideCurWindow = cur.timestamp > curWindowStart
-    // 1. cur is inside current window
-    insideCurWindow ||
-      // 2. needLastSample and cur is the last sample because next sample is inside window
-      (rangeFunction.needsLastSample && rows.hasNext && rows.head.timestamp > curWindowStart) ||
-      // 3. needLastSample and no more rows after cur
-      (rangeFunction.needsLastSample && !rows.hasNext)
+    // cur is inside current window
+    cur.timestamp > curWindowStart
   }
 
   /**
@@ -400,11 +394,7 @@ class SlidingWindowIterator(raw: RangeVectorCursor,
       false
     } else {
       val headIsOutsideWindow = windowQueue.head.timestamp <= curWindowStart
-      // One of the following two conditions need to hold true:
-      // 1. if no need for last sample, and head is outside the window
-      (!rangeFunction.needsLastSample && headIsOutsideWindow) ||
-        // 2. if needs last sample, then ok to remove window's head only if there is more than one item in window
-        (rangeFunction.needsLastSample && windowQueue.size > 1 && headIsOutsideWindow)
+      headIsOutsideWindow
     }
   }
 }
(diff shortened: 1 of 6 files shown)


kvpetrov:
The tests are now working fine but should we be catching  0 window stuff for when the aggregation function is not provided or last is provided explicitely and populate window with queryConfig.staleSampleAfterMs value?
Whoever makes a  call on SlidingWindowIterator, how does he know that he is supposed to pass in stale value instead of a window in a case of last or no aggregate function? The case is indeed very unusual and not like all the other functions. What do you think?

query/src/test/scala/filodb/query/exec/LastSampleFunctionSpec.scala
```diff
@@ -50,7 +50,7 @@ class LastSampleFunctionSpec extends RawDataWindowingSpec {
       val start = now + ThreadLocalRandom.current().nextLong(80000)
       val end = start
       val step = 1
-      val lastSamplesIter = new SlidingWindowIterator(rv.rows, start, step, end, 0, LastSampleFunction, queryConfig)
+      val lastSamplesIter = new SlidingWindowIterator(rv.rows, start, step, end, queryConfig.staleSampleAfterMs, LastSampleFunction, queryConfig)
```
kvpetrov, line 53: what if we do 0 window check in SlidingWindowIterator? would it be more encapsulated/logical?

query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
```diff
@@ -400,11 +394,7 @@ class SlidingWindowIterator(raw: RangeVectorCursor,
       false
     } else {
       val headIsOutsideWindow = windowQueue.head.timestamp <= curWindowStart
-      // One of the following two conditions need to hold true:
-      // 1. if no need for last sample, and head is outside the window
-      (!rangeFunction.needsLastSample && headIsOutsideWindow) ||
-        // 2. if needs last sample, then ok to remove window's head only if there is more than one item in window
-        (rangeFunction.needsLastSample && windowQueue.size > 1 && headIsOutsideWindow)
+      headIsOutsideWindow
```
vishramachandran, line 397: One liner:
`windowQueue.nonEmpty && windowQueue.head.timestamp <= curWindowStart`

Can make it inline condition

query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
```diff
@@ -412,21 +407,23 @@ object RangeFunction {
...
-    } else {
-      sampleToEmit.setValues(endTimestamp, window.head.getDouble(1))
+    for (i <- (window.size - 1) to 0 by -1) {
+      val row = window.apply(i)
+      val rowValue = row.getDouble(1)
+      //val rowTimestampMs = row.getLong(0)
+      if (!rowValue.isNaN ) {
+        sampleToEmit.setValues(endTimestamp, rowValue)
+        return
+      }
```
vishramachandran, line 424: Consider using cForRange or a while loop. `(window.size - 1) to 0 by -1` is much slower

query/src/test/scala/filodb/query/exec/LastSampleFunctionSpec.scala
vishramachandran: Lets keep consistent with ChunkedWindowIterator

vishramachandran (approved):
Approved with comments

query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
kvpetrov: ack

PR #1208, opened 2021-09-02
feat (query): support for absent_over_time for subqueries

kvpetrov:
**Pull Request checklist**

**New behavior :**

**BREAKING CHANGES**

Breaking changes may include:

**Other information**:

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
@@ -257,28 +257,74 @@ class SingleClusterPlanner(dsRef: DatasetRef,
     // is optimal, if there is no overlap and even worse significant gap between the individual subqueries, retrieving
     // the entire range might be suboptimal, this still might be a better option than issuing and concatenating numerous
     // subqueries separately
-    val rangeFn = InternalRangeFunction.lpToInternalFunc(sqww.functionId)
-    var innerPeriodicSeriesPlan = sqww.innerPeriodicSeries
-    val paramsExec = materializeFunctionArgs(sqww.functionArgs, qContext)
+    var innerPlan = sqww.innerPeriodicSeries
     val window = Some(sqww.subqueryWindowMs)
-    val rangeVectorTransformer =
-      PeriodicSamplesMapper(
-        sqww.startMs, sqww.stepMs, sqww.endMs,
-        window,
-        Some(rangeFn),
-        qContext,
-        false,
-        paramsExec,
-        sqww.offsetMs,
-        false,
-        true
-      )
-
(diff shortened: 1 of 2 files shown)


query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
```diff
@@ -405,7 +405,11 @@ class AvgOverTimeFunction(var sum: Double = Double.NaN, var count: Int = 0) exte
   override def apply(startTimestamp: Long, endTimestamp: Long, window: Window,
                      sampleToEmit: TransientRow,
                      queryConfig: QueryConfig): Unit = {
-    sampleToEmit.setValues(endTimestamp, sum/count)
+    if (count == 0) {
+      sampleToEmit.setValues(endTimestamp, Double.NaN)
+    } else {
+      sampleToEmit.setValues(endTimestamp, sum/count)
+    }
```
vishramachandran, line 412: This should be reverted. There is no divide-by-zero exception when Double is involved. Result is Infinity. 

```
scala> 5d/0d
res0: Double = Infinity
```

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
```diff
@@ -257,28 +257,74 @@ class SingleClusterPlanner(dsRef: DatasetRef,
...
-        paramsExec,
-        sqww.offsetMs,
-        false,
-        true
+        Seq(),
+        offsetMs,
+        false
+      ))
+    )
+    val aggregate = Aggregate(AggregationOperator.Sum, innerPlan, Nil, Seq("job"))
```
vishramachandran, line 308: Can you remove code duplication? I think we may have the same thing for regular absent-over-time function as well.

vishramachandran (approved):
Approved with comments.
Please wait for @TanviBhavsar to review as well. She is aware of Absent intricacies better than me.

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
amolnayak311: Hi @vishramachandran What would be the expected behavior? If sum is NaN the division will give NaN irrespective of whether count is 0 or not, but will be Infinity when sum is non NaN and count is 0. The code change above will give NaN in all scenarios.

kvpetrov: the problem is that it does not match with the Prometheus, the reason, I had to put this in is to match with the Prom behavior. I don't think conceptually -infinity is correct either.

TanviBhavsar (approved):
Looks good to me. Can you also add a unit test?

PR #1216, opened 2021-09-15
bug(query, core): Shared lock was being acquired twice in one query

vishramachandran:
**Pull Request checklist**

**Current behavior :** 

We acquire shared lock in `lookupPartitions` which is a deadly side-effect not known to callers. For a recent feature a new call was added to invoke `lookupPartitions`. Since the query path includes that call elsewhere, the shared lock for shard got acquired twice. However we release it only once causing problems. Load testing brought the shared to stand-still.

In addition, we do not have a timeout during shared lock acquisition.

**New behavior :**

* Refactor out shared lock into separate method, and call it from specific plans that need them.
* While at it, I have also refactored the duplicate code to check if shard is ready after bootstrapping.

### DIFF
--- a/memory/src/main/scala/filodb.memory/EvictionLock.scala
+++ b/memory/src/main/scala/filodb.memory/EvictionLock.scala
@@ -67,9 +67,9 @@ class EvictionLock(trackQueriesHoldingEvictionLock: Boolean = false,
 
   def releaseExclusive(): Unit = reclaimLock.releaseExclusive()
 
-  def acquireSharedLock(holderId: String, promQL: String): Unit = {
+  def acquireSharedLock(timeoutMs: Long, holderId: String, promQL: String): Boolean = {
     if (trackQueriesHoldingEvictionLock) runningQueries.put(holderId, promQL)
-    reclaimLock.lock()
+    reclaimLock.tryAcquireSharedNanos(timeoutMs * 1000000)
   }
 
   def releaseSharedLock(holderId: String): Unit = {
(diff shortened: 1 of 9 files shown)


memory/src/main/scala/filodb.memory/EvictionLock.scala
```diff
@@ -67,9 +67,9 @@ class EvictionLock(trackQueriesHoldingEvictionLock: Boolean = false,
 
   def releaseExclusive(): Unit = reclaimLock.releaseExclusive()
 
-  def acquireSharedLock(holderId: String, promQL: String): Unit = {
+  def acquireSharedLock(timeoutMs: Long, holderId: String, promQL: String): Boolean = {
     if (trackQueriesHoldingEvictionLock) runningQueries.put(holderId, promQL)
-    reclaimLock.lock()
+    reclaimLock.tryAcquireSharedNanos(timeoutMs * 1000000)
```
whizkido, line 72: Can we use a constant or a TimeUnit conversion here.

query/src/main/scala/filodb/query/exec/MultiSchemaPartitionsExec.scala
```diff
@@ -126,13 +126,8 @@ final case class MultiSchemaPartitionsExec(queryContext: QueryContext,
...
                 querySession: QuerySession)
                (implicit sched: Scheduler): ExecResult = {
-    if (!source.isReadyForQuery(dataset, shard)) {
-      if (!queryContext.plannerParams.allowPartialResults) {
-        throw new ServiceUnavailableException(s"Unable to answer query since shard $shard is still bootstrapping")
-      }
-      querySession.resultCouldBePartial = true
-      querySession.partialResultsReason = Some("Result may be partial since some shards are still bootstrapping")
-    }
+    source.checkReadyForQuery(dataset, shard, querySession)
```
whizkido, line 129: This is nice .

sherali42 (approved):
👍

whizkido (approved):
Minor comments and approved.

PR #1232, opened 2021-09-30
bug(query) Fix issues with multi-partition aggregation operation

amolnayak311:
**Review only PR**
**Pull Request checklist**

- [X] Docs have been added / updated (for bug fixes / features) ?

**Current behavior :** 

Multi partition queries behave incorrectly . Consider the following multi-partition query 

``count(sum by(metric_ws_tag)(foo{_ws_ = "demo", _ns_ =~ "App-.*"}))``

The inner aggregation ``sum by(metric_ws_tag)(foo{_ws_ = "demo", _ns_ =~ "App-.*"})`` may have same`` metric_ws_tag``  and the sum operation will reduce them to 1 row with the corresponding sums added up. The ``count`` of this result should return the same number as the number of rows returned by the ``sum1` aggregation operation.  Current behavior pushes the high level ``count(sum by(metric_ws_tag)(foo{...}))`` aggregation to all partitions and sums the counts which double counts results giving an incorrect count than expected

**New behavior :**

The new behavior is to check from the leaf of the ``LogicalPlan`` and materialize the query the first time we encounter at any level an aggregation that accesses data across different partitions. All other aggregation operations then need to be done in memory of the calling process. Continuing with above given example, only the ``sum by(metric_ws_tag)(foo{_ws_ = "demo", _ns_ =~ "App-.*"})`` operation will be materialized and executed in various partitions, the aggregated sum is then counted in process of the process executing the ``count`` operation.

**BREAKING CHANGES**

None

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/queryplanner/ShardKeyRegexPlanner.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryplanner/ShardKeyRegexPlanner.scala
@@ -126,19 +126,20 @@ class ShardKeyRegexPlanner(val dataset: Dataset,
     PlanResult(Seq(execPlan))
   }
 
-  private def hasChildAggregate(lp: LogicalPlan): Boolean = lp match {
-    case _: Aggregate => true
-    case nonLeaf: NonLeafLogicalPlan => nonLeaf.children.exists(hasChildAggregate(_))
-    case _ => false
-  }
-
   /***
    * For aggregate queries like sum(test{_ws_ = "demo", _ns_ =~ "App.*"})
    * It will be broken down to sum(test{_ws_ = "demo", _ns_ = "App-1"}), sum(test{_ws_ = "demo", _ns_ = "App-2"}) etc
    * Sub query could be across multiple partitions so aggregate using MultiPartitionReduceAggregateExec
    * */
   private def materializeAggregate(aggregate: Aggregate, queryContext: QueryContext): PlanResult = {
-    val plan = if (hasChildAggregate(aggregate.vectors)) {
+    // Pushing down aggregates to run on individual partitions is most efficient, however, if we have multiple
+    // nested aggregation we should be pushing down the lowest aggregate for correctness. Consider the following query
+    // count(sum by(foo)(test1{_ws_ = "demo", _ns_ =~ "App-.*"})), doing a count of the counts received from individual
+    // partitions may not give the correct answer, we should push down the sum to multiple partitions, perform a reduce
+    // locally and then run count on the results. The following implementation checks for descendant aggregates, if
+    // there are any, the provided aggregation needs to be done using inProcess, else we can materialize the aggregate
+    // using the wrapped planner
+    val plan = if (LogicalPlanUtils.hasDescendantAggregate(aggregate.vectors)) {
         val childPlan = materialize(aggregate.vectors, queryContext)
         val reducer = LocalPartitionReduceAggregateExec(queryContext, inProcessPlanDispatcher,
           Seq(childPlan), aggregate.operator, aggregate.params)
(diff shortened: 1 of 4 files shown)


vishramachandran (changes requested):
@amolnayak311 Welcome to the FiloDB repo!
Excellent work on your first complete PR. 
Appreciate the attention to detail and the multiple iterations. Thank you for opening yourself up to multiple solutions.

I have called out some minor change requests. Once addressed we will be ready to merge.

coordinator/src/main/scala/filodb.coordinator/queryplanner/ShardKeyRegexPlanner.scala
```diff
@@ -115,45 +115,60 @@ class ShardKeyRegexPlanner(val dataset: Dataset,
...
-         SetOperatorExec(qContext, inProcessPlanDispatcher, Seq(lhsExec), Seq(rhsExec), logicalPlan.operator,
-          LogicalPlanUtils.renameLabels(onKeysReal, datasetMetricColumn),
-          LogicalPlanUtils.renameLabels(logicalPlan.ignoring, datasetMetricColumn), datasetMetricColumn)
-      else
-         BinaryJoinExec(qContext, inProcessPlanDispatcher, Seq(lhsExec), Seq(rhsExec), logicalPlan.operator,
-          logicalPlan.cardinality, LogicalPlanUtils.renameLabels(onKeysReal, datasetMetricColumn),
-          LogicalPlanUtils.renameLabels(logicalPlan.ignoring, datasetMetricColumn),
-          LogicalPlanUtils.renameLabels(logicalPlan.include, datasetMetricColumn), datasetMetricColumn)
-     PlanResult(Seq(execPlan))
+      SetOperatorExec(qContext, inProcessPlanDispatcher, Seq(lhsExec), Seq(rhsExec), logicalPlan.operator,
```
vishramachandran, line 118: Can we call our as a FIXME the work to optimize the push-down of predicate in the case where both LHS and RHS land up in the same partition ?

amolnayak311: Done

coordinator/src/test/scala/filodb.coordinator/queryplanner/ShardKeyRegexPlannerSpec.scala
```diff
@@ -481,22 +489,459 @@ class ShardKeyRegexPlannerSpec extends AnyFunSpec with Matchers with ScalaFuture
...
+          plan1.queryContext.origQueryParams.asInstanceOf[PromQlQueryParams].promQl :: Nil toSet) shouldEqual
+          Set("count(test1{_ws_=\"demo\",_ns_=\"App2\"}) by (foo)",
+            "count(test1{_ws_=\"demo\",_ns_=\"App1\"}) by (foo)")
+        plan1.isInstanceOf[LocalPartitionReduceAggregateExec] shouldEqual true
+      case _ => fail("Expected two children")
+    }
+  }
+
+
+  it("should handle push down aggregation to right level case 2") {
```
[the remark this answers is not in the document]
amolnayak311: Thanks, I hope we have covered most of the cases we can think of.

coordinator/src/main/scala/filodb.coordinator/queryplanner/ShardKeyRegexPlanner.scala
```diff
@@ -88,72 +88,95 @@ class ShardKeyRegexPlanner(val dataset: Dataset,
...
-      reducer
+    // Pushing down aggregates to run on individual partitions is most efficient, however, if we have multiple
+    // nested aggregation we should be pushing down the lowest aggregate for correctness. Consider the following query
+    // count(sum by(foo)(test1{_ws_ = "demo", _ns_ =~ "App-.*"})), doing a count of the counts received from individual
+    // partitions may not give the correct answer, we should push down the sum to multiple partitions, perform a reduce
+    // locally and then run count on the results. The following implementation checks for descendant aggregates, if
+    // there are any, the provided aggregation needs to be done using inProcess, else we can materialize the aggregate
+    // using the wrapped planner
+    val plan = if (LogicalPlanUtils.hasDescendantAggregate(aggregate.vectors)) {
+        val childPlan = materialize(aggregate.vectors, queryContext)
```
TanviBhavsar, line 145: Can you add test case for this case?

TanviBhavsar (approved):
Congrats on your first PR!
Approved, after the single comment is addressed

coordinator/src/main/scala/filodb.coordinator/queryplanner/ShardKeyRegexPlanner.scala
amolnayak311: Thanks Tanvi, the test case ``it("should handle push down aggregation to right level case 1") `` on line 498 of the spec has multi level aggregation ``sum(count by(foo)(test1{_ws_ = "demo", _ns_ =~ "App-.*"}))`` we assert that count aggregation  is pushed down and sum happens locally.

vishramachandran (approved):

PR #1256, opened 2021-10-18
fix(query): correctly handle empty by/without aggregation clauses

alextheimer:
**Pull Request checklist**

**Current behavior :**
Empty `without()` clause, as in:
```
sum without()(node_cpu_seconds_total)
```
yields a single aggregate value instead of all values for all time series.

This is because aggregate `LogicalPlan`/`ExecPlan` data structures store without/by labels as `Sequence`s-- information about whether-or-not an empty clause was specified in the query is lost.

For example, both of the queries:
```
sum(metric)
sum without()(metric)
```
would yield relevant `LogicalPlan`/`ExecPlan` instances similar to:
```
Aggregate(Sum, metric, by=[], without=[])
```

**New behavior :**
Aggregate clause information is now stored in a single `AggregateClause(type: ClauseType, labels: Seq[String])` class. `Aggregate` and `AggregateMapReduce` have each had their `by: Seq[String]` and `without: Seq[String]` fields replaced with a single `Option[AggregateClause]`, where the value `None` implies no clause was specified.


alextheimer:
test this please

sherali42 (approved):
👍 looks good @alextheimer

PR #1285, opened 2021-11-08
fix(query): Label values/names query to use String column type in ResultSchema

sherali42:
**Pull Request checklist**

- Metadata queries LabelValues/LabelNames have result type as String where as `Series` has Map type. This is causing MetadataRemoteExec to break as it is assumed that all metadata queries return map type result. The changes in PR make sure that depending on result type, appropriate result schema is set at the ExecPlan level. In MetadataRemoteExec, appropriate response is generated based on the query/result schema,

### DIFF
--- a/query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
+++ b/query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
@@ -12,7 +12,7 @@ import filodb.core.DatasetRef
 import filodb.core.binaryrecord2.{BinaryRecordRowReader, MapItemConsumer}
 import filodb.core.memstore.{MemStore, TimeSeriesMemStore}
 import filodb.core.metadata.Column.ColumnType
-import filodb.core.metadata.Column.ColumnType.MapColumn
+import filodb.core.metadata.Column.ColumnType.{MapColumn, StringColumn}
 import filodb.core.query._
 import filodb.core.query.NoCloseCursor.NoCloseCursor
 import filodb.core.store.ChunkSource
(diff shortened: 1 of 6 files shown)


query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
```diff
@@ -164,7 +164,56 @@ final case class TopkCardPresenter(k: Int) extends RangeVectorTransformer {
...
+            rv.head match {
+              case srv: SerializedRangeVector =>
+                srv.schema.toStringPairs (binaryRowReader.recordBase, binaryRowReader.recordOffset)
+                  .map (pair => pair._1.utf8 -> pair._2.utf8).toMap
+              case _ => throw new UnsupportedOperationException("Metadata query currently needs SRV results")
+            }
+          }
+        }
+        IteratorBackedRangeVector(new CustomRangeVectorKey(Map.empty),
+          UTF8MapIteratorRowReader(metadataResult.toIterator), None)
```
tjackpaul, line 195: Since this is a duplicate code, if possible move this to a common method and use the same in MetadataDistConcatExec as well.

query/src/main/scala/filodb/query/exec/MetadataRemoteExec.scala
```diff
@@ -39,17 +44,42 @@ case class MetadataRemoteExec(queryEndpoint: String,
   }
 
   def toQueryResponse(response: MetadataSuccessResponse, id: String, parentSpan: kamon.trace.Span): QueryResponse = {
-    val iteratorMap = response.data.map { r => r.map { v => (v._1.utf8, v._2.utf8) }}
+      if (response.data.isEmpty) mapTypeQueryResponse(response, id)
```
tjackpaul, line 47: Can any DistConcat fail somewhere if we consider empty response as mapTypeResponse ?

tjackpaul (approved):
lgtm, left few minor comments.. in general can you add some java comments in the class around major operations.. thanks for the PR @sherali42

PR #1289, opened 2021-11-12
feat(query):  add hierarchical cardinality query infrastructure

alextheimer:
**Pull Request checklist**

**New behavior :**
This infrastructure supports hierarchical cardinality queries of the abstract form:
```
Find (active, total) cardinality pairs for all time series with <shard-key-prefix>,
  then group them by { key[:1], key[:2], key[:3], ... }.
```

## `TsCardinalities extends LogicalPlan`
Initialized with a `shardKeyPrefix` and an integer `groupDepth`, which specifies the "hierarchical depth" at which to aggregate the cardinalities.

Examples:
```
{ prefix=[], groupDepth=1 } -> {
    prefix=["ws_a", "ns_a"] -> (4, 6),
    prefix=["ws_a", "ns_b"] -> (2, 4),
    prefix=["ws_b", "ns_c"] -> (3, 5) }

{ prefix=["ws_a", "ns_a"], groupDepth=2 } -> {
    prefix=["ws_a", "ns_a", "met_a"] -> (4, 6),
    prefix=["ws_a", "ns_a", "met_b"] -> (3, 5) }

{ prefix=["ws_a"], groupDepth=0 } -> {
    prefix=["ws_a"] -> (3, 5) }
```

Note that `groupDepth` must specify a hierarchical depth that's:

1. at least as deep as shardKeyPrefix
2. less than '2' when the prefix does not contain values for all lesser depths.

Example (if shard keys specify a ws, ns, and metric):
```
  shardKeyPrefix     groupDepth
  []                 { WS, NS }
  [ws]               { WS, NS }
  [ws, ns]           { NS, METRIC }
  [ws, ns, metric]   { METRIC }                                                         
```

## `TsCardExec extends ExecPlan` (and `TsCardReduceExec`)

The `RangeVector` produced by `execute` has the schema:
```
[group:String, active:Int, total:Int]
```
where:
- the "group" column contains the group-by prefix values concatenated by a `,` delimiter.
- the "active" and "total" columns give the cardinalities for (1) only the actively-ingesting timeseries and (2) all timeseries, respectively.

Note that, during aggregation, an aggregator `TsCardReduceExec` with a pending result of size `MAX_RESPONSE_SIZE=5000` will count subsequent cardinalities of previously-unseen prefixes into an overflow bucket. The count for this bucket is specified by the result row with `group=_overflow_`.

## `tscard` CLI command

Invoke the CLI command as:
```
--host <hostname/IP> [--port ...] --command tscard --dataset <dataset> --shardkeyprefix <shard-key-prefix> --groupdepth {0, 1, 2}
```

Examples:
```
$ ./filo-cli --host 127.0.0.1 --dataset prometheus --command tscard --shardkeyprefix --groupdepth ws
Output schema: ResultSchema(List(ColumnInfo(group,StringColumn), ColumnInfo(active,IntColumn), ColumnInfo(total,IntColumn)),1,Map(),None,List())
Number of Range Vectors: 1
/shard:/Map()
	demo-1	500	500
	demo-0	500	500

$ ./filo-cli --host 127.0.0.1 --dataset prometheus --command tscard --shardkeyprefix --groupdepth 1
Output schema: ResultSchema(List(ColumnInfo(group,StringColumn), ColumnInfo(active,IntColumn), ColumnInfo(total,IntColumn)),1,Map(),None,List())
Number of Range Vectors: 1
/shard:/Map()
	demo-0,App-3	124	124
	demo-1,App-1	124	124
	demo-1,App-3	124	124
	demo-0,App-1	124	124
	demo-0,App-2	124	124
	demo-1,App-2	124	124
	demo-0,App-0	128	128
	demo-1,App-0	128	128

$ ./filo-cli --host 127.0.0.1 --dataset prometheus --command tscard --shardkeyprefix demo-0 --groupdepth 0
Output schema: ResultSchema(List(ColumnInfo(group,StringColumn), ColumnInfo(active,IntColumn), ColumnInfo(total,IntColumn)),1,Map(),None,List())
Number of Range Vectors: 1
/shard:/Map()
	demo-0	500	500

$ ./filo-cli --host 127.0.0.1 --dataset prometheus --command tscard --shardkeyprefix demo-0 --groupdepth 1
Output schema: ResultSchema(List(ColumnInfo(group,StringColumn), ColumnInfo(active,IntColumn), ColumnInfo(total,IntColumn)),1,Map(),None,List())
Number of Range Vectors: 1
/shard:/Map()
	demo-0,App-3	124	124
	demo-0,App-2	124	124
	demo-0,App-0	128	128
	demo-0,App-1	124	124

$ ./filo-cli --host 127.0.0.1 --dataset prometheus --command tscard --shardkeyprefix demo-0 App-0 --groupdepth 2
Output schema: ResultSchema(List(ColumnInfo(group,StringColumn), ColumnInfo(active,IntColumn), ColumnInfo(total,IntColumn)),1,Map(),None,List())
Number of Range Vectors: 1
/shard:/Map()
	demo-0,App-0,heap_usage-3	32	32
	demo-0,App-0,heap_usage-1	32	32
	demo-0,App-0,heap_usage-2	32	32
	demo-0,App-0,heap_usage-0	32	32
```

### Important Notes:
- The `LogicalPlan` and `ExecPlan` infrastructure to support `topkcard` (i.e. all `ItemsSketch`-related code) has been removed.

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/client/QueryOps.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/client/QueryOps.scala
@@ -51,7 +51,10 @@ trait QueryOps extends ClientBase with StrictLogging {
                      k: Int,
                      addInactive: Boolean,
                      timeout: FiniteDuration = 15.seconds): Seq[CardinalityRecord] =
-    askCoordinator(GetTopkCardinality(dataset, shards, shardKeyPrefix, k, addInactive), timeout) {
+    askCoordinator(
+      GetTopkCardinality(dataset, shards, shardKeyPrefix,
+                         shardKeyPrefix.size + 1, k, addInactive),
+      timeout) {
       case s: Seq[CardinalityRecord] @unchecked => s
       case e: QueryError => throw e.t
     }
--- a/coordinator/src/main/scala/filodb.coordinator/queryplanner/LogicalPlanUtils.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryplanner/LogicalPlanUtils.scala
@@ -66,8 +66,7 @@ object LogicalPlanUtils extends StrictLogging {
       case lp: LabelValues                 => TimeRange(lp.startMs, lp.endMs)
       case lp: LabelCardinality            => TimeRange(lp.startMs, lp.endMs)
       case lp: LabelNames                  => TimeRange(lp.startMs, lp.endMs)
-      case lp: TopkCardinalities           => throw new IllegalArgumentException(
-                                                          "no time params for TopKCardinalitites")
+      case lp: TsCardinalities             => throw new IllegalArgumentException("no time params for TsCardinalitites")
       case lp: SeriesKeysByFilters         => TimeRange(lp.startMs, lp.endMs)
       case lp: ApplyInstantFunctionRaw     => getTimeFromLogicalPlan(lp.vectors)
       case lp: ScalarBinaryOperation       => TimeRange(lp.rangeParams.startSecs * 1000, lp.rangeParams.endSecs * 1000)
@@ -93,7 +92,7 @@ object LogicalPlanUtils extends StrictLogging {
       case lp: LabelValues              => lp.copy(startMs = timeRange.startMs, endMs = timeRange.endMs)
       case lp: LabelNames               => lp.copy(startMs = timeRange.startMs, endMs = timeRange.endMs)
       case lp: LabelCardinality         => lp.copy(startMs = timeRange.startMs, endMs = timeRange.endMs)
-      case lp: TopkCardinalities        => lp.copy()
+      case lp: TsCardinalities          => lp.copy()
       case lp: SeriesKeysByFilters      => lp.copy(startMs = timeRange.startMs, endMs = timeRange.endMs)
     }
   }
@@ -344,7 +343,7 @@ object LogicalPlanUtils extends StrictLogging {
       case lp: RawChunkMeta                => None
       case sq: SubqueryWithWindowing       => getPeriodicSeriesPlan(sq.innerPeriodicSeries)
       case tlsq: TopLevelSubquery          => getPeriodicSeriesPlan(tlsq.innerPeriodicSeries)
-      case lp: TopkCardinalities           => None
+      case lp: TsCardinalities             => None
     }
   }
 
--- a/coordinator/src/main/scala/filodb.coordinator/queryplanner/LogicalPlanUtils.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryplanner/LogicalPlanUtils.scala
@@ -92,7 +92,7 @@ object LogicalPlanUtils extends StrictLogging {
       case lp: LabelValues              => lp.copy(startMs = timeRange.startMs, endMs = timeRange.endMs)
       case lp: LabelNames               => lp.copy(startMs = timeRange.startMs, endMs = timeRange.endMs)
       case lp: LabelCardinality         => lp.copy(startMs = timeRange.startMs, endMs = timeRange.endMs)
-      case lp: TsCardinalities          => lp.copy()
+      case lp: TsCardinalities          => lp  // immutable & no members need to be updated
       case lp: SeriesKeysByFilters      => lp.copy(startMs = timeRange.startMs, endMs = timeRange.endMs)
     }
   }
--- a/coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
@@ -140,7 +140,7 @@ class MultiPartitionPlanner(partitionLocationProvider: PartitionLocationProvider
       case lp: ApplyAbsentFunction        => super.materializeAbsentFunction(qContext, lp)
       case lp: ScalarBinaryOperation      => super.materializeScalarBinaryOperation(qContext, lp)
       case lp: ApplyLimitFunction         => super.materializeLimitFunction(qContext, lp)
-      case _: TopkCardinalities           => throw new IllegalArgumentException("TopkCardinalities unexpected here")
+      case _: TsCardinalities             => throw new IllegalArgumentException("TsCardinalities unexpected here")
 
       // Imp: At the moment, these two cases for subquery will not get executed, materialize is already
       // Checking if the plan is a TopLevelSubQuery or any of the descendant is a SubqueryWithWindowing and
@@ -429,7 +429,8 @@ class MultiPartitionPlanner(partitionLocationProvider: PartitionLocationProvider
     PlanResult(execPlan::Nil)
   }
 
-  private def materializeTopkCardinalities(qContext: QueryContext, lp: TopkCardinalities) : PlanResult = {
+  private def materializeTsCardinalities(qContext: QueryContext, lp: TsCardinalities) : PlanResult = {
+    // TODO(a_theimer): outdated?
     // This code is nearly identical to materializeMetadataQueryPlan, but it prevents some
     //   boilerplate code clutter that results when TopkCardinalities extends MetadataQueryPlan.
     //   If this code's maintenance isn't worth some extra stand-in lines of code (i.e. at matchers that
@@ -447,8 +448,8 @@ class MultiPartitionPlanner(partitionLocationProvider: PartitionLocationProvider
           localPartitionPlanner.materialize(lp.copy(), qContext)
         else ??? // TODO: handle remote
       }.toSeq
-      val reducer = TopkCardReduceExec(qContext, inProcessPlanDispatcher, leafPlans, lp.k)
-      reducer.addRangeVectorTransformer(TopkCardPresenter(lp.k))
+      val reducer = TsCardReduceExec(qContext, inProcessPlanDispatcher, leafPlans)
+      reducer.addRangeVectorTransformer(TsCardPresenter())
       reducer
     }
     PlanResult(execPlan::Nil)
--- a/coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
@@ -429,32 +429,6 @@ class MultiPartitionPlanner(partitionLocationProvider: PartitionLocationProvider
     PlanResult(execPlan::Nil)
   }
 
-  private def materializeTsCardinalities(qContext: QueryContext, lp: TsCardinalities) : PlanResult = {
-    // TODO(a_theimer): outdated?
-    // This code is nearly identical to materializeMetadataQueryPlan, but it prevents some
-    //   boilerplate code clutter that results when TopkCardinalities extends MetadataQueryPlan.
-    //   If this code's maintenance isn't worth some extra stand-in lines of code (i.e. at matchers that
-    //   take MetadataQueryPlan instances), then just extend TopkCardinalities from MetadataQueryPlan.
-    val queryParams = qContext.origQueryParams.asInstanceOf[PromQlQueryParams]
-    val partitions = partitionLocationProvider.getAuthorizedPartitions(
-      TimeRange(queryParams.startSecs * 1000, queryParams.endSecs * 1000))
-    val execPlan = if (partitions.isEmpty) {
-      logger.warn(s"No partitions found for ${queryParams.startSecs}, ${queryParams.endSecs}")
-      localPartitionPlanner.materialize(lp, qContext)
-    } else {
-      val leafPlans = partitions.map { p =>
-        logger.debug(s"partitionInfo=$p; queryParams=$queryParams")
-        if (p.partitionName.equals(localPartitionName))
-          localPartitionPlanner.materialize(lp.copy(), qContext)
-        else ??? // TODO: handle remote
-      }.toSeq
-      val reducer = TsCardReduceExec(qContext, inProcessPlanDispatcher, leafPlans)
-      reducer.addRangeVectorTransformer(TsCardPresenter())
-      reducer
-    }
-    PlanResult(execPlan::Nil)
-  }
-
   private def createMetadataRemoteExec(qContext: QueryContext, partitionAssignment: PartitionAssignment,
                                        urlParams: Map[String, String]) = {
     val finalQueryContext = generateRemoteExecParams(
--- a/coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
@@ -134,6 +134,7 @@ class SingleClusterPlanner(val dataset: Dataset,
         many.head match {
           case _: LabelValuesExec => LabelValuesDistConcatExec(qContext, targetActor, many)
           case _: LabelNamesExec => LabelNamesDistConcatExec(qContext, targetActor, many)
+          case _: TsCardExec => TsCardReduceExec(qContext, targetActor, many)
           case _: LabelCardinalityExec => {
             val reduceExec = LabelCardinalityReduceExec(qContext, targetActor, many)
             // Presenter here is added separately which use the bytes representing the sketch to get an estimate
@@ -145,11 +146,6 @@ class SingleClusterPlanner(val dataset: Dataset,
               reduceExec.addRangeVectorTransformer(new LabelCardinalityPresenter())
             }
             reduceExec
-          }
-          case lce: TsCardExec => {
-            val reducer = TsCardReduceExec(qContext, targetActor, many)
-            reducer.addRangeVectorTransformer(TsCardPresenter(lce.groupDepth))
-            reducer
           }
           case ske: PartKeysExec => PartKeysDistConcatExec(qContext, targetActor, many)
           case ep: ExecPlan =>
--- a/coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
@@ -146,9 +146,9 @@ class SingleClusterPlanner(val dataset: Dataset,
             }
             reduceExec
           }
-          case lce: TopkCardExec => {
-            val reducer = TopkCardReduceExec(qContext, targetActor, many, lce.k)
-            reducer.addRangeVectorTransformer(TopkCardPresenter(lce.k))
+          case lce: TsCardExec => {
+            val reducer = TsCardReduceExec(qContext, targetActor, many)
+            reducer.addRangeVectorTransformer(TsCardPresenter())
             reducer
           }
           case ske: PartKeysExec => PartKeysDistConcatExec(qContext, targetActor, many)
@@ -240,7 +240,7 @@ class SingleClusterPlanner(val dataset: Dataset,
       case lp: ScalarVectorBinaryOperation => materializeScalarVectorBinOp(qContext, lp)
       case lp: LabelValues                 => materializeLabelValues(qContext, lp)
       case lp: LabelNames                  => materializeLabelNames(qContext, lp)
-      case lp: TopkCardinalities           => materializeTopkCardinalities(qContext, lp)
+      case lp: TsCardinalities             => materializeTsCardinalities(qContext, lp)
       case lp: SeriesKeysByFilters         => materializeSeriesKeysByFilters(qContext, lp)
       case lp: ApplyMiscellaneousFunction  => materializeApplyMiscellaneousFunction(qContext, lp)
       case lp: ApplySortFunction           => materializeApplySortFunction(qContext, lp)
@@ -509,11 +509,11 @@ class SingleClusterPlanner(val dataset: Dataset,
     PlanResult(metaExec, false)
   }
 
-  private def materializeTopkCardinalities(qContext: QueryContext,
-                                           lp: TopkCardinalities): PlanResult = {
+  private def materializeTsCardinalities(qContext: QueryContext,
+                                           lp: TsCardinalities): PlanResult = {
     val metaExec = shardMapperFunc.assignedShards.map{ shard =>
       val dispatcher = dispatcherForShard(shard)
-      exec.TopkCardExec(qContext, dispatcher, dsRef, shard, lp.shardKeyPrefix, lp.k, lp.addInactive)
+      exec.TsCardExec(qContext, dispatcher, dsRef, shard, lp.shardKeyPrefix)
     }
     PlanResult(metaExec, false)
   }
--- a/coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
@@ -148,7 +148,7 @@ class SingleClusterPlanner(val dataset: Dataset,
           }
           case lce: TsCardExec => {
             val reducer = TsCardReduceExec(qContext, targetActor, many)
-            reducer.addRangeVectorTransformer(TsCardPresenter())
+            reducer.addRangeVectorTransformer(TsCardPresenter(lce.groupDepth))
             reducer
           }
           case ske: PartKeysExec => PartKeysDistConcatExec(qContext, targetActor, many)
@@ -510,10 +510,10 @@ class SingleClusterPlanner(val dataset: Dataset,
   }
 
   private def materializeTsCardinalities(qContext: QueryContext,
-                                           lp: TsCardinalities): PlanResult = {
+                                         lp: TsCardinalities): PlanResult = {
     val metaExec = shardMapperFunc.assignedShards.map{ shard =>
       val dispatcher = dispatcherForShard(shard)
-      exec.TsCardExec(qContext, dispatcher, dsRef, shard, lp.shardKeyPrefix)
+      exec.TsCardExec(qContext, dispatcher, dsRef, shard, lp.shardKeyPrefix, lp.groupDepth)
     }
     PlanResult(metaExec, false)
   }
--- a/core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityStore.scala
+++ b/core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityStore.scala
@@ -5,14 +5,14 @@ import java.io.Closeable
 /**
  * The data stored in each node of the Cardinality Store trie
  *
- * @param name name of the item in shardKeyPrefix
+ * @param prefix name of the item in shardKeyPrefix
  * @param tsCount total number of timeSeries under this shardKeyPrefix (example, number of timeseries under ws,ns)
  * @param activeTsCount number of actively ingesting timeSeries under this shardKeyPrefix
  *                      (example, number of timeseries under ws,ns)
  * @param childrenCount number of immediate children for this shardKey (example, number of ns under ws)
  * @param childrenQuota quota for number of immediate children
  */
-case class Cardinality(name: String, tsCount: Int, activeTsCount: Int, childrenCount: Int, childrenQuota: Int)
+case class Cardinality(prefix: Seq[String], tsCount: Int, activeTsCount: Int, childrenCount: Int, childrenQuota: Int)
 
 /**
  *
@@ -57,7 +57,7 @@ trait CardinalityStore {
    * This method will be called for each shard key prefix when a new time series is added
    * to the index.
    */
-  def store(shardKeyPrefix: Seq[String], card: Cardinality): Unit
+  def store(card: Cardinality): Unit
 
   /**
    * Read existing cardinality value, if one does not exist return the zero value
--- a/core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityStore.scala
+++ b/core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityStore.scala
@@ -5,6 +5,7 @@ import java.io.Closeable
 /**
  * The data stored in each node of the Cardinality Store trie
  *
+ * @param shard the shard ID for which this CardinalityStore contains data
  * @param prefix the shard key prefix that corresponds with the cardinality data
  * @param tsCount total number of timeSeries under this shardKeyPrefix (example, number of timeseries under ws,ns)
  * @param activeTsCount number of actively ingesting timeSeries under this shardKeyPrefix
@@ -12,17 +13,8 @@ import java.io.Closeable
  * @param childrenCount number of immediate children for this shardKey (example, number of ns under ws)
  * @param childrenQuota quota for number of immediate children
  */
-case class Cardinality(prefix: Seq[String], tsCount: Int, activeTsCount: Int, childrenCount: Int, childrenQuota: Int)
-
-/**
- * Interface for CardinalityStore iterator.
- */
-trait CardIter extends Iterator[Cardinality] with Closeable {
-  /**
-   * Must be called before iteration.
-   */
-  def seek(): Unit
-}
+case class CardinalityRecord(shard: Int, prefix: Seq[String], tsCount: Int,
+                             activeTsCount: Int, childrenCount: Int, childrenQuota: Int)
 
 /**
  *
@@ -67,13 +59,13 @@ trait CardinalityStore {
    * This method will be called for each shard key prefix when a new time series is added
    * to the index.
    */
-  def store(card: Cardinality): Unit
+  def store(card: CardinalityRecord): Unit
 
   /**
    * Read existing cardinality value, if one does not exist return the zero value
    * indicated
    */
-  def getOrZero(shardKeyPrefix: Seq[String], zero: Cardinality): Cardinality
+  def getOrZero(shardKeyPrefix: Seq[String], zero: CardinalityRecord): CardinalityRecord
 
   /**
    * Remove entry from store. Need to call for each shard key prefix to fully remove shard key.
@@ -85,7 +77,7 @@ trait CardinalityStore {
    * Fetch children of the node for the given shard key prefix.
    * @param depth: only children of this size will be scanned.
    */
-  def scanChildren(shardKeyPrefix: Seq[String], depth: Int): CardIter
+  def scanChildren(shardKeyPrefix: Seq[String], depth: Int): Iterator[CardinalityRecord] with Closeable
 
   /**
    * Close store. Data will be thrown away
--- a/core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityStore.scala
+++ b/core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityStore.scala
@@ -15,6 +15,16 @@ import java.io.Closeable
 case class Cardinality(prefix: Seq[String], tsCount: Int, activeTsCount: Int, childrenCount: Int, childrenQuota: Int)
 
 /**
+ * Interface for CardinalityStore iterator.
+ */
+trait CardIter extends Iterator[Cardinality] with Closeable {
+  /**
+   * Must be called before iteration.
+   */
+  def seek(): Unit
+}
+
+/**
  *
  * Abstracts storage of cardinality for each shard prefix.
  *
@@ -75,7 +85,7 @@ trait CardinalityStore {
    * Fetch children of the node for the given shard key prefix.
    * @param depth: only children of this size will be scanned.
    */
-  def scanChildren(shardKeyPrefix: Seq[String], depth: Int): Iterator[Cardinality] with Closeable
+  def scanChildren(shardKeyPrefix: Seq[String], depth: Int): CardIter
 
   /**
    * Close store. Data will be thrown away
--- a/core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityStore.scala
+++ b/core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityStore.scala
@@ -1,7 +1,10 @@
 package filodb.core.memstore.ratelimit
 
+import java.io.Closeable
+
 /**
  * The data stored in each node of the Cardinality Store trie
+ *
  * @param name name of the item in shardKeyPrefix
  * @param tsCount total number of timeSeries under this shardKeyPrefix (example, number of timeseries under ws,ns)
  * @param activeTsCount number of actively ingesting timeSeries under this shardKeyPrefix
@@ -72,7 +75,7 @@ trait CardinalityStore {
    * Fetch children of the node for the given shard key prefix.
    * @param depth: only children of this size will be scanned.
    */
-  def scanChildren(shardKeyPrefix: Seq[String], depth: Int): Seq[Cardinality]
+  def scanChildren(shardKeyPrefix: Seq[String], depth: Int): Iterator[Cardinality] with Closeable
 
   /**
    * Close store. Data will be thrown away
--- a/core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityStore.scala
+++ b/core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityStore.scala
@@ -69,11 +69,9 @@ trait CardinalityStore {
   def remove(shardKeyPrefix: Seq[String]): Unit
 
   /**
-   * Fetch immediate children of the node for the given shard key prefix.
+   * Fetch children of the node for the given shard key prefix.
+   * @param depth: only children of this size will be scanned.
    */
-  def scanImmediateChildren(shardKeyPrefix: Seq[String]): Seq[Cardinality]
-
-  // TODO(a_theimer)
   def scanChildren(shardKeyPrefix: Seq[String], depth: Int): Seq[Cardinality]
 
   /**
--- a/core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityStore.scala
+++ b/core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityStore.scala
@@ -1,7 +1,5 @@
 package filodb.core.memstore.ratelimit
 
-import java.io.Closeable
-
 /**
  * The data stored in each node of the Cardinality Store trie
  *
@@ -77,7 +75,7 @@ trait CardinalityStore {
    * Fetch children of the node for the given shard key prefix.
    * @param depth: only children of this size will be scanned.
    */
-  def scanChildren(shardKeyPrefix: Seq[String], depth: Int): Iterator[CardinalityRecord] with Closeable
+  def scanChildren(shardKeyPrefix: Seq[String], depth: Int): Seq[CardinalityRecord]
 
   /**
    * Close store. Data will be thrown away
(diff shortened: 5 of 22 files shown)


query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
```diff
@@ -66,25 +66,59 @@ trait MetadataDistConcatExec extends NonLeafExecPlan {
...
+  def deSerialize[T](bytes: Array[Byte]): T = {
+    val byte_is = new ByteArrayInputStream(bytes)
+    val obj_is = new ObjectInputStream(byte_is)
+    val res = obj_is.readObject().asInstanceOf[T]
+    byte_is.close()
+    obj_is.close()
+    res
+  }
+}
+
```
alextheimer, line 88: Not 100% sure this is the best way to go about things, but it seems to work

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
```diff
@@ -395,75 +419,214 @@ final case class LabelCardinalityExec(queryContext: QueryContext,
...
-  //   https://datasketches.apache.org/docs/Frequency/FrequentItemsErrorTable.html
-  val MAX_ITEMSKETCH_MAP_SIZE = 256;
+  val MAX_RESPONSE_SIZE = 5000
+
+  // just to keep memStore.topKCardinality calls consistent
+  val ADD_INACTIVE = true
+
+  val OVERFLOW_NAME = "_overflow_".utf8
+
+  // TODO(a_theimer): Int? Long?
```
alextheimer, line 434: TODO: remove this

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
```diff
@@ -395,75 +419,214 @@ final case class LabelCardinalityExec(queryContext: QueryContext,
...
+   *   prefixes with the specified size.
+   *
+   * Note: next() will always return the same array instance.
+   *
+   * @param tsMemStore contains the shard keys over which to iterate
+   * @param initPrefix shard key prefix to resolve
+   * @param extendToSize the length of the prefix children to iterate through.
+   *   Must be at (1) least initPrefix.size and (2) greater than zero.
+   */
+  class PrefixIterator (tsMemStore: TimeSeriesMemStore,
```
alextheimer, line 466: This is repurposed from an older piece of code. Its original intent was to keep as little in memory as possible while expanding shard key prefixes. It will do that same job here, but its benefit is probably less substantial since every prefix needs to be kept in-memory, now, anyways. Let me know if I should trash this in favor of something less complicated.

memory/src/main/scala/filodb.memory/format/ZeroCopyBinary.scala
```diff
@@ -115,7 +115,8 @@ object ZeroCopyBinary {
  */
 // scalastyle:off
 // FIXME Needs to use MemoryAccessor
-final class ZeroCopyUTF8String(val base: Any, val offset: Long, val numBytes: Int) extends ZeroCopyBinary {
+final class ZeroCopyUTF8String(val base: Any, val offset: Long, val numBytes: Int
+                  ) extends ZeroCopyBinary with Serializable {  // TODO(a_theimer): does this actually work?
```
alextheimer, line 119: Is this secretly breaking something?

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
```diff
@@ -96,70 +130,60 @@ final case class TopkCardReduceExec(queryContext: QueryContext,
...
       case (QueryError(_, _, ex), _)         => throw ex
     }.flatten
-      // fold all child sketches into a single sketch
-      .foldLeftL(new ItemsSketch[String](TopkCardExec.MAX_ITEMSKETCH_MAP_SIZE))(sketchFold)
-      .map{ sketch =>
-        val serSketch = ZeroCopyUTF8String(sketch.toByteArray(new ArrayOfStringsSerDe))
-        val it = Seq(SingleValueRowReader(serSketch)).iterator
+      .foldLeftL(new mutable.HashMap[Seq[ZeroCopyUTF8String], CardCounts])(mapFold)
+      .map{ aggMap =>
+        val serMap = ZeroCopyUTF8String(SerializeUtils.serialize(aggMap.toMap))  // TODO(a_theimer): toMap needed?
```
alextheimer, line 135: TODO: figure this out

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
```diff
@@ -395,75 +419,214 @@ final case class LabelCardinalityExec(queryContext: QueryContext,
...
+    //   previous queue.
+    private val spaceQueues = new Array[mutable.Queue[String]](extendToSize)
+
+    // Stores the prefix returned by next().
+    private val currPrefix = new mutable.ArrayBuffer[String](extendToSize)
+
+    private var iFirstEmptyQueue = -1
+
+    // TODO(a_theimer): this feels gross
+    initialize()
```
alextheimer, line 486: This is like a pseudo-constructor, but it would seem gratuitous to initialize this in a companion object.

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
```diff
@@ -395,75 +419,214 @@ final case class LabelCardinalityExec(queryContext: QueryContext,
...
+      } else {
+        currPrefix.reduceToSize(iFirstEmptyQueue)
+        if (iFirstEmptyQueue > 0) {
+          // this space name is still unused
+          currPrefix(iFirstEmptyQueue - 1) = spaceQueues(iFirstEmptyQueue - 1).front
+        }
+        // Fill each successive queue with the topKCardinality result such that
+        //   its argument prefix is defined by the preceding queue front labels.
+        for (iempty <- iFirstEmptyQueue until extendToSize) {
+          // TODO(a_theimer): MAX_RESPONSE_SIZE works, but not exactly the same meaning
```
alextheimer, line 544: TODO: probably just remove the TODO, since MAX_RESPONSE_SIZE is sufficient enough.

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
```diff
@@ -395,75 +419,214 @@ final case class LabelCardinalityExec(queryContext: QueryContext,
...
-final case object TopkCardExec {
+/**
+ * Contains utilities for all TsCardinality materialize() derivatives.
+ */
+final case object TsCardExec {
   // TODO: tune this
-  // See here for help choosing a size:
-  //   https://datasketches.apache.org/docs/Frequency/FrequentItemsErrorTable.html
-  val MAX_ITEMSKETCH_MAP_SIZE = 256;
+  val MAX_RESPONSE_SIZE = 5000
```
alextheimer, line 427: This trips a `require` in `RowReader` that limits blob sizes to 64kb. Unsure yet how to bypass that...

coordinator/src/main/scala/filodb.coordinator/queryplanner/LogicalPlanUtils.scala
```diff
@@ -93,7 +92,7 @@ object LogicalPlanUtils extends StrictLogging {
       case lp: LabelValues              => lp.copy(startMs = timeRange.startMs, endMs = timeRange.endMs)
       case lp: LabelNames               => lp.copy(startMs = timeRange.startMs, endMs = timeRange.endMs)
       case lp: LabelCardinality         => lp.copy(startMs = timeRange.startMs, endMs = timeRange.endMs)
-      case lp: TopkCardinalities        => lp.copy()
+      case lp: TsCardinalities          => lp.copy()
```
amolnayak311, line 95: No need to copy, we don't mutate the state anyways and there is no change in the input logical plan and output logical plan.

coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
```diff
@@ -140,7 +140,7 @@ class MultiPartitionPlanner(partitionLocationProvider: PartitionLocationProvider
       case lp: ApplyAbsentFunction        => super.materializeAbsentFunction(qContext, lp)
       case lp: ScalarBinaryOperation      => super.materializeScalarBinaryOperation(qContext, lp)
       case lp: ApplyLimitFunction         => super.materializeLimitFunction(qContext, lp)
-      case _: TopkCardinalities           => throw new IllegalArgumentException("TopkCardinalities unexpected here")
+      case _: TsCardinalities             => throw new IllegalArgumentException("TsCardinalities unexpected here")
```
amolnayak311, line 143: When we provide just the ws and we need to get all ws/ns under the given ws, it will be a multi partition call. We will need this implementation in that case.

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
```diff
@@ -144,9 +144,9 @@ class SingleClusterPlanner(val dataset: Dataset,
             reduceExec.addRangeVectorTransformer(new LabelCardinalityPresenter())
             reduceExec
           }
-          case lce: TopkCardExec => {
-            val reducer = TopkCardReduceExec(qContext, targetActor, many, lce.k)
-            reducer.addRangeVectorTransformer(TopkCardPresenter(lce.k))
+          case lce: TsCardExec => {
+            val reducer = TsCardReduceExec(qContext, targetActor, many)
+            reducer.addRangeVectorTransformer(TsCardPresenter(lce.groupDepth))
```
amolnayak311, line 149: Add presenter only if no further reduction is necessary, we will need it when we implement handling multi partition calls (previous comment).

memory/src/main/scala/filodb.memory/format/ZeroCopyBinary.scala
amolnayak311: Why ``Serializable``?

query/src/main/scala/filodb/query/LogicalPlan.scala
```diff
@@ -29,7 +29,7 @@ sealed trait LogicalPlan {
       case l: LabelValues              => l.copy(filters = filters)
       case n: LabelNames               => n.copy(filters = filters)
       case s: SeriesKeysByFilters      => s.copy(filters = filters)
-      case c: TopkCardinalities        => c.copy()
+      case c: TsCardinalities          => c.copy()
```
amolnayak311, line 32: No need to copy the object as its immutable and there are no changes to make.

coordinator/src/main/scala/filodb.coordinator/queryplanner/LogicalPlanUtils.scala
alextheimer: Updated

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
alextheimer: Updated

memory/src/main/scala/filodb.memory/format/ZeroCopyBinary.scala
alextheimer: Just so I can use byte/object input/output streams to serialize the `Map[ZeroCopyUTF8String, CardCounts]` I use to store cardinality data.

query/src/main/scala/filodb/query/LogicalPlan.scala
alextheimer: Updated

coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
```diff
@@ -429,6 +429,33 @@ class MultiPartitionPlanner(partitionLocationProvider: PartitionLocationProvider
...
+    val execPlan = if (partitions.isEmpty) {
+      logger.warn(s"No partitions found for ${queryParams.startSecs}, ${queryParams.endSecs}")
+      localPartitionPlanner.materialize(lp, qContext)
+    } else {
+      val leafPlans = partitions.map { p =>
+        logger.debug(s"partitionInfo=$p; queryParams=$queryParams")
+        if (p.partitionName.equals(localPartitionName))
+          localPartitionPlanner.materialize(lp, qContext)
+        else {
+          createMetadataRemoteExec(qContext, p, Map("match[]" -> queryParams.promQl))
```
alextheimer, line 449: This is following suit with `LabelNames` and `LabelCardinality`. I'm not 100% sure this is right, though, and I need to figure out how to test whether-or-not it works

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
```diff
@@ -134,6 +134,7 @@ class SingleClusterPlanner(val dataset: Dataset,
         many.head match {
           case _: LabelValuesExec => LabelValuesDistConcatExec(qContext, targetActor, many)
           case _: LabelNamesExec => LabelNamesDistConcatExec(qContext, targetActor, many)
+          case _: TsCardExec => TsCardReduceExec(qContext, targetActor, many)
```
alextheimer, line 137: Where should I cap this off with a presenter? It I do it inside `SingleClusterPlanner`, then `MultiPartitionPlanner` is broken. If I do it in `MultiPartitionPlanner`, then `TsCardinalities` can't be materialized by `SingleClusterPlanners`.

coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
alextheimer: deferred to later commit

coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
alextheimer: deferred to later PR

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
alextheimer: deferred to later PR

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
```diff
@@ -96,70 +130,60 @@ final case class TopkCardReduceExec(queryContext: QueryContext,
...
       case (QueryError(_, _, ex), _)         => throw ex
     }.flatten
-      // fold all child sketches into a single sketch
-      .foldLeftL(new ItemsSketch[String](TopkCardExec.MAX_ITEMSKETCH_MAP_SIZE))(sketchFold)
-      .map{ sketch =>
-        val serSketch = ZeroCopyUTF8String(sketch.toByteArray(new ArrayOfStringsSerDe))
-        val it = Seq(SingleValueRowReader(serSketch)).iterator
+      .foldLeftL(new mutable.HashMap[Seq[ZeroCopyUTF8String], CardCounts])(mapFold)
+      .map{ aggMap =>
+        val serMap = ZeroCopyUTF8String(SerializeUtils.serialize(aggMap.toMap))
```
alextheimer, line 135: Any clever way to bypass that `toMap`?

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
```diff
@@ -395,75 +419,212 @@ final case class LabelCardinalityExec(queryContext: QueryContext,
...
+    //   contains labels for the "subspace" of the label at the front of the
+    //   previous queue.
+    private val spaceQueues = new Array[mutable.Queue[String]](extendToSize)
+
+    // Stores the prefix returned by next().
+    private val currPrefix = new mutable.ArrayBuffer[String](extendToSize)
+
+    private var iFirstEmptyQueue = -1
+
+    initialize()
```
alextheimer, line 484: This is like a pseudo-constructor, but it would seem gratuitous to initialize this in a companion object.

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
alextheimer: I see now that this actually encodes the fully-qualified names of every object-- working on a better fix, now

alextheimer: I realize now that this encodes a fully-qualified type with every single reference. Fixing now...

alextheimer: Fixed in recent commit

memory/src/main/scala/filodb.memory/format/ZeroCopyBinary.scala
alextheimer: Removed this change in most recent commit

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
```diff
@@ -395,75 +397,281 @@ final case class LabelCardinalityExec(queryContext: QueryContext,
...
+    result.toMap
+  }
+
+  def serializeMap(myMap: Map[Seq[ZeroCopyUTF8String], CardCounts]): Array[Byte] = {
+    // Need a quick way to access the underlying array.
+    val builder = new ArrayBuilder.ofByte
+
+    // Pre-allocate a space to store the bytes for use with a
+    //   Java ByteBuffer. Then we'll addAll() to the ArrayBuilder.
+    // TODO: unsafe casts to bypass individual byte writes?
```
alextheimer, line 460: This would probably warrant a new `ArrayBuffer`-esque class with `appendSize` and `appendBlob` methods, and that feels like it would be overkill.

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
```diff
@@ -395,75 +397,283 @@ final case class LabelCardinalityExec(queryContext: QueryContext,
...
+                  totalPrefix -> CardCounts(card.activeTsCount, card.tsCount)
+                }
+            }.toMap
           }
-          // serialize the sketch; pack it into a RangeVector
-          val serSketch = ZeroCopyUTF8String(sketch.toByteArray(new ArrayOfStringsSerDe))
-          val it = Seq(SingleValueRowReader(serSketch)).iterator
+
+          // serialize the map; pack it into a RangeVector
+          val serMap = ZeroCopyUTF8String(serializeMap(countMap.toMap))
```
vishramachandran, line 663: The maximum string length that can be added is 65536.  See RecordBuilder.scala line 182. 
Can you think of adding each entry of the map as one record ?


cc @amolnayak311  Please check if your PR has this limitation.

vishramachandran (changes requested):
@alextheimer Thanks for the quick changes!

I have added some comments for you to work on. I am still reviewing the grouping of results. Will post another set of comments when I am done.

query/src/main/scala/filodb/query/LogicalPlan.scala
```diff
@@ -126,11 +126,56 @@ case class SeriesKeysByFilters(filters: Seq[ColumnFilter],
...
-case class TopkCardinalities(shardKeyPrefix: Seq[String], k: Int, addInactive: Boolean) extends LogicalPlan
+object GroupDepth extends Enumeration {
+  protected case class GroupDepthVal(prefixIndex: Int) extends super.Val
+
+  import scala.language.implicitConversions
+  implicit def valueToGroupDepthVal(v: Value): GroupDepthVal = v.asInstanceOf[GroupDepthVal]
+
+  val WS = GroupDepthVal(0)
+  val NS = GroupDepthVal(1)
+  val METRIC = GroupDepthVal(2)
```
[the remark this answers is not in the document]
alextheimer: Updated! But let me know if documentation/assertions need to be even more general (they still heavily imply shard keys are always size 3).

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
vishramachandran: Alex, consider returning N rows of this schema: (group, activeInt, totalInt)
where group is a concatenation of strings with a delimiter.

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
```diff
@@ -395,75 +397,281 @@ final case class LabelCardinalityExec(queryContext: QueryContext,
...
-  val MAX_ITEMSKETCH_MAP_SIZE = 256;
+  val MAX_RESPONSE_SIZE = 5000
+
+  // just to keep memStore.topKCardinality calls consistent
+  val ADD_INACTIVE = true
+
+  val OVERFLOW_NAME = "_overflow_".utf8
+
+  case class CardCounts(active: Int, total: Int) {
+    require(total >= active, "total must be at least as large as active")
```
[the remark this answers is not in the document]
alextheimer: Updated

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
```diff
@@ -71,20 +69,34 @@ final case class PartKeysDistConcatExec(queryContext: QueryContext,
...
+      map.foreach{ case (names, counts) =>
+        val accCountsOpt = acc.get(names)
+        // check if we either (1) won't increase the size or (2) have enough room for another
+        val accCounts = if (accCountsOpt.nonEmpty || acc.size < MAX_RESPONSE_SIZE) {
+          accCountsOpt.getOrElse(CardCounts(0, 0))
+        } else {
+          acc.getOrElseUpdate(names.map(_ => OVERFLOW_NAME).toSeq, CardCounts(0, 0))
+        }
+        val sumCounts = CardCounts(accCounts.active + counts.active,
+                                   accCounts.total + counts.total)
```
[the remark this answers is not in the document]
alextheimer: Updated

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
```diff
@@ -395,75 +397,283 @@ final case class LabelCardinalityExec(queryContext: QueryContext,
...
+    val bytes = java.nio.ByteBuffer.wrap(base, offset, NUM_INT_BYTES)
+    bytes.putInt(num)
+  }
+
+  private def bytesToInt(base: Array[Byte], offset: Int): Int = {
+    val reader = java.nio.ByteBuffer.wrap(base, offset, NUM_INT_BYTES)
+    reader.getInt()
+  }
+
+  def deSerializeMap(bytes: Array[Byte]): Map[Seq[ZeroCopyUTF8String], CardCounts] = {
```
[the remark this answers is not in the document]
alextheimer: No longer applicable-- ExecPlans return a row for each group.

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
alextheimer: Updated: presenter removed, serialization removed.

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
alextheimer: Resolved after schema change

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
alextheimer: Resolved

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
alextheimer: Resolved after schema change

coordinator/src/main/scala/filodb.coordinator/client/QueryOps.scala
```diff
@@ -51,7 +51,10 @@ trait QueryOps extends ClientBase with StrictLogging {
                      k: Int,
                      addInactive: Boolean,
                      timeout: FiniteDuration = 15.seconds): Seq[CardinalityRecord] =
-    askCoordinator(GetTopkCardinality(dataset, shards, shardKeyPrefix, k, addInactive), timeout) {
+    askCoordinator(
+      GetTopkCardinality(dataset, shards, shardKeyPrefix,
+                         shardKeyPrefix.size + 1, k, addInactive),
```
alextheimer, line 56: This just preserves the original intent.

core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityStore.scala
```diff
@@ -54,7 +57,7 @@ trait CardinalityStore {
    * This method will be called for each shard key prefix when a new time series is added
    * to the index.
    */
-  def store(shardKeyPrefix: Seq[String], card: Cardinality): Unit
+  def store(card: Cardinality): Unit
```
alextheimer, line 60: `Cardinality` contains all the info we need

core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityStore.scala
```diff
@@ -14,6 +14,16 @@ import java.io.Closeable
...
+/**
+ * Interface for CardinalityStore iterator.
+ */
+trait CardIter extends Iterator[Cardinality] with Closeable {
+  /**
+   * Must be called before iteration.
+   */
+  def seek(): Unit
+}
+
```
alextheimer: The intent is to force the client to acknowledge any I/O errors-- the previous setup just ate any errors that occurred while `Cardinality`s were scanned. The client should be more aware of the backend, now, anyways, since `close()` will be left uncalled if the iterator is not exhausted.

core/src/main/scala/filodb.core/memstore/ratelimit/RocksDbCardinalityStore.scala
```diff
@@ -144,108 +160,129 @@ class RocksDbCardinalityStore(ref: DatasetRef, shard: Int) extends CardinalitySt
...
     logger.debug(s"Storing shard=$shard dataset=$ref ${new String(key)} with $card")
-    db.put(key, cardinalityToBytes(card))
+    db.put(key, cardinalityNodeToBytes(CardinalityNode.fromCardinality(card)))
   }
 
   def getOrZero(shardKeyPrefix: Seq[String], zero: Cardinality): Cardinality = {
-    val value = db.get(toStringKey(shardKeyPrefix, false).getBytes(StandardCharsets.UTF_8))
-    if (value == NotFound) zero else bytesToCardinality(value)
+    val value = db.get(toStringKey(shardKeyPrefix).getBytes(StandardCharsets.UTF_8))
+    if (value == NotFound) zero else CardinalityNode.toCardinality(bytesToCardinalityNode(value), shardKeyPrefix)
```
vishramachandran, line 241: Not a very big deal but, can do in one step and avoid allocation.

core/src/main/scala/filodb.core/memstore/ratelimit/RocksDbCardinalityStore.scala
```diff
@@ -144,108 +160,129 @@ class RocksDbCardinalityStore(ref: DatasetRef, shard: Int) extends CardinalitySt
...
+    b.toString()
   }
 
-  private def cardinalityToBytes(card: Cardinality): Array[Byte] = {
+  /**
+   * Builds keys as:
+   *   <keyDepth>{KeySeparator}<name-1>{KeySeparator}<name-2> ... <name-n>
+   * The final {KeySeparator} is always omitted.
+   */
+  private def toStringKey(shardKeyPrefix: Seq[String]): String = {
```
vishramachandran, line 215: Document parameter as: fully specified RocksDB key. Result cannot be used for prefix search. It is used to lookup a single key.

core/src/main/scala/filodb.core/memstore/ratelimit/RocksDbCardinalityStore.scala
```diff
@@ -144,108 +160,129 @@ class RocksDbCardinalityStore(ref: DatasetRef, shard: Int) extends CardinalitySt
...
+      val it_ = db.newIterator()
+      var nextCard_ = Cardinality(Seq("should be overwritten before exposed"), -1, -1, -1, -1)
+      val strPrefix_ = toStringKeyPrefix(shardKeyPrefix, depth)
+      var isSeeked = false
+
+      override def seek(): Unit = {
+        logger.debug(s"Scanning shard=$shard dataset=$ref ${new String(strPrefix_)}")
+        it_.seek(strPrefix_.getBytes(StandardCharsets.UTF_8))
+        isSeeked = true
+      }
```
vishramachandran, line 262: Can't you avoid the `seek` method call requirement by doing the seek in the constructor ?

core/src/main/scala/filodb.core/memstore/ratelimit/RocksDbCardinalityStore.scala
```diff
@@ -144,108 +160,129 @@ class RocksDbCardinalityStore(ref: DatasetRef, shard: Int) extends CardinalitySt
...
+
+      // note: must be called exactly once before each next() call
+      override def hasNext: Boolean = {
+        assert(isSeeked, "must call seek() before use")
+        if (it_.isValid) {
+          // store the next matching key and increment the iterator
+          val key = new String(it_.key(), StandardCharsets.UTF_8)
+          if (key.startsWith(strPrefix_)) {
+            val node = bytesToCardinalityNode(it_.value())
+            val prefix = key.split(KeySeparator).drop(1)
```
vishramachandran, line 272: Why are we dropping the first one ? Since it is "" ? Can you document ?

core/src/main/scala/filodb.core/memstore/ratelimit/RocksDbCardinalityStore.scala
```diff
@@ -144,108 +160,129 @@ class RocksDbCardinalityStore(ref: DatasetRef, shard: Int) extends CardinalitySt
...
-      import scala.util.control.Breaks._
-
-      breakable {
-        while (it.isValid()) {
-          val key = new String(it.key(), StandardCharsets.UTF_8)
-          if (key.startsWith(searchPrefix)) {
-            buf += bytesToCardinality(it.value())
-          } else break // dont continue beyond valid results
-          it.next()
+  override def scanChildren(shardKeyPrefix: Seq[String], depth: Int): CardIter = {
```
vishramachandran, line 248: Need to document in BOLD that the iterator needs to be closed after consumption. Else there will be an off-heap leak that is extremely hard to identify. Were you able to verify if it will be closed?

IIRC, this was one reason I did not choose iterators for this. But if we can be assured that we will safely close then I am ok with this.

core/src/main/scala/filodb.core/memstore/ratelimit/RocksDbCardinalityStore.scala
```diff
@@ -144,108 +160,129 @@ class RocksDbCardinalityStore(ref: DatasetRef, shard: Int) extends CardinalitySt
...
         }
+        it_.close()
+        false
       }
-    } finally {
-      it.close();
+
+      override def next(): Cardinality = nextCard_
+
+      override def close(): Unit = it_.close()
```
vishramachandran, line 284: Does it need an "if-not-already-closed" check?

core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityTracker.scala
```diff
@@ -169,30 +169,48 @@ class CardinalityTracker(ref: DatasetRef,
...
+   * Use this method to query the top-k cardinality consumers
+   *   under a provided shard key prefix.
    *
-   * @param k
+   * @param depth cardinalities are returned for all prefixes of this size
    * @param shardKeyPrefix zero or more elements that form a valid shard key prefix
    * @return Top-K records, can the less than K if fewer children
    */
-  def topk(k: Int, shardKeyPrefix: Seq[String], addInactive: Boolean): Seq[CardinalityRecord] = {
+  def topk(k: Int, shardKeyPrefix: Seq[String], depth: Int, addInactive: Boolean): Seq[CardinalityRecord] = {
```
vishramachandran, line 179: I wonder if we should get rid of 'k' and use a server configured limit instead. This can remove the priority queue too. The method would be renamed as `scan`. Thoughts ?

core/src/main/scala/filodb.core/memstore/ratelimit/RocksDbCardinalityStore.scala
vishramachandran: After looking at the details, the benefit of the iterator would be reduced if we constructed `CardinalityRecord` here per my other suggestion. We can simplify and return a Seq here instead of the iterator close complexity.

vishramachandran (changes requested):
Thank you for diving into details of RocksDB card tracker and pushing capability down into it.

Some aesthetic and possibly small performance fixes requested. Generally looks good!

core/src/main/scala/filodb.core/memstore/ratelimit/RocksDbCardinalityStore.scala
```diff
@@ -20,25 +19,42 @@ import spire.syntax.cfor._
 import filodb.core.{DatasetRef, GlobalScheduler}
 import filodb.memory.format.UnsafeUtils
 
-class CardinalitySerializer extends Serializer[Cardinality] {
-  def write(kryo: Kryo, output: Output, card: Cardinality): Unit = {
+/**
+ * Stored as values in the RocksDb database.
+ * Identical to Cardinality, except that only the least-significant prefix name is stored.
+ */
+case class CardinalityNode(name: String, tsCount: Int, activeTsCount: Int,
```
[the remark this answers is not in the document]
alextheimer: I'll implement this idea, but as an alternative:

Why not have some `CardinalityCounts` class that contains all the `tsCount`, `activeTsCount`, etc. values that are common to `CardinalityNode`, `CardinalityRecord`, and `Cardinality`? I think the three classes serve different-enough purposes to keep separate, and a `CardinalityCounts` class would make more clear what those purposes are (map a name to cardinality data, map a prefix to cardinality data, and map a prefix/shard to cardinality data). Plus, shard data isn't relevant to the `CardinalityStore` API.

core/src/main/scala/filodb.core/memstore/ratelimit/RocksDbCardinalityStore.scala
alextheimer: Done!

core/src/main/scala/filodb.core/memstore/ratelimit/RocksDbCardinalityStore.scala
alextheimer: Ah, I assumed any `RocksIterator` I/O calls could throw errors, but I see that was wrong. I removed the seek requirement.

core/src/main/scala/filodb.core/memstore/ratelimit/RocksDbCardinalityStore.scala
alextheimer: Added!

core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityTracker.scala
alextheimer: I like it 100x better than `topk` (unless there's something magic `CardinalityTracker` can do to make `topk` happen super fast, but that doesn't seem to be the case). The caller can find the top k if they want, but that's out-of-scope for the `CardinalityTracker`.

Maybe there shouldn't be a `k` at all? Would this be a good spot for an iterator?

core/src/main/scala/filodb.core/memstore/ratelimit/RocksDbCardinalityStore.scala
alextheimer: Nope. From RocksDB:
```
  @Override
  public void close() {
    if (owningHandle_.compareAndSet(true, false)) {
      disposeInternal();
    }
  }
```
But this is not explicitly documented as "`close` is guaranteed idempotent". Want me to add a check anyways (if we keep the iterator)?

core/src/main/scala/filodb.core/memstore/ratelimit/RocksDbCardinalityStore.scala
alextheimer: Done!

core/src/main/scala/filodb.core/memstore/ratelimit/RocksDbCardinalityStore.scala
alextheimer: Irrelevant now-- the iterator is replaced by a Seq

core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityTracker.scala
alextheimer: I replaced `topk` with `scan(shardKeyPrefix: Seq[String], depth: Int): Seq[CardinalityRecord]`. `NMAX` is configurable within the `CardinalityStore`, and a warning is logged if the result is clipped.

core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityTracker.scala
```diff
@@ -169,50 +167,22 @@ class CardinalityTracker(ref: DatasetRef,
...
-        else x.activeTsCount - y.activeTsCount
-      }
-    }.reverse
-
     if (shardKeyPrefix.size == depth) {
-      // directly fetch the cardinality and return
-      val card = getCardinality(shardKeyPrefix)
-      return Seq(CardinalityRecord(
-        shard, card.prefix, card.tsCount, card.activeTsCount,
-        if (shardKeyPrefix.length == shardKeyLen - 1) card.tsCount else card.childrenCount,
```
alextheimer: I removed this `if` and the one formerly at line 208. No tests fail, but I'm not 100% certain of why they were there. Should I push this `if` down into the `CardinalityStore`? Put them back here?

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
```diff
@@ -395,75 +348,96 @@ final case class LabelCardinalityExec(queryContext: QueryContext,
...
+                 total + other.total)
+    }
+  }
+
+  /**
+   * Convenience class for converting between RowReader and TsCardExec data.
+   */
+  case class RowData(group: ZeroCopyUTF8String, counts: CardCounts) {
+    def toRowReader(): RowReader = {
+      SeqRowReader(Seq(group, counts.active, counts.total))
```
vishramachandran, line 374: This is ok, but you can consider creating a custom implementation of RowReader to avoid the Seq allocation. There are examples in code.

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
```diff
@@ -395,75 +348,96 @@ final case class LabelCardinalityExec(queryContext: QueryContext,
...
+    }
+  }
+  object RowData {
+    def fromRowReader(rr: RowReader): RowData = {
+      val group = rr.getAny(0).asInstanceOf[ZeroCopyUTF8String]
```
vishramachandran, line 379: Curious to know what RowReader implementation it gets deserialized into when reading.

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
```diff
@@ -395,75 +348,96 @@ final case class LabelCardinalityExec(queryContext: QueryContext,
...
-final case object TopkCardExec {
-  // TODO: tune this
-  // See here for help choosing a size:
-  //   https://datasketches.apache.org/docs/Frequency/FrequentItemsErrorTable.html
-  val MAX_ITEMSKETCH_MAP_SIZE = 256;
+/**
+ * Contains utilities for all TsCardinality materialize() derivatives.
+ */
+final case object TsCardExec {
+  val MAX_RESPONSE_SIZE = 5000  // TODO: tune this
```
vishramachandran, line 355: There are multiple constants NMAX, and this one. Can we consolidate ?

core/src/main/scala/filodb.core/memstore/ratelimit/RocksDbCardinalityStore.scala
```diff
@@ -144,101 +168,131 @@ class RocksDbCardinalityStore(ref: DatasetRef, shard: Int) extends CardinalitySt
...
-            buf += bytesToCardinality(it.value())
-          } else break // dont continue beyond valid results
+            val node = bytesToCardinalityValue(it.value())
+            // Drop the first element, since it's just the size of the prefix.
+            // Ex: 2{KeySeparator}A{KeySeparator}B ==(split)==> [2, A, B] ==(drop)==> [A, B]
+            val prefix = key.split(KeySeparator).drop(1)
+            buf += CardinalityValue.toCardinalityRecord(node, prefix, shard)
+          } else {
+            if (buf.size == NMAX) {
+              logger.warn(s"scanChildren result clipped")
```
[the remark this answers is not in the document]
vishramachandran: On second thoughts, I think we should push the "OTHERS" logic into here, and consolidate the two constants NMAX and MAX_RESPONSE_SIZE. Feel free to defer to another PR if you'd like.

vishramachandran (approved):
Great work! Thanks for the attention to detail.
I have added some comments. Feel free to defer some of these to another PR as appropriate.

core/src/main/scala/filodb.core/memstore/ratelimit/RocksDbCardinalityStore.scala
```diff
@@ -144,101 +168,131 @@ class RocksDbCardinalityStore(ref: DatasetRef, shard: Int) extends CardinalitySt
...
   }
 
-  override def scanChildren(shardKeyPrefix: Seq[String]): Seq[Cardinality] = {
+  override def scanChildren(shardKeyPrefix: Seq[String],
+                            depth: Int): Seq[CardinalityRecord] = {
+
+    require(depth > shardKeyPrefix.size,
+      s"scan depth $depth must be greater than the size of the prefix ${shardKeyPrefix.size}")
+
+    val NMAX = 10000
```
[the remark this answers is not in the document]
alextheimer: Done!

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
alextheimer: `MAX_RESPONSE_SIZE` is network-limited, and (formerly-named) `NMAX` is memory-limited. If those two are similar enough, I'll just use `NMAX` in both places. Does that sound good?

core/src/main/scala/filodb.core/memstore/ratelimit/RocksDbCardinalityStore.scala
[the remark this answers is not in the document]
alextheimer: Done!

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
alextheimer: Ended up doing this, so marking resolved

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
alextheimer: It's a BinaryRecordRowReader. Is that a bad thing?

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
alextheimer: Done!

PR #1294, opened 2021-11-30
feat(query): namespace cardinality publisher

alextheimer:
**Pull Request checklist**

**New behavior :**
Adds a `TenantIngestionMetering` instance to the leader `NodeCoordinatorActor` (via the `ShardManager`). This periodically queries a `NodeCoordinatorActor` for namespace-level cardinality data, then it publishes the results as `Kamon` metrics.


coordinator/src/main/scala/filodb.coordinator/NamespaceCardinalityPublisher.scala
```diff
@@ -0,0 +1,103 @@
...
+  // All "_TU" constants quantify units of TIME_UNIT.
+  private val TIME_UNIT = TimeUnit.SECONDS
+  private val SCHED_INIT_DELAY_TU = 10
+  private val SCHED_DELAY_TU = 10
+  private val ASK_TIMEOUT_TU = 30
+
+  private val METRIC_ACTIVE = "active"
+  private val METRIC_TOTAL = "total"
+  private val NS = "ns_agg"
+  private val WS = "ns_agg"
```
alextheimer, line 37: Do these look alright?

Should `NS` and `WS` be taken as argument?

coordinator/src/main/scala/filodb.coordinator/NamespaceCardinalityPublisher.scala
```diff
@@ -0,0 +1,103 @@
...
+
+  // (future, dataset) pairs
+  private val futQueue_ = new java.util.concurrent.ConcurrentLinkedQueue[(Future[Any], String)]()
```
alextheimer, line 40: This is thread-safe mostly just-in-case something goes catastrophically wrong with the scheduler.

coordinator/src/main/scala/filodb.coordinator/NamespaceCardinalityPublisher.scala
```diff
@@ -0,0 +1,103 @@
...
+ * The intent is to publish a low-cardinality metric such that namespace
+ * cardinality queries can be efficiently answered.
+ */
+case class NamespaceCardinalityPublisher(dsIterProducer: () => Iterator[DatasetRef],
+                                         coordProducer: () => ActorRef) {
+  // All "_TU" constants quantify units of TIME_UNIT.
+  private val TIME_UNIT = TimeUnit.SECONDS
+  private val SCHED_INIT_DELAY_TU = 10
+  private val SCHED_DELAY_TU = 10
+  private val ASK_TIMEOUT_TU = 30
```
alextheimer, line 32: Whoops, these are still really small.

Is 15 mins still the delay we're sticking with?

coordinator/src/main/scala/filodb.coordinator/NamespaceCardinalityPublisher.scala
```diff
@@ -0,0 +1,103 @@
...
+        case Some(Success(QueryResult(_, _, rv, _, _, _))) =>
+          rv.foreach(_.rows().foreach{ rr =>
+            // publish a cardinality metric for each namespace
+            val data = RowData.fromRowReader(rr)
+            val dataset = futOpt.get._2
+            val tags = Map("_ws_" -> WS, "_ns_" -> NS, "ds" -> dataset, "prefix" -> data.group)
+            Kamon.gauge(METRIC_ACTIVE).withTags(TagSet.from(tags)).update(data.counts.active.toDouble)
+            Kamon.gauge(METRIC_TOTAL).withTags(TagSet.from(tags)).update(data.counts.total.toDouble)
+          })
+        case _ => ???
```
alextheimer, line 82: How should I handle a timeout/error?

coordinator/src/main/scala/filodb.coordinator/NamespaceCardinalityPublisher.scala
```diff
@@ -0,0 +1,103 @@
...
+   */
+  private def queryAndSchedulePublish() : Unit = {
+    val groupDepth = 1
+    val prefix = Nil
+    dsIterProducer().foreach { dsRef =>
+      val fut = Client.asyncAsk(
+        coordProducer(),
+        LogicalPlan2Query(dsRef, TsCardinalities(prefix, groupDepth)),
+        FiniteDuration(ASK_TIMEOUT_TU, TIME_UNIT))
+      futQueue_.add((fut, dsRef.dataset))
```
vishramachandran, line 99: You can simply do this on the future. It will happen asynchronously on the thread you ask it to run the lambda. No need for queue.
```
implicit val sched = scheduler
future.OnComplete {
   case res: Seq[CardinalityRecord] => // update the Kamon gauges
   case ex: Exception => logger.error("Could not fetch cardinality metrics", ex)
} 
```

coordinator/src/main/scala/filodb.coordinator/NamespaceCardinalityPublisher.scala
```diff
@@ -0,0 +1,103 @@
...
+        return
+      }
+
+      futOpt.get._1.value match {
+        case Some(Success(QueryResult(_, _, rv, _, _, _))) =>
+          rv.foreach(_.rows().foreach{ rr =>
+            // publish a cardinality metric for each namespace
+            val data = RowData.fromRowReader(rr)
+            val dataset = futOpt.get._2
+            val tags = Map("_ws_" -> WS, "_ns_" -> NS, "ds" -> dataset, "prefix" -> data.group)
```
vishramachandran, line 78: Cannot use the `_ws_` and `_ns_` tags here since those tags are used for the filodb metric. They would be set globally at one place for all metrics. Need to use something like `metric_ws` and `metric_ns` for the prefix. For now use the first value of prefix for ws and second for ns. OK to hardcode for now.

coordinator/src/main/scala/filodb.coordinator/NamespaceCardinalityPublisher.scala
```diff
@@ -0,0 +1,103 @@
...
+case class NamespaceCardinalityPublisher(dsIterProducer: () => Iterator[DatasetRef],
+                                         coordProducer: () => ActorRef) {
+  // All "_TU" constants quantify units of TIME_UNIT.
+  private val TIME_UNIT = TimeUnit.SECONDS
+  private val SCHED_INIT_DELAY_TU = 10
+  private val SCHED_DELAY_TU = 10
+  private val ASK_TIMEOUT_TU = 30
+
+  private val METRIC_ACTIVE = "active"
+  private val METRIC_TOTAL = "total"
```
vishramachandran, line 35: Use more descriptive names. `active_timeseries_by_tenant` and `total_timeseries_by_tenant` could be some names.

coordinator/src/main/scala/filodb.coordinator/NamespaceCardinalityPublisher.scala
vishramachandran: Need to add clusterType as one more tag. This is important.

coordinator/src/main/scala/filodb.coordinator/NamespaceCardinalityPublisher.scala
vishramachandran: Log and move on. See my code snippet in other comment.

coordinator/src/main/scala/filodb.coordinator/NamespaceCardinalityPublisher.scala
vishramachandran: Ask timeout and publish interval could be same. 

We should make the publish interval configurable so it can be changed easily in the future.

coordinator/src/main/scala/filodb.coordinator/NamespaceCardinalityPublisher.scala
vishramachandran: See other comment.

coordinator/src/main/scala/filodb.coordinator/NamespaceCardinalityPublisher.scala
vishramachandran: This queue is not needed. There is a simpler way to achieve the goal. See other comment.

coordinator/src/main/scala/filodb.coordinator/ShardManager.scala
```diff
@@ -37,6 +37,14 @@ private[coordinator] final class ShardManager(settings: FilodbSettings,
   private val _coordinators = new mutable.LinkedHashMap[Address, ActorRef]
   private val _errorShardReassignedAt = new mutable.HashMap[DatasetRef, mutable.HashMap[Int, Long]]
 
+  private val _shardMetricAggregator = NamespaceCardinalityPublisher(
+    () => { _datasetInfo.map{ case (dsRef, _) => dsRef}.toIterator },
+    () => { _coordinators.head._2 }
+  )
+  if (settings.config.getBoolean("shard-key-level-ingestion-metrics-enabled")) {
+    _shardMetricAggregator.schedulePeriodicPublishJob()
+  }
```
vishramachandran, line 46: `NamespaceCardinalityPublisher` could be renamed to `TenantIngestionMetering`. We dont want concept of Namespace in class names.

Also, we should instantiate the publisher only if `shard-key-level-ingestion-metrics-enabled` is true. It could be done using an option.

vishramachandran (changes requested):
We should wait until the other PR goes through.
Comments added for your consideration.

coordinator/src/main/scala/filodb.coordinator/NamespaceCardinalityPublisher.scala
alextheimer: Done!

coordinator/src/main/scala/filodb.coordinator/NamespaceCardinalityPublisher.scala
alextheimer: Done!

coordinator/src/main/scala/filodb.coordinator/NamespaceCardinalityPublisher.scala
alextheimer: Removed.

coordinator/src/main/scala/filodb.coordinator/NamespaceCardinalityPublisher.scala
alextheimer: Done

coordinator/src/main/scala/filodb.coordinator/ShardManager.scala
alextheimer: Done!

coordinator/src/main/scala/filodb.coordinator/NamespaceCardinalityPublisher.scala
alextheimer: Done!

coordinator/src/main/scala/filodb.coordinator/NamespaceCardinalityPublisher.scala
alextheimer: Resolved

coordinator/src/main/scala/filodb.coordinator/NamespaceCardinalityPublisher.scala
alextheimer: Done

alextheimer:
Rebased with the most recent `develop` updates, and confirmed to work locally.

alextheimer:
352597a smuggles in a test that should have been included in the last PR

core/src/main/resources/filodb-defaults.conf
```diff
@@ -211,6 +211,9 @@ filodb {
   shard-manager {
     # Minimum time required between successive automatic shard reassignments done by ShardManager
     reassignment-min-interval = 2 hours
+
+    # Time between each TenantIngestionMetering query/publish job
+    metering-query-interval = 15 minutes
```
vishramachandran, line 216: Move this config as a sibling of `shard-key-level-ingestion-metrics-enabled` config

coordinator/src/main/scala/filodb.coordinator/TenantIngestionMetering.scala
```diff
@@ -0,0 +1,89 @@
...
+ */
+case class TenantIngestionMetering(settings: FilodbSettings,
+                                   dsIterProducer: () => Iterator[DatasetRef],
+                                   coordActorProducer: () => ActorRef) extends StrictLogging{
+
+  private val ASK_TIMEOUT = FiniteDuration(
+    settings.config.getDuration("shard-manager.metering-query-interval").toSeconds,
+    TimeUnit.SECONDS)
+  private val SCHED_INIT_DELAY = ASK_TIMEOUT  // time until first job is scheduled
+  private val SCHED_DELAY = ASK_TIMEOUT  // time between all jobs after the first
```
vishramachandran, line 38: Let's call out a TODO to fix this in different PR. If for whatever reason the query takes longer than timeout consistently, metric may never be published. Instead, the we could slow down the queries and publish less often. Need not be done in this PR.

vishramachandran (approved):
👍 Please address the two minor suggestions before pushing

core/src/main/resources/filodb-defaults.conf
alextheimer: Done!

coordinator/src/main/scala/filodb.coordinator/TenantIngestionMetering.scala
alextheimer: Noted-- I'll define a throttling function in my next PR.

alextheimer: Included in #1310.

PR #1304, opened 2021-12-10
fix(core): check for UnknownSchema before shardkey retrieval

sherali42:

- check for UnknownSchema before shardKey retrieval. It is moved to code block where it is being used.


tjackpaul (approved):
thanks for the quick PR @sherali42

PR #1305, opened 2021-12-13
bug(prometheus): Enforce maximum regex length

vishramachandran:
**Pull Request checklist**

Upon very long regular expression filters, lucene returns a StackOverflowError on every shard, and this has the potential to degrade performance of the cluster severely. The shard lock is not released and subsequent queries fail. The StackOverflowError has also been seen on ElasticSearch.

The stack overflow error we saw was for a regex filter of length 17k characters, and it was an unintentional regex which was not per intent. It used comma instead of pipe character to delimit list of possibilities. 

**New behavior :**
Since it is very hard to dive into the regex to explicitly allow and disallow certain regexes, we have to resort to enforcing a regular expression filter length limit - currently set to 1KB, which is same as ElasticSearch default.

**BREAKING CHANGES**

This can potentially affect existing queries that have regex > 1KB.

### DIFF
--- a/prometheus/src/main/scala/filodb/prometheus/query/PrometheusModel.scala
+++ b/prometheus/src/main/scala/filodb/prometheus/query/PrometheusModel.scala
@@ -7,6 +7,7 @@ import filodb.core.binaryrecord2.BinaryRecordRowReader
 import filodb.core.metadata.Column.ColumnType
 import filodb.core.metadata.PartitionSchema
 import filodb.core.query.{Result => _, _}
+import filodb.prometheus.parse.Parser.REGEX_MAX_LEN
 import filodb.query.{QueryResult => FiloQueryResult, _}
 import filodb.query.AggregationOperator.Avg
 import filodb.query.exec.{ExecPlan, HistToPromSeriesMapper}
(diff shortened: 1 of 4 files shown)


prometheus/src/main/scala/filodb/prometheus/query/PrometheusModel.scala
```diff
@@ -40,7 +40,9 @@ object PrometheusModel {
         val filter = m.getType match {
           case MatchType.EQUAL => Filter.Equals(m.getValue)
           case MatchType.NOT_EQUAL => Filter.NotEquals(m.getValue)
-          case MatchType.REGEX_MATCH => Filter.EqualsRegex(m.getValue)
+          case MatchType.REGEX_MATCH =>
+            require(m.getValue.length < 1000, "Regular expression filters should be < 1000 characters")
```
amolnayak311, line 44: 1000 is hardcoded here

amolnayak311 (changes requested):

amolnayak311 (approved):

tjackpaul (approved):
thanks for the quick PR @vishramachandran

veermaha (approved):

PR #1312, opened 2021-12-20
feat(query): add Parser.queryToEqualLabelMap

alextheimer:
# Update

See #1314 for the alternative fix.

# Original Text

**Pull Request checklist**

**New behavior :**
Adds the method `Parser.queryToLabelMap(query: String): Map[String, String]` which parses an instant selector into a map of (label, value) pairs (including `__name__`).

### DIFF
--- a/prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
+++ b/prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
@@ -1,7 +1,8 @@
 package filodb.prometheus.ast
 
 import scala.util.Try
-import filodb.core.{GlobalConfig, query}
+
+import filodb.core.{query, GlobalConfig}
 import filodb.core.query.{ColumnFilter, RangeParams}
 import filodb.prometheus.parse.Parser
 import filodb.query._
@@ -11,7 +12,6 @@ object Vectors {
   val TypeLabel       = "_type_"
   val BucketFilterLabel = "_bucket_"
 }
-
 
 object WindowConstants {
   val conf = GlobalConfig.systemConfig
--- a/prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
+++ b/prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
@@ -301,16 +301,20 @@ case class InstantExpression(metricName: Option[String],
   }
 
   /**
-   * Returns a mapping of labels to values.
+   * Returns a mapping from selector labels (including an explicit/implicit __name__) to values.
+   * Throws an IllegalArgumentException if any of the values are not matched by equality without a regex.
    *
    * Note: this InstantExpression does not require a metric name in order
    *   for this method to run to completion.
    */
-  def toLabelMap() : Map[String, String] = {
+  def toEqualLabelMap() : Map[String, String] = {
     // See Vector::mergeNameToLabels for details about why we're ignoring the member variable.
-    val requireName = false
-    val mergeNameToLabelsLocal = makeMergeNameToLabels(requireName)
+    val mergeNameToLabelsLocal = makeMergeNameToLabels(requireName = false)
     mergeNameToLabelsLocal.map{ labelMatch =>
+      if (labelMatch.labelMatchOp != EqualMatch) {
+        throw new IllegalArgumentException(
+          "cannot convert to label map if any values aren't matched by equality")
+      }
       labelMatch.label -> labelMatch.value
     }.toMap
   }
--- a/prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
+++ b/prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
@@ -139,9 +139,30 @@ sealed trait Vector extends Expression {
   val regexColumnName: String = "::(?=[^::]+$)" //regex pattern to extract ::columnName at the end
 
   // Convert metricName{labels} -> {labels, __name__="metricName"} so it's uniform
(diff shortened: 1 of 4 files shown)


prometheus/src/test/scala/filodb/prometheus/parse/ParserSpec.scala
```diff
@@ -726,6 +726,29 @@ class ParserSpec extends AnyFunSpec with Matchers {
...
+    val queryMapPairs = Seq(
+      ("foo", Map("__name__" -> "foo")),
+      ("""{__name__="foo"}""" -> Map("__name__" -> "foo")),
+      ("""{__name__="foo", bar="baz"}""" -> Map("__name__" -> "foo", "bar" -> "baz")),
+      ("""{bar="baz"}""" -> Map("bar" -> "baz")),
+      ("""foo{bar="baz", bog="bah"}""" -> Map("__name__" -> "foo", "bar" -> "baz", "bog" -> "bah")),
+    )
+
+    val queriesShouldFail = Seq(
+      """foo{__name__="bar"}""",
```
alextheimer, line 739: `foo{__name__="foo"}` will fail, but I probably shouldn't explicitly enforce that here

prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
```diff
@@ -139,9 +138,30 @@ sealed trait Vector extends Expression {
...
+   *
+   * If you need to make use of a name-optional Vector (i.e. as a temporary vessel for parsed PromQL),
+   *   the hacky workaround is as follows:
+   *     (1) lazily initialize mergeNameToLabels so a metric name is not explicitly required
+   *         at the Vector's initialization, and
+   *     (2) when needed, locally initialize a "mergeNameToLabels-like" value without
+   *         a name requirement.
+   *
+   * This method exists only to consolidate code for the initialization of these values.
+   */
```
alextheimer, line 161: I've been riffing off this idea on a local branch, but I keep running into the same two problems:
1. There is no guarantee that the validation will ever run.
2. The validation can't be manually run for tests (especially when `InstantExpression`s are nested inside of other `Expression` types) without spaghetti code.

Please, please, please see TODO:PR for a much safer fix. I'll add comments to argue my case there.

alextheimer:
Closing since some tests don't pass, and the clearest way forward is to remove them.

prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
alextheimer: This branch still isn't as clean/straightforward as I had hoped-- let the record show that I still think #1314 is substantially safer.

prometheus/src/test/scala/filodb/prometheus/parse/ParserSpec.scala
```diff
@@ -184,10 +184,6 @@ class ParserSpec extends AnyFunSpec with Matchers {
     parseError("foo::b{gibberish}")
     parseError("foo{1}")
     parseError("{}")
-    parseError("{x=\"\"}")
-    parseError("{x=~\".*\"}")
-    parseError("{x!~\".+\"}")
-    parseError("{x!=\"a\"}")
```
alextheimer, line 190: Since it's alright to parse nameless selectors, now.

prometheus/src/test/scala/filodb/prometheus/parse/ParserSpec.scala
```diff
@@ -726,6 +722,30 @@ class ParserSpec extends AnyFunSpec with Matchers {
...
+  it("should correctly build a label map") {
+    val queryMapPairs = Seq(
+      ("foo", Map("__name__" -> "foo")),
+      ("""{__name__="foo"}""" -> Map("__name__" -> "foo")),
+      ("""{__name__="foo", bar="baz"}""" -> Map("__name__" -> "foo", "bar" -> "baz")),
+      ("""{bar="baz"}""" -> Map("bar" -> "baz")),
+      ("""foo{bar="baz", bog="bah"}""" -> Map("__name__" -> "foo", "bar" -> "baz", "bog" -> "bah")),
+    )
+
+    val queriesShouldFail = Seq(
```
alextheimer, line 734: `foo{__name__="foo"}` will also fail, but I probably shouldn't explicitly enforce that here.

prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
```diff
@@ -139,12 +138,57 @@ sealed trait Vector extends Expression {
...
+  lazy val mergeNameToLabels: Seq[LabelMatch] = makeMergeNameToLabels()
+
+  // These constraints are also applied within makeMergeNameToLabels, but they are
+  //   *immediately* applied here (i.e. not lazily as with mergeNameToLabels).
+  applyPromqlConstraints()
+
+  /**
+   * Apply PromQL-scope constraints to the vector selector:
+   *   (1) only one __name__ specifier (implicit or otherwise) exists.
+   *   (2) at least one label matcher exists.
```
alextheimer, line 151: This should technically be "at least one label matcher does not match the empty string," but there isn't a clean way to enforce this with this setup-- to use a `ColumnFilter`, `mergeNameToLabels` needs to be evaluated.

alextheimer: (this validation exists to pass a `parseError("{}")` test)

vishramachandran (approved):

amolnayak311 (approved):

PR #1314, opened 2021-12-21
==WIP== // feat(query): add Parser.queryToLabelMap via visitor

alextheimer:
This is an alternative to #1312.

It is safer than relying on evaluation of a lazy value for validation, it allows name validation to be manually triggered, and it adds a layer of separation between the `Parser` and FiloDB-specific query constraints.

See comments for additional justification.

**Pull Request checklist**

**New behavior :**

Adds the method Parser.queryToLabelMap(query: String): Map[String, String] which parses an instant selector into a map of (label, value) pairs (including __name__).

Additionally adds `Expression.requireMetricNames` to give better control over when the FiloDB-specific constraint is enforced.


prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
```diff
@@ -139,10 +139,8 @@ sealed trait Vector extends Expression {
...
 
   // Convert metricName{labels} -> {labels, __name__="metricName"} so it's uniform
-  lazy val mergeNameToLabels: Seq[LabelMatch] = {
+  val mergeNameToLabels: Seq[LabelMatch] = {
     val nameLabel = labelSelection.find(_.label == PromMetricLabel)
-    if (nameLabel.isEmpty && metricName.isEmpty)
-      throw new IllegalArgumentException("Metric name is not present")
     if (metricName.nonEmpty) {
       if (nameLabel.nonEmpty) throw new IllegalArgumentException("Metric name should not be set twice")
       // metric name specified but no __name__ label.  Add it
```
alextheimer, line 146: Now, an exception is always thrown during construction if the metric name is set twice.

prometheus/src/main/scala/filodb/prometheus/parse/Parser.scala
```diff
@@ -31,8 +31,7 @@ object Parser extends StrictLogging {
     }
   }
 
-  // Only called directly by tests.
-  def parseQuery(query: String): Expression = {
+  private def parseQueryWithoutMetricNameRequirement(query: String): Expression = {
```
alextheimer, line 34: The validations "setup" by each pre-PR `Parser` method (i.e. that exactly one metric name exists for any selector) remain the same as before. The only difference is that the validations are guaranteed to run before the `Parser` returns a result.

Only `parseQueryWithoutMetricNameRequirement`-- a private method-- can return a parsed result where a metric name has not been enforced. Its only current use is within `queryToLabelMap` below.

prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
alextheimer: Also, the FiloDB constraint-- that a metric name is present-- is not required at construction.

prometheus/src/test/scala/filodb/prometheus/parse/ParserSpec.scala
```diff
@@ -745,14 +768,14 @@ class ParserSpec extends AnyFunSpec with Matchers {
   }
 
   private def parseSuccessfully(query: String) = {
-    val result = LegacyParser.parseQuery(query)
+    val result = LegacyParser.parseQuery(query).requireMetricNames()
```
alextheimer, line 771: You'd need to be careful to call `requireMetricNames` only when `LegacyParser` and `AntlrParser` methods are directly accessed. Otherwise, all public methods of `Parser` (except `parseQueryWithoutMetricNameRequirement`) will call it automatically.

alextheimer: The general idea here is that `Parser` enforces `filodb` constraints when it wants to, but the other two parsers don't.

alextheimer:
closed since the alternative #1312 was merged

PR #1319, opened 2022-01-21
Merging changes from develop to 0.9.16-integration

alextheimer:


alextheimer:
Is that merge commit supposed to be there?

alextheimer:
Re-doing PR to double-check that the commits are correct

PR #1323, opened 2022-01-21
misc(query): optimize single label value metadata lookup query execution

sherali42:
**Pull Request checklist**

Existing behavior is to construct a map of key/values even if the query is only for single column/label. With this change, only a set of de-duplicated values will be returned.

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
+++ b/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
@@ -66,7 +66,8 @@ class TimeSeriesShardStats(dataset: DatasetRef, shardNum: Int) {
   val numDirtyPartKeysFlushed = Kamon.counter("memstore-index-num-dirty-keys-flushed").withTags(TagSet.from(tags))
   val indexRecoveryNumRecordsProcessed = Kamon.counter("memstore-index-recovery-partkeys-processed").
     withTags(TagSet.from(tags))
-  val indexPartkeyLookups = Kamon.counter("memstore-index-partkey-lookup").withTags(TagSet.from(tags))
+  val indexPartkeyLookups = Kamon.counter("memstore-index-partkey-lookups").withTags(TagSet.from(tags))
+  val partkeyLabelScans = Kamon.counter("memstore-labels-partkeys-scanned").withTags(TagSet.from(tags))
   val downsampleRecordsCreated = Kamon.counter("memstore-downsample-records-created").withTags(TagSet.from(tags))
 
   /**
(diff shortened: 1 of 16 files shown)


vishramachandran (approved):
@sherali42 approved. We need to add query metrics to stats. I'll let you decide whether it should go in this PR or another one

query/src/main/scala/filodb/query/PromCirceSupport.scala
```diff
@@ -116,8 +116,9 @@ object PromCirceSupport {
         timeSeriesScanned <- c.downField("timeSeriesScanned").as[Long]
         dataBytesScanned     <- c.downField("dataBytesScanned").as[Long]
         resultBytes     <- c.downField("resultBytes").as[Long]
+        partkeysVisited <- c.downField("partkeysVisited").as[Long]
       } yield {
-        QueryStatistics(group, timeSeriesScanned, dataBytesScanned, resultBytes)
+        QueryStatistics(group, timeSeriesScanned, dataBytesScanned, resultBytes, partkeysVisited)
```
vishramachandran, line 121: We already have `timeSeriesScanned` which can be used for this. We don't need a new stat

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -692,19 +693,45 @@ class TimeSeriesShard(val ref: DatasetRef,
...
-  def labelValuesWithFilters(filter: Seq[ColumnFilter],
+  def labelValuesWithFilters(filters: Seq[ColumnFilter],
                              labelNames: Seq[String],
                              endTime: Long,
                              startTime: Long,
+                             querySession: QuerySession,
                              limit: Int): Iterator[Map[ZeroCopyUTF8String, ZeroCopyUTF8String]] = {
-    LabelValueResultIterator(partKeyIndex.partIdsFromFilters(filter, startTime, endTime), labelNames, limit)
+    val metricShardKeys = schemas.part.options.shardKeyColumns
+    val metricGroupBy = deploymentPartitionName +: clusterType +: shardKeyValuesFromFilter(metricShardKeys, filters)
```
vishramachandran, line 710: Lets discuss this in person in more detail.
I have questions around when the namespace for accounting will be identified correctly, and when it will not

query/src/main/scala/filodb/query/PromCirceSupport.scala
sherali42: ack, reverted this change

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
sherali42: made `multiple` as default for missed shardKeyValue

PR #1328, opened 2022-02-01
core(misc): implement target-schema notion to identify the source of the data

sherali42:
**Pull Request checklist**

### Target Schema
Implement target-schema to identify the resource/service/node from which data is published. Target schema is specified as set of labels that can identify the source. The combined values of these labels lets us place data in the target shard, collocating the relevant data together. For e.g  if a resource is publishing metadata metric in addition to the telemetry data, most of the queries end up joining the time-series data with metadata. If the data is sharded by resource_id instead of using all the labels, `join` queries will be answered efficiently.
#### Changes in this PR
- shardKeyHash 
   - if target-schema is defined and it doesn't contain metric then metric will not be used to calculate the hash
   - if target-schema is not defined, shardKey hash will be generated by combining hashes of all the shard keys including metric (current behavior)
- partitionHash
   - if target-schema is defined, then partitionHash is generated by combining only targetSchema label values
   - if target-schema is not defined, then all the nonShardKeyLabels are used to calculate partitionHash (current behavior)
- Above methods are implemented so that metrics-gateway and query-service implementations can utilize these for identifying the shards for ingestion/querying
- TargetSchemaProvider is implemented to provide `TargetSchemaChanges` during the query-window
   - If the TargetSchema change happens with in the query-range, then route the query to all (spread) shards and stitch the results
   - if the TargetSchema change happened beyond query-range, pick the target-schema that is defined before `startTime` of the query-window and use it to find the target shard to route the query.
   - if TargetSchema is not defined before the `startTime`, then route thee query to all (spread) shards.
- Refactoring due to API changes
- Test cases


core/src/main/scala/filodb.core/TargetSchemaProvider.scala
```diff
@@ -0,0 +1,11 @@
+package filodb.core
+
+import filodb.core.query.ColumnFilter
+
+trait TargetSchemaProvider {
+  def targetSchemaFunc(filter: Seq[ColumnFilter]): Seq[String]
```
vishramachandran, line 6: Target schema can change over time. The response should have a time dimension

coordinator/src/main/scala/filodb.coordinator/client/QueryCommands.scala
```diff
@@ -59,6 +59,11 @@ object QueryCommands {
     }
   }
 
+  final case class FunctionalTargetSchemaProvider(f: Seq[ColumnFilter] => Seq[TargetSchema] = { _ => Seq.empty})
+    extends TargetSchemaProvider {
+    def targetSchemaFunc(filter: Seq[ColumnFilter]): Seq[TargetSchema] = f(filter)
```
sherali42, line 64: rename TargetSchema to TargetSchemaChange, review comment.

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
```diff
@@ -185,11 +187,55 @@ class SingleClusterPlanner(val dataset: Dataset,
...
+      val targetSchema = if (targetSchemaChange.nonEmpty) targetSchemaChange.last.schema else Seq.empty
+      val shardHash = RecordBuilder.shardKeyHash(shardValues, dsOptions.metricColumn, metric, targetSchema)
+      if(useTargetSchemaForShards) {
+        val nonShardKeyLabelPairs = filters.filter(f => !shardColumns.contains(f.column)
+                                                          && f.filter.isInstanceOf[Filter.Equals])
+                                            .map(cf => cf.column ->
+                                              cf.filter.asInstanceOf[Filter.Equals].value.toString).toMap
+        val partitionHash = RecordBuilder.partitionKeyHash(nonShardKeyLabelPairs, shardVals.toMap, targetSchema,
+          dsOptions.metricColumn, metric)
+        Seq(shardMapperFunc.ingestionShard(shardHash, partitionHash, spreadProvToUse.spreadFunc(filters).last.spread))
```
sherali42, line 202: add comment explaining - we go to single shard since all target-schema labels are provided in query filters

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
```diff
@@ -433,12 +483,33 @@ class SingleClusterPlanner(val dataset: Dataset,
...
+
+    val execPlans = shardsFromFilters(renamedFilters, qContext, useTSForQueryShards).map { shard =>
       val dispatcher = dispatcherForShard(shard)
       MultiSchemaPartitionsExec(qContext, dispatcher, dsRef, shard, renamedFilters,
         toChunkScanMethod(rangeSelectorWithOffset), dsOptions.metricColumn, schemaOpt, colName)
     }
-    PlanResult(execPlans, needsStitch)
+    // Stitch only if spread and/or target-schema (all labels provided in the query), changes during the query-window
+    PlanResult(execPlans, needsStitch ||
+      (tsChangeExists && allTSLabelsPresent))
```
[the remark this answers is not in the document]
sherali42: fixed the condition to apply stitch-plan when the target-schema changes during the query-window. as the data might be routed to different shards after the target-schema change.

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
sherali42: added documentation

coordinator/src/main/scala/filodb.coordinator/client/QueryCommands.scala
sherali42: renamed it to `TargetSchemaChange`

core/src/main/scala/filodb.core/TargetSchemaProvider.scala
sherali42: updated the code to consider target-schema changes over time.

vishramachandran (approved):

PR #1330, opened 2022-02-02
fix(prometheus) Metric name should not be set twice

amolnayak311:
**Pull Request checklist**

- [x ] Tests for the changes have been added (for bug fixes / features) ?

**Current behavior :**  Following prom expressions are valid

```foo{_metric_="foo"}``` but ```foo{__name__="foo"}``` will fail to parse

**New behavior :**
both ```foo{_metric_="foo"}```  and ``` foo{__name__="foo"}``` will fail to parse

### DIFF
--- a/prometheus/src/test/scala/filodb/prometheus/parse/ParserSpec.scala
+++ b/prometheus/src/test/scala/filodb/prometheus/parse/ParserSpec.scala
@@ -185,6 +185,9 @@ class ParserSpec extends AnyFunSpec with Matchers {
     parseError("foo{1}")
     parseError("{}")
     parseError("foo{__name__=\"bar\"}")
+    parseError("foo{__name__=\"foo\"}")
+    parseError("foo{_metric_=\"bar\"}")
+    parseError("foo{_metric_=\"foo\"}")
 
     parseSuccessfully("test{a=\"b\"}[5y] OFFSET 3d")
     parseSuccessfully("test{a=\"b\"}[5y] LIMIT 3")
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: prometheus/src/test/scala/filodb/prometheus/parse/ParserSpec.scala [hunk 1]
```diff
@@ -185,6 +185,9 @@ class ParserSpec extends AnyFunSpec with Matchers {
     parseError("foo{1}")
     parseError("{}")
     parseError("foo{__name__=\"bar\"}")
+    parseError("foo{__name__=\"foo\"}")
+    parseError("foo{_metric_=\"bar\"}")
+    parseError("foo{_metric_=\"foo\"}")
```
alextheimer, line 190: Does something like `{_metric_="foo"}` still parse successfully? I'm not 100% sure if `metricName` accounts for both `foo{~}` and `{_metric_="foo"}`

alextheimer (dismissed):
lgtm if the test in the comment passes

prometheus/src/test/scala/filodb/prometheus/parse/ParserSpec.scala
amolnayak311: I'll add tests for those too but yes it parses. I'm not sure is ```{_metric_="foo"}``` is equivalent to ```foo{~}``` as ```_metric_``` is more internal to FiloDB

amolnayak311: Pushed more tests

alextheimer: Ack, lgtm. Good point about `_metric_`-- I'm also not sure.

vishramachandran: `{__name__="foo", _metric_="foo"}` should fail.

PR #1331, opened 2022-02-02
***WIP*** fix(gateway): limit InputRecord label value size

alextheimer:
**Pull Request checklist**

**Current behavior :**
Label values too large will cause Lucene to throw exceptions.

**New behavior :**
Label values are capped at a specific size. When the gateway receives a record that exceeds this limit:
1. A warning is logged.
2. A counter is incremented for the ws/ns pair.
3. The record is dropped.

### DIFF
--- a/gateway/src/main/scala/filodb/gateway/GatewayServer.scala
+++ b/gateway/src/main/scala/filodb/gateway/GatewayServer.scala
@@ -59,9 +59,6 @@ import filodb.timeseries.TestTimeseriesProducer
 object GatewayServer extends StrictLogging {
   Kamon.init
 
-  val MAX_RECORD_TAG_VALUE_SIZE_BYTES = 16384  // 2^14; 2^15 is lucene limit
-  val RECORD_TAG_VALUE_BYTES_PER_CHAR = 2;
-
   // Get global configuration using universal FiloDB/Akka-based config
   val settings = new FilodbSettings()
   val config = settings.allConfig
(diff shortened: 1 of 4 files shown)


gateway/src/main/scala/filodb/gateway/GatewayServer.scala
```diff
@@ -58,6 +59,9 @@ import filodb.timeseries.TestTimeseriesProducer
 object GatewayServer extends StrictLogging {
   Kamon.init
 
+  val MAX_RECORD_TAG_VALUE_SIZE_BYTES = 16384  // 2^14; 2^15 is lucene limit
+  val RECORD_TAG_VALUE_BYTES_PER_CHAR = 2;
```
alextheimer, line 63: Any better way to do this? Doesn't look like a `Charset.bytesPer` exists.
Also: this isn't necessarily "2"-- I still need to look through the agent code to find the encoding.

gateway/src/main/scala/filodb/gateway/GatewayServer.scala
```diff
@@ -234,14 +239,37 @@ object GatewayServer extends StrictLogging {
...
     // While there are still messages in the queue and there aren't containers to send, pull and build
     while (!queue.isEmpty && builder.allContainers.length <= 1) {
-      queue.poll().addToBuilder(builder)
-      // TODO: add metrics
+      val rec = queue.poll();
+      if (!validateRecord(rec)) {
+        logger.warn("record validation failed-- dropping invalid record")
+        numInvalidRecords.withTags(TagSet.from(Map(
+          "_ws_" -> rec.tags("_ws_"),
+          "_ns_" -> rec.tags("_ns_")
```
alextheimer, line 267: Suggested in the radar, but this again hard-codes "ws/ns" mindset.

alextheimer:
Closing for now-- this is less trivial than originally planned, and is probably not worth the effort.

PR #1333, opened 2022-02-04
fix(query): prevent serialization of empty RangeVectors

alextheimer:
**Pull Request checklist**

Adds a single filter that removes empty `SerializedRangeVectors` before they're sent over-the-wire.

### DIFF
--- a/query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
+++ b/query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
@@ -46,7 +46,7 @@ trait MetadataDistConcatExec extends NonLeafExecPlan {
       case (QueryError(_, _, ex), _)         => throw ex
     }.toListL.map { resp =>
       val metadataResult = scala.collection.mutable.Set.empty[Map[ZeroCopyUTF8String, ZeroCopyUTF8String]]
-      resp.foreach { rv =>
+      resp.filter(!_.isEmpty).foreach { rv =>
         metadataResult ++= rv.head.rows.map { rowReader =>
           val binaryRowReader = rowReader.asInstanceOf[BinaryRecordRowReader]
           rv.head match {
--- a/query/src/main/scala/filodb/query/exec/ExecPlan.scala
+++ b/query/src/main/scala/filodb/query/exec/ExecPlan.scala
@@ -213,6 +213,7 @@ trait ExecPlan extends QueryCommand {
                   sampleLimit} samples. Try applying more filters or reduce time range.")
               srv
           }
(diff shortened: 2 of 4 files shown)


core/src/main/scala/filodb.core/query/RangeVector.scala
```diff
@@ -137,6 +137,15 @@ trait RangeVector {
...
   def prettyPrint(formatTime: Boolean = true): String = "RV String Not supported"
+
+  /**
+   * To prevent empty SerializedRangeVectors from being sent over-the-wire, this boolean
+   * can be used to filter out confirmed-empty RangeVectors.
+   *
+   * For the sake of performance, it is not the responsibility of RangeVector/RangeVectorCursor
+   * to set/unset this value; it should be maintained only where an empty RangeVector might arise.
+   */
+  var confirmedToContainAllEmptyRows = false
```
vishramachandran, line 148: No dont add var in trait. Add it to SerializedRangeVector only

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
```diff
@@ -46,7 +46,7 @@ trait MetadataDistConcatExec extends NonLeafExecPlan {
       case (QueryError(_, _, ex), _)         => throw ex
     }.toListL.map { resp =>
       val metadataResult = scala.collection.mutable.Set.empty[Map[ZeroCopyUTF8String, ZeroCopyUTF8String]]
-      resp.foreach { rv =>
+      resp.filter(!_.isEmpty).foreach { rv =>
```
vishramachandran, line 49: Not needed here. Everything eventually flows through ExecPlan.scala, so it will eventually be filtered.

vishramachandran (approved):
Approved with comments.

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
alextheimer: Some metadata tests fail if I undo this-- `rv` is a (possibly empty) sequence of `RangeVector`s.

vishramachandran: @sherali42  Can you help confirm if removing RV when no rows will not cause regression for metadata queries ?

sherali42 (approved):
👍 thanks for the quick PR @alextheimer

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
sherali42: nope, we should be good

PR #1335, opened 2022-02-08
perf(query) Changes for LongTimeRangePlanner binary join push down

amolnayak311:
**Pull Request checklist**

**Current behavior :**
Binary joins are not pushed down whenever possible in ```LongTimeRangePlanner ```. This can cause raw data to be exported from the nodes leading to inefficient joins

**New behavior :**
 Joins are pushed down wherever possible.


coordinator/src/main/scala/filodb.coordinator/queryplanner/LongTimeRangePlanner.scala
```diff
@@ -180,31 +180,58 @@ import filodb.query.exec._
...
+      case lp: TopLevelSubquery            => super.materializeTopLevelSubquery(qContext, lp)
       case lp: ApplyLimitFunction          => rawClusterMaterialize(qContext, lp)
       case lp: LabelNames                  => rawClusterMaterialize(qContext, lp)
       case lp: LabelCardinality            => materializeLabelCardinalityPlan(lp, qContext)
     }
     // scalastyle:on cyclomatic.complexity
   }
 
+  override def materializeBinaryJoin(qContext: QueryContext, logicalPlan: BinaryJoin): PlanResult =
+    if (LogicalPlanUtils.getOffsetMillis(logicalPlan).foldLeft(0L)(_.max(_)) == 0) {
```
kvpetrov, line 209: not sure I understood, why do we care if there is an offset? ie offset means that the actual start of the expression is shifted in relation with the start/end of query context. In other words, in that case we need to calculate the actual range involved in the query and deal accordingly?

coordinator/src/main/scala/filodb.coordinator/queryplanner/LongTimeRangePlanner.scala
```diff
@@ -180,31 +180,58 @@ import filodb.query.exec._
...
+      // No offset, can push down binary join to LT and Raw planner and stitch
+      // Binary joins can be pushed down to respective planners, exporting results to QS is not required
+      val (startTime, endTime) = (logicalPlan.startMs, logicalPlan.endMs)
+      val execPlan = if (startTime > latestDownsampleTimestampFn) {
+        rawClusterPlanner.materialize(logicalPlan, qContext)
+      } else if (endTime < earliestRawTimestampFn) {
+        downsampleClusterPlanner.materialize(logicalPlan, qContext)
+      } else {
+        val dsLogicalPlan = copyLogicalPlanWithUpdatedTimeRange(logicalPlan,
```
kvpetrov, line 218: this is the trickiest part, in general there can be anything under the binary join, for example subquery with overlapping window. Splitting subquery into two expressions one stitching them potentially can bring back invalid results.  This is why I was excited originally when we stopped splitting/stitching high level expressions such as binary join and started doing them on the leave raw series. Stitching raw series from downsample and raw clusters can always be done perfectly and every single expression on top of it would be calculated properly. In that case push down of the aggregation is not possible but the logic is simple and correct. That brings me to the point that probably you meant lookback, not the offset in the above "if" statement. If you read documentation of the LogicalPlanUtils (getLookBackMillis), then this might be what you want (instead  of getOffsetMillis). So, if we don't have aggregation intervals (lookbacks) then it's safe to split the plan. The issue of stitching across the time is the major headache for the subqueries in general. We could do the splits fairly efficiently because there should be an overlap between the downsample and raw cluster and IF this overlap is bigger than the lookback we can still calculate perfectly AND push down aggregation.

kvpetrov (commented):
I think there is an error in the logic, put a couple comments.

coordinator/src/main/scala/filodb.coordinator/queryplanner/LongTimeRangePlanner.scala
amolnayak311: Please refer to comment on line 229  on why we need to check offsets.

coordinator/src/main/scala/filodb.coordinator/queryplanner/LongTimeRangePlanner.scala
amolnayak311: Can you please provide a concrete PromQL expression for your scenario. That would be helpful for more discussion.

kvpetrov: avg_over_time(rate(foo[1d])[10d:1h])+ avg_over_time(rate(bar[1d])[10d:1h])
here we have a sliding window of 1d over 10 days every hour. Lookback should be 1 day here. 
It would be interesting to see what logic above would do to this query? What would be the start/end here? I assume it will probably send it all the the raw cluster if "now" is in the raw cluster range.

vishramachandran (approved):
Looks great! Thanks for digging into the details. Small comments.

PR #1336, opened 2022-02-10
feat(core): Lucene faceting for 300x more efficient label-value queries

vishramachandran:
**Pull Request checklist**

**Current behavior :**

For metadata searches that need to enumerate values of fields, we end up iterating through matching documents and build a hash set of field values. This is inefficient. It eats up CPU creating bottlenecks in query execution, especially in high cardinality clusters and reduces overall query throughput.

**New behavior :**

This PR uses Lucene faceting which tremendously improves performance (by 300x) of label-value queries which is a major CPU consumer in the system today.

* Faceting is enabled in raw cluster and disabled in downsample cluster
* Faceting can be disabled in raw cluster via configuration
* If faceting is enabled, label-value queries automatically choose the efficient execution path. Otherwise, they fall back to existing approach.
* Faceting requires 10% more disk space. So raw clusters will need more disk space.
* Faceting increases indexing time by 2x, so node bootstrap will take more time.

### With Synthetic Data 
Benchmarks performed with `PartKeyIndexBenchmark`
Num part keys: 1 mil
Part Keys generated using TestTimeSeriesProducer
All time series have same metric
Query to collect metric names has a ws/ns filter.

#### Prior to this PR, without faceting 
Index Size: 227MB
Time to index: 17s
All time series have about 25 unique metric names
Throughput of getting list of metric names (iterative): 0.281 ops/s

#### With this PR, with faceting all labels
Index Size: 242MB
Time to index: 35s
Throughput of getting list of metric names (facet collector): 113.276 ops/s

### With Production Data 
Data cannot be uploaded to git since it is not sanitized.
Benchmarks performed with `PartKeyIndexBenchmark`
Num part keys: 574k
Query to collect metric names has a ws/ns filter.

#### Prior to this PR, without faceting 
Index Size: 310MB
Time to index: 17s
Throughput of getting list of metric names (iterative): 0.065 ops/s

#### With this PR, with faceting all labels
Index Size: 337MB
Time to index: 39s
Throughput of getting list of metric names (facet collector): 26.4 ops/s

### Summary: 
Index size increased by 6% to 8%
Time to index increased by around 2x
Throughput of label values query increased by 300 to 400 times

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
+++ b/core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
@@ -17,6 +17,9 @@ import kamon.metric.MeasurementUnit
 import org.apache.lucene.analysis.standard.StandardAnalyzer
 import org.apache.lucene.document._
 import org.apache.lucene.document.Field.Store
+import org.apache.lucene.facet.{FacetsCollector, FacetsConfig}
+import org.apache.lucene.facet.sortedset.{DefaultSortedSetDocValuesReaderState,
+                     SortedSetDocValuesFacetCounts, SortedSetDocValuesFacetField}
 import org.apache.lucene.index._
 import org.apache.lucene.search._
 import org.apache.lucene.search.BooleanClause.Occur
@@ -123,6 +126,11 @@ class PartKeyLuceneIndex(ref: DatasetRef,
   //start this thread to flush the segments and refresh the searcher every specific time period
   private var flushThread: ControlledRealTimeReopenThread[IndexSearcher] = _
 
+  val facetsConfig = new FacetsConfig
+  schema.options.shardKeyColumns.foreach { c =>
+    facetsConfig.setIndexFieldName(c, s"facet_$c")
+  }
+
   case class ReusableLuceneDocument() {
     val document = new Document()
     private val partIdField = new StringField(PART_ID, "0", Store.NO)
@@ -135,6 +143,9 @@ class PartKeyLuceneIndex(ref: DatasetRef,
     private val otherFields = mutable.WeakHashMap[String, StringField]() // weak so that it is GCed when unused
 
     def addField(name: String, value: String): Unit = {
+      if (schema.options.shardKeyColumns.contains(name)) {
+        document.add(new SortedSetDocValuesFacetField(name, value))
+      }
       val field = otherFields.getOrElseUpdate(name, new StringField(name, "", Store.NO))
       field.setStringValue(value)
       document.add(field)
@@ -272,6 +283,31 @@ class PartKeyLuceneIndex(ref: DatasetRef,
     indexWriter.close()
   }
 
+  private val readerStateCache = mutable.WeakHashMap[(IndexReader, String), DefaultSortedSetDocValuesReaderState]()
+  def shardKeyColValues(colFilters: Seq[ColumnFilter], startTime: Long, endTime: Long,
+                        colName: String, limit: Int = 100): Seq[String] = {
+    require(schema.options.shardKeyColumns.contains(colName), s"The provided column $colName is not a shard key column")
+    val labelValues = mutable.ArrayBuffer[String]()
+    withNewSearcher { searcher =>
+      val reader = searcher.getIndexReader()
+      // the state object is expensive to create, so reuse across queries. It is associated with reader
+      val state = readerStateCache.getOrElseUpdate((reader, colName),
+                    new DefaultSortedSetDocValuesReaderState(searcher.getIndexReader, s"facet_$colName"))
+      val fc = new FacetsCollector
+      val query = colFiltersToQuery(colFilters, startTime, endTime)
+      FacetsCollector.search(searcher, query, 10, fc)
+      val facets = new SortedSetDocValuesFacetCounts(state, fc)
+      val result = facets.getTopChildren(limit, colName)
+      if (result != null) {
+        for (i <- 0 until result.childCount) {
+          val lv = result.labelValues(i)
+          labelValues += lv.label
+        }
+      }
+    }
+    labelValues
+  }
+
   /**
     * Fetch values/terms for a specific column/key/field, in order from most frequent on down.
    * Note that it iterates through all docs up to a certain limit only, so if there are too many terms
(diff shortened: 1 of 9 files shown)


core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -71,6 +77,7 @@ final case class PartKeyLuceneIndexRecord(partKey: Array[Byte], startTime: Long,
 
 class PartKeyLuceneIndex(ref: DatasetRef,
                          schema: PartitionSchema,
+                         facetEnabled: Boolean,
```
amolnayak311, line 80: Nit pick, Personally I prefer adding new fields at the end with default values, ``false`` in this case.

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -135,13 +145,20 @@ class PartKeyLuceneIndex(ref: DatasetRef,
     private val otherFields = mutable.WeakHashMap[String, StringField]() // weak so that it is GCed when unused
 
     def addField(name: String, value: String): Unit = {
+      // Use PartKeyIndexBenchmark to measure indexing performance before changing this
+      if (facetEnabled && name.nonEmpty && value.nonEmpty) {
+        facetsConfig.setRequireDimensionDrillDown(name, false)
+        facetsConfig.setIndexFieldName(name, FACET_FIELD_PREFIX + name)
+        document.add(new SortedSetDocValuesFacetField(name, value))
```
amolnayak311, line 152: From ``SortedSetDocValuesFacetField`` the documentation, Faceting is 25% slower and more importantly there is an additional cost for ``IndexReader`` open at the cost of lower RAM usage. Something to keep in mind.

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -356,7 +407,9 @@ class PartKeyLuceneIndex(ref: DatasetRef,
     val document = makeDocument(partKeyOnHeapBytes, partKeyBytesRefOffset, partKeyNumBytes, partId, startTime, endTime)
     logger.debug(s"Upserting document ${partKeyString(partId, partKeyOnHeapBytes, partKeyBytesRefOffset)} " +
                  s"with startTime=$startTime endTime=$endTime into dataset=$ref shard=$shardNum")
-    indexWriter.updateDocument(new Term(PART_ID, partId.toString), document)
+    val doc = luceneDocument.get()
```
amolnayak311, line 410: This is same as ``document`` we have in previous line.

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -464,11 +517,14 @@ class PartKeyLuceneIndex(ref: DatasetRef,
       logger.warn(s"Could not find in Lucene startTime for partId=$partId in dataset=$ref. Using " +
         s"$startTime instead.", new IllegalStateException()) // assume this time series started retention period ago
     }
-    val updatedDoc = makeDocument(partKeyOnHeapBytes, partKeyBytesRefOffset, partKeyNumBytes,
+    makeDocument(partKeyOnHeapBytes, partKeyBytesRefOffset, partKeyNumBytes,
       partId, startTime, endTime)
+
+    val doc = luceneDocument.get()
```
amolnayak311, line 523: Any reason why the return value of ``makeDocument`` is ignored and retrieved from ``ThreadLocal`` again?

amolnayak311 (approved):

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -272,6 +289,38 @@ class PartKeyLuceneIndex(ref: DatasetRef,
...
+      try {
+        val state = readerStateCache.getOrElseUpdate((reader, colName),
+            new DefaultSortedSetDocValuesReaderState(reader, FACET_FIELD_PREFIX + colName))
+        val fc = new FacetsCollector
+        val query = colFiltersToQuery(colFilters, startTime, endTime)
+        FacetsCollector.search(searcher, query, 10, fc)
+        val facets = new SortedSetDocValuesFacetCounts(state, fc)
+        val result = facets.getTopChildren(limit, colName)
+        if (result != null) {
+          for (i <- 0 until Math.min(result.childCount, limit)) {
```
sherali42, line 310: doesn't topk make sure result count has an upper bound of `limit`?

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
sherali42: yeah its not consistent too, at some places return value is still used.

sherali42 (approved):
👍 thanks @vishramachandran for taking this up.

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
vishramachandran: I did not use default param values since I really wanted callers to think about faceting for performance.

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
vishramachandran: I have now removed assignment of makeDocument return value, which returns the actual lucene document. The problem is we need the FacetsConfig object too.

The thread local (`ReusableDocument` object) stores both the lucene `Document` and the `FacetsConfig` object which are both needed to build the final document that will be added to index.

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
vishramachandran: I have changed this to `result.labelValues.length`

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
vishramachandran: Pls see other comment

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
vishramachandran: As we discovered, that javadoc is from older lucene version. Latest 8.x does not have that in the javadoc. Things have probably changed.

PR #1337, opened 2022-02-10
fix(misc): fix formatCode for non increasing histogram buckets

sherali42:
**Pull Request checklist**

**Issue** 
Downsample job while applying sum based aggregate function on Histogram chunks (of type non-increasing bucket values), `formatCode` is being set to `HistFormat_Geometric_XOR`. To be precise in MutableHistogram.serialize. Later during ingestion these histogram chunks are marked as `InvalidHistogram`, because HistFormat_Geometric_XOR is not in valid format list (`isValidFormatCode`). and these chunks are ignored and never persisted. causing failures during query time.
**Fix**
for non-increasing geometric buckets, format code is set to `HistFormat_Geometric1_Delta` in BinaryHistogram.writeDoubles method.


vishramachandran (approved):
Thanks for digging in.

1. PRB is still failing.
2. Please confirm that this does not affect regular ingestion or query path

sherali42:
> Thanks for digging in.
> 
> 1. PRB is still failing.
> 2. Please confirm that this does not affect regular ingestion or query path

fixed PRB. Fix is applicable to only `minus-1` geometric buckets. It should not affect regular ingestion/query paths.

PR #1362, opened 2022-04-06
fix(query): handle labelvalue query serialization for backward compatibility

sherali42:
**Pull Request checklist**

Single label value query backward compatibility
- older version returns Map type where as newer version works with List type, this change handles Map type results returned by older version and converts it into StringList.


tjackpaul (approved):
lgtm, thanks for the quick PR

PR #1370, opened 2022-04-13
feat(coord) Stateful shard assignment

amolnayak311:
**Pull Request checklist**

**Current behavior :**

Shard assignment is random, when deployed on Kubernetes  Deployment, when a pod(s) go down, the reassignment to shards to pods is random. This prevents using any state.

**New behavior :**

To use Kubernetes StatefulSets where we need state to be retained across pod restarts, we need a deterministic way to assign shards to pods. In stateful sets, the pod name has an ordinal at the end which can be used to assign shards to pods.

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/FilodbCluster.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/FilodbCluster.scala
@@ -77,7 +77,8 @@ final class FilodbCluster(val system: ExtendedActorSystem, overrideConfig: Confi
     * All actions are idempotent. It manages the underlying lifecycle of all node actors.
     */
   private[coordinator] lazy val guardian = system.actorOf(NodeGuardian.props(
-    settings, metaStore, memStore, new K8sStatefulSetShardAssignmentStrategy(true)), guardianName)
+    settings, metaStore, memStore, new K8sStatefulSetShardAssignmentStrategy(useHostNameToResolveShards =
+        settings.config.getBoolean("shard-manager.enable-k8s-stateful-shard-strategy"))), guardianName)
   // TODO: Make parameter driven
 
   def isInitialized: Boolean = _isInitialized.get
--- a/coordinator/src/main/scala/filodb.coordinator/FilodbCluster.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/FilodbCluster.scala
@@ -79,7 +79,7 @@ final class FilodbCluster(val system: ExtendedActorSystem, overrideConfig: Confi
   private[coordinator] lazy val guardian = system.actorOf(NodeGuardian.props(
     settings, metaStore, memStore,
     if (settings.config.getBoolean("shard-manager.enable-k8s-stateful-shard-strategy"))
-      K8sStatefulSetShardAssignmentStrategy else DefaultShardAssignmentStrategy)
+      new K8sStatefulSetShardAssignmentStrategy else DefaultShardAssignmentStrategy)
     , guardianName)
 
   def isInitialized: Boolean = _isInitialized.get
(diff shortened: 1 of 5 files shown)


coordinator/src/main/scala/filodb.coordinator/FilodbCluster.scala
```diff
@@ -77,7 +77,8 @@ final class FilodbCluster(val system: ExtendedActorSystem, overrideConfig: Confi
     * All actions are idempotent. It manages the underlying lifecycle of all node actors.
     */
   private[coordinator] lazy val guardian = system.actorOf(NodeGuardian.props(
-    settings, metaStore, memStore, DefaultShardAssignmentStrategy), guardianName)
+    settings, metaStore, memStore, new K8sStatefulSetShardAssignmentStrategy(useHostNameToResolveShards =
+        settings.config.getBoolean("shard-manager.enable-k8s-stateful-shard-strategy"))), guardianName)
```
vishramachandran, line 81: Can we pass in  `K8sStatefulSetShardAssignmentStrategy` or `DefaultShardAssignmentStrategy` based on the config?

amolnayak311: Yes that should be possible but the strategy already uses the ``DefaultShardAssignmentStrategy`` as fallback? Do you still want to based on flag pass one over other?

amolnayak311: Changes made

coordinator/src/main/scala/filodb.coordinator/ShardAssignmentStrategy.scala
```diff
@@ -31,8 +33,67 @@ trait ShardAssignmentStrategy {
...
+      .flatMap(hostName => pat.findFirstIn(hostName).map(ordinal => (hostName, -Integer.parseInt(ordinal))))
   }
 
+  override def shardAssignments(coord: ActorRef,
+                                dataset: DatasetRef,
+                                resources: DatasetResourceSpec,
+                                mapper: ShardMapper): Seq[Int] =
+      getOrdinalFromActorRef(coord) match {
+        case Some((hostName, ordinal)) =>
+          val numShardsPerHost = resources.numShards / resources.minNumNodes
```
vishramachandran, line 66: This needs tweaking. If there are 10 shards, and minNodes = 3, one shard will never be assigned.

Ceil division is an alternative, but it would result in 4,4,2 as opposed to the ideal 4,3,3.

vishramachandran (changes requested):
There is an important fix needed. Please see comments.

Could you add some tests to the `ShardAssignmentStrategySpec` class ?

coordinator/src/main/scala/filodb.coordinator/ShardAssignmentStrategy.scala
amolnayak311: I have called this out in the comments of the class

```
  The implementation assumes the number of shards is a multiple of number of nodes and there are no
  uneven assignments of shards to nodes
  ```

IMO its a misconfiguration to have unequal shards on each pod making imbalanced use of attached persistent volume if any in any Kubernetes stateful set. I will not complicate the setup rather have a warning and fallback to default strategy in case ```numShards % minNodes```  is not equal to 0

amolnayak311:
> There is an important fix needed. Please see comments.
> 
> Could you add some tests to the `ShardAssignmentStrategySpec` class ?

I thought of adding tests but not sure how to mock the hostname resolution

vishramachandran (approved):
Thanks for the iterations and spec. Looks good 👍

PR #1344, opened 2022-03-04
**WIP**: perf(query): BinaryJoin pushdown optimization

alextheimer:
**Pull Request checklist**

- [x] The commit(s) message(s) follows the contribution [guidelines](CONTRIBUTING.md) ?
- [x] Tests for the changes have been added (for bug fixes / features) ?
- [ ] Docs have been added / updated (for bug fixes / features) ?

------------------------

When both children of a BinaryJoin are a PeriodicSeries, the result ExecPlan would typically be of the form:

```
E~BinaryJoinExec(binaryOp=ADD) on ActorPlanDispatcher(actor=0)
-T~PeriodicSamplesMapper()
--E~MultiSchemaPartitionsExec(shard=0) on ActorPlanDispatcher(actor=0)  // lhs
-T~PeriodicSamplesMapper()
--E~MultiSchemaPartitionsExec(shard=0) on ActorPlanDispatcher(actor=0)  // rhs
-T~PeriodicSamplesMapper()
--E~MultiSchemaPartitionsExec(shard=1) on ActorPlanDispatcher(actor=1)  // lhs
-T~PeriodicSamplesMapper()
--E~MultiSchemaPartitionsExec(shard=1) on ActorPlanDispatcher(actor=1)  // rhs
```

Data is pulled from each shard, sent to the BinaryJoin actor, then that single actor needs to process all of this data.

When (1) a target-schema is defined and (2) every join-key fully-specifies the target-schema columns, we can relieve much of this single-actor pressure. Lhs/rhs values will never be joined across shards, so the following ExecPlan would yield the same result as the above plan:

```
E~LocalPartitionDistConcatExec() on ActorPlanDispatcher(actor=0)
-E~BinaryJoinExec(binaryOp=ADD) on ActorPlanDispatcher(actor=0)
--T~PeriodicSamplesMapper()
---E~MultiSchemaPartitionsExec(shard=0) on InProcessPlanDispatcher()  // lhs
--T~PeriodicSamplesMapper()
---E~MultiSchemaPartitionsExec(shard=0) on InProcessPlanDispatcher()  // rhs
-E~BinaryJoinExec(binaryOp=ADD) on ActorPlanDispatcher(actor=1)
--T~PeriodicSamplesMapper()
---E~MultiSchemaPartitionsExec(shard=1) on InProcessPlanDispatcher()  // lhs
--T~PeriodicSamplesMapper()
---E~MultiSchemaPartitionsExec(shard=1) on InProcessPlanDispatcher()  // rhs
[...]

PR #1371, opened 2022-04-13
feat(query): add BinaryJoin pushdown optimization

alextheimer:
**Pull Request checklist**

This is the "`LogicalPlan` counterpart" of #1344. Here, plans are materialized directly into their optimized state.

Pending optimization cases (in the order to be completed):
- [x] Raw data pushdowns (no mappers/aggregators)
- [x] Nested pushdowns
- [x] Pushdowns with function mappers
- [ ] Pushdowns with aggregators
- [ ] Exploit commutativity and pushdown to grandchildren

The general idea:
```
  /**
   * Suppose we had the following LogicalPlan:
   *
   * BinaryJoin(lhs=[PeriodicSeries on shards 0,1], rhs=[PeriodicSeries on shards 0,1])
   *
   * This might be materialized as follows:
   *
   * E~BinaryJoinExec(binaryOp=ADD)
   * -T~PeriodicSamplesMapper()
   * --E~MultiSchemaPartitionsExec(shard=0)  // lhs
   * -T~PeriodicSamplesMapper()
   * --E~MultiSchemaPartitionsExec(shard=1)  // lhs
   * -T~PeriodicSamplesMapper()
   * --E~MultiSchemaPartitionsExec(shard=0)  // rhs
   * -T~PeriodicSamplesMapper()
   * --E~MultiSchemaPartitionsExec(shard=1)  // rhs
   *
   * Data is pulled from two shards, sent to the BinaryJoin actor, then that single actor
   *   needs to process all of this data.
   *
   * When (1) a target-schema is defined for all leaf-level filters, (2) all of these
   *   target schemas are identical, and (3) the join-key fully-specifies the
   *   target-schema labels, we can relieve much of this single-actor pressure.
   *   Lhs/rhs values will never be joined across shards, so the following ExecPlan
   *   would yield the same result as the above plan:
   *
   * E~LocalPartitionDistConcatExec()
   * -E~BinaryJoinExec(binaryOp=ADD)
   * --T~PeriodicSamplesMapper()
   * ---E~MultiSchemaPartitionsExec(shard=0)
   * --T~PeriodicSamplesMapper()
   * ---E~MultiSchemaPartitionsExec(shard=0)
   * -E~BinaryJoinExec(binaryOp=ADD)
   * --T~PeriodicSamplesMapper()
   * ---E~MultiSchemaPartitionsExec(shard=1)
   * --T~PeriodicSamplesMapper()
   * ---E~MultiSchemaPartitionsExec(shard=1)
   *
   * Now, data is joined locally and in smaller batches.
   */
```


query/src/main/scala/filodb/query/exec/ExecPlan.scala
```diff
@@ -264,17 +264,20 @@ trait ExecPlan extends QueryCommand {
...
   /**
     * Prints the ExecPlan and RangeVectorTransformer execution flow as a tree
     * structure, useful for debugging
     *
     * @param useNewline pass false if the result string needs to be in one line
     */
-  final def printTree(useNewline: Boolean = true, level: Int = 0): String = {
+  final def printTree(useNewline: Boolean = true,
+                      level: Int = 0,
+                      childSort: (ExecPlan) => Int = (ep) => 0): String = {
```
alextheimer, line 276: I need to add documentation here for this `childSort` parameter. It's just for tests, so string comparisons between different `printTree` calls are more predictable.

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
```diff
@@ -269,45 +304,143 @@ class SingleClusterPlanner(val dataset: Dataset,
...
+    }.toSet)
+  }
+
+  // scalastyle:off method.length
+  // scalastyle:off cyclomatic.complexity
+  /**
+   * Returns the set of shards to which this BinaryJoin can be pushed down.
+   * @return an occupied Option iff it is valid to perform the a pushdown optimization on this BinaryJoin.
+   */
+  private def getBinaryJoinPushdownShards(qContext: QueryContext,
```
alextheimer, line 392: If we end up adding aggregator pushdowns, this method is pretty easily converted to a general "getPushdownShards" for either pushdown optimization.

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
```diff
@@ -22,8 +22,68 @@ import filodb.query.LogicalPlan._
...
+   * @param filters Query Column Filters
+   * @param targetSchema TargetSchema
+   * @return useTargetSchema - use target-schema to calculate query shards
+   */
+  private def useTargetSchemaForShards(filters: Seq[ColumnFilter], targetSchema: Option[TargetSchemaChange]): Boolean =
+    targetSchema match {
+      case Some(ts) if ts.schema.nonEmpty => ts.schema
+        .forall(s => filters.exists(cf => cf.column == s && cf.filter.isInstanceOf[Filter.Equals]))
+      case _ => false
+    }
```
alextheimer, line 57: This is all Sher's code (just moved into the class object).

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
```diff
@@ -269,45 +304,143 @@ class SingleClusterPlanner(val dataset: Dataset,
...
+            case _ => throw new IllegalArgumentException(s"unhandled type: ${rs.rangeSelector.getClass}")
+          }
+          case _ => throw new IllegalArgumentException(s"unhandled type: ${leafPlan.getClass}")
+        }
+        val tsLabels = findTargetSchema(tsFunc, start, stop).map(_.schema)
+        LeafInfo(leafPlan, filters.head, tsLabels)
+      } else {
+        LeafInfo(leafPlan, filters.head, None)
+      }
+    }
```
alextheimer, line 371: I'll try to make this logic less confusing (and, at the very least, not check that the same TSP is defined for each plan).

coordinator/src/main/scala/filodb.coordinator/queryplanner/DefaultPlanner.scala
```diff
@@ -459,4 +486,18 @@ object PlannerUtil extends StrictLogging {
...
+    def helper(ep: ExecPlan, set: mutable.HashSet[Int]): Unit = ep match {
+      case mspe: MultiSchemaPartitionsExec => set.add(mspe.shard)
+      case nl: NonLeafExecPlan => nl.children.foreach(helper(_, set))
+      case _ => throw new IllegalArgumentException(s"unhandled type: ${ep.getClass}")
+    }
+    val set = new mutable.HashSet[Int]
+    helper(ep, set)
+    set.toSet
+  }
 }
```
alextheimer, line 503: I think this can be removed, now.

alextheimer: Nope, since the spec uses it to sort printTree output.

query/src/main/scala/filodb/query/exec/ExecPlan.scala
alextheimer: Added

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
```diff
@@ -22,8 +22,68 @@ import filodb.query.LogicalPlan._
...
+  /**
+   * Returns an occupied option with target schema labels iff they are defined and identical
+   *   for every leaf LogicalPlan.
+   */
+  private def getUniversalTargetSchemaLabels(lp: LogicalPlan,
+                                    tsp: TargetSchemaProvider): Option[Seq[String]] = {
+    lp match {
+      case rs: RawSeries => {
+        val rangeSelectorOpt: Option[(Long, Long)] = rs.rangeSelector match {
+          // TODO(a_theimer): other selectors?
```
alextheimer, line 68: Am I missing something here?

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
```diff
@@ -269,45 +304,147 @@ class SingleClusterPlanner(val dataset: Dataset,
...
+    // construct a LeafInfo for each leaf
+    val targetSchemaProvider = qContext.plannerParams.targetSchemaProvider
+    val leaves = findLeafLogicalPlans(lp).map{ leafPlan =>
+      val filters = getColumnFilterGroup(leafPlan)
+      assert(filters.size == 1, s"expected leaf plan to yield single filter group, but got ${filters.size}")
+      if (targetSchemaProvider.isDefined) {
+        // store the target schema labels (if they're defined) into the LeafInfo
+        val tsFunc = targetSchemaProvider.get.targetSchemaFunc(filters.head.toSeq)
+        val (start, stop) = leafPlan match {
+          // TODO(a_theimer): other leaf plan types? Range selector types?
```
alextheimer, line 360: Missing anything here?

query/src/main/scala/filodb/query/exec/ExecPlan.scala
```diff
@@ -264,17 +264,22 @@ trait ExecPlan extends QueryCommand {
     */
   protected def args: String
 
+  // TODO(a_theimer): more generic way to handle lambda param with default?
```
alextheimer, line 267: If you've got a more scala-thonic way to do this, let me know-- Google has not been super helpful.
I want something like:
```
ExecPlan => T = (ep) => 0
```

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
```diff
@@ -269,45 +304,147 @@ class SingleClusterPlanner(val dataset: Dataset,
...
+          case rs: RawSeries => rs.rangeSelector match {
+            case is: IntervalSelector => (is.from, is.to)
+            case _ => throw new IllegalArgumentException(s"unhandled type: ${rs.rangeSelector.getClass}")
```
alextheimer, line 363: I see I've handled this differently than in `getUniversalTargetSchema`. I'll changes this to instead return None if the IntervalSelector is unrecognized. Maybe I should do this for the unrecognized LogicalPlan case, as well?

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
```diff
@@ -269,45 +305,149 @@ class SingleClusterPlanner(val dataset: Dataset,
...
+              case is: IntervalSelector => findTargetSchema(tsFunc, is.from, is.to).map(_.schema)
+              case _ => None
+            }
+          } else None
+          leafInfos.append(LeafInfo(leafPlan, filters.head, tsLabels))
+        }
+        // Do nothing; not pulling data from any shards.
+        case sc: ScalarPlan => {}
+        // Note!! If an unrecognized plan type is encountered, this just pessimistically returns None.
+        case _ => return None
```
alextheimer, line 373: This applies to a couple matchers in this code: I could instead (1) add a case for all possible types, or (2) throw an exception when an unrecognized type is encountered. (1) seems like clutter, and (2) seems unsafe.

coordinator/src/test/scala/filodb.coordinator/queryplanner/SingleClusterPlannerSpec.scala
```diff
@@ -449,13 +450,205 @@ class SingleClusterPlannerSpec extends AnyFunSpec with Matchers with ScalaFuture
...
+          |---E~BinaryJoinExec(binaryOp=ADD, on=List(job, app), ignoring=List()) on ActorPlanDispatcher(Actor[akka://default/system/testProbe-1#-852082083],raw)
+          |----T~PeriodicSamplesMapper(start=20000000, step=100000, end=30000000, window=None, functionId=None, rawSource=true, offsetMs=None)
+          |-----E~MultiSchemaPartitionsExec(dataset=timeseries, shard=17, chunkMethod=TimeRangeChunkScan(19700000,30000000), filters=List(ColumnFilter(job,Equals(baz)), ColumnFilter(__name__,Equals(foo))), colName=None, schema=None) on InProcessPlanDispatcher(filodb.core.query.QueryConfig@7c3e4b1a)
+          |----T~PeriodicSamplesMapper(start=20000000, step=100000, end=30000000, window=None, functionId=None, rawSource=true, offsetMs=None)
+          |-----E~MultiSchemaPartitionsExec(dataset=timeseries, shard=17, chunkMethod=TimeRangeChunkScan(19700000,30000000), filters=List(ColumnFilter(job,Equals(baz)), ColumnFilter(__name__,Equals(bat))), colName=None, schema=None) on InProcessPlanDispatcher(filodb.core.query.QueryConfig@7c3e4b1a)""".stripMargin),
+      // scalars should not prevent pushdown
+      ("""foo{job="baz"} + 5""",
```
vishramachandran, line 625: Excellent that we cover this.
Do we also do pushdown for query such as `foo or vector(0)` ? This is a very common query that is issued today, and would be nice to optimize. If yes, please add a test case. If no, let us add a todo and a ticket to cover this in a separate PR. I dont want for us to further bloat this one. 🙂

coordinator/src/test/scala/filodb.coordinator/queryplanner/SingleClusterPlannerSpec.scala
```diff
@@ -449,13 +450,205 @@ class SingleClusterPlannerSpec extends AnyFunSpec with Matchers with ScalaFuture
...
+          |--FA1~StaticFuncArgs(5.0,RangeParams(20000,100,30000))
+          |--T~PeriodicSamplesMapper(start=20000000, step=100000, end=30000000, window=None, functionId=None, rawSource=true, offsetMs=None)
+          |---E~MultiSchemaPartitionsExec(dataset=timeseries, shard=1, chunkMethod=TimeRangeChunkScan(19700000,30000000), filters=List(ColumnFilter(job,Equals(baz)), ColumnFilter(__name__,Equals(foo))), colName=None, schema=None) on ActorPlanDispatcher(Actor[akka://default/system/testProbe-1#1813367789],raw)
+          |-T~ScalarOperationMapper(operator=ADD, scalarOnLhs=false)
+          |--FA1~StaticFuncArgs(5.0,RangeParams(20000,100,30000))
+          |--T~PeriodicSamplesMapper(start=20000000, step=100000, end=30000000, window=None, functionId=None, rawSource=true, offsetMs=None)
+          |---E~MultiSchemaPartitionsExec(dataset=timeseries, shard=17, chunkMethod=TimeRangeChunkScan(19700000,30000000), filters=List(ColumnFilter(job,Equals(baz)), ColumnFilter(__name__,Equals(foo))), colName=None, schema=None) on ActorPlanDispatcher(Actor[akka://default/system/testProbe-1#1813367789],raw)""".stripMargin)
+    )
```
vishramachandran, line 635: Can we add a test case where push down needs to happen for small portion of large query ?

coordinator/src/test/scala/filodb.coordinator/queryplanner/SingleClusterPlannerSpec.scala
```diff
@@ -449,13 +450,205 @@ class SingleClusterPlannerSpec extends AnyFunSpec with Matchers with ScalaFuture
...
+    def spread(filter: Seq[ColumnFilter]): Seq[SpreadChange] = {
+      Seq(SpreadChange(0, 1))
+    }
+    def targetSchema(filter: Seq[ColumnFilter]): Seq[TargetSchemaChange] = {
+      Seq(TargetSchemaChange(0, Seq("job", "app")))
+    }
+
+    // Note: all expected plan strings are generated by sorting children on the
+    //   smallest shard ID from which data is pulled.
+    val queryExpectedPairs = Seq(
```
vishramachandran, line 471: I love the extensive list of test cases and docs for each test case. Thanks!

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
```diff
@@ -22,8 +23,68 @@ import filodb.query.LogicalPlan._
...
+          val tsChanges = tsp.targetSchemaFunc(rs.filters)
+          findTargetSchema(tsChanges, fromMs, toMs).map(_.schema)
+        }
+        if (targetSchemaOpt.isDefined) targetSchemaOpt.get else None
+      }
+      case sc: ScalarPlan => Some(Seq.empty)
+      case nl: NonLeafLogicalPlan => {
+        val tsLabelOpts = nl.children.map(getUniversalTargetSchemaLabels(_, tsp))
+        val allDefinedAndSame = tsLabelOpts.forall(_.isDefined) &&
+          tsLabelOpts.drop(1).forall(_ == tsLabelOpts.head)
```
vishramachandran, line 82: nice trick! some English words to explain to reader what is happening will be useful 🙂

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
```diff
@@ -269,45 +305,149 @@ class SingleClusterPlanner(val dataset: Dataset,
...
 
-  override def materializeBinaryJoin(qContext: QueryContext,
-                                     lp: BinaryJoin): PlanResult = {
-    val lhs = walkLogicalPlanTree(lp.lhs, qContext)
+  /**
+   * Returns the shards spanned by a LogicalPlan's data.
+   * @return an occupied Option iff shards could be determined for all leaf-level plans.
+   */
+  private def getShardSpanFromLp(lp: LogicalPlan,
+                                 qContext: QueryContext): Option[Set[Int]] = {
```
vishramachandran, line 346: The docs can convey that it will return a `Some` only iff the entire exec plan subtree for this logical plan can be pushed down to individual shards and concatenated, as opposed to an exec plan with distributed execution followed by some form of reduction at the top.

vishramachandran (commented):
@alextheimer I love the attention to detail in your work. I look at the thorough test cases and feel quite assured that you are paying attention to fine details on this. Thank you for taking the time to get surgical precision here (if I may use this analogy). 🤝 

Once again, thank you for the multiple iterations in simplifying and paying the cost of improving the understandability of code. 

There are some quick comments and suggestions. After that, we are good to merge.

coordinator/src/main/scala/filodb.coordinator/queryplanner/DefaultPlanner.scala
```diff
@@ -459,4 +486,18 @@ object PlannerUtil extends StrictLogging {
     childTargets.iterator.drop(rnd.nextInt(childTargets.size)).next
    }
   }
+
+  /**
+   * Returns the set of shards that contain an ExecPlan's data.
+   */
+   def getShardSpanFromEp(ep: ExecPlan): Set[Int] = {
```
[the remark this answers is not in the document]
alextheimer: It was used to order children of an ExecPlan during `printTree`, but it's now removed (along with the new `printTree` parameter) since the print order is no longer random.

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
```diff
@@ -22,8 +23,68 @@ import filodb.query.LogicalPlan._
...
+          case IntervalSelector(fromMs, toMs) => Some((fromMs, toMs))
+          case _ => None
+        }
+        val targetSchemaOpt = rangeSelectorOpt.map{ case (fromMs, toMs) =>
+          val tsChanges = tsp.targetSchemaFunc(rs.filters)
+          findTargetSchema(tsChanges, fromMs, toMs).map(_.schema)
+        }
+        if (targetSchemaOpt.isDefined) targetSchemaOpt.get else None
+      }
+      case sc: ScalarPlan => Some(Seq.empty)
```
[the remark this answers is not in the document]
alextheimer: Fixed and added comments. I see `ScalarVaryingDoublePlan` is the issue-- cases have been re-arranged so all non-leaf plans are processed before other cases.

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
```diff
@@ -269,45 +305,149 @@ class SingleClusterPlanner(val dataset: Dataset,
     */
   // scalastyle:off cyclomatic.complexity
   override def walkLogicalPlanTree(logicalPlan: LogicalPlan,
-                                  qContext: QueryContext): PlanResult = {
+                                  qContext: QueryContext,
+                                   forceInProcess: Boolean): PlanResult = {
```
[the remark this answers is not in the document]
alextheimer: > that it will cause the exec plan subtree for the entire logical plan to be pushed down fully into individual shards and concatenated

Not true-- if some`ExecPlan` tree pulls data from multiple shards, materialization with `forceInProcess=true` does not affect whether-or-not a pushdown occurs, and it will blindly dispatch everything with an `InProcessDispatcher`. 

I had added a small blurb about it with the other param docs in `DefaultPlanner::defaultWalkLogicalPlanTree`, but I've now moved all of these descriptions to (hopefully more appropriately) the abstract `DefaultPlanner::walkLogicalPlanTree`.

(The `forceInProcess` description is:
```
@param forceInProcess if true, all materialized plans will dispatch via an InProcessDispatcher
```
Let me know if that is insufficient.)

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
```diff
@@ -269,45 +305,149 @@ class SingleClusterPlanner(val dataset: Dataset,
...
+                        targetSchemaLabels: Option[Seq[String]])
+
+    // construct a LeafInfo for each leaf...
+    val targetSchemaProvider = qContext.plannerParams.targetSchemaProvider
+    val leafInfos = new ArrayBuffer[LeafInfo]()
+    for (leafPlan <- findLeafLogicalPlans(lp)) {
+      val filters = getColumnFilterGroup(leafPlan)
+      assert(filters.size == 1, s"expected leaf plan to yield single filter group, but got ${filters.size}")
+      leafPlan match {
+        case rs: RawSeries => {
```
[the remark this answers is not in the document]
alextheimer: Done!

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
```diff
@@ -333,13 +473,92 @@ class SingleClusterPlanner(val dataset: Dataset,
     PlanResult(joined)
   }
 
+  // scalastyle:off method.length
```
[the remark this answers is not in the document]
alextheimer: Nope-- removed

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
alextheimer: Done!

coordinator/src/test/scala/filodb.coordinator/queryplanner/SingleClusterPlannerSpec.scala
alextheimer: Good catch. I've added a case in `getBinaryPushdownShards` to prevent `VectorPlan` pushdowns. The reasoning is:
```
      // VectorPlans can't currently be pushed down. Consider:
      //     foo{...} or vector(0)
      //   If foo{...} is sharded with spread=1, but both shards contain no foo{...} data,
      //   then both pushed-down plans will yield vector(0). These will be concatenated via a DistConcatExec.
```

coordinator/src/test/scala/filodb.coordinator/queryplanner/SingleClusterPlannerSpec.scala
alextheimer: Added the following (where only that last, inner-most join should be pushed down):
```
       """(
         |  (
         |    foo{job="baz"} + on(app, inst, tag) bar{job="baz"}
         |  ) + on(job, inst) (
         |    rate(foo{job="baz"}[1m]) + ignoring(inst) ceil(bar{job="baz"})
         |  )
         |) + on(job, inst, tag) (
         |  (
         |    floor(foo{job="baz"}) + on(job) bar{job="baz"}
         |  ) + on(job, inst) (
         |    sgn(rate(foo{job="baz"}[1m])) + on(job, app) sqrt(bar{job="baz"})
         |  )
         |)""".stripMargin,
```

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
alextheimer: I think you've described this separate function (which uses `getShardSpanFromLp`) (new comments have been added since you last looked):
```
  /**
   * Returns the set of shards to which this BinaryJoin can be pushed down.
   * When this function returns a Some, the argument BinaryJoin can be materialized for each shard
   *   in the result, and the concatenation of these shard-local joins will, on execute(), yield the
   *   same RangeVectors as the non-pushed-down plan.
   * See materializeBinaryJoinWithPushdown for details about this pushdown optimization.
   * @return an occupied Option iff it is valid to perform the a pushdown optimization on this BinaryJoin.
   */
  private def getBinaryJoinPushdownShards(qContext: QueryContext,
                                          bj: BinaryJoin): Option[Set[Int]]
```

`getShardSpanFromLp` is pushdown-agnostic; it just walks the tree and returns a `Some` iff it can determine which shards contain the plan's data.

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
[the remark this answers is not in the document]
alextheimer: See my other [comment](https://github.com/filodb/FiloDB/pull/1371#discussion_r862327200) about `foo or vector(0)`-- roughly the same idea applies to `ScalarVaryingDoublePlan`. Since SVDP's should contain only one row, it doesn't make sense to push them down.

For example:
```
foo + scalar(bar{<all-target-schema-labels>, unique_timeseries_tag="tag1"})
```
where spread=1. If the scalar lies on one shard-- say, shard 0-- and `foo` lies on 0 and 1, then the scalar could only be pushed down with `foo`-0, but we'd still need to add it to `foo`-1. If the scalar lies on multiple shards, then it plainly cannot be pushed down with `foo`.

coordinator/src/test/scala/filodb.coordinator/queryplanner/SingleClusterPlannerSpec.scala
alextheimer: Ticket is created.

coordinator/src/main/scala/filodb.coordinator/queryplanner/DefaultPlanner.scala
```diff
@@ -163,7 +172,13 @@ trait  DefaultPlanner {
     }
 
    // scalastyle:off method.length
-   def addAggregator(lp: Aggregate, qContext: QueryContext, toReduceLevel: PlanResult):
+
+   /**
+    * @param forceDispatcher forces the reducer exec plan to dispatch with the provided dispatcher
```
sherali42, line 177: this doc is not needed here?

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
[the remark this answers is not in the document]
vishramachandran: sounds good. Add 'for this logical plan'

```
@param forceInProcess if true, all materialized plans for this entire logical plan will dispatch via an InProcessDispatcher
```

sherali42 (approved):
@alextheimer excellent 💯 , thank you!!

coordinator/src/test/scala/filodb.coordinator/queryplanner/SingleClusterPlannerSpec.scala
vishramachandran: Thank, lets create a unit test that ensures that push-down does not happen for vector(0) binary joins.

vishramachandran (approved):
Approved with some comments!

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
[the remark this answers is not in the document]
alextheimer: added

coordinator/src/test/scala/filodb.coordinator/queryplanner/SingleClusterPlannerSpec.scala
alextheimer: Done (glad you suggested this-- it caught a bug).

coordinator/src/main/scala/filodb.coordinator/queryplanner/DefaultPlanner.scala
alextheimer: Good catch-- removed

PR #1373, opened 2022-04-21
perf(query): Label cardinality uses faceting improvement

vishramachandran:
**Pull Request checklist**

LabelCardinality now uses lucene faceting to count label cardinalities, as opposed to iterating each time series that match filters.

Benchmarks show that faceting call is several orders of magnitude faster than iteration. So we make multiple calls to iterate over each label's values and add to the sketch instead of loading each partKey into heap.


amolnayak311 (approved):
LGTM, Just to reiterate, we have chosen to give up the ability to find label cardinality for multiple shard keys for now, perhaps in future if needed, we will add that capability

PR #1368, opened 2022-04-12
bug(core): If purgeRate > evictRate, partIdsToEvict list can grow without size cap

vishramachandran:
**Pull Request checklist**

- [x] The commit(s) message(s) follows the contribution [guidelines](CONTRIBUTING.md) ?
- [ ] Tests for the changes have been added (for bug fixes / features) ?
- [ ] Docs have been added / updated (for bug fixes / features) ?

If time series partitions in shared are purged before they are evicted, then partIdsToEvict list can grow uncontrollably big. We need to prune this list regularly.

Thinking how to unit test this. Will add if I have an approach to do this.

PR #1377, opened 2022-05-04
cherry-pick commit for 0.9.16 hotfix

alextheimer:


vishramachandran:
Please revert the "bug(core): If purgeRate > evictRate, partIdsToEvict list can grow without size cap (#1368)" from the integration branch in this PR. We may need more time to test it. It can go in next round

vishramachandran (approved):

PR #1380, opened 2022-05-06
fix(query): absent with sum_over_time result should be consistent

v267:
**Pull Request checklist**

**Current behavior :**
```
% curl -i --location -g --request GET 'http://localhost:9900/ws-bar/api/v1/query_range?query=absent(sum_over_time(foo{_ns_="ns-bar"}[5m]))&start=1651815600&end=1651816200&step=60'
HTTP/1.1 200 OK
content-type: application/json; charset=UTF-8
Content-Type: application/json
content-length: 515
Access-Control-Allow-Methods: GET

{
  "status" : "success",
  "data" : {
    "resultType" : "matrix",
    "result" : [ {
      "metric" : {
        "_ns_" : "ns-bar",
        "_ws_" : "ws-bar"
      },
      "values" : [ [ 1651815600, "1.0" ], [ 1651815660, "1.0" ], [ 1651815720, "1.0" ], [ 1651815780, "1.0" ], [ 1651815840, "1.0" ], [ 1651815900, "1.0" ], [ 1651815960, "1.0" ], [ 1651816020, "1.0" ], [ 1651816080, "1.0" ], [ 1651816140, "1.0" ], [ 1651816200, "1.0" ] ]
    } ]
  },
  "errorType" : null,
  "error" : null
}
```

**New behavior :**
```
% curl -i --location -g --request GET 'http://localhost:9900/ws-bar/api/v1/query_range?query=absent(sum_over_time(foo{_ns_="ns-bar"}[5m]))&start=1651815600&end=1651816200&step=60'
HTTP/1.1 200 OK
content-type: application/json; charset=UTF-8
Content-Type: application/json
content-length: 435
Access-Control-Allow-Methods: GET

{
  "status" : "success",
  "data" : {
    "resultType" : "matrix",
    "result" : [ {
      "metric" : { },
      "values" : [ [ 1651815600, "1.0" ], [ 1651815660, "1.0" ], [ 1651815720, "1.0" ], [ 1651815780, "1.0" ], [ 1651815840, "1.0" ], [ 1651815900, "1.0" ], [ 1651815960, "1.0" ], [ 1651816020, "1.0" ], [ 1651816080, "1.0" ], [ 1651816140, "1.0" ], [ 1651816200, "1.0" ] ]
    } ]
  },
  "errorType" : null,
  "error" : null
}
```

### DIFF
--- a/coordinator/src/test/scala/filodb.coordinator/queryplanner/ShardKeyRegexPlannerSpec.scala
+++ b/coordinator/src/test/scala/filodb.coordinator/queryplanner/ShardKeyRegexPlannerSpec.scala
@@ -1000,6 +1000,20 @@ class ShardKeyRegexPlannerSpec extends AnyFunSpec with Matchers with ScalaFuture
 
   it("should materialize absent function mapper correctly with implicit WS") {
     {
+      // absent of sum_over_time: _ws_ in query
+      val lp = Parser.queryToLogicalPlan(
+        "absent(sum_over_time(test{_ws_ = \"demo\", _ns_ = \"App-1\", instance = \"Inst-1\" }[5m]))",
+        1000, 1000)
+      val shardKeyMatcherFn = (shardColumnFilters: Seq[ColumnFilter]) => {
+        Seq(Seq(ColumnFilter("_ns_", Equals("App-1")), ColumnFilter("_ws_", Equals("demo"))))
+      }
+      val engine = new ShardKeyRegexPlanner(dataset, localPlanner, shardKeyMatcherFn, queryConfig)
+      val execPlan = engine.materialize(lp, QueryContext(origQueryParams = promQlQueryParams))
+      execPlan.asInstanceOf[LocalPartitionReduceAggregateExec].rangeVectorTransformers.head
+        .asInstanceOf[AbsentFunctionMapper].columnFilter.isEmpty shouldEqual true
+    }
+    {
+      // absent of sum_over_time: _ws_ NOT in query
       val lp = Parser.queryToLogicalPlan("absent(sum_over_time(test{_ns_ = \"App-1\", instance = \"Inst-1\" }[5m]))",
         1000, 1000)
       val shardKeyMatcherFn = (shardColumnFilters: Seq[ColumnFilter]) => {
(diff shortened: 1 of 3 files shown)


coordinator/src/test/scala/filodb.coordinator/queryplanner/ShardKeyRegexPlannerSpec.scala
```diff
@@ -997,4 +997,30 @@ class ShardKeyRegexPlannerSpec extends AnyFunSpec with Matchers with ScalaFuture
       case _ => fail("Expected two children")
     }
   }
-}
\ No newline at end of file
+
+  it("should materialize absent function mapper correctly with implicit WS") {
```
boyapatichandu, line 1001: nit: Can we add one more test for absent without over_time function .. e.x `absent(test{_ns_ = \"App-1\", instance = \"Inst-1\" })` This should return column filters with _ws_ included.

boyapatichandu (approved):
One minor nitpick. Apart from changes looks good to me.

coordinator/src/test/scala/filodb.coordinator/queryplanner/ShardKeyRegexPlannerSpec.scala
alextheimer: Wait, I don't think that's right. Absent either returns an empty result or a single RangeVector with an empty key.

alextheimer: I agree that we should add similar test cases for `absent` and `absent_over_time`, though.

alextheimer: I think just asserting that these should give empty column filters is sufficient:
```
absent(test{_ws_=\"demo\", _ns_ = \"App-1\", instance = \"Inst-1\" })
absent(test{_ns_ = \"App-1\", instance = \"Inst-1\" })
absent_over_time(test{_ws_=\"demo\", _ns_ = \"App-1\", instance = \"Inst-1\" }[5m])
absent_over_time(test{_ns_ = \"App-1\", instance = \"Inst-1\" }[5m])
```

(and I would be curious to know if anything different happens...)

boyapatichandu (approved):
Looks good to me

coordinator/src/test/scala/filodb.coordinator/queryplanner/ShardKeyRegexPlannerSpec.scala
alextheimer: Correction to the above: if there are non-metric labels given with the query, they are included in the absent result.

The prometheus docs show the following:
- absent(<selector>): all non-metric labels should be in the result iff rows are present
- absent_over_time(<range-selector>): all non-metric labels should be in the result iff rows are present
- absent(sum_over_time(<range-selector>)): `sum_over_time` will either return rows (with full keys), or not. If it does, absent returns nothing. If it does not, then there are no selector labels to even include as a RangeVector key.

I'll approve the PR, since these look good. Could you add some functional tests for each of these cases just to be safe? (I see only one `absent(sum_over_time)` test, for example).

alextheimer (approved):
left some comments-- lgtm, but please add functional tests just to be safe

PR #1386, opened 2022-05-11
fix(query): propagate ApplyInstantFunction column filters to FunctionArgsPlans

alextheimer:
**Pull Request checklist**

The `ApplyInstantFunction` implementation of `replacePeriodicSeriesFilters` did not propagate filters to its `FunctionArgsPlan` children. Now, it recurses for these children, too.

### DIFF
--- a/coordinator/src/test/scala/filodb.coordinator/queryplanner/SingleClusterPlannerSpec.scala
+++ b/coordinator/src/test/scala/filodb.coordinator/queryplanner/SingleClusterPlannerSpec.scala
@@ -7,7 +7,7 @@ import filodb.coordinator.ShardMapper
 import filodb.coordinator.client.QueryCommands.{FunctionalSpreadProvider, FunctionalTargetSchemaProvider, StaticSpreadProvider}
 import filodb.core.{GlobalScheduler, MetricsTestData, SpreadChange, TargetSchemaChange}
 import filodb.core.metadata.Schemas
-import filodb.core.query._
+import filodb.core.query.{ColumnFilter, _}
 import filodb.core.query.Filter.Equals
 import filodb.core.store.TimeRangeChunkScan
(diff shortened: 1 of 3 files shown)


amolnayak311:
- Thanks @alextheimer  for the PR. As we spoke, lets look at other ExecPlans, specifically ``ApplyInstantFunctionRaw``, ``SubqueryWithWindowing`` and ``PeriodicSeriesWithWindowing`` to see if we need to make a similar change.
- Lets try to come up with a representative PromQL for each of these exec plans and assert our changes work as expected.

coordinator/src/test/scala/filodb.coordinator/queryplanner/SingleClusterPlannerSpec.scala
```diff
@@ -1312,4 +1313,104 @@ class SingleClusterPlannerSpec extends AnyFunSpec with Matchers with ScalaFuture
...
+      initialFilters,
+      Nil
+    )
+    val periodicSeries = PeriodicSeries(
+      rawSeries,
+      1, 10, 100
+    )
+    val funcArg = ScalarVaryingDoublePlan(periodicSeries, ScalarFunctionId.Scalar, Nil)
+
+    val plans = Seq(
```
alextheimer, line 1369: These are manually constructed because:
- the parser forbids `-WithWindowing` types with non-fixed args, so we can't otherwise test whether-or-not `scalar()`-wrapped vector args are updated by `replaceFilters`. (99% sure something like `quantile_over_time(scalar(foo{}), bar{}[])` is valid PromQL).
- `ApplyInstantFunctionRaw` and `ScalarVaryingDoublePlan`s are never constructed with `ColumnFilter`-containing args when these plans are parsed out of a query string, so we wouldn't be able to test `replaceFilters` for them, either.

tjackpaul (approved):
lgtm

PR #1387, opened 2022-05-14
fix(query): add histogram case to ScalarOperationMapper

alextheimer:
**Pull Request checklist**

Previously, histograms were not handled by `ScalarOperationMapper`. Histogram-handler code has now been added.

Example: if `hist_metric` is a histogram, then
```
hist_metric{label="foo"} + 5
```
would throw an exception because `ScalarOperationMapper` would call `getDouble` on a `TransientHistRow`.

### DIFF
--- a/query/src/main/scala/filodb/query/exec/RangeVectorTransformer.scala
+++ b/query/src/main/scala/filodb/query/exec/RangeVectorTransformer.scala
@@ -2,8 +2,10 @@ package filodb.query.exec
 
 import scala.collection.Iterator
 import scala.collection.mutable.ListBuffer
+
 import monix.reactive.Observable
 import spire.syntax.cfor._
+
 import filodb.core.metadata.Column.ColumnType
 import filodb.core.metadata.PartitionSchema
 import filodb.core.query._
--- a/query/src/main/scala/filodb/query/exec/RangeVectorTransformer.scala
+++ b/query/src/main/scala/filodb/query/exec/RangeVectorTransformer.scala
@@ -214,7 +214,7 @@ final case class ScalarOperationMapper(operator: BinaryOperator,
       val resultIterator: RangeVectorCursor = new WrappedCursor(rv.rows) {
         private val rows = rv.rows
         private val result = valueColumnType(sourceSchema) match {
-          case ColumnType.DoubleColumn => new TransientRow(0, 0d)
+          case ColumnType.DoubleColumn => new TransientRow(timestamp = 0, value = 0d)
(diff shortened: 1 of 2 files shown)


query/src/main/scala/filodb/query/exec/RangeVectorTransformer.scala
```diff
@@ -197,36 +196,66 @@ final case class ScalarOperationMapper(operator: BinaryOperator,
...
+            case ColumnType.DoubleColumn =>
+              val doubleResRow = result.asInstanceOf[TransientRow]
+              val oldVal = next.getDouble(1)
+              val newVal = if (scalarOnLhs) operatorFunction.calculate(sclrVal, oldVal)
+                           else operatorFunction.calculate(oldVal, sclrVal)
+              doubleResRow.setValues(timestamp, newVal)
+            case ColumnType.HistogramColumn =>
+              val histResRow = result.asInstanceOf[TransientHistRow]
+              val oldHist = next.getHistogram(1).asInstanceOf[HistogramWithBuckets]
+              if (oldHist.numBuckets > histResRow.getHistogram(1).numBuckets) {
```
alextheimer, line 237: This fix is not pretty, but the alternative is to read a single row, buffer it, and initialize the `MutableHistogram` to the correct size. This code is substantially less straightforward to read. If `ColumnType`s weren't `object` classes, maybe we could add a `numBuckets` field to `HistogramColumn`?

query/src/main/scala/filodb/query/exec/RangeVectorTransformer.scala
```diff
@@ -197,36 +198,66 @@ final case class ScalarOperationMapper(operator: BinaryOperator,
...
+                // If we don't have enough buckets in the TransientHistRow, we initialize and store
+                //   an appropriately-sized histogram. This should only happen once.
+                histResRow.setValues(0, MutableHistogram(oldHist))
```
vishramachandran, line 242: You `histResRow.getHistogram` on 0th index, but invoke `histResRow.setValues` on 1st index. Intentional ?

vishramachandran (approved):
Approved with important comment

query/src/main/scala/filodb/query/exec/RangeVectorTransformer.scala
alextheimer: Intellij displays those parameter labels that make it easy to forget when magic numbers don't actually have any name/description-- I'll add a "timestamp" label; that is not an index.

PR #1402, opened 2022-06-09
test (core): Updated test for durable index storage

kvpetrov:
**Pull Request checklist**

**New behavior :**

**BREAKING CHANGES**

Breaking changes may include:

**Other information**:

### DIFF
--- a/core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
+++ b/core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
@@ -71,11 +71,21 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
 
   private val stats = new DownsampledTimeSeriesShardStats(rawDatasetRef, shardNum)
 
+  private val indexMetadataStore : Option[IndexMetadataStore] = {
+    downsampleConfig.indexMetastoreImplementation match {
+      case IndexMetastoreImplementation.NoImp => None
+      case IndexMetastoreImplementation.File => Some(
+        new FileCheckpointedIndexMetadataStore(downsampleConfig.indexLocation.get)
+      )
+      case IndexMetastoreImplementation.Ephemeral => Some(new EphemeralIndexMetadataStore())
+    }
+  }
+
   private val partKeyIndex = new PartKeyLuceneIndex(indexDataset, schemas.part, false,
     false, shardNum, indexTtlMs,
     downsampleConfig.indexLocation.map(new java.io.File(_)),
-    if (downsampleConfig.enablePersistentIndexing)
-      Some(new FileCheckpointedIndexMetadataStore(downsampleConfig.indexLocation.get)) else None)
+    indexMetadataStore
+  )
 
   private val indexUpdatedHour = new AtomicLong(0)
 
@@ -148,15 +158,20 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
           recoverIndexInternal(None)
         case (IndexState.Synced, checkpointMillis)             =>
           logger.info("Found index state empty, bootstrapping downsample index")
+          // TODO
+          // although here we provide the logic to invoke recoverIndexInternal from a particular
+          // checkpoint, recoverIndexInternal does not recover from a checkpoint properly because
+          // it would not restart from an appropriate partId and instead would start with 0.
+          // So, this code is the skeleton for the future implementation and allows us to put
+          // a unit test in the code.
           recoverIndexInternal(checkpointMillis)
         case _                                  =>
           // FIXME: Part ids in case of indexRefresh are assumed to monotonically increase, however, in case recovery
           //  is done from an existing timestamp, we need a way not only to store the state of the ids recycled but
           //  latest allocated partId to avoid assigning the same id again.
-          //logger.info(s"Nothing to recover the index for dataset=$indexDataset shard=$shardNum" +
-          //            s" starting index refresh thread")
-          //indexRefresh().runToFuture(housekeepingSched)
-          ???
+          logger.info(s"Nothing to recover the index for dataset=$indexDataset shard=$shardNum" +
+                      s" starting index refresh thread")
+          indexRefresh().runToFuture(housekeepingSched)
       }
     } else {
       // Index persistence is not enabled, this will simply follow the path for existing index recovery
@@ -164,6 +179,14 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
     }
   }
 
+  // TODO
+  // This method provides in its signature the way to recover from a particular checkpoint.
+  // Currently however, it should always be called with None as it will not recover properly
+  // given a checkpoint. The reason is that it will start producing partIds from 0 while there
+  // will already be an index on disk with some number of partIds. We cannot utilize the
+  // existing logic of refreshWithDownsamplePartKey either as it will grow partIds indefinitely
+  // without any restarts and the rebuilding will be slow. Most likely we want to completely
+  // get rid of partIds for downsample shards.
   private def recoverIndexInternal(checkpointMillis: Option[Long]): Future[Long] = {
     indexBootstrapper
       .bootstrapIndexDownsample(
@@ -222,7 +245,7 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
     }
   }
 
-  def indexRefresh(): Task[Unit] = {
+  def indexRefresh(): Task[Long] = {
     // Update keys until hour()-2 hours ago. hour()-1 hours ago can cause missed records if
     // refresh was triggered exactly at end of the hour. All partKeys for the hour would need to be flushed
     // before refresh happens because we will not revist the hour again.
@@ -239,7 +262,7 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
     indexRefresh(toHour, fromHour)
   }
 
-  def indexRefresh(toHour: Long, fromHour: Long): Task[Unit] = {
+  def indexRefresh(toHour: Long, fromHour: Long): Task[Long] = {
     indexRefresher.refreshWithDownsamplePartKeys(
       partKeyIndex, shardNum, rawDatasetRef, fromHour, toHour, schemas)(lookupOrCreatePartId)
       .map { count =>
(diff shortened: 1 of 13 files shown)


core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
```diff
@@ -74,7 +74,8 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
   private val partKeyIndex = new PartKeyLuceneIndex(indexDataset, schemas.part, false,
     false, shardNum, indexTtlMs,
     downsampleConfig.indexLocation.map(new java.io.File(_)),
-    if (downsampleConfig.enablePersistentIndexing) Some(new EphemeralIndexMetadataStore()) else None)
+    if (downsampleConfig.enablePersistentIndexing)
```
amolnayak311, line 77: Please do not remove existing code, use config to instantiate right store. We may have a different implementation of store later

core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
```diff
@@ -139,47 +140,51 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
...
       }
     } else {
       // Index persistence is not enabled, this will simply follow the path for existing index recovery
-      recoverIndexInternal
+      recoverIndexInternal(None)
     }
   }
 
-  private def recoverIndexInternal: Future[Unit] = {
+  private def recoverIndexInternal(checkpointMillis: Option[Long]): Future[Long] = {
```
amolnayak311, line 167: Ideally we use refresh to update the index since last sync time and thus recover, which is called from bootstrap will be called only if full rebuild is needed. Check out the implementation of ``createPartitionID`` it simply increments a counter where as increment update of index can not  use an incrementing counter for partId as we had discussed. Given this information  of ``createPartitionID``, ``recoverIndexInternal``  can only rebuild the index from scratch and not from any given timestamp and ``refreshWithDownsamplePartKeys`` will be the way to do that

core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
```diff
@@ -139,47 +140,51 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
...
         case _                                  =>
           // FIXME: Part ids in case of indexRefresh are assumed to monotonically increase, however, in case recovery
           //  is done from an existing timestamp, we need a way not only to store the state of the ids recycled but
           //  latest allocated partId to avoid assigning the same id again.
-          logger.info(s"Nothing to recover the index for dataset=$indexDataset shard=$shardNum" +
-                      s" starting index refresh thread")
-          indexRefresh().runToFuture(housekeepingSched)
+          //logger.info(s"Nothing to recover the index for dataset=$indexDataset shard=$shardNum" +
+          //            s" starting index refresh thread")
+          //indexRefresh().runToFuture(housekeepingSched)
```
amolnayak311, line 158: Why comment out code? This is needed to start the index refresh thread

core/src/main/scala/filodb.core/memstore/IndexBootstrapper.scala
```diff
@@ -56,21 +56,24 @@ class IndexBootstrapper(colStore: ColumnStore) {
...
     // state, start - ttlMs)
-
     val recoverIndexLatency = Kamon.gauge("shard-recover-index-latency", MeasurementUnit.time.milliseconds)
       .withTag("dataset", ref.dataset)
       .withTag("shard", shardNum)
     val start = System.currentTimeMillis()
     //here we need to adjust ttlMs
     //because we might already have index available on disk
-    val checkpointTime = start - ttlMs
+    var checkpointTime = checkpointMillis.getOrElse(0L)
```
amolnayak311, line 73: Check previous comment on what bootstrap can not accept a start time to refresh from yet.  Unless dependency on partIds is removed.

core/src/main/scala/filodb.core/memstore/IndexMetadataStore.scala
```diff
@@ -106,4 +110,38 @@ class EphemeralIndexMetadataStore extends IndexMetadataStore {
...
+      (IndexState.TriggerRebuild, None)
+    } else {
+      (IndexState.Synced, Some(checkpointTime))
+    }
+    currentState((datasetRef, shard)) = state
+    state
+  }
+
+  // NOT sure I understand why we need this call?
+  override def updateInitState(datasetRef: DatasetRef, shard: Int, state: IndexState.Value, time: Long): Unit = {
```
amolnayak311, line 143: How do you trigger index refresh externally? It is important to distinguish between initial state and state when shards are running. If the state (not init state) is updated with TriggerRebuild what would be the behavior when shards are running and next ``refreshIndex`` is scheduled?

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -143,6 +143,8 @@ class PartKeyLuceneIndex(ref: DatasetRef,
         notifyLifecycleListener(IndexState.TriggerRebuild, System.currentTimeMillis)
         throw new IllegalStateException("Unable to clean up index directory", t)
     }
+  } else {
```
amolnayak311, line 146: Remove non empty else block, feel free to leave the comment

core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
```diff
@@ -139,47 +140,51 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
...
+        val endHour = hour() - indexJobIntervalInHours - 1
+        indexUpdatedHour.set(endHour)
+        // The time set here in synced state should match the time we set in indexUpdatedHour
+        partKeyIndex.notifyLifecycleListener(IndexState.Synced, endHour * 3600 * 1000L)
+        stats.shardTotalRecoveryTime.update(System.currentTimeMillis() - creationTime)
+        startHousekeepingTask()
+        startStatsUpdateTask()
+        logger.info(s"Shard now ready for query dataset=$indexDataset shard=$shardNum")
+        isReadyForQuery = true
+        count
```
amolnayak311, line 186: Where are you using this count?

core/src/main/scala/filodb.core/memstore/IndexMetadataStore.scala
```diff
@@ -106,4 +110,38 @@ class EphemeralIndexMetadataStore extends IndexMetadataStore {
     initStateMap((datasetRef, shard)) = (state, Some(time))
 }
 
+/**
+ * Non thread safe in memory index state metadata store
+ */
+class FileCheckpointedIndexMetadataStore(val indexLocation: String) extends IndexMetadataStore {
```
amolnayak311, line 116: What is the point of storing the timestamp on file and state in memory? Either both be stored on disk or both in memory. On restart how would you know the state this time in file corresponds to?

amolnayak311 (changes requested):

core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
kvpetrov: put configuration

core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
kvpetrov: log but don't remove it

core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
```diff
@@ -139,47 +140,51 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
...
+  def recoverIndex(): Future[Long] = {
     if (downsampleConfig.enablePersistentIndexing) {
       partKeyIndex.getCurrentIndexState() match {
         case (IndexState.Empty, _)              =>
           logger.info("Found index state empty, bootstrapping downsample index")
-          recoverIndexInternal
+          recoverIndexInternal(None)
+        case (IndexState.Synced, checkpointMillis)             =>
+          logger.info("Found index state empty, bootstrapping downsample index")
+          recoverIndexInternal(checkpointMillis)
```
kvpetrov, line 151: probably it should be renamed to recoverIndexEmpty() which does not take millis, (it will figure out millis on its own)

core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
```diff
@@ -139,47 +140,51 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
...
 
-  def recoverIndex(): Future[Unit] = {
+  def recoverIndex(): Future[Long] = {
     if (downsampleConfig.enablePersistentIndexing) {
       partKeyIndex.getCurrentIndexState() match {
         case (IndexState.Empty, _)              =>
           logger.info("Found index state empty, bootstrapping downsample index")
-          recoverIndexInternal
+          recoverIndexInternal(None)
+        case (IndexState.Synced, checkpointMillis)             =>
```
kvpetrov, line 149: should be called with checkpointMillis but it's not recoverIndexInternal, should be something else, for ex recoverIndexSync, internally it should do logic similar to recover index fresh. (refreshWithDownsamplePartKey)

core/src/main/scala/filodb.core/memstore/IndexMetadataStore.scala
kvpetrov: I assume we will address it later with the "command" concept, leaving it "as is" for now.

core/src/main/scala/filodb.core/memstore/IndexBootstrapper.scala
kvpetrov: Added a comment. The code provides the capability to recover from a checkpoint when we add missing functionality. Essentially, here we are adding a skeleton from the  future implementation.

core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
kvpetrov: so, I put a comment in the method, essentially telling that currently in production settings one should call the method only with None.

amolnayak311 (approved):
Left couple of comments. Thanks for the PR

PR #1403, opened 2022-06-10
misc(core): add support for multi-field facets to partition-key columns

sherali42:
**Pull Request checklist**

Add support for multi-field facets to partition-key columns. This will optimize the metadata query lookups, If the dataset is hierarchical and there are heavy volume of multi-label value lookups.
- configured/enabled per dataset
- supports only partition-key string type columns

### DIFF
--- a/core/src/main/resources/filodb-defaults.conf
+++ b/core/src/main/resources/filodb-defaults.conf
@@ -55,7 +55,7 @@ filodb {
       # lucene facet feature.
       #
       # Note: only "string" column type supported.
-      # additionalFacets = [
+      # multiColumnFacets = [
       #     {
       #       all-hosts-availZone = ["avail_zone", "rack", "host"]
       #     }
--- a/core/src/main/resources/filodb-defaults.conf
+++ b/core/src/main/resources/filodb-defaults.conf
@@ -40,13 +40,27 @@ filodb {
       # Eg: For copyTags `"_ns_" = [ "exporter", "job" ]`, when an incoming TimeSeries already has both `exporter`
       #     and `job`, then it will pickup only the value from `exporter` for `_ns_`.
       copyTags = {
-        "_ns_" = [ "_ns", "exporter", "job" ]
+        "_ns_" = ["_ns", "exporter", "job"]
       }
-      ignoreShardKeyColumnSuffixes = { "_metric_" = ["_bucket", "_count", "_sum"] }
+      ignoreShardKeyColumnSuffixes = {"_metric_" = ["_bucket", "_count", "_sum"]}
       ignoreTagsOnPartitionKeyHash = ["le"]
       metricColumn = "_metric_"
       # shard key columns should be in hierarchical order
       shardKeyColumns = ["_ws_", "_ns_", "_metric_"]
+
+      # Multi-column facets to be created on "partition-schema" columns.
+      # Eg: below config creates a facet "all-hosts-availZone" by concatenating
+      # the column values ("avail_zone", "rack", "host") "\u03C0" as delimiter
+      # If the data is hierarchical in nature, This feature lets us fetch the metadata very efficiently by using
+      # lucene facet feature.
+      #
+      # Note: only "string" column type supported.
+      # additionalFacets = [
+      #     {
+      #       all-hosts-availZone = ["avail_zone", "rack", "host"]
+      #     }
+      #   ]
+      multiColumnFacets = []
     }
   }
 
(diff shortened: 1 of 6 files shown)


core/src/main/resources/filodb-defaults.conf
```diff
@@ -40,13 +40,27 @@ filodb {
...
       shardKeyColumns = ["_ws_", "_ns_", "_metric_"]
+
+      # Multi-column facets to be created on "partition-schema" columns.
+      # Eg: below config creates a facet "all-hosts-availZone" by concatenating
+      # the column values ("avail_zone", "rack", "host") "\u03C0" as delimiter
+      # If the data is hierarchical in nature, This feature lets us fetch the metadata very efficiently by using
+      # lucene facet feature.
+      #
+      # Note: only "string" column type supported.
+      # additionalFacets = [
```
amolnayak311, line 58: What's is ``additionalFacets`` vs ``multiColumnFacets``?

Just calling out, I believe this concatenation must be less than ``FACET_FIELD_MAX_LEN``

core/src/main/scala/filodb.core/metadata/Schemas.scala
```diff
@@ -58,6 +58,7 @@ final case class DataSchema private(name: String,
 final case class PartitionSchema(columns: Seq[Column],
                                  predefinedKeys: Seq[String],
                                  options: DatasetOptions) {
+  val columnIdxLookup = columns.zipWithIndex.map { case(col, idx) => col.name -> Tuple2(col, idx) }.toMap
```
amolnayak311, line 61: ``Tuple2(col, idx)`` can be written as ``(col, idx)``

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -565,12 +569,40 @@ class PartKeyLuceneIndex(ref: DatasetRef,
...
     luceneDocument.get().document
   }
 
+  // Multi-column facets to be created on "partition-schema" columns
+  private def createMultiColumnFacets(partKeyOnHeapBytes: Array[Byte], partKeyBytesRefOffset: Int): Unit = {
+    schema.options.multiColumnFacets.foreach(facetCols => {
+      facetCols match {
+        case (name, cols) => {
+          val concatFacetValue = cols.flatMap { col =>
+            val (columnInfo, pos) = schema.columnIdxLookup(col)
```
amolnayak311, line 588: If the column is missing it will throw ``java.util.NoSuchElementException``. Better option would be invoke ``get`` to get an ``Option``  and when ``None``, either throw a specific exception with meaningful message or log a warning and return empty string. Throwing an exception will fail adding the document to index.

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -565,12 +569,40 @@ class PartKeyLuceneIndex(ref: DatasetRef,
...
+            if (columnInfo.columnType == StringColumn) {
+              val base = partKeyOnHeapBytes
+              val offset = bytesRefToUnsafeOffset(partKeyBytesRefOffset)
+              val strOffset = schema.binSchema.blobOffset(base, offset, pos)
+              val numBytes = schema.binSchema.blobNumBytes(base, offset, pos)
+              Seq(new String(base, strOffset.toInt - UnsafeUtils.arayOffset,
```
amolnayak311, line 594: Why flatMap on a ``Seq`` of length 1 or 0, can we use ``map`` and either return a string, either empty or the value?

amolnayak311 (approved):
LGTM, have left few comments if you can address them please?

core/src/main/resources/filodb-defaults.conf
sherali42: ah good catch, i refactored later and forgot to update the example. will fix it

core/src/main/scala/filodb.core/metadata/Schemas.scala
sherali42: IDE is complaining when i was running test cases, just to please IDE I added this.

core/src/main/resources/filodb-defaults.conf
sherali42: `FACET_FIELD_MAX_LEN` this is being checked in `addFacet`. so field will be omitted if it crosses the threshold.

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
sherali42: 👍 fixed it

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
sherali42: I wanted to exclude the field altogether in case the field type is not `string`. But fixed it anyway 👍

PR #1404, opened 2022-06-10
feat(query): add resultByteLimit

alextheimer:
**Pull Request checklist**

Adds `resultByteLimit` to `PlannerParams`. If a query result size exceeds this, the query will fail.

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
@@ -23,6 +23,7 @@ import filodb.core.memstore.ratelimit.CardinalityRecord
 import filodb.core.metadata.{Dataset, Schemas}
 import filodb.core.query.{QueryConfig, QueryContext, QuerySession, QueryStats}
 import filodb.core.store.CorruptVectorException
+import filodb.memory.data.Shutdown
 import filodb.query._
 import filodb.query.exec.ExecPlan
 
@@ -107,6 +108,10 @@ final class QueryActor(memStore: TimeSeriesStore,
       override def uncaughtException(t: Thread, e: Throwable): Unit = {
         logger.error("Uncaught Exception in Query Scheduler", e)
         uncaughtExceptions.increment()
+        e match {
+          case ie: InternalError => Shutdown.haltAndCatchFire(ie)
+          case _ => { /* Do nothing. */ }
+        }
       }
     }
     val threadFactory = new ForkJoinPool.ForkJoinWorkerThreadFactory {
--- a/memory/src/main/scala/filodb.memory/data/Shutdown.scala
+++ b/memory/src/main/scala/filodb.memory/data/Shutdown.scala
@@ -8,8 +8,9 @@ import kamon.Kamon
 object Shutdown extends StrictLogging {
 
   val forcedShutdowns = Kamon.counter("forced-shutdowns").withoutTags()
-  def haltAndCatchFire(e: Exception, unitTest: Boolean = false): Unit = {
+  def haltAndCatchFire(e: Throwable, unitTest: Boolean = false): Unit = {
     forcedShutdowns.increment()
+    Thread.sleep(60000)  // Sleep 1m to make sure counter is published.
     if (unitTest) throw e
     logger.error(s"Shutting down process since it may be in an unstable/corrupt state", e)
     import scala.concurrent.duration._


core/src/main/scala/filodb.core/query/QueryContext.scala
```diff
@@ -26,10 +24,16 @@ case class PlannerParams(applicationId: String = "filodb",
...
+                         queryTimeoutMillis: Int = GlobalConfig.systemConfig
+                           .getDuration("filodb.query.ask-timeout").toMillis.toInt,
+                         sampleLimit: Int = GlobalConfig.systemConfig
+                           .getInt("filodb.query.sample-limit"),
+                         groupByCardLimit: Int = GlobalConfig.systemConfig
+                           .getInt("filodb.query.group-by-cardinality-limit"),
+                         joinQueryCardLimit: Int = GlobalConfig.systemConfig
+                           .getInt("filodb.query.join-cardinality-limit"),
+                         resultByteLimit: Long = GlobalConfig.systemConfig
+                           .getLong("filodb.query.result-byte-limit"),
```
alextheimer, line 36: I can change these all back if we want to keep the number of changes minimal.

query/src/main/scala/filodb/query/exec/ExecPlan.scala
```diff
@@ -221,11 +221,18 @@ trait ExecPlan extends QueryCommand {
...
               .record(Math.max(0, System.currentTimeMillis - startExecute))
+            // Does not include sizes of RV's that arrived at this node as SerializableRangeVectors.
             val numDataBytes = builder.allContainers.map(_.numBytes).sum
             val numKeyBytes = r.foldLeft(0)(_ + _.key.keySize)
             val resultSize = numDataBytes + numKeyBytes
             SerializedRangeVector.queryResultBytes.record(resultSize)
-            querySession.queryStats.getResultBytesCounter(Nil).addAndGet(resultSize)
+            // Includes sizes of all RV's in this subtree.
+            val totalSize = querySession.queryStats.getResultBytesCounter(Nil).addAndGet(resultSize)
+            if (totalSize > queryContext.plannerParams.resultByteLimit) {
```
alextheimer, line 231: It would be nice if this check could be done with each RV (i.e. keep a running byte total), but there isn't really a great way to get a `SerializableRangeVector`'s size without iterating through its builder's containers each time, ala:
```
startBytes = builder.allContainers.map(_.numBytes).sum
SerializedRangeVector(builder, other_args)
endBytes = builder.allContainers.map(_.numBytes).sum
srvBytes = endBytes - totalBytes
```

If that is preferable to a single check at the end, then I would also need to count sizes of the RV's that are already SRV's before they're processed. I haven't yet disproven the assumption that "any already-SRV we process must have its size included in the `QuerySession` counter", so I could just check that `counter + sizes_of_other_RVs` doesn't exceed the limit. I'm not sure how contentious the counter lock might become, though.

The other option would be to just give each SRV a `size` attribute and calculate on construction. This would get weird with scalars, though. I could probably just ignore things like fixed scalars, but I would need to handle the "varying" flavors like:
```
final case class ScalarVaryingDouble(private val timeValueMap: Map[Long, Double],
                                     override val outputRange: Option[RvRange]) extends ScalarRangeVector {
```
I could ballpark it?

All of this is mostly to say: I think this single check is good enough.

amolnayak311 (commented):
While I understand the intent behind this PR, we do not know what query that works currently stops working. Failing the query that previously worked isn't a pleasant experience and I would go on to log potential queries that would fail, giving enough time to users to fix them.  Since there isn't a good way to send back additional information with the result, logging a warning seems like the best choice. This is in same spirit of deprecation of APIs where we keep the warning at least for a one or two versions.

core/src/main/resources/filodb-defaults.conf
```diff
@@ -171,18 +171,21 @@ filodb {
...
     # Binary Join Cardinality limit
-    join-cardinality-limit = 25000
+    join-cardinality-limit = 100000
 
     # Group by Cardinality limit
-    group-by-cardinality-limit = 1000
+    group-by-cardinality-limit = 100000
+
+    # Maximum result size returnable from any ExecPlan::execute call (including nested calls)
+    result-byte-limit = 1000000000
```
alextheimer, line 188: These were previously unused-- their now-used values are pulled from the `PlannerParams` constructor.

amolnayak311: What would happen to an instant aggregated query with high churn?I understand this is a very large number but something to callout for? Rather than throwing an exception, I will give a flexibility to either throw an exception or log a warning. This would make it easy to see which unexpected queries which worked previously now start (or would start) failing. This would give a change to fine tune these limits without impacting the end users across versions.

alextheimer: Excellent point-- I've added a config flag that indicates whether-or-not the limit should be enforced. Regardless of enforcement, there is now always a warning logged if the limit is breached.

core/src/main/scala/filodb.core/query/QueryContext.scala
vishramachandran: Lets not change this please

vishramachandran (changes requested):
Thanks for this important PR!

core/src/main/resources/filodb-defaults.conf
```diff
@@ -171,18 +171,24 @@ filodb {
...
     # Maximum number of samples to return in a query
     sample-limit = 1000000
 
     # Binary Join Cardinality limit
-    join-cardinality-limit = 25000
+    join-cardinality-limit = 100000
 
     # Group by Cardinality limit
-    group-by-cardinality-limit = 1000
+    group-by-cardinality-limit = 100000
```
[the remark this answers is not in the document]
alextheimer: reverted

core/src/main/resources/filodb-defaults.conf
```diff
@@ -171,18 +171,24 @@ filodb {
...
+
+    # Maximum result size returnable from any ExecPlan::execute call (including nested calls)
+    result-byte-limit = 1000000000
```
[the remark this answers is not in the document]
alextheimer: I had it ultra-high so things wouldn't break-by-default back when this PR enforced the limit by default. Bumped down to 18MB now.

core/src/main/scala/filodb.core/query/QueryContext.scala
alextheimer: reverted

query/src/main/scala/filodb/query/exec/ExecPlan.scala
```diff
@@ -221,11 +221,22 @@ trait ExecPlan extends QueryCommand {
...
               .withTag("plan", getClass.getSimpleName)
               .record(Math.max(0, System.currentTimeMillis - startExecute))
+            // Does not include sizes of RV's that arrived at this node as SerializableRangeVectors.
             val numDataBytes = builder.allContainers.map(_.numBytes).sum
             val numKeyBytes = r.foldLeft(0)(_ + _.key.keySize)
             val resultSize = numDataBytes + numKeyBytes
             SerializedRangeVector.queryResultBytes.record(resultSize)
-            querySession.queryStats.getResultBytesCounter(Nil).addAndGet(resultSize)
+            // Includes sizes of all RV's in this subtree.
+            val totalSize = querySession.queryStats.getResultBytesCounter(Nil).addAndGet(resultSize)
```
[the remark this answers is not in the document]
alextheimer: I've made the change, but I'm still skeptical that `resultSize` has the count we want, since (1) `numDataBytes` is initialized like this:
```
val numDataBytes = builder.allContainers.map(_.numBytes).sum
```
and (2) `builder` is only mutated by the `RangeVector` case (not `SerializableRangeVector`) of an earlier `map` over the RV's of the `ExecResult`.

For example, `numDataBytes` would undercount for `LocalPartitionDistConcatExec`  because its `ExecResult` plans would (in most cases) be `SerializableRangeVectors`. I can demonstrate this undercount locally.

alextheimer:
test

query/src/main/scala/filodb/query/exec/ExecPlan.scala
```diff
@@ -221,11 +221,22 @@ trait ExecPlan extends QueryCommand {
...
+            if (totalSize > querySession.qContext.plannerParams.resultByteLimit) {
+              qLogger.warn(s"Maximum result size exceeded (${queryContext.plannerParams.resultByteLimit} bytes). " +
```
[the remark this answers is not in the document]
alextheimer: fixed

query/src/main/scala/filodb/query/exec/ExecPlan.scala
```diff
@@ -221,11 +221,22 @@ trait ExecPlan extends QueryCommand {
...
               .record(Math.max(0, System.currentTimeMillis - startExecute))
+            // Does not include sizes of RV's that arrived at this node as SerializableRangeVectors.
             val numDataBytes = builder.allContainers.map(_.numBytes).sum
             val numKeyBytes = r.foldLeft(0)(_ + _.key.keySize)
             val resultSize = numDataBytes + numKeyBytes
             SerializedRangeVector.queryResultBytes.record(resultSize)
-            querySession.queryStats.getResultBytesCounter(Nil).addAndGet(resultSize)
+            // Includes sizes of all RV's in this subtree.
+            val totalSize = querySession.queryStats.getResultBytesCounter(Nil).addAndGet(resultSize)
+            if (totalSize > querySession.qContext.plannerParams.resultByteLimit) {
```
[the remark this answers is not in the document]
alextheimer: I (incorrectly) dismissed this idea [here](https://github.com/filodb/FiloDB/pull/1404#discussion_r894965474)-- I see now that SRV's maintain a sequence of only the `RecordContainers` they occupy. The latest commit uses these to (1) compute a running `resultSize` and (2) exit early if the maximum size is breached.

Again, I'm skeptical that `resultSize` should only contain incoming-`SerializableRangeVector` sizes: please see https://github.com/filodb/FiloDB/pull/1404#discussion_r912547337

PR #1405, opened 2022-06-12
fix(core): add periodic TSShard index reader refresh

alextheimer:
**Pull Request checklist**

Adds periodic Lucene `IndexReader` refreshes in `TimeSeriesShard` to prevent faceted label-value queries from including now-deleted labels.

### DIFF
--- a/core/src/main/resources/filodb-defaults.conf
+++ b/core/src/main/resources/filodb-defaults.conf
@@ -465,6 +465,9 @@ filodb {
     # If all of the index-faceting-enabled-* properties are false, faceting is fully disabled.
     # Disable if performance cost of faceting all labels is too high
     index-faceting-enabled-for-all-labels = true
+
+    # Lucene IndexReader refresh frequency.
+    reader-refresh-interval = 6 hours
   }
 
   # for standalone worker cluster configuration, see akka-bootstrapper
--- a/core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
+++ b/core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
@@ -192,6 +192,8 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
     // records with same part key. This is true since we update part keys only once per flush interval in raw dataset.
     logger.info(s"Starting housekeeping for downsample cluster of dataset=$rawDatasetRef shard=$shardNum " +
       s"every ${rawStoreConfig.flushInterval}")
+    require(!downsampleStoreConfig.readerRefreshInterval.isDefined,
+      "cannot define the reader refresh interval for downsaple config; raw flush interval is used")
     houseKeepingFuture = Observable.intervalWithFixedDelay(rawStoreConfig.flushInterval,
       rawStoreConfig.flushInterval).mapEval { _ =>
       purgeExpiredIndexEntries()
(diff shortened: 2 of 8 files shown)


core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -463,6 +463,12 @@ class TimeSeriesShard(val ref: DatasetRef,
 
   private[memstore] val addPartitionsDisabled = AtomicBoolean(false)
 
+  private[memstore] val housekeepingSched = Scheduler.computation(
+    name = "housekeeping",
+    reporter = UncaughtExceptionReporter(logger.error("Uncaught Exception in Housekeeping Scheduler", _)))
+
+  private var houseKeepingFuture: CancelableFuture[Unit] = initHousekeeping()
```
alextheimer, line 470: Will probably rename anything "housekeeping-interval"-related to "reader-refresh-interval"

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -463,6 +463,12 @@ class TimeSeriesShard(val ref: DatasetRef,
 
   private[memstore] val addPartitionsDisabled = AtomicBoolean(false)
 
+  private[memstore] val housekeepingSched = Scheduler.computation(
```
alextheimer, line 466: I'm just mirroring the DS setup-- unsure yet if this is the right way to go about things.

core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
```diff
@@ -192,6 +192,8 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
     // records with same part key. This is true since we update part keys only once per flush interval in raw dataset.
     logger.info(s"Starting housekeeping for downsample cluster of dataset=$rawDatasetRef shard=$shardNum " +
       s"every ${rawStoreConfig.flushInterval}")
+    require(downsampleStoreConfig.readerRefreshInterval.isEmpty,
```
amolnayak311, line 195: Why put this assertion? Currently refresh in downsample shard is performed as part of housekeeping task which purges expired data.

core/src/main/resources/filodb-defaults.conf
```diff
@@ -465,6 +465,9 @@ filodb {
     # If all of the index-faceting-enabled-* properties are false, faceting is fully disabled.
     # Disable if performance cost of faceting all labels is too high
     index-faceting-enabled-for-all-labels = true
+
+    # Lucene IndexReader refresh frequency.
+    reader-refresh-interval = 6 hours
```
amolnayak311, line 470: 6 hours default IMO is too high. Refresh in Lucene is not as expensive operation like commit and I think ``purgeExpiredPartitions`` is a good place to perform it. [ElasticSearch](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html#index-refresh-interval-setting) for example performs this refresh each second by default on its underlying index.  Our existing ``refreshReadersBlocking`` is not ideal as it will block all threads performing the refresh waiting for the refresh to finish. Instead we should use ``maybeRefresh`` in ``SearcherManager`` which will not block other threads while one of them performs the refresh.  We can also run performance benchmarks to assert our assumptions than have a pessimistic refresh interval. @vishramachandran  what's your view?

amolnayak311 (changes requested):
Left some changes, if we perform refresh along with purge we will have an index consistent with the data.

amolnayak311:
I looked at ``PartKeyLuceneIndex`` and the ``flushThread: ControlledRealTimeReopenThread[IndexSearcher]`` is what performs index refresh periodically. It would be interesting to see why this doesn't work for you in  reading purged partitions.

core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
alextheimer: That's right-- I'm just being explicit here to prevent someone in the future from assuming "readerRefreshInterval" will affect the downsample store.

alextheimer:
Closing since @amolnayak311 is correct: I can demonstrate locally that `flushThread` sufficiently refreshes the readers.

PR #1415, opened 2022-07-01
fix(exec): mark shard query-ready after index recovery

alextheimer:
**Pull Request checklist**

Currently, a new shard is marked "ready for queries" immediately before initializing an asynchronous index recovery. Until the index is completely built, queries may return partial results.

This PR delays the "ready for queries" mark until after the index has been completely built.

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/TimeSeriesMemStore.scala
+++ b/core/src/main/scala/filodb.core/memstore/TimeSeriesMemStore.scala
@@ -128,14 +128,17 @@ extends TimeSeriesStore with StrictLogging {
                      flushSched: Scheduler,
                      cancelTask: Task[Unit]): CancelableFuture[Unit] = {
     val shard = getShardE(dataset, shardNum)
-    shard.isReadyForQuery = true
     logger.info(s"Shard now ready for query dataset=$dataset shard=$shardNum")
     shard.shardStats.shardTotalRecoveryTime.update(System.currentTimeMillis() - shard.creationTime)
     shard.startIngestion(stream, cancelTask, flushSched)
   }
 
-  def recoverIndex(dataset: DatasetRef, shard: Int): Future[Long] =
-    getShardE(dataset, shard).recoverIndex()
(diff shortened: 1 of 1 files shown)


core/src/main/scala/filodb.core/memstore/TimeSeriesMemStore.scala
```diff
@@ -128,14 +128,17 @@ extends TimeSeriesStore with StrictLogging {
...
     shard.shardStats.shardTotalRecoveryTime.update(System.currentTimeMillis() - shard.creationTime)
     shard.startIngestion(stream, cancelTask, flushSched)
   }
 
-  def recoverIndex(dataset: DatasetRef, shard: Int): Future[Long] =
-    getShardE(dataset, shard).recoverIndex()
+  def recoverIndex(dataset: DatasetRef, shard: Int): Future[Long] = {
+    val shard = getShardE(dataset, shard)
+    val fut = shard.recoverIndex()
+    fut.onComplete(_ => shard.isReadyForQuery = true)
```
amolnayak311: This is currently set in ``startIngestion`` in ``TimeSeriesMemStore`` and for non downsample this is the flow


```
for {
        _ <- tsStore.recoverIndex(ref, shard)
        checkpoints <- tsStore.metastore.readCheckpoints(ref, shard)
      } yield {
        if (checkpoints.isEmpty) {
          normalIngestion(...)
        } else {
          // Figure out recovery end watermark and intervals.  The reportingInterval is the interval at which
          // offsets come back from the MemStore for us to report progress.
           ...
          for {lastOffset <- doRecovery(...)}
            yield {
              normalIngestion(...)
            }
        }
```

This ``normalIngestion`` calls  ``startIngestion`` in ``TimeSeriesMemStore`` and that's where this flag is set to ``true``. By setting this flag from ``onComplete`` of recover we might be setting this bit too early.

CC @vishramachandran what's you take?

alextheimer:
Closing since @amolnayak311 is correct; confirmed locally this does not fix the issue.

PR #1416, opened 2022-07-03
*WIP* fix(core): add Lucene post-bootstrap commit

alextheimer:
**Pull Request checklist**

# \*\*\*\*\*WIP\*\*\*\*\*\*

Forces a `commit` immediately after the index bootstraps (and before readers are refreshed).

From the Lucene docs ([_IndexWriter::commit_](https://lucene.apache.org/core/7_4_0/core/org/apache/lucene/index/IndexWriter.html#commit--)):
> Commits all pending changes (added and deleted documents, segment merges, added indexes, etc.) to the index, and **syncs all referenced index files, such that a reader will see the changes** and the index updates will survive an OS or machine crash or power loss.

This prevents post-bootstrap query responses from excluding time-series that had not yet been committed to the index.

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/IndexBootstrapper.scala
+++ b/core/src/main/scala/filodb.core/memstore/IndexBootstrapper.scala
@@ -41,6 +41,7 @@ class IndexBootstrapper(colStore: ColumnStore) {
       }
       .countL
       .map { count =>
+        index.commit()
         index.refreshReadersBlocking()
         recoverIndexLatency.update(System.currentTimeMillis() - start)
         count


core/src/main/scala/filodb.core/memstore/IndexBootstrapper.scala
```diff
@@ -41,6 +41,7 @@ class IndexBootstrapper(colStore: ColumnStore) {
       }
       .countL
       .map { count =>
+        index.commit()
```
amolnayak311, line 44: I am not sure ``commit`` will help. While commit ensures readers see the latest data, refreshing the index (next line of code) will achieve that too. We do not care about durability of index yet, which is what `` commit `` guarantees.

amolnayak311 (commented):
Left a comment inline to code. I am not sure this will help with the ``absent`` query

PR #1417, opened 2022-07-04
*WIP* feat(query): add result byte size limit

alextheimer:
**Pull Request checklist**

TEMP


query/src/main/scala/filodb/query/exec/ExecPlan.scala
```diff
@@ -188,13 +188,14 @@ trait ExecPlan extends QueryCommand {
       rv : Observable[RangeVector], recordSchema: RecordSchema, resultSchema: ResultSchema
     ): Task[QueryResult] = {
         @volatile var numResultSamples = 0 // BEWARE - do not modify concurrently!!
+        @volatile var resultSize = 0L
```
alextheimer, line 191: Unsure why `volatile` was needed above (afaik `map` is single-threaded), so I'm mirroring it here just-in-case.

query/src/main/scala/filodb/query/exec/ExecPlan.scala
```diff
@@ -208,9 +209,23 @@ trait ExecPlan extends QueryCommand {
 
               numResultSamples += srv.numRowsSerialized
               // fail the query instead of limiting range vectors and returning incomplete/inaccurate results
-              if (enforceLimit && numResultSamples > queryContext.plannerParams.sampleLimit)
+              if (enforceSampleLimit && numResultSamples > queryContext.plannerParams.sampleLimit)
                 throw new BadQueryException(s"This query results in more than ${queryContext.plannerParams.
                   sampleLimit} samples. Try applying more filters or reduce time range.")
+
+              resultSize += srv.numBytes() + srv.key.keySize
```
alextheimer, line 216: Please see: https://github.com/filodb/FiloDB/pull/1404#discussion_r912584492

core/src/main/scala/filodb.core/query/RangeVector.scala
```diff
@@ -324,6 +324,7 @@ final class SerializedRangeVector(val key: RangeVectorKey,
                                   override val outputRange: Option[RvRange]) extends RangeVector with
                                           SerializableRangeVector with java.io.Serializable {
 
+  def numBytes(): Int = containers.map(_.numBytes).sum
```
vishramachandran, line 327: Can you please confirm that this does not alter state of record builders, and this is an O(1) call within each RecordContainer ? Looks like this is being called while the result is being built. Want to be sure.

core/src/main/resources/filodb-defaults.conf
```diff
@@ -184,6 +184,12 @@ filodb {
     # Group by Cardinality limit
     group-by-cardinality-limit = 1000
 
+    # Maximum result size returnable from any ExecPlan::execute call (including nested calls)
+    result-byte-limit = 18MB
```
vishramachandran, line 188: You may want to document that this setting needs to be tweaked in tandem with Akka message size limit. Increasing just this will not do it. Speaking of which, the Akka message size limit of 25MB can also be reduced to 18MB now.

query/src/main/scala/filodb/query/exec/ExecPlan.scala
```diff
@@ -208,9 +209,23 @@ trait ExecPlan extends QueryCommand {
...
+              if (resultSize > querySession.qContext.plannerParams.resultByteLimit) {
+                qLogger.warn(s"Reached maximum result size (final or intermediate) for data serialized out of a host " +
+                  s"(${queryContext.plannerParams.resultByteLimit / 1e6} MB). " +
```
vishramachandran, line 219: This may be a hard to read decimal in error message. Could be formatted for better readability.

Also "for data serialized out of a host" could be "for data serialized out of a host or shard"

vishramachandran (approved):
Approved with comment. Lets make sure the `numBytes` call is not performance intensive.

core/src/main/resources/filodb-defaults.conf
alextheimer: done!

query/src/main/scala/filodb/query/exec/ExecPlan.scala
alextheimer: updated

PR #1423, opened 2022-07-21
fix(query): prevent lookback across partition splits

alextheimer:
**Pull Request checklist**

This PR changes the `MultiPartitionPlanner` to return an empty result for any non-metadata evaluation times where lookback windows cross partition splits.

High-level idea is:
1. Walk the tree to a leaf.
2. Find the partitions/time-ranges that contain the leaf's data.
3. Build a list of "invalid" time-ranges where each describes a lookback window immediately after a partition split. All evaluation times within the ranges are now considered invalid.
4. Propagate these ranges up the tree. Each node updates the ranges such that it passes a result to its parent that indicates its own invalid evaluation times.
5. After the root returns its invalid ranges, invert the ranges to find "valid" query ranges.
6. For each query range, materialize an `ExecPlan`.
7. If necessary, stitch the materialized plans together.

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
@@ -93,7 +93,7 @@ class MultiPartitionPlanner(partitionLocationProvider: PartitionLocationProvider
         //   stale-data lookbacks, etc) do not cross partitions.
         val timeRanges = {
           val invalidRanges = getInvalidRanges(lp, queryParams)
-          val mergedInvalidRanges = mergeRanges(invalidRanges)
+          val mergedInvalidRanges = mergeAndSortRanges(invalidRanges)
           val totalTimeRange = TimeRange(1000 * queryParams.startSecs, 1000 * queryParams.endSecs)
           invertRanges(mergedInvalidRanges, totalTimeRange)
         }
@@ -106,8 +106,15 @@ class MultiPartitionPlanner(partitionLocationProvider: PartitionLocationProvider
           } else {
             copyLogicalPlanWithUpdatedTimeRange(lp, range)
           }
+          // Adjust start seconds in case integer division (during millis-to-seconds conversion)
+          //   gives a timestamp outside the valid range.
+          val startSec: Long = {
+            val msPastSec = range.startMs % 1000
+            (range.startMs / 1000) + { if (msPastSec > 0) 1 else 0 }
+          }
+          val endSec: Long = range.endMs / 1000
           val updQueryContext = qContext.copy(origQueryParams =
-            queryParams.copy(startSecs = range.startMs / 1000, endSecs = range.endMs / 1000))
+            queryParams.copy(startSecs = startSec, endSecs = endSec))
           walkLogicalPlanTree(updLogicalPlan, updQueryContext).plans.head
         }
         if (plans.isEmpty) {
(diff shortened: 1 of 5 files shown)


kvpetrov (commented):
I'm no so sure this will work overall. Suppose you have something like this (taken from unit tests):

"sum_over_time(heap_usage[10m:1m] offset 3m)[2m:1m] offset 1m",
TopLevelSubquery(
  SubqueryWithWindowing(
    PeriodicSeries(
      RawSeries(IntervalSelector(1524855060000,1524855720000),List(ColumnFilter(__name__,Equals(heap_usage))),List(),Some(300000),None),
      1524855060000,60000,1524855720000,None
    ),
    1524855840000,60000,1524855900000,SumOverTime,List(),600000,60000,Some(180000)
  ),
  1524855840000,
  60000,
  1524855900000,120000,Some(60000)
)

The leaf RawSeries plan won't have any lookback and because of that it would get the default 5 minutes (Generally speaking lookback for RawSeries does not make much sense), 5 minutes is an interesting rule that if we don't have data, we are free to grab anything within 5 minute range  Then you would say that because there is a split here, you would want to make the interval of the split + 5 minutes to be invalid. After that you would request that the query at the very top be modified, so, that it would have the original start until the invalid interval starts and then starting from the end of the original interval to the end. But changing the original query start and end has nothing to do with hole that we have in the RawSeries. The problem is twofold:
1) I don't think lookback of rawseries are conceptually lookbacks that you're looking for. In the query above, conceptually you're looking for 2m or maybe 10m? Maybe, you would say I don't care for either, I'm trying to find a real hole that any RawSeries cannot cover. For this one proceed to the second point.
2) start and end of the original query, subquery especially might have very little to do to the start and end raw series underneath it, while at the end of the game you would be modifying start and end of the top query based on the start and end of the leaf RawSeries. In the example above, the original start/end of the query as passed by the user is = 1524 855 988 000, then it opens up to be: 1 524 855 840 000 - 1524 855 900 000 but these are still drastically different from underlying RawSeries 1 524 855 060 000 - 1524 855 720 000. However, if I understood the code properly, we will end up changing the top level based on the holes in the leaf.
 
 Overall, I feel what we are trying is difficult to accomplish properly. I would have rather dealt with the issue purely on the level of RawSeries and send raw data (maybe with some pushdown/aggregation) from one of the clusters to the other one but paying close attention to the amount of data being sent with some limits established. In other words I see what we are doing is a very difficult maybe impossible planner work while instead we should be doing query quota work (which we need badly) and very understandable/easy to test/correct result stitching of RawSeries from different clusters. We can discuss on Monday, I myself don't see easy fixes. Previous iterations of subquery code did not bother with stitching at all and when the data was in two clusters, it was just intelligently picking one of them NOT trying to stitch/query both clusters.

coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
```diff
@@ -80,11 +84,60 @@ class MultiPartitionPlanner(partitionLocationProvider: PartitionLocationProvider
...
+          val mergedInvalidRanges = mergeAndSortRanges(invalidRanges)
+          val totalTimeRange = TimeRange(1000 * queryParams.startSecs, 1000 * queryParams.endSecs)
+          invertRanges(mergedInvalidRanges, totalTimeRange)
+        }
+        // Generate (logicalPlan, qContext) pairs for each of the above time-ranges, then materialize.
+        // If only one valid time range exists, it spans only a single second, and the range matches
+        //   the queryParams, we assume this is intentionally an instant query and do not update
+        //   the plan via "copyLogicalPlanWithUpdatedTimeRange".
+        // copyLogicalPlanWithUpdatedTimeRange would set all nested plan ranges to exactly one second,
+        //   which would break the instant query.
```
alextheimer, line 106: TODO: this is not true, but it does cause issues with `TopLevelSubquery` (and possibly others).

alextheimer:
Merged instead: https://github.com/filodb/FiloDB/pull/1436

PR #1429, opened 2022-08-02
fix(memory) BinaryVector fixes for M1 Mac

amolnayak311:
**Issue**
On Mac M1, ``incNumBytes`` has a quirk. The operation fails with SIGBUS when a ``getAndAddInt`` is invoked on an address offset that is one of the three preceding every multiple of 16. For e.g when addrs is 13, 14, 15, 29, 30, 31, 45, 46, 47 byte offset from the base address, the operation ``UnsafeUtils.unsafe.getAndAddInt(UnsafeUtils.ZeroPointer, addr, inc)`` fails.  However, when we do a get and put non atomically, it works.  The fix is done to currently unblock developers from using M1 Mac. However, the fix is far from ideal as the data written to the block of memory by FiloDB is application specific and the block of memory is opaque to the operating system, in other words, the OS should not be interpreting this data as long as the application is working on the block it is allocated. 

**IMPORTANT**
The PR is built on top of this PR https://github.com/filodb/FiloDB/pull/1428 by @kvpetrov. 

**Next steps**
Once the issue with ``UnsafeUtils.unsafe.getAndAddInt(UnsafeUtils.ZeroPointer, addr, inc)``  is identified, the changes to ``BinaryVector.incNumBytes``  will be reverted.

### DIFF
--- a/memory/src/main/scala/filodb.memory/format/BinaryVector.scala
+++ b/memory/src/main/scala/filodb.memory/format/BinaryVector.scala
@@ -23,6 +23,8 @@ object BinaryVector {
    */
   type BinaryVectorPtr = BinaryRegion.NativePointer
 
+  val isArm64 = System.getProperty("os.arch").equalsIgnoreCase("aarch64")
+
   /**
    * Returns the vector type and subtype from the WireFormat bytes of a BinaryVector
    */
@@ -265,8 +267,6 @@ trait BinaryAppendableVector[@specialized(Int, Long, Double, Boolean) A] {
   // Native address of the appendable vector
   def addr: BinaryRegion.NativePointer
 
-  val arch = System.getProperty("os.arch")
-
   /** Max size that current buffer can grow to */
   def maxBytes: Int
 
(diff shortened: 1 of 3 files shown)


vishramachandran (approved):
Approved, Thanks for digging into this!

memory/src/main/scala/filodb.memory/format/BinaryVector.scala
```diff
@@ -343,7 +345,20 @@ trait BinaryAppendableVector[@specialized(Int, Long, Double, Boolean) A] {
   def checkSize(need: Int, have: Int): AddResponse =
     if (!(need <= have)) VectorTooSmall(need, have) else Ack
 
-  @inline final def incNumBytes(inc: Int): Unit = UnsafeUtils.unsafe.getAndAddInt(UnsafeUtils.ZeroPointer, addr, inc)
+  @inline final def incNumBytes(inc: Int): Unit = arch match {
+    case "aarch64"  =>
```
vishramachandran, line 349: Instead of comparing a string in this method, I would compare with a boolean that is set once in a companion class.

amolnayak311: Good point, I will change it.

whizkido: Nice findings

PR #1432, opened 2022-08-04
fix(query): simplify split-partition logic in MPP

alextheimer:
**Pull Request checklist**

High-level idea: if a leaf-level `LogicalPlan` contains data in multiple partitions (i.e. its data has at some point been migrated between partitions), then issue the same query to each and stitch the results. When the results are stitched, timestamps with unequal values will be dropped.

This is an alternative to https://github.com/filodb/FiloDB/pull/1423. It does not accomplish exactly what that PR does (lookbacks between partitions aren't explicitly prevented), but its logic is substantially simpler.

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
@@ -154,7 +154,7 @@ class MultiPartitionPlanner(partitionLocationProvider: PartitionLocationProvider
       case lp: ApplyAbsentFunction        => super.materializeAbsentFunction(qContext, lp)
       case lp: ScalarBinaryOperation      => super.materializeScalarBinaryOperation(qContext, lp)
       case lp: ApplyLimitFunction         => super.materializeLimitFunction(qContext, lp)
-      case lp: TsCardinalities            => materializeTsCardinalities(lp, qContext)
+      case lp: TsCardinalities            => throw new IllegalArgumentException("TsCardinalities unexpected here")
       case lp: SubqueryWithWindowing       => super.materializeSubqueryWithWindowing(qContext, lp)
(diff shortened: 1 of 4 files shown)


coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
```diff
@@ -190,41 +190,21 @@ class MultiPartitionPlanner(partitionLocationProvider: PartitionLocationProvider
...
-    } else {
-      // 3a. Get the start and end time is ms based on the lookback, offset and the user provided start and end time
-      val (maxOffsetMs, minOffsetMs) = LogicalPlanUtils.getOffsetMillis(logicalPlan)
-        .foldLeft((Long.MinValue, Long.MaxValue)) {
-          case ((accMax, accMin), currValue) => (accMax.max(currValue), accMin.min(currValue))
-        }
-
-      val periodicSeriesTimeWithOffset = TimeRange((queryParams.startSecs * 1000) - maxOffsetMs,
-        (queryParams.endSecs * 1000) - minOffsetMs)
-      val lookBackMs = getLookBackMillis(logicalPlan).max
```
alextheimer, line 218: This is something I found while writing the other PR-- the same lookback and offset are assumed for all top/nested plans.

prometheus/src/main/scala/filodb/prometheus/parse/Parser.scala
```diff
@@ -160,14 +160,14 @@ object Parser extends StrictLogging {
     }
   }
 
-  def queryToLogicalPlan(query: String, queryTimestamp: Long, step: Long, mode: Mode = Shadow): LogicalPlan = {
+  def queryToLogicalPlan(query: String, queryTimestamp: Long, step: Long, mode: Mode = mode): LogicalPlan = {
```
alextheimer, line 163: Sneaking this in with this PR.

alextheimer:
Merged instead: https://github.com/filodb/FiloDB/pull/1436

PR #1434, opened 2022-08-07
maint(core): Upgrade kamon 2.2.2 -> 2.5.6

Scottmitch:
kamon library version upgrade which brings new transative dependencies
such as opentracing and protobuf-java.


tjackpaul (approved):
lgtm.. thanks for the PR

PR #1444, opened 2022-08-25
fix(coord) Stateful set ShardManager bug fix

amolnayak311:
**Pull Request checklist**

**Current behavior :** 

Currently ``shardAssignments`` method given a coordinator returns the list of shards the provided coordinator are assigned to. However, the method should just return just the unassigned shards and not a static list. The current implementation returns a static list which works at startup but when nodes go down, all nodes are reported as unassigned despite those nodes serving the data. This the ``num_active_shards`` metrics incorrectly reports the shards as assigned and not active.

**New behavior :**

The ``shardAssignments``'s implementation is now changed to return a subset of the unassigned shards the provide coordinator should serve. If there are none, an empty list if returned. The unit tests are updated accordingly.


tjackpaul (approved):
lgtm.. thanks for the quick PR @amolnayak311

vishramachandran (approved):

PR #1452, opened 2022-09-10
fix(query) Metadata queries with one valid and one invalid tag returns garbled content

amolnayak311:
**Pull Request checklist**

**Current behavior :**

``MetadataRemoteExec`` as part of optimization  PR https://github.com/filodb/FiloDB/pull/1362, if the data response size is a singleton map, the key name is skipped and the values are returned as a list of strings. However, when a column given is incorrect and does not exist in index, the response has to be a map. 

**New behavior :**

The change checks if the labels given are more than 1 and responds with a string of lists or a singleton map accordingly.


sherali42 (approved):
👍 thanks @amolnayak311 for the PR

PR #1456, opened 2022-09-21
fix(spark-job): make index job run only for downsample chunk window

sherali42:
**Pull Request checklist**

Since we read (for staleness check) before updating index, we don't have to catch up to current time.
Job run is idempotent now.  With this change, job will be run in cadence with Chunk Downsampler job by default.


tjackpaul (approved):
thanks for the quick PR

vishramachandran (approved):

PR #1454, opened 2022-09-16
*WIP* feat(downsample): add data export

alextheimer:
# WIP

**Pull Request checklist**

- [ ] The commit(s) message(s) follows the contribution [guidelines](CONTRIBUTING.md) ?
- [ ] Tests for the changes have been added (for bug fixes / features) ?
- [ ] Docs have been added / updated (for bug fixes / features) ?

PR #1457, opened 2022-09-22
feat(downsample): add data export

alextheimer:
**Pull Request checklist**

Adds an "export" step to the chunk downsampler Spark job.

A time-series is exported to a configured bucket address only if all of the following conditions are met:
- labels match (by equality; without a regex) at least one export rule key
- labels don't match any set of the rule's `exclude` filters
- the `include` filters are empty, or labels match at least one set of the rule's `include` filters

The export path is also configurable:
```
      # Specifies how to generate a path to an exported time-series.
      # The sequence of key-value pairs describe a directory path, where the
      #   time-series are exported to the final directory.
      # For example, if the path-spec is:
      #     path-spec = [
      #       "key1", "value1",
      #       "key2", "value2"
      #     ]
      # then the exported time-series will be stored at the path:
      #     ~/key1=value1/key2=value2/<file>
      # All {{label-name}} strings will be replaced with the time-series label's value.
      # All <<time-spec>> strings will be replaced with the result of formatting
      #   the export window's end time with the time-spec string.
      path-spec = [
        "ws", "{{_ws_}}-foobar",
        "year", "hello-<<YYYY>>",
        "api", "v1",
        "weird-ns-label", "{{_ns_}}"
      ]
```

This PR is a simpler alternative to: https://github.com/filodb/FiloDB/pull/1454

### DIFF
--- a/spark-jobs/src/main/scala/filodb/downsampler/Utils.scala
+++ b/spark-jobs/src/main/scala/filodb/downsampler/Utils.scala
@@ -1,7 +1,8 @@
 package filodb.downsampler
 
-import filodb.core.store.{AllChunkScan, ChunkSetInfoReader, ReadablePartition}
 import kamon.Kamon
+
+import filodb.core.store.{AllChunkScan, ChunkSetInfoReader, ReadablePartition}
 
 object Utils {
 
--- a/spark-jobs/src/main/scala/filodb/downsampler/Utils.scala
+++ b/spark-jobs/src/main/scala/filodb/downsampler/Utils.scala
@@ -1,8 +1,12 @@
 package filodb.downsampler
 
 import filodb.core.store.{AllChunkScan, ChunkSetInfoReader, ReadablePartition}
+import kamon.Kamon
 
 object Utils {
+
+  lazy val numRawChunksSkipped = Kamon.counter("num-raw-chunks-skipped").withoutTags()
+
   /**
    * Specifies a range of a chunk's rows.
    */
@@ -39,8 +43,7 @@ object Utils {
       .filter{ chunkRange =>
         val isValidChunk = chunkRange.istartRow <= chunkRange.iendRow
         if (!isValidChunk) {
-          // TODO(a_theimer): numRawChunksSkipped.increment()
-          // TODO(a_theimer): use DsContext?
+          numRawChunksSkipped.increment()
           DownsamplerContext.dsLogger.warn(s"Skipping chunk of partition since startRow lessThan endRow " +
             s"hexPartKey=${readablePart.hexPartKey} " +
             s"startRow=${chunkRange.istartRow} " +
--- a/spark-jobs/src/main/scala/filodb/downsampler/Utils.scala
+++ b/spark-jobs/src/main/scala/filodb/downsampler/Utils.scala
@@ -11,10 +11,10 @@ object Utils {
                         istartRow: Int,
                         iendRow: Int)
 
-  case class ExportRow(partKeyMap: Map[String, String],
-                       timestamp: Long,
-                       value: Double,
-                       partitionStrings: Iterator[String])
+  case class ExportData(partKeyMap: Map[String, String],
+                        timestamp: Long,
+                        value: Double,
+                        partitionStrings: Iterator[String])
 
   /**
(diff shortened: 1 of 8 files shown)


spark-jobs/src/main/scala/filodb/downsampler/Utils.scala
```diff
@@ -0,0 +1,57 @@
+package filodb.downsampler
+
+import kamon.Kamon
+
+import filodb.core.store.{AllChunkScan, ChunkSetInfoReader, ReadablePartition}
+
+object Utils {
+
+  lazy val numRawChunksSkipped = Kamon.counter("num-raw-chunks-skipped").withoutTags()
```
alextheimer, line 9: FYI: this will double the count, since both the downsampler and exporter use `getChunkRangeIter`.

spark-jobs/src/main/scala/filodb/downsampler/chunk/BatchExporter.scala
```diff
@@ -0,0 +1,223 @@
...
+    val valueCol = columns.indexWhere(_.name == partition.schema.data.valueColName)
+    val timestampIter = getChunkColIter(partition, chunkReader, timestampCol, irowStart).asLongIt
+    columns(valueCol).columnType match {
+      case DoubleColumn =>
+        val valueIter = getChunkColIter(partition, chunkReader, valueCol, irowStart).asDoubleIt
+        (irowStart until irowEnd).iterator.map{ _ =>
+          (timestampIter.next, valueIter.next)
+        }
+      case colType =>
+        DownsamplerContext.dsLogger.warn(
```
alextheimer, line 113: I use this `DownsamplerContext` for logging in `BatchExporter`. Should it be something different?

spark-jobs/src/main/scala/filodb/downsampler/chunk/BatchExporter.scala
```diff
@@ -0,0 +1,223 @@
...
+   */
+  private def exportDataToRow(exportData: ExportRowData): Row = {
+    val dataSeq = new mutable.ArrayBuffer[Any](3 + downsamplerSettings.exportPathSpecPairs.size)
+    dataSeq.append(
+      exportData.partKeyString,
+      exportData.timestamp,
+      exportData.value
+    )
+    dataSeq.appendAll(exportData.partitionStrings)
+    Row.fromSeq(dataSeq)
```
alextheimer, line 168: I should add a `NOTE` here: immutable `Seq`s were a huge headache. When I'd just do something like:
```
Seq(val, val, val) ++ exportData.partitionStrings.toSeq
```
I'd get `ArrayIndexOutOfBounds` exceptions during the export, when `Row`s were compared against the `DataFrame` schema. This saves us two allocations, I guess 🤷

spark-jobs/src/main/scala/filodb/downsampler/chunk/BatchExporter.scala
```diff
@@ -0,0 +1,223 @@
...
+      StructField("labels", StringType),
+      StructField("timestamp", LongType),
+      StructField("value", DoubleType)
+    )
+    // append all partitioning columns as strings
+    fields.appendAll(downsamplerSettings.exportPathSpecPairs.map(f => StructField(f._1, StringType)))
+    StructType(fields)
+  }
+  val partitionByCols = {
+    val cols = new mutable.ArrayBuffer[String](3 + downsamplerSettings.exportPathSpecPairs.size)
```
alextheimer, line 65: TODO: extra +3 is unneeded here

spark-jobs/src/main/scala/filodb/downsampler/chunk/BatchExporter.scala
alextheimer: Note added.

spark-jobs/src/main/scala/filodb/downsampler/chunk/BatchExporter.scala
alextheimer: Removed.

sherali42 (approved):
thanks @alextheimer, looks good 👍

PR #1459, opened 2022-09-23
feat(core) Persistent index checkpointing

amolnayak311:
**Pull Request checklist**

**Current behavior :** 

Currently on each restart, downsample index is rebuilt from source. This is an issue especially with large downsample stores where index rebuilds can take tens several minutes if not hours. Index is not something we should rebuild from scratch and we can possibly look at updating just the delta from the last sync time. 

**New behavior :**
We introduce a new implementation of ``IndexMetadataStore`` which uses file system for storing the metadata about index sync. Additionally we introduce the concept of ``INDEX_GENERATION`` which is environment variable used to trigger rebuild of index. It is designed to ensure the index is rebuilt only when a new, higher ``INDEX_GENERATION`` is seen and not otherwise.

### DIFF
--- a/core/src/main/scala/filodb.core/downsample/DownsampleConfig.scala
+++ b/core/src/main/scala/filodb.core/downsample/DownsampleConfig.scala
@@ -34,6 +34,8 @@ final case class DownsampleConfig(config: Config) {
                 else Seq.empty
   val indexLocation = config.getOrElse[Option[String]]("index-location", None)
 
+  val maxRefreshHours = config.getOrElse("max-refresh-hours", Long.MaxValue)
+
   val enablePersistentIndexing = indexLocation.isDefined
 
   val indexMetastoreImplementation = if (config.hasPath("index-metastore-implementation")) {
--- a/core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
+++ b/core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
@@ -6,6 +6,7 @@ import java.util.concurrent.atomic.AtomicLong
 import scala.collection.mutable
 import scala.concurrent.{ExecutionContext, Future}
 import scala.concurrent.duration._
+import scala.util.{Failure, Success, Try}
 
 import com.typesafe.config.Config
 import com.typesafe.scalalogging.StrictLogging
@@ -73,11 +74,20 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
 
   private val indexMetadataStore : Option[IndexMetadataStore] = {
     downsampleConfig.indexMetastoreImplementation match {
-      case IndexMetastoreImplementation.NoImp => None
-      case IndexMetastoreImplementation.File => Some(
-        new FileCheckpointedIndexMetadataStore(downsampleConfig.indexLocation.get)
-      )
-      case IndexMetastoreImplementation.Ephemeral => Some(new EphemeralIndexMetadataStore())
+      case IndexMetastoreImplementation.NoImp       => None
+      case IndexMetastoreImplementation.File        =>
+        val expectedVersion = System.getenv(FileSystemBasedIndexMetadataStore.expectedGenerationEnv)
+        val expectedVersionOption = if (expectedVersion != null) {
+            logger.info("Expected version for the index from env is {}", expectedVersion)
+            Try(Integer.parseInt(expectedVersion)) match {
+              case Success(parsedVersion)   => Some(parsedVersion)
+              case Failure(e)               => logger.warn("Failed while parsing index generation", e)
+                                               None
+            }
+        } else None
+        Some(new FileSystemBasedIndexMetadataStore(downsampleConfig.indexLocation.get,
+                                                    expectedVersionOption, downsampleConfig.maxRefreshHours))
+      case IndexMetastoreImplementation.Ephemeral   => Some(new EphemeralIndexMetadataStore())
     }
   }
 
@@ -151,7 +161,8 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
   def recoverIndex(): Future[Long] = {
     if (downsampleConfig.enablePersistentIndexing) {
       partKeyIndex.getCurrentIndexState() match {
-        case (IndexState.Empty, _)                             =>
+        case (IndexState.Empty, _)   |
+             (IndexState.TriggerRebuild, _)                    =>
           logger.info("Found index state empty, bootstrapping downsample index")
           recoverIndexInternal(None)
         case (IndexState.Synced, checkpointMillis)             =>
@@ -170,29 +181,22 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
 
   private def recoverIndexInternal(checkpointMillis: Option[Long]): Future[Long] = {
     // By passing -1 for partId, numeric partId will not be persisted in the index
+    // Since we are recovering, always start from last synced time and update till current timestamp.
+    val endHour = hour()
+
     (checkpointMillis match {
       case Some(tsMillis)   => // We know we have to refresh only since this hour, it is possible we do not
                                 // find any data for this refresh as the index is already updated
-                                val (startHour, endHour) = (hour(tsMillis), hour())
+                                val startHour = hour(tsMillis) - 1
                                 // do not notify listener as the map operation will be updating the state
-                                indexRefresh(endHour, startHour, notifyListener = false)
+                                indexRefresh(endHour, startHour, periodicRefresh = false)
       case None             => // No checkpoint time found, start refresh from scratch
                                 indexBootstrapper
                                   .bootstrapIndexDownsample(
                                     partKeyIndex, shardNum, indexDataset, indexTtlMs)
     }).map { count =>
         logger.info(s"Bootstrapped index for dataset=$indexDataset shard=$shardNum with $count records")
-        // need to start recovering 6 hours prior to now since last index migration could have run 6 hours ago
-        // and we'd be missing entries that would be migrated in the last 6 hours.
-        // Hence indexUpdatedHour should be: currentHour - 6
-        val indexJobIntervalInHours = (downsampleStoreConfig.maxChunkTime.toMinutes + 59) / 60 // for ceil division
-        // the checkpoint updated should be at least what we had previously, subsequently, the refresh thread will
-        // take care of periodically updating this last synced value
-        val endHour = checkpointMillis.getOrElse(0L).max(hour() - indexJobIntervalInHours - 1)
         indexUpdatedHour.set(endHour)
-        // The time set here in synced state should match the time we set in indexUpdatedHour
-        // It is possible when we syn from a last timestamp, we dont update any data in index and as such, this
-        // synced time set is the pessimistic time to ensure no data sync to index is lost
         partKeyIndex.notifyLifecycleListener(IndexState.Synced, endHour * 3600 * 1000L)
         stats.shardTotalRecoveryTime.update(System.currentTimeMillis() - creationTime)
         startHousekeepingTask()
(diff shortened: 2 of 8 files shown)


core/src/main/scala/filodb.core/downsample/DownsampleConfig.scala
```diff
@@ -34,6 +34,8 @@ final case class DownsampleConfig(config: Config) {
                 else Seq.empty
   val indexLocation = config.getOrElse[Option[String]]("index-location", None)
 
+  val maxRefreshHours = config.getOrElse("max-refresh-hours", Long.MaxValue)
```
sandeep6189, line 37: NIT: Should we add comment here on how this new variable `maxRefreshHours` will be used ? ( purely a nitpick comment since I am not sure what is the commenting convention we follow for this codebase )

core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
```diff
@@ -73,11 +74,20 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
...
-      case IndexMetastoreImplementation.Ephemeral => Some(new EphemeralIndexMetadataStore())
+      case IndexMetastoreImplementation.NoImp       => None
+      case IndexMetastoreImplementation.File        =>
+        val expectedVersion = System.getenv(FileSystemBasedIndexMetadataStore.expectedGenerationEnv)
+        val expectedVersionOption = if (expectedVersion != null) {
+            logger.info("Expected version for the index from env is {}", expectedVersion)
+            Try(Integer.parseInt(expectedVersion)) match {
+              case Success(parsedVersion)   => Some(parsedVersion)
+              case Failure(e)               => logger.warn("Failed while parsing index generation", e)
+                                               None
```
sandeep6189, line 85: NIT: Wondering if we should move this to a separate function and add a UT for this ?

Quick Question: what is the difference between the expectedVersion vs expectedVersionOption ?

core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
```diff
@@ -151,7 +161,8 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
   def recoverIndex(): Future[Long] = {
     if (downsampleConfig.enablePersistentIndexing) {
       partKeyIndex.getCurrentIndexState() match {
-        case (IndexState.Empty, _)                             =>
+        case (IndexState.Empty, _)   |
+             (IndexState.TriggerRebuild, _)                    =>
           logger.info("Found index state empty, bootstrapping downsample index")
```
sandeep6189, line 166: NIT: Update logging here since we have a new case `IndexState.TriggerRebuild` ?

sandeep6189, line 166: NIT: Update logging here since we have a new case IndexState.TriggerRebuild ?

core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
```diff
@@ -170,29 +181,22 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
...
   private def recoverIndexInternal(checkpointMillis: Option[Long]): Future[Long] = {
     // By passing -1 for partId, numeric partId will not be persisted in the index
+    // Since we are recovering, always start from last synced time and update till current timestamp.
+    val endHour = hour()
+
     (checkpointMillis match {
       case Some(tsMillis)   => // We know we have to refresh only since this hour, it is possible we do not
                                 // find any data for this refresh as the index is already updated
-                                val (startHour, endHour) = (hour(tsMillis), hour())
+                                val startHour = hour(tsMillis) - 1
```
sandeep6189, line 190: Is the -1 here is added to avoid rounding errors and to make sure we are always looking back ?

sherali42 (approved):
looks good 👍

core/src/main/scala/filodb.core/memstore/IndexMetadataStore.scala
```diff
@@ -86,6 +91,249 @@ trait IndexMetadataStore {
...
+        if (magicByte != FileSystemBasedIndexMetadataStore.snapFileV1Magic) {
+          // if first byte is not the expected magicByte then the file is not as expected and possibly garbled
+          logger.warn(" Snap {} exist and possibly corrupt, this will trigger a rebuild", snapFile.toString)
+          Failure(new IllegalStateException("Invalid magic bytes in snap file"))
+        } else {
+          // If the first byte is magicByte read just the next 8 bytes
+          val bytes = Array[Byte](0, 0, 0, 0, 0, 0, 0, 0)
+          if (fin.read(bytes) != 8) {
+            // Not able to read 8 bytes, the file is truncated and we will trigger a rebuild
+            logger.warn("Snap file {} with right magic bytes but less than 8 for time," +
```
sandeep6189, line 252: NIT: wondering if we should use logger.error here ? (since currentState throws exception, which means the initState could also throw exception, but I am not sure if that will lead to some fatal error)

core/src/main/scala/filodb.core/memstore/IndexMetadataStore.scala
```diff
@@ -86,6 +91,249 @@ trait IndexMetadataStore {
...
+            val lastSyncTime = ByteBuffer.wrap(bytes).getLong()
+            Success(lastSyncTime)
+          }
+        }
+   }
+  /**
+   *  Updates the state of the index, the snap file will be updated with the correct timestamp in case the state
+   *  is Synced, current implementation of updateState ignores other index states. Also when the index is updated
+   *  and synced, it is important to ensure the gen file is updated with the current generation. This guarantees that
+   *  next time the on bootstrap, the initState should return Synced and not TriggerRebuild
```
sandeep6189, line 265: NIT: on the

core/src/main/scala/filodb.core/memstore/IndexMetadataStore.scala
```diff
@@ -86,6 +91,249 @@ trait IndexMetadataStore {
...
+            Failure(new IllegalStateException("Snap file truncated"))
+          } else {
+            val lastSyncTime = ByteBuffer.wrap(bytes).getLong()
+            Success(lastSyncTime)
+          }
+        }
+   }
+  /**
+   *  Updates the state of the index, the snap file will be updated with the correct timestamp in case the state
+   *  is Synced, current implementation of updateState ignores other index states. Also when the index is updated
```
sandeep6189, line 263: updateState does some checks/deletes etc on state TriggerBuild and Empty ?

core/src/main/scala/filodb.core/memstore/IndexMetadataStore.scala
```diff
@@ -86,6 +91,249 @@ trait IndexMetadataStore {
...
+                                      expectedGeneration.map{
+                                        generation =>
+                                          Using(new FileOutputStream(genFile)) { genFos =>
+                                            logger.info("Updating genFile " +
+                                              "of dataset={}, shard={} with generation={}",
+                                              datasetRef.toString, shard, generation)
+                                            genFos.write(genFileV1Magic)
+                                            val buff = ByteBuffer.allocate(4)
+                                            buff.putInt(generation)
+                                            genFos.write(buff.array())
```
sandeep6189, line 302: separate function and UT for genFile write  ? ( UT would call this block and checks if it is written to file correctly )

core/src/main/scala/filodb.core/memstore/IndexMetadataStore.scala
```diff
@@ -86,6 +91,249 @@ trait IndexMetadataStore {
...
+   */
+  // scalastyle:off method.length
+  override def updateState(datasetRef: DatasetRef, shard: Int, state: IndexState.Value, time: Long): Unit = {
+    state match {
+      case IndexState.Synced           =>  // Only handle cases where state is Synced
+                                    val snapFile = new File(snapFilePathFormat.format(datasetRef.dataset, shard))
+                                    val genFile = new File(genFilePathFormat.format(datasetRef.dataset, shard))
+                                    Using(new FileOutputStream(snapFile)){
+                                      // First write the last synced time to the snap file
+                                      snapFos =>
```
sandeep6189, line 280: separate function and UT for snapshot file write  ? ( UT would call this block and checks if it is written to file correctly )

core/src/main/scala/filodb.core/downsample/DownsampleConfig.scala
sandeep6189: understood the usage later in the PR. thank you :)

core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
amolnayak311: Yes, its to account for cases where the last refresh happened exactly at the same time we snapped the time last time.

core/src/main/scala/filodb.core/memstore/IndexBootstrapper.scala
```diff
@@ -80,7 +81,7 @@ class IndexBootstrapper(colStore: ColumnStore) {
         // Note that we do not set an end time for the Synced here, instead
         // we will do it from DownsampleTimeSeriesShard
         index.refreshReadersBlocking()
-        recoverIndexLatency.update(System.currentTimeMillis() - start)
+        recoverIndexLatency.update(System.currentTimeMillis() - startCheckpoint)
         count
```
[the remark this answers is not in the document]
amolnayak311: Its the number of part keys that we get in ``colStore.scanPartKeys`` call.

core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
amolnayak311: Good point, let me add one

core/src/main/scala/filodb.core/memstore/IndexMetadataStore.scala
```diff
@@ -86,6 +91,249 @@ trait IndexMetadataStore {
...
+              if (expectedGen > generation) {
+                logger.info("Gen file {} has generation {} and expected generation is {}, triggering rebuild",
+                  genFile.toString, generation, expectedGen)
+                (IndexState.TriggerRebuild, None)
+              } else if (shouldRebuildIndexForOldSnap(datasetRef, shard, snapFile)) {
+                logger.info("lastSnapTime is more than {} hours from now, " +
+                  "for dataset={} and shard={}. Triggering index rebuild instead of syncing the delta",
+                  maxRefreshHours, datasetRef.toString, shard)
+                (IndexState.TriggerRebuild, None)
+              } else
```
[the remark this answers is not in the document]
amolnayak311: I agree its bit complicated. Let me leave a note and no touch this for now if that's ok.

core/src/main/scala/filodb.core/memstore/IndexMetadataStore.scala
```diff
@@ -86,6 +91,249 @@ trait IndexMetadataStore {
...
+                currentState(datasetRef, shard)
+            }
+          }
+        }.get
+      }
+    }).getOrElse(
```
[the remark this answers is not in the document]
amolnayak311: ``getOrElse`` is on the ``Option`` ``expectedGeneration`` the part in `` map`` will be invoked only when the value is `` Some(..)`` and we will get the tuple we expect, if this ``expectedGeneration`` is ``None`` the alternate code will be invoked to return the tuple we expect.

core/src/main/scala/filodb.core/memstore/IndexMetadataStore.scala
amolnayak311: Changing the log level would not cause any issue. I just logged a warn since I am returning a failure anyways,

core/src/main/scala/filodb.core/memstore/IndexMetadataStore.scala
amolnayak311: Good catch.

core/src/main/scala/filodb.core/memstore/IndexMetadataStore.scala
amolnayak311: We have unit tests in ``FileSystemBasedIndexMetadataStoreSpec`` testing the expected functionality at higher level than reading the actual files written.

core/src/main/scala/filodb.core/memstore/IndexMetadataStore.scala
amolnayak311: Same as above

PR #1462, opened 2022-10-04
fix(coord) Stateful shard assignment strategy should fail if expected hostname resolution fails

amolnayak311:
**Pull Request checklist**

**Current behavior :** 
 ``K8sStatefulSetShardAssignmentStrategy`` currently falls back to default strategy if the host name resolution does not work. 

**New behavior :**
The current default strategy is not correct and silently will reassign shards to random pods defeating the purpose of having prebuilt index. We need to retry with backoff and see if we can assign shards meant for that pod and fail if its not possible to do so. 

**BREAKING CHANGES**
Previously working  ``K8sStatefulSetShardAssignmentStrategy`` configured with deployment (which isn't the right configuration) will now fail instead of silently falling back to default strategy.

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/ShardAssignmentStrategy.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/ShardAssignmentStrategy.scala
@@ -46,7 +46,7 @@ trait ShardAssignmentStrategy {
  * The implementation assumes the number of shards is a multiple of number of nodes and there are no
  * uneven assignments of shards to nodes
  */
-class K8sStatefulSetShardAssignmentStrategy(val maxAssignmentAttempts: Int)
+class K8sStatefulSetShardAssignmentStrategy(val maxAssignmentAttempts: Int = 8)
       extends ShardAssignmentStrategy with StrictLogging {
 
   private val pat = "-\\d+$".r
--- a/coordinator/src/main/scala/filodb.coordinator/ShardAssignmentStrategy.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/ShardAssignmentStrategy.scala
@@ -7,6 +7,10 @@ import com.typesafe.scalalogging.StrictLogging
 
 import filodb.coordinator.NodeClusterActor.DatasetResourceSpec
 import filodb.core.DatasetRef
+import filodb.memory.data.Shutdown
+
+
+
 
 trait ShardAssignmentStrategy {
 
@@ -42,7 +46,8 @@ trait ShardAssignmentStrategy {
  * The implementation assumes the number of shards is a multiple of number of nodes and there are no
  * uneven assignments of shards to nodes
  */
-class K8sStatefulSetShardAssignmentStrategy extends ShardAssignmentStrategy with StrictLogging {
+class K8sStatefulSetShardAssignmentStrategy(val maxAssignmentAttempts: Int)
+      extends ShardAssignmentStrategy with StrictLogging {
 
   private val pat = "-\\d+$".r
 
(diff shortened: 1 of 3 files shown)


coordinator/src/main/scala/filodb.coordinator/ShardAssignmentStrategy.scala
```diff
@@ -88,31 +102,53 @@ class K8sStatefulSetShardAssignmentStrategy extends ShardAssignmentStrategy with
...
+
+  private def retryWithBackoff[T](coord: ActorRef, dataset: DatasetRef, resources: DatasetResourceSpec,
+                               mapper: ShardMapper, hostName: String,
+                               currentAttempt: Int, maxAssignmentAttempts: Int)
+                              (f: (ActorRef, DatasetRef, DatasetResourceSpec, ShardMapper, Int, Int) => T): T= {
+    // Host name does not have the ordinal at the end like a stateful set needs to have, retry we we have
+    // retries left
+    if (currentAttempt > maxAssignmentAttempts) {
+      logger.error(s"Unable to resolve host names after $currentAttempt attempts, terminating now")
+      sys.exit(1)
```
vishramachandran, line 127: Use Shutdown companion class to perform this. There are common actions there.

coordinator/src/main/scala/filodb.coordinator/ShardAssignmentStrategy.scala
```diff
@@ -88,31 +102,53 @@ class K8sStatefulSetShardAssignmentStrategy extends ShardAssignmentStrategy with
...
+    } else {
+      val retryAfter = 1 << currentAttempt
```
vishramachandran, line 129: Doesn't linear backoff work better here ? DNS lookup resulting in a denial of service is less likely. We want the issue to resolve asap.

coordinator/src/main/scala/filodb.coordinator/ShardAssignmentStrategy.scala
```diff
@@ -53,16 +57,26 @@ class K8sStatefulSetShardAssignmentStrategy extends ShardAssignmentStrategy with
...
-     getOrdinalFromActorRef(coord) match {
-        case Some((hostName, ordinal)) =>
+                       dataset: DatasetRef,
+                       resources: DatasetResourceSpec,
+                       mapper: ShardMapper): Seq[Int] =
+    shardAssignmentsWithRetry(coord, dataset, resources, mapper, currentAttempt = 1, maxAssignmentAttempts)
+
+
+  // scalastyle:off method.length
+  private def shardAssignmentsWithRetry(coord: ActorRef,
```
vishramachandran, line 72: Can you refactor the ordinal discovery with retry into a separate method and reuse from `FiloDbClusterDiscovery` class as well ? This exact capability will be needed there as well.

vishramachandran (changes requested):
Some small changes requested. Thank you.

coordinator/src/main/scala/filodb.coordinator/ShardAssignmentStrategy.scala
amolnayak311: The reason I kept exponential backoff is we don't need to configure the fixed interval. It will try in 2, 4, 8, 16, 32, 64.. seconds and only number of attempts is to be configured. What can be done is cap this retry interval to a fixed hardcoded number like 60 seconds.

coordinator/src/main/scala/filodb.coordinator/ShardAssignmentStrategy.scala
amolnayak311: I don't see a clear way to reuse for following reasons. 

1. ``FiloDbClusterDiscovery`` doesn't perform host lookup by actor name
2. Only place we of a host lookup in  ``FiloDbClusterDiscovery`` is here, which looks up local host name and not a remote actor hostname. 

```  lazy val ordinalOfLocalhost: Int = {
    if (settings.localhostOrdinal.isDefined) settings.localhostOrdinal.get
    else {
      val fullHostname = InetAddress.getLocalHost.getHostName
      val hostName = if (fullHostname.contains(".")) fullHostname.substring(0, fullHostname.indexOf('.'))
      else fullHostname

      val pat = "-\\d+$".r
      // extract ordinal at the end from hostname such as filodb-host-0, filodb-host-10
      val ordinal = pat.findFirstIn(hostName).map(ordinal => -Integer.parseInt(ordinal))
        .getOrElse(throw new IllegalArgumentException(s"hostname $hostName does not appear like a stateful sets host"))
      ordinal
    }
  }```

``ShardAssignmentStrategy`` has predefined methods used by our existing Clustering based up Akka and the implementation is around these methods.

vishramachandran (approved):

PR #1463, opened 2022-10-05
fix(query) Configurable record container size in Metadata and PartKeysExec plans

amolnayak311:
Time series with label/values exceeding the default 4K container limit fail. The PR introduces a configurable container size based on ``ExecPlan`` providing flexibility in changing the container size instead of globally bumping the limits up to a higher value.

**Pull Request checklist**

**Current behavior :** 

Creating a ``SerializedRangeVector`` for a timeseries with label name value pair length exceeding 4K fails. 

**New behavior :**

Creating a ``SerializedRangeVector`` for a timeseries with label name value pair length exceeding 8K will fail.  But the size of the container now can be tweaked at ``ExecPlan`` level instead of a globally set value.

### DIFF
--- a/query/src/main/scala/filodb/query/exec/ExecPlan.scala
+++ b/query/src/main/scala/filodb/query/exec/ExecPlan.scala
@@ -53,12 +53,12 @@ trait ExecPlan extends QueryCommand {
   def enforceSampleLimit: Boolean = true
 
   /**
-   * Default max size of the containers created for the ExecPlan. Specific ExecPlans may choose to override and
+   * Default max size of the record containers created for the ExecPlan. Specific ExecPlans may choose to override and
    * use different values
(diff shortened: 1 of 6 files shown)


query/src/main/scala/filodb/query/exec/ExecPlan.scala
```diff
@@ -52,6 +52,14 @@ trait ExecPlan extends QueryCommand {
     */
   def enforceSampleLimit: Boolean = true
 
+  /**
+   * Default max size of the containers created for the ExecPlan. Specific ExecPlans may choose to override and
```
vishramachandran, line 56: name as RecordContainers when writing query results to the wire

vishramachandran (approved):

PR #1465, opened 2022-10-14
feat(downsample): exclude labels from export

alextheimer:
**Pull Request checklist**

## Excluding Labels

Adds a configurable `drop-labels` sequence to export rules. If an export rule is applied to a row, any labels specified in `drop-labels` will be excluded from the exported data.

This PR also changes how rules are defined. Keys now apply to groups of rules; the group is iterated until an applicable rule is found. The benefit of this setup is that rule "modifiers" (i.e. `drop-labels`) can be specified for different sets of time-series.

```
      # Each row's labels are compared against all rule-group keys. If a match is found,
      #   the row's labels are compared *sequentially* against each of the group's rules until
      #   a rule meets both of the following criteria:
      #       (1) No "block" filters are matched.
      #       (2) The "allow" filters are either matched or empty.
      # This rule will be applied to the row. This means that row data is/isn't exported
      #   depending on the matched filters, and any other rule effects are also applied (such as dropped labels).
      # If a matching rule (or group) are not found, the data will not be exported.
      groups = [
        {
          key = ["ws-foo"]
          rules = [
            {
              allow-filters = [
                ["_ns_=\"my_ns\"", "bar!=\"baz\""],
                ["_ns_=\"my_other_ns\""]
              ]
              block-filters = [
                ["_metric_=~\".*bad_suffix\""]
              ]
              drop-labels = ["label1", "other_label"]
            }
          ]
        }
      ]
```

## Other Improvements
- Configurable output file format.
- JSON label-value pairs in the output.
- Various performance optimizations.
- Finds a configured factory class via reflection to make the SparkSession.

### DIFF
--- a/spark-jobs/src/main/scala/filodb/downsampler/chunk/BatchExporter.scala
+++ b/spark-jobs/src/main/scala/filodb/downsampler/chunk/BatchExporter.scala
@@ -1,8 +1,5 @@
 package filodb.downsampler.chunk
 
-import com.fasterxml.jackson.databind.ObjectMapper
-import com.fasterxml.jackson.module.scala.DefaultScalaModule
-
 import java.security.MessageDigest
 import java.text.SimpleDateFormat
 import java.util.Date
@@ -18,7 +15,7 @@ import filodb.core.query.ColumnFilter
 import filodb.core.store.{ChunkSetInfoReader, ReadablePartition}
 import filodb.downsampler.DownsamplerContext
 import filodb.downsampler.Utils._
-import filodb.downsampler.chunk.BatchExporter.{DATE_REGEX_MATCHER, LABEL_REGEX_MATCHER, MAP_TO_JSON_SER}
+import filodb.downsampler.chunk.BatchExporter.{DATE_REGEX_MATCHER, LABEL_REGEX_MATCHER}
 import filodb.memory.format.{TypedIterator, UnsafeUtils}
 
 case class ExportRule(allowFilterGroups: Seq[Seq[ColumnFilter]],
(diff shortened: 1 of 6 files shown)


spark-jobs/src/main/scala/filodb/downsampler/chunk/BatchExporter.scala
```diff
@@ -187,21 +189,24 @@ case class BatchExporter(downsamplerSettings: DownsamplerSettings) {
         })}
   }
 
-  def shouldExport(partKeyMap: Map[String, String]): Boolean = {
-    rules.find { rule =>
-      // find the rule with a key that matches these label-values
+  def getRuleIfShouldExport(partKeyMap: Map[String, String]): Option[ExportRule] = {
+    keyToRules.find { case (key, rules) =>
```
amolnayak311, line 193: Is it possible that the part key matches multiple ``ExportRule``?

spark-jobs/src/main/scala/filodb/downsampler/chunk/BatchExporter.scala
```diff
@@ -211,9 +216,15 @@ case class BatchExporter(downsamplerSettings: DownsamplerSettings) {
    */
   private def getExportDatas(readablePartition: ReadablePartition,
                              partKeyMap: Map[String, String],
+                             rule: ExportRule,
                              userStartTime: Long,
                              userEndTime: Long): Iterator[ExportRowData] = {
-    val partKeyString = partKeyMap.map(pair => s"${pair._1}=${pair._2}").toSeq.sorted.mkString(",")
+    // Drop unwanted labels from the exported "labels" column.
+    val partKeyString = partKeyMap.filterKeys { label =>
```
amolnayak311, line 223: If multiple export rules are matched,  ``getRuleIfShouldExport`` returns the first match the drop label for the matched rule would be used.

alextheimer: [see other comment]

amolnayak311 (approved):

PR #1490, opened 2022-12-09
fix(query): override RemoteExec::doExecute instead of execute()

alextheimer:
**Pull Request checklist**

Currently, `RemoteExec` plans override the usual `execute` logic; this means that common tasks like `RangeVectorTransformer` application will be missed if they are not explicitly performed in the overridden `execute`.

This PR moves the overridden logic into `doExecute`, so all common `execute` tasks will be performed automatically.

### DIFF
--- a/core/src/main/resources/filodb-defaults.conf
+++ b/core/src/main/resources/filodb-defaults.conf
@@ -284,6 +284,9 @@ filodb {
     # Choices are "legacy", "antlr", and "shadow". Shadow mode uses legacy but also checks antlr for errors.
     parser = "antlr"
 
+    # Executes RemoteExec plans without invoking super.execute().
+    enable-legacy-remote-execute = false
+
     routing {
       # not currently used
     }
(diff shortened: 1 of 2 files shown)


query/src/test/scala/filodb/query/exec/RemoteMetadataExecSpec.scala
```diff
@@ -167,16 +167,10 @@ class RemoteMetadataExecSpec extends AnyFunSpec with Matchers with ScalaFutures
 
     val resp = exec.execute(memStore, querySession).runToFuture.futureValue
     val result = (resp: @unchecked) match {
-      case QueryResult(id, _, response, _, _, _) => {
-        val rv = response(0)
-        rv.rows.size shouldEqual 0
-        rv.rows.map { row =>
-          val record = row.asInstanceOf[BinaryRecordRowReader]
-          rv.asInstanceOf[SerializedRangeVector].schema.toStringPairs(record.recordBase, record.recordOffset)
-        }
```
alextheimer, line 176: Sketched out by this-- @sherali42 , do we need the result to contain an empty RV?

amolnayak311:
What requires this change? Currently since execute is overridden, no RVTs are applied. How do we ensure now that the transformation is not applied twice? Once on remote side, after the response from remote partition is received?

PR #1496, opened 2022-12-18
feat(coord): Circuit Breaker to protect query engine from overload

vishramachandran:
**Pull Request checklist**

## Motivation
Reject queries before they are accepted into the queue if the system is overloaded thereby reducing latencies for queries that come later, and increasing overall throughput and availability of system. Reduce query timeouts and improve client experience.

## Library used:
https://monix.io/docs/current/catnap/circuit-breaker.html

## Circuit Breaker Configuration

```
    circuit-breaker {
      # query circuit breaker enabled or not
      enabled = false
      # open the circuit when number of failures reaches this count
      open-when-num-failures = 5
      # move to half-open after this time
      reset-timeout = 30 seconds
      # backoff-factor when trying to half-open
      exp-backoff-factor = 1.5
      # maximum time in open state
      max-reset-timeout = 5 minutes
    }
```

## Perf Test Setup

* 1 FiloDB node with 8 shards, 4GB Heap, on M1-Max dev machine
* Ingestion completed. Ingestion not running during query load.
* 5k cardinality 1-to-1 binary join + aggregate query for 3 hours with step of 1min. sum(foo+bar) type query.
* 20 queries per second
* 3 minutes long
* Queries hit FiloDB directly. In a deployment setup where FiloDB is not responsible for serializing results to json, performance will be better.
* Query result streaming not enabled.

## Without Circuit Breaker
without1
without2
without3
without4

## With Circuit Breaker
with1
with2
with3
with4

## Analysis

* Mean, 50th and 75%ile latencies are much better with circuit breaker since queries dont sit in the queue for long time. Also, clients don’t wait for a long time eventually getting a timeout.
* Total number of OK queries are lesser when Circuit breaker is on (with high latency though). This is because the breaker rejects very early a small percentage of queries that may have OKed with high latency.
* How does FiloDB OK more queries without the circuit breaker - this may be surprising? There is a mechanism to check for query timeout before spending any time on the query, and at regular periods during query execution. This mechanism prevents the engine from spending time on wasteful queries that may have already timed out.
* Without circuit breaker, many requests don’t get a response and timeout. At these times the node is “unavailable”.
* With circuit breaker, we didnt see connection issues on the client. The system was available.

## Conclusion

* If the goal is to reduce latencies at the cost of possibly rejecting/shedding queries (that may have had a small chance of succeeding with high latency) early on, then enabling circuit breaker is a good idea.
* Circuit Breaker is needed for High-Availability. It is also important to protect the system from thundering herd type incessant query workloads.

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
@@ -170,9 +170,9 @@ final class QueryActor(memStore: TimeSeriesStore,
             .onErrorHandle { t =>
               StreamQueryError(q.queryContext.queryId, querySession.queryStats, t)
             }.map { resp =>
-            // Avoiding the assert when the InProcessPlanDispatcher is used. As it runs
-            // the query on the current/Actor thread instead of the scheduler
-            if (!q.dispatcher.isInstanceOf[InProcessPlanDispatcher]) {
+              // Avoiding the assert when the InProcessPlanDispatcher is used. As it runs
+              // the query on the current/Actor thread instead of the scheduler
+              if (!q.dispatcher.isInstanceOf[InProcessPlanDispatcher]) {
                 FiloSchedulers.assertThreadName(QuerySchedName)
               }
               replyTo ! resp
@@ -181,6 +181,7 @@ final class QueryActor(memStore: TimeSeriesStore,
                   logQueryErrors(e.t)
                   queryErrors.increment()
                   queryExecuteSpan.fail(e.t.getMessage)
+                  // rethrow so circuit beaker can block queries when there is a surge of such exceptions
                   if (e.t.isInstanceOf[QueryTimeoutException] || e.t.isInstanceOf[AskTimeoutException]) throw e.t
                 case _ =>
               }
@@ -204,6 +205,7 @@ final class QueryActor(memStore: TimeSeriesStore,
                   logQueryErrors(e.t)
                   queryErrors.increment()
                   queryExecuteSpan.fail(e.t.getMessage)
+                  // rethrow so circuit beaker can block queries when there is a surge of such exceptions
                   if (e.t.isInstanceOf[QueryTimeoutException] || e.t.isInstanceOf[AskTimeoutException]) throw e.t
                 case _ =>
               }
--- a/coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
@@ -86,22 +86,25 @@ final class QueryActor(memStore: TimeSeriesStore,
                                               functionalSpreadProvider)
   val queryScheduler = createInstrumentedQueryScheduler()
 
-  private val numRejectedPlans = Kamon.counter("circuit-breaker-rejected-plans").withTags(TagSet.from(tags))
-  private val circuitBreakerNumFailures = config.getInt("circuit-breaker.num-failures")
-  private val circuitBreakerResetTimeout = config.as[FiniteDuration]("circuit-breaker.reset-timeout")
-  private val circuitBreakerExpBackOffFactor = config.getDouble("circuit-breaker.exp-backoff-factor")
-  private val circuitBreakerMaxTimeout = config.as[FiniteDuration]("circuit-breaker.max-reset-timeout")
-  private val circuitBreaker = CircuitBreaker[Task].unsafe(circuitBreakerNumFailures,
-    circuitBreakerResetTimeout, circuitBreakerExpBackOffFactor, circuitBreakerMaxTimeout,
-    onRejected = Task.eval(numRejectedPlans.increment()))
-
-
   private val tags = Map("dataset" -> dsRef.toString)
   private val lpRequests = Kamon.counter("queryactor-logicalPlan-requests").withTags(TagSet.from(tags))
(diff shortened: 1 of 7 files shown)


core/src/main/resources/filodb-defaults.conf
```diff
@@ -613,7 +626,7 @@ gateway {
 
   # Maximum size of queue for each shard's records.  Must be a power of 2.  When the max queue size is reached
   # then the connection will stop accepting incoming records.
-  max-queue-size = 16384
+  max-queue-size = 262144
```
sandeep6189, line 629: quick question for my understanding: why are we increasing the max-queue-size 16 times at the gateway ?

sandeep6189 (commented):
Should we add a UT to test if Circuit breaker is working as expected ?

core/src/main/resources/filodb-defaults.conf
vishramachandran: My M1-mac was filling the queues too quickly before the Kafka-writer could write them to the disk. Hence needed a larger queue.

coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
```diff
@@ -87,9 +89,22 @@ final class QueryActor(memStore: TimeSeriesStore,
   private val tags = Map("dataset" -> dsRef.toString)
   private val lpRequests = Kamon.counter("queryactor-logicalPlan-requests").withTags(TagSet.from(tags))
   private val epRequests = Kamon.counter("queryactor-execplan-requests").withTags(TagSet.from(tags))
-  private val resultVectors = Kamon.histogram("queryactor-result-num-rvs").withTags(TagSet.from(tags))
```
alextheimer, line 90: Is this intentional?

coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
```diff
@@ -155,46 +170,62 @@ final class QueryActor(memStore: TimeSeriesStore,
...
                 FiloSchedulers.assertThreadName(QuerySchedName)
               }
               replyTo ! res
               res match {
-                case QueryResult(_, _, vectors, _, _, _) => resultVectors.record(vectors.length)
                 case e: QueryError =>
                   logQueryErrors(e.t)
                   queryErrors.increment()
                   queryExecuteSpan.fail(e.t.getMessage)
+                  if (e.t.isInstanceOf[QueryTimeoutException] || e.t.isInstanceOf[AskTimeoutException]) throw e.t
```
alextheimer, line 207: Can this be a matcher, instead? Or are these the only two we will ever want to throw?

alextheimer (approved):
Left a couple comments, but in-general: lgtm 👍

coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
```diff
@@ -155,46 +170,62 @@ final class QueryActor(memStore: TimeSeriesStore,
...
+                case _ =>
```
alextheimer, line 208: Can this match `QueryResult`, instead?

coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
vishramachandran: yes - intentional. Not being plotted in dashboard. Not helpful with detailed debugging. Result size already tracked with other metrics.

coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
vishramachandran: Only these exceptions for now. I'll leave it as is and we can refactor if this list grows.

sherali42 (approved):
👍 thanks for this PR @vishramachandran

coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
```diff
@@ -203,11 +236,11 @@ final class QueryActor(memStore: TimeSeriesStore,
       t match {
         case _: BadQueryException => // dont log user errors
         case _: AskTimeoutException => // dont log ask timeouts. useless - let it simply flow up
-        case e: QueryTimeoutException => // log just message, no need for stacktrace
-          logger.error(s"queryId: ${q.queryContext.queryId} QueryTimeoutException: " +
-            s"${q.queryContext.origQueryParams} ${e.getMessage}")
+        case _: QueryTimeoutException | _: ExecutionRejectedException => // log just message, no need for stacktrace
+          logger.error(s"Query Error ${t.getClass.getSimpleName} queryId=${q.queryContext.queryId} " +
+            s"${q.queryContext.origQueryParams} ${t.getMessage}")
```
amolnayak311, line 241: This is  ``ExecutionRejectedException``

```
class ExecutionRejectedException(val message: String, cause: Throwable)
  extends RuntimeException(message, cause) with Serializable {

  def this(message: String) = this(message, null)
  def this(cause: Throwable) = this(null, cause)
}
```

If it is constructed using just root cause, we will miss the exception in logs as the message null and root cause is not logged. Im not sure where and how it might be instantiated this way.

amolnayak311 (approved):
LGTM left a couple of minor comments.

coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
vishramachandran: ExecutionRejectedException typically does not have an inner cause. This is what I have seen. Will skip for now - can add later if needed.

PR #1501, opened 2023-01-26
fix(core) Conform to Lucene regex by removing leading ^ and ending $.

yu-shipit:
Lucene regex does not support anchor operator such as ^ and $. https://www.elastic.co/guide/en/elasticsearch/reference/current/regexp-syntax.html#regexp-unsupported-operators Remove those characters from users' regex so that Lucene can handle it.

**Pull Request checklist**

**New behavior :**

**BREAKING CHANGES**

Breaking changes may include:

**Other information**:

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
+++ b/core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
@@ -5,6 +5,7 @@ import java.io.IOException
 import java.nio.charset.StandardCharsets
 import java.util
 import java.util.{Base64, PriorityQueue}
+import java.util.regex.Pattern
 
 import scala.collection.JavaConverters._
 import scala.collection.immutable.HashSet
@@ -88,6 +89,29 @@ object PartKeyLuceneIndex {
     md.update(bytes, offset, length)
     val strDigest = Base64.getEncoder.encodeToString(md.digest())
     strDigest
+  }
+
+   private def removeRegexTailDollarSign(regex: String): String = {
+    // avoid unnecessary calculation when $ is not present at the end of the regex.
+    if (regex.nonEmpty && regex.last == '$' && Pattern.matches("""^(|.*[^\\])(\\\\)*\$$""", regex)) {
+      // (|.*[^\\]) means either empty or a sequence end with a character not \.
+      // (\\\\)* means any number of \\.
+      // remove the last $ if it is not \$.
+      // $ at locations other than the end will not be removed.
+      regex.substring(0, regex.length - 1)
+    } else {
+      regex
+    }
+  }
+
+  /**
+   * Remove leading anchor &#94; and ending anchor $.
+   *
+   * @param regex the orignal regex string.
+   * @return the regex string without anchors.
+   */
+  def removeRegexAnchors(regex: String): String = {
+    removeRegexTailDollarSign(regex.stripPrefix("^"))
   }
 }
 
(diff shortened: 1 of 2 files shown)


core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -792,13 +792,22 @@ class PartKeyLuceneIndex(ref: DatasetRef,
...
   }
 
+
+  /**
+   * Remove leading anchor &#94; and ending anchor $.
+   * @param regex the orignal regex string.
+   * @return the regex string without anchors.
+   */
+  private def removeRegexAnchors(regex : String) : String = {
+    regex.stripPrefix("^").stripSuffix("$")
```
alextheimer, line 802: Should we be careful about escape characters? I forget the exact methods, but maybe something like:
```
regex.stripPrefix("^")
// match (possibly zero) escaped backslashes, then a final "$" anchor; remove the anchor
.strip(matchRegex="(\\\\)*(\$)", replaceSecondGroupWith="")
```

(Let me know if I'm overthinking things and there's some reason we shouldn't do this.)

yu-shipit: Thank you. 
fixed the code. So it can handle case like """$"""  """\$"""  """\\$"""
The corresponding regex without anchor $ are """"""    """\$"""  """\\"""

yu-shipit: tested against Prometheus. '\$' is a dollar sign, whereas '$' is not.

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -792,13 +792,32 @@ class PartKeyLuceneIndex(ref: DatasetRef,
     logger.info(s"Refreshed index searchers to make reads consistent for dataset=$ref shard=$shardNum")
   }
 
+
+  private def removeRegexTailDollarSign(regex : String) : String = {
+    // avoid unnecessary calculation when $ is not present at the end of the regex.
+    if (regex.nonEmpty && regex.last == '$' && !regex.dropRight(1).foldLeft(false)((v, c) => !v && c == '\\')) {
```
alextheimer, line 798: This takes a second to understand-- if you could match with a regex, it would be more straightforward, I think.

core/src/test/scala/filodb.core/memstore/PartKeyLuceneIndexSpec.scala
```diff
@@ -786,6 +786,28 @@ class PartKeyLuceneIndexSpec extends AnyFunSpec with Matchers with BeforeAndAfte
     keyIndex.singlePartKeyFromFilters(Seq(filter2), 4, 10) shouldBe None
   }
 
+  it("should get a single match for part keys by a regex filter") {
```
alextheimer, line 789: Could you add a couple tests that just call `removeRegexAnchors` to confirm visually that things like:
```
^$foo.*$\$$
foo.*\\\$
foo.*$\\\\$
```
etc. are updated as expected?

alextheimer (approved):
Approved (with a couple requests). Thanks @yu-shipit for the PR!

core/src/test/scala/filodb.core/memstore/PartKeyLuceneIndexSpec.scala
yu-shipit: Added a test for them

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
yu-shipit: updated the code with a regex version. Hopefully it is easier to understand.

PR #1506, opened 2023-01-30
fix(query): ignore isCumulative field when comparing ResultSchemas

sherali42:
**Pull Request checklist**

When comparing ResultSchemas, `isCumulative` field will be ignored. This field is used only at the leaf node to pick the right `rate` function depending on whether the ColumnType is cumulative type or not.

### DIFF
--- a/core/src/main/scala/filodb.core/query/ResultTypes.scala
+++ b/core/src/main/scala/filodb.core/query/ResultTypes.scala
@@ -1,5 +1,7 @@
 package filodb.core.query
 
+import java.util.Objects
+
 import scala.reflect.runtime.universe._
 
 import com.typesafe.scalalogging.StrictLogging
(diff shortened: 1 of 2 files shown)


core/src/main/scala/filodb.core/query/ResultTypes.scala
```diff
@@ -22,9 +24,16 @@ final case class PartitionInfo(schema: RecordSchema, base: Array[Byte], offset:
...
 
 /**
- * Describes column/field name and type
+ * Describes column/field name and type.
+ * isCumulative is not considered for equality/Hashcode.
  */
-final case class ColumnInfo(name: String, colType: Column.ColumnType, isCumulative: Boolean = true)
+final case class ColumnInfo(name: String, colType: Column.ColumnType, isCumulative: Boolean = true) {
+  override def equals(obj: Any): Boolean = obj match {
+    case ColumnInfo(n: String, ct: Column.ColumnType, _) => n == name && ct == colType
```
amolnayak311, line 32: singe its java String, lets user ``equals`` instead of ``==`` other wise lgtm

sandeep6189: Quick Question: Does java's String.equals work on null String objects ? In C#, String.equals failed on null strings ( NullReferenceException )

sandeep6189: (Super new to Java runtime, so still learning these things )

sherali42: @amolnayak311 it does invoke equals on String. unlike Java it doesn't do ref comparison.

sandeep6189 (approved):

PR #1508, opened 2023-01-31
fix(parser): add support for zero-duration time range units

alextheimer:
**Pull Request checklist**

Currently, queries with bracket-notation range durations such as:
```
foo{label1="value1"}[2d0h15m]
```
will fail to parse because
- `prometheus.ast.Duration` was used to find the millisecond-duration of each unit-value pair.
- `Duration`s cannot have zero value.

This PR adds logic to skip construction of a `Duration` when a unit's value is zero.

### DIFF
--- a/prometheus/src/main/scala/filodb/prometheus/parse/AntlrParser.scala
+++ b/prometheus/src/main/scala/filodb/prometheus/parse/AntlrParser.scala
@@ -103,9 +103,11 @@ class AntlrParser extends PromQLBaseVisitor[Object] {
 
   override def visitSubquery(ctx: PromQLParser.SubqueryContext): SubqueryClause = {
     val list = ctx.DURATION
-    val window = parseDuration(list.get(0))
-    val step = if (list.size() > 1) Some(parseDuration(list.get(1))) else None
-    SubqueryClause(window, step)
+    val windowMs = parseDuration(list.get(0))
+    val stepMs = if (list.size() > 1) Some(parseDuration(list.get(1))) else None
+    SubqueryClause(
+      Duration(windowMs / 1000, Second),
+      stepMs.map(ms => Duration(ms / 1000, Second)))
   }
 
   override def visitBinaryOperation(ctx: PromQLParser.BinaryOperationContext): BinaryExpression = {
@@ -209,11 +211,11 @@ class AntlrParser extends PromQLBaseVisitor[Object] {
   }
 
   override def visitWindow(ctx: PromQLParser.WindowContext): Duration = {
-    parseDuration(ctx.DURATION)
+    Duration(parseDuration(ctx.DURATION) / 1000, Second)
   }
 
   override def visitOffset(ctx: PromQLParser.OffsetContext): Duration = {
-    parseDuration(ctx.DURATION)
+    Duration(parseDuration(ctx.DURATION) / 1000, Second)
   }
 
   override def visitLimitOperation(ctx: PromQLParser.LimitOperationContext) = {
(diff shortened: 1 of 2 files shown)


prometheus/src/main/scala/filodb/prometheus/parse/AntlrParser.scala
```diff
@@ -103,9 +103,11 @@ class AntlrParser extends PromQLBaseVisitor[Object] {
 
   override def visitSubquery(ctx: PromQLParser.SubqueryContext): SubqueryClause = {
     val list = ctx.DURATION
-    val window = parseDuration(list.get(0))
-    val step = if (list.size() > 1) Some(parseDuration(list.get(1))) else None
-    SubqueryClause(window, step)
+    val windowMs = parseDuration(list.get(0))
```
yu-shipit, line 106: is it possible the list is empty?

prometheus/src/main/scala/filodb/prometheus/parse/AntlrParser.scala
```diff
@@ -103,9 +103,11 @@ class AntlrParser extends PromQLBaseVisitor[Object] {
...
+    val stepMs = if (list.size() > 1) Some(parseDuration(list.get(1))) else None
+    SubqueryClause(
+      Duration(windowMs / 1000, Second),
```
yu-shipit, line 109: since parseDuration returns a Duration with a millisecond as its unit. Do we need to encapsulate it again?
I think Duration allows operation like 1s + 1ms.

prometheus/src/test/scala/filodb/prometheus/parse/ParserSpec.scala
```diff
@@ -754,24 +754,55 @@ class ParserSpec extends AnyFunSpec with Matchers {
...
+            (3 * 24 * 60 * 60) + (25 * 60),
+            (12 * 60 * 60) + (15 * 60) + 30
+      )
+    )
+    specs.foreach{ spec =>
+      Parser.parseQuery(spec.query) match {
         case RangeExpression(_, _, window, offset) =>
-          window.millis(0) shouldEqual 1000 * windowSec
-          offset.get.millis(0) shouldEqual 1000 * offsetSec
+          window.millis(0) shouldEqual 1000 * spec.windowSec
```
yu-shipit, line 778: looks good to me.
However, a minor thing I want to bring up.
The test for the original code would be
window.millis(0) / 1000  shouldEqual spec.windowSec.
There is no difference for the test because the milliseconds are divisible by 1000

prometheus/src/main/scala/filodb/prometheus/parse/AntlrParser.scala
alextheimer: Nope, since it wouldn't otherwise be parsed as a `DURATION`: https://github.com/filodb/FiloDB/blob/7ba04ffc47385d165b9733393f5787dcedbd9274/prometheus/src/main/java/filodb/prometheus/antlr/PromQL.g4#L184

prometheus/src/main/scala/filodb/prometheus/parse/AntlrParser.scala
alextheimer: `parseDuration` used to return a `Duration`, but it now just returns the total elapsed milliseconds in a `DURATION`-- that name is probably just confusing now, so I'll update it.

prometheus/src/test/scala/filodb/prometheus/parse/ParserSpec.scala
alextheimer: You're saying it would be more correct to use this, instead?
```
window.millis(0) / 1000 shouldEqual spec.windowSec
```

alextheimer: Also in hindsight, it would be way clearer to use `Duration`s here-- I'll do that

prometheus/src/main/scala/filodb/prometheus/parse/AntlrParser.scala
alextheimer: I've changed the return type back to `Duration`.

yu-shipit (approved):
it looks good to me

PR #1509, opened 2023-01-31
fix(query): treat empty join key as any other during AND join

alextheimer:
**Pull Request checklist**

Currently, a join such as
```
sum(foo{}) AND bar{}
```
will erroneously return `sum(foo{})` if `bar{}` returns no data.

The logic in `setOpAnd` can be described as two sequential parts:
1. Create a set of all join keys on the RHS.
2. Add a RangeVector from LHS to the result iff its join key appears in the RHS set.

The bug is two-fold:
- At the first step, the logic skips a join key if its vector contains no rows; the key is not added to the RHS set.
- At the second step, we check that the RHS contains each LHS join key. For each key it doesn't contain, the logic falls into an `else if` condition: the LHS vector is added to the result if its join key is empty.

Since the first step won't add the key to the RHS set, the second step will fall into the `else if` and add the LHS vector to the result.

The fix is also two-fold:
- Remove the `else if` condition at the second step. It was setup specifically to handle queries such as `up AND ON (not_a_label) vector(1)`.
- Remove a check during the first step that prevents an empty join key from getting placed into the RHS set. Now, empty join keys are treated like any other (and queries like `up AND ON (not_a_label) vector(1)`) will succeed.

### DIFF
--- a/query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
+++ b/query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
@@ -93,15 +93,6 @@ final case class SetOperatorExec(queryContext: QueryContext,
     else rvk.labelValues.filterNot(lv => ignoringLabels.contains(lv._1))
   }
 
-  private def joinKeys2(rvk: RangeVectorKey): Option[Map[Utf8Str, Utf8Str]] = {
-    if (onLabels.nonEmpty) {
-      val labels = rvk.labelValues.filter(lv => onLabels.contains(lv._1))
-      Some(labels)
-    } else {
-      Some(rvk.labelValues.filterNot(lv => ignoringLabels.contains(lv._1)))
-    }
-  }
-
   /***
     * Returns true when range vector does not have any values
     */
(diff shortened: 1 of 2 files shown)


query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
```diff
@@ -118,8 +118,8 @@ final case class SetOperatorExec(queryContext: QueryContext,
 
     rhsRvs.foreach { rv =>
       val jk = joinKeys(rv.key)
-      // Don't add range vector if it is empty
-      if (jk.nonEmpty && !isEmpty(rv, rhsSchema)) {
```
amolnayak311, line 122: If RHS is empty, ``jk`` will be empty and nothing gets executed. What will happen to ``sum(foo{}) OR bar{}``  where ``bar{}`` is empty. The expectation is to have LHS results as [here](http://demo.robustperception.io:9090/graph?g0.expr=up%7B%7D%20OR%20bar%7B%7D&g0.tab=1&g0.stacked=0&g0.range_input=1h)

amolnayak311 (approved):

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
amolnayak311: Missed this one, the change is in AND only.

PR #1513, opened 2023-02-08
fix(query): pushdown joins only if lhs/rhs shard keys are identical

alextheimer:
**Pull Request checklist**

Currently, queries such as:
```
// target-schema labels:  shard_key, label1
foo{shard_key="aaa"} + on(label1) bar{shard_key="bbb"}
// Note: shard-key labels are implicit in the on() clause.
```
will be pushed down because planner logic only requires that the join operands have target-schemas defined against the same set of labels.

This PR adds logic to also require that shard keys are identical on each side of the join.

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
@@ -365,6 +365,21 @@ class SingleClusterPlanner(val dataset: Dataset,
   }
   // scalastyle:on cyclomatic.complexity
 
+  private def getTargetSchemaLabelsFromRawSeries(rs: RawSeries, qContext: QueryContext): Option[Seq[String]] = {
+    val tsp = targetSchemaProvider(qContext)
+    val filters = getColumnFilterGroup(rs).map(_.toSeq)
+    // Get time params from the RangeSelector, and use them to identify a TargetSchemaChanges.
+    val startEndMsPairOpt = rs.rangeSelector match {
+      case is: IntervalSelector => Some((is.from, is.to))
+      case _ => None
+    }
+    val tsFunc = tsp.targetSchemaFunc(filters.head)
+    startEndMsPairOpt match {
+      case Some((startMs, endMs)) => findTargetSchema(tsFunc, startMs, endMs).map(_.schema)
+      case _ => None
+    }
+  }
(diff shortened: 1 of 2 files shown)


coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
```diff
@@ -387,18 +405,11 @@ class SingleClusterPlanner(val dataset: Dataset,
...
             case _ => None
           }
-          val tsLabels: Option[Seq[String]] = {
-            val tsFunc = tsp.targetSchemaFunc(filters.head)
-            startEndMsPairOpt match {
-              case Some((startMs, endMs)) => findTargetSchema(tsFunc, startMs, endMs).map(_.schema)
-              case _ => None
-            }
-          }
+          val tsLabels: Option[Seq[String]] = getTargetSchemaLabelsFromRawSeries(rs, qContext)
```
amolnayak311, line 412: Why a method? This lead to code duplication and doing ``map``  instead of pattern match will be concise as well

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
```diff
@@ -461,49 +475,61 @@ class SingleClusterPlanner(val dataset: Dataset,
...
-                              targetSchemaLabels.get.toSet.diff(shardColumns.toSet).subsetOf(bj.on.toSet)
-                          }
-        // union lhs/rhs shards, since one might be empty (if it's a scalar)
-        if (canPushdown) Some(lhsShards.get.union(rhsShards.get)) else None
+        //   discoverable, equal, and preserved by join keys.
+        val lhsData = helper(bj.lhs)
+        val rhsData = helper(bj.rhs)
+        val canPushdown = lhsData.isDefined && rhsData.isDefined && {
+          val hasScalar = lhsData.get.shards.isEmpty || rhsData.get.shards.isEmpty
+          val canPushdownJoin = lhsData.get.shards == rhsData.get.shards &&
```
amolnayak311, line 483: Is the order of ``Seq`` containing shards granted to be ordered consistently? ``Seq(1, 2, 3) != Seq(3, 2, 1)`` though you want to push down.

Also too much going on here, some comments would really be helpful to understand the intent.

PR #1531, opened 2023-03-13
fix(coord) Alternative host lookup strategy to avoid reverse lookup of ip address

amolnayak311:
**Pull Request checklist**

**Current behavior :** 
Current strategy of assigning shards to pods is to look at the ordinal of the host name when using stateful sets. The hostname is looked up using reverse IP lookup of the  coordinator Actor. 

**New behavior :**

The strategy will fail when reverse lookup from ip cannot be done reliably. To eliminate this reverse lookup, we add a way to query the coordinator actor to provide its hostname.

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/ShardAssignmentStrategy.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/ShardAssignmentStrategy.scala
@@ -46,12 +46,28 @@ trait ShardAssignmentStrategy {
  * The implementation assumes the number of shards is a multiple of number of nodes and there are no
  * uneven assignments of shards to nodes
  */
-class K8sStatefulSetShardAssignmentStrategy(val maxAssignmentAttempts: Int = 8)
+class K8sStatefulSetShardAssignmentStrategy(val maxAssignmentAttempts: Int = 8,
+                                            val headlessServiceName: Option[String] = None)
       extends ShardAssignmentStrategy with StrictLogging {
 
   private val pat = "-\\d+$".r
 
-  private[coordinator] def getOrdinalFromActorRef(coord: ActorRef): Option[(String, Int)] =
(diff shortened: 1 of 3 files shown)


coordinator/src/main/scala/filodb.coordinator/ShardAssignmentStrategy.scala
```diff
@@ -52,15 +56,23 @@ class K8sStatefulSetShardAssignmentStrategy(val maxAssignmentAttempts: Int = 8)
...
-      .orElse(Some(InetAddress.getLocalHost.getHostName))
-      .map(name => if (name.contains(".")) name.substring(0, name.indexOf('.')) else name)
-      .flatMap(hostName =>
-        Some((hostName, pat.findFirstIn(hostName).map(ordinal => -Integer.parseInt(ordinal)).getOrElse(-1))))
-
+    Try {
+      implicit val timeout = Timeout(5.seconds)
+      val coordinatorHostName = coord.path.address.host
+        .map(_ => Await.result((coord ? MiscCommands.GetHostName).mapTo[String],
+          FilodbSettings.globalOrDefault.DefaultTaskTimeout)).getOrElse(InetAddress.getLocalHost.getHostName)
```
vishramachandran, line 63: The Await will block the actor dispatcher thread. This is not good, and accumulates tech debt. 
Prefer to move to Clustering V2, but I understand this interim solution may be needed.

amolnayak311: Thanks @vishramachandran, agreed this isn't an ideal solution. I have added a comment and called it out.

vishramachandran (approved):
This is tech debt. We are doing network IO (actor call) on actor dispatcher thread.

Understand the need for doing this to workaround immediate problems.

Clustering V2 should be the proper solution.

PR #1538, opened 2023-03-22
fix(query) Support HA Planner to use gRPC endpoints and exception serialization

amolnayak311:
**Pull Request checklist**

The PR addresses two issues with the gRPC based QS

- Current implementation of gRPC based query service does not support HA planner to use gRPC service on the other cluster in case the current cluster's shards are unavailable. The PR adds support to use either gRPC or PromQLRemoteExec based routing for HA.  HA supports using cluster specific planners in the same gRPC service using a pre-configured selector.

- Exceptions across gRPC remoting are serialized but they aren't deserialized to the specific exception type to keep the implementation simple, however this makes client applications looking for specific exceptions to respond with appropriate error codes difficult. The PR handles some common exception classes to deserialize them to the specific type than use the generic type.

### DIFF
--- a/grpc/src/main/protobuf/query_service.proto
+++ b/grpc/src/main/protobuf/query_service.proto
@@ -53,7 +53,7 @@ message StreamingErrorResponse {
 
 message StreamingBodyResponse {
   string queryId                                 = 1;
-  SerializableRangeVector result                 = 2;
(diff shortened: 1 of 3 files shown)


vishramachandran (approved):
LGTM. Thanks for the iterations.

PR #1539, opened 2023-03-23
fix(core) message use histogram_quantile for max(histogram) queries.

yu-shipit:
max(histogram) is not supported histogram_quantile(1, histogram) may be what users want.

**Pull Request checklist**

**New behavior :**

**BREAKING CHANGES**

Breaking changes may include:

**Other information**:

### DIFF
--- a/query/src/main/scala/filodb/query/exec/aggregator/RowAggregator.scala
+++ b/query/src/main/scala/filodb/query/exec/aggregator/RowAggregator.scala
@@ -121,22 +121,26 @@ object RowAggregator {
   def apply(aggrOp: AggregationOperator, params: Seq[Any], schema: ResultSchema): RowAggregator = {
     val valColType = ResultSchema.valueColumnType(schema)
     aggrOp match {
-      case Min      => MinRowAggregator
-      case Max      => MaxRowAggregator
+      case Min if valColType == ColumnType.DoubleColumn => MinRowAggregator
+      case Max if valColType == ColumnType.DoubleColumn => MaxRowAggregator
       case Sum if valColType == ColumnType.DoubleColumn => SumRowAggregator
       case Sum if isHistMax(valColType, schema)            => HistMaxSumAggregator
       case Sum if valColType == ColumnType.HistogramColumn => HistSumRowAggregator
       case Count if valColType == ColumnType.DoubleColumn    => CountRowAggregator.double
       case Count if valColType == ColumnType.HistogramColumn => CountRowAggregator.hist
-      case Group    => GroupRowAggregator
-      case Avg      => AvgRowAggregator
-      case TopK     => new TopBottomKRowAggregator(params(0).asInstanceOf[Double].toInt, false)
-      case BottomK  => new TopBottomKRowAggregator(params(0).asInstanceOf[Double].toInt, true)
-      case Quantile => new QuantileRowAggregator(params(0).asInstanceOf[Double])
-      case Stdvar   => StdvarRowAggregator
-      case Stddev   => StddevRowAggregator
+      case Group if valColType == ColumnType.DoubleColumn   => GroupRowAggregator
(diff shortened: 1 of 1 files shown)


yu-shipit:
max(histogram) will the the message below.
```
{
  "status" : "error",
  "data" : null,
  "errorType" : "requirement_failed",
  "error" : "max(histogram expr) is not supported. You may use histogram_quantile(1, histogram expr) instead.",
  "queryStats" : [ {
    "group" : [ "local", "raw", "aci-telemetry", "Test-data-1679598622_p1", "my_histogram" ],
    "timeSeriesScanned" : 2,
    "dataBytesScanned" : 259,
    "resultBytes" : 0,
    "cpuNanos" : 0
  } ]
}
```

alextheimer (approved):
Thanks @yu-shipit for digging into this unexpectedly-complex problem 👍

amolnayak311:
1. If the behavior differs from Prometheus we need to call it out explicitly, in the PR and also a comment in the code as we generally strive to maintain compatibility as far as possible. 
2. It is important for the user to understand that histogram_quantile will give the approximation  which is dependent on their buckets, while the user might be expecting to see the max value recorded.

query/src/main/scala/filodb/query/exec/aggregator/RowAggregator.scala
```diff
@@ -121,8 +121,8 @@ object RowAggregator {
   def apply(aggrOp: AggregationOperator, params: Seq[Any], schema: ResultSchema): RowAggregator = {
     val valColType = ResultSchema.valueColumnType(schema)
     aggrOp match {
-      case Min      => MinRowAggregator
-      case Max      => MaxRowAggregator
+      case Min if valColType == ColumnType.DoubleColumn => MinRowAggregator
```
amolnayak311, line 124: Do you see any case where `LongColumn` and `IntColumn` would need this? I understand we don't want to use this for histograms but by restricting to DoubleColumn only we also don't support Long and Int columns

amolnayak311 (approved):

query/src/main/scala/filodb/query/exec/aggregator/RowAggregator.scala
```diff
@@ -121,22 +121,26 @@ object RowAggregator {
...
+      case BottomK if valColType == ColumnType.DoubleColumn
+                  => new TopBottomKRowAggregator(params(0).asInstanceOf[Double].toInt, true)
+      case Quantile if valColType == ColumnType.DoubleColumn
+          => new QuantileRowAggregator(params(0).asInstanceOf[Double])
+      case Stdvar if valColType == ColumnType.DoubleColumn  => StdvarRowAggregator
+      case Stddev if valColType == ColumnType.DoubleColumn  => StddevRowAggregator
       case CountValues => new CountValuesRowAggregator(params(0).asInstanceOf[String])
-      case _     => ???
+      case _ => throw new IllegalArgumentException(s"This operation ${aggrOp} is not supported for ${valColType}." +
+      s" Because it does not result in a meaningful query. Please contact us if you have a use case.")
```
vishramachandran, line 143: Let us not use this message as fallback for all.
For histograms:
> "The ${aggrOp} operation is not supported directly for histogram types. Check if you need to resolve to the individual histogram buckets, or calculate the rate and histogram_quantile before applying the aggregation. If you have a genuine use case for this query, please get in touch."


For doubles, it should be an illegal state exception. It should be unreachable code.

vishramachandran (changes requested):
Asking for a small change in error message

query/src/main/scala/filodb/query/exec/aggregator/RowAggregator.scala
vishramachandran: Count_values should also be applied only to doubles.

vishramachandran (approved):
LGTM. thank you!!

PR #1543, opened 2023-03-28
fix(query): vector(static_value) on rhs returns TS of static_value when lhs returns empty

sandeep6189:
vector(static_value) on rhs returns TS of static_value when lhs returns empty. The fix is added to re-enable the failing `OperatorPrecedence` integration tests.

**Pull Request checklist**

**Current behavior :** 

In a binary join query of format "lhs or vector(static_value)",  FiloDB returns empty TS when lhs evaluates to empty result. However, prometheus returns a TS of static_value based on the range parameters given. This is a corner case scenario we found for or vector(static_val) scenario.

**New behavior :**
FiloDB is compliant to the prometheus results on the given above scenario.

### DIFF
--- a/query/src/main/scala/filodb/query/exec/RangeVectorTransformer.scala
+++ b/query/src/main/scala/filodb/query/exec/RangeVectorTransformer.scala
@@ -1,6 +1,5 @@
 package filodb.query.exec
 
-import scala.collection.Iterator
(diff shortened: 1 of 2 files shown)


sherali42 (approved):
👍

vishramachandran (approved):
Thank you for finding and picking up the regression! Appreciate it.

PR #1545, opened 2023-03-30
fix(core) fix count_values query.

yu-shipit:
The column for count_values aggregator is StringColumn.

**Pull Request checklist**

### DIFF
--- a/query/src/main/scala/filodb/query/exec/aggregator/RowAggregator.scala
+++ b/query/src/main/scala/filodb/query/exec/aggregator/RowAggregator.scala
@@ -120,36 +120,30 @@ object RowAggregator {
     */
   def apply(aggrOp: AggregationOperator, params: Seq[Any], schema: ResultSchema): RowAggregator = {
     val valColType = ResultSchema.valueColumnType(schema)
-    if (valColType == ColumnType.HistogramColumn) {
-      aggrOp match {
-        case Sum if isHistMax(valColType, schema) => HistMaxSumAggregator
-        case Sum   => HistSumRowAggregator
-        case Count => CountRowAggregator.hist
-        case _ => throw new IllegalArgumentException(s"The ${aggrOp} operation is not supported directly for histogram"
-          + s" types. Check if you need to resolve to the individual histogram buckets,"
-          + s"or calculate the rate and histogram_quantile before applying the aggregation. "
-          + s"If you have a genuine use case for this query, please get in touch.")
-      }
-    } else if (valColType == ColumnType.DoubleColumn) {
-      aggrOp match {
-        case Min => MinRowAggregator
-        case Max => MaxRowAggregator
-        case Sum => SumRowAggregator
-        case Count => CountRowAggregator.double
-        case Group => GroupRowAggregator
-        case Avg => AvgRowAggregator
-        case TopK => new TopBottomKRowAggregator(params(0).asInstanceOf[Double].toInt, false)
-        case BottomK => new TopBottomKRowAggregator(params(0).asInstanceOf[Double].toInt, true)
-        case Quantile => new QuantileRowAggregator(params(0).asInstanceOf[Double])
-        case Stdvar => StdvarRowAggregator
-        case Stddev => StddevRowAggregator
-        case CountValues => new CountValuesRowAggregator(params(0).asInstanceOf[String])
-      }
-    } else {
-      if (aggrOp == CountValues) {
-        return new CountValuesRowAggregator(params(0).asInstanceOf[String])
-      }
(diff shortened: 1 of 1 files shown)


query/src/main/scala/filodb/query/exec/aggregator/RowAggregator.scala
```diff
@@ -130,7 +130,7 @@ object RowAggregator {
           + s"or calculate the rate and histogram_quantile before applying the aggregation. "
           + s"If you have a genuine use case for this query, please get in touch.")
       }
-    } else if (valColType == ColumnType.DoubleColumn) {
+    } else if (valColType == ColumnType.DoubleColumn || valColType == ColumnType.StringColumn) {
```
amolnayak311, line 133: Why do we need `else if`?  Original intent of the PR is to handle histograms which is already done just let everything else follow the same path as before. By restricting the types we don't know additional regression we may see.

amolnayak311 (changes requested):

query/src/main/scala/filodb/query/exec/aggregator/RowAggregator.scala
yu-shipit: good point. The patch removed the default case so that compile would fail if missing any cases.

amolnayak311: @vishramachandran , requesting your feedback.

amolnayak311: The current patch doesn't work for topk/bottomk as Im getting 
``topk(2, foo{..})`` fails with `The TopK operation is not supported for StringColumn`
thus I strongly feel as I originally recommended, minimize the blast radius and restrict your changes to only histograms, which was the original intention. Everything else should take the same code path as before.

yu-shipit: you are right.

I think the problem may be caused by this function
`  def valueColumnType(schema: ResultSchema): ColumnType = {
    require(schema.isTimeSeries, s"Schema $schema is not time series based, cannot continue query")
    require(schema.columns.size >= 2, s"Schema $schema has less than 2 columns, cannot continue query")
    schema.columns(1).colType
  }
`
For schema more than 2 columns schema.columns(1).colType may be always be the valueColumnType

query/src/main/scala/filodb/query/exec/aggregator/RowAggregator.scala
```diff
@@ -120,33 +120,30 @@ object RowAggregator {
...
+      case Sum if valColType == ColumnType.HistogramColumn => HistSumRowAggregator
+      case Count if valColType == ColumnType.DoubleColumn => CountRowAggregator.double
+      case Count if valColType == ColumnType.HistogramColumn => CountRowAggregator.hist
+      case Group if valColType == ColumnType.DoubleColumn => GroupRowAggregator
+      case Avg if valColType == ColumnType.DoubleColumn => AvgRowAggregator
+      case TopK => new TopBottomKRowAggregator(params(0).asInstanceOf[Double].toInt, false)
+      case BottomK => new TopBottomKRowAggregator(params(0).asInstanceOf[Double].toInt, true)
+      case Quantile => new QuantileRowAggregator(params(0).asInstanceOf[Double])
+      case Stdvar if valColType == ColumnType.DoubleColumn => StdvarRowAggregator
+      case Stddev if valColType == ColumnType.DoubleColumn => StddevRowAggregator
```
amolnayak311, line 137: @vishramachandran I see `Group`, `Avg`, `StdVar` and `StdDev` are now restricted to column type double only. Would that be ok? I mentioned a couple of times [here](https://github.com/filodb/FiloDB/pull/1545#discussion_r1153543524) that we should try and minimize the blast radius by just fixing the count and sum for histograms only and let everything else continue to work the way it is. We already found a regression in count and luckily I found regression in topk/bottomk during my local testing. I would like to get your views. I'll be holding off my approval for the PR till then.Thanks

cc @kvpetrov @sandeep6189

vishramachandran: I agree on containing the blast radius and minimizing regressions. At the same time we need to send out a clearer error message that helps the user esp if the query is not meaningful. I don't see us supporting some of the aggregations (like min, max, stddev, stdvar etc) on histograms. Do you see that possibility ?

I think we are seeing this regression because of absence of unit tests esp for this function. Improving unit tests by running this function through various types could make it more solid, but it comes at the cost of increasing this PR scope.

amolnayak311 (approved):
As discussed over the call, this looks good now.

PR #1548, opened 2023-04-01
feat(query): Adding quota parameters for individual query limits and further reporting.

kvpetrov:
Adding quota parameters for individual query limits and further reporting.

**Current behavior :** 
Currently, there are some limits established for query processing, however, these limits are either defaults in the scala code or properties of FiloDB runtime. 

**New behavior :**
We want to be able to pass these limits as part of the query utilizing PlannerParams

**BREAKING CHANGES**
Protos defining PlannerParams are redefined in a manner which is not backwards compatible. This should be ok as our scala pojos normally are not backwards compatible. 

**Other information**:


core/src/main/scala/filodb.core/query/QueryContext.scala
```diff
@@ -22,26 +22,50 @@ case class PromQlQueryParams(promQl: String, startSecs: Long, stepSecs: Long, en
...
+        execPlanResultBytes = 15000000,
+        groupByCardinality = 50000,
+        joinQueryCardinality = 50000,
+        timeSeriesSamplesScanBytes = 150000
+      )
+    } else {
+      IndividualQuota()
+    }
+  }
+}
```
alextheimer, line 46: Are these separate "warning" thresholds needed? Is it sufficient to warn at X% of the limit?

If the separate thresholds are needed, a `Threshold(warning, limit)` class is another option to consider (this way we wouldn't need something like `apply(warnDefaults)`).

core/src/main/scala/filodb.core/query/QueryContext.scala
```diff
@@ -22,26 +22,50 @@ case class PromQlQueryParams(promQl: String, startSecs: Long, stepSecs: Long, en
...
                          splitSizeMs: Long = 1.day.toMillis,
                          skipAggregatePresent: Boolean = false,
                          processFailure: Boolean = true,
                          processMultiPartition: Boolean = false,
                          allowPartialResults: Boolean = false)
+
 object PlannerParams {
   def apply(constSpread: Option[SpreadProvider], sampleLimit: Int): PlannerParams =
-    PlannerParams(spreadOverride = constSpread, sampleLimit = sampleLimit)
+    PlannerParams(spreadOverride = constSpread, enforcedQuota = IndividualQuota(execPlanSamples = sampleLimit))
```
alextheimer, line 68: I'm not sure where we use this specific `apply`, but it might construct `PlannerParams` with some warning limits less than the enforced limits.

query/src/main/scala/filodb/query/exec/BinaryJoinExec.scala
```diff
@@ -73,9 +73,11 @@ final case class BinaryJoinExec(queryContext: QueryContext,
...
       case (QueryResult(_, _, result, _, _, _), _)
-        if (result.size  > queryContext.plannerParams.joinQueryCardLimit && cardinality == Cardinality.OneToOne) =>
-        throw new BadQueryException(s"The join in this query has input cardinality of ${result.size} which" +
-          s" is more than limit of ${queryContext.plannerParams.joinQueryCardLimit}." +
+        if (
+          result.size  > queryContext.plannerParams.enforcedQuota.joinQueryCardinality &&
+          cardinality == Cardinality.OneToOne
+        ) => throw new BadQueryException(s"The join in this query has input cardinality of ${result.size} which" +
+          s" is more than limit of ${queryContext.plannerParams.enforcedQuota.joinQueryCardinality}." +
           s" Try applying more filters or reduce time range.")
```
alextheimer, line 81: Where will `warnQuota` be used?

alextheimer (dismissed):
Gotta love those `ExecPlan::printTree` comparison tests 😛
Approved-- please consider some of the comments I left before you merge.

query/src/main/scala/filodb/query/ProtoConverters.scala
```diff
@@ -180,13 +180,28 @@ object ProtoConverters {
...
       val builder = GrpcMultiPartitionQueryService.PlannerParams.newBuilder()
       builder.setApplicationId(pp.applicationId)
       builder.setQueryTimeoutMillis(pp.queryTimeoutMillis)
-      builder.setSampleLimit(pp.sampleLimit)
-      builder.setGroupByCardLimit(pp.groupByCardLimit)
-      builder.setJoinQueryCardLimit(pp.joinQueryCardLimit)
-      builder.setResultByteLimit(pp.resultByteLimit)
+      builder.setEnforcedQuota(enforcedQuota)
+      builder.setWarnQuota(warnQuota)
+      pp.queryOrigin.map(qo => builder.setQueryOrigin(qo))
```
amolnayak311, line 202: anti pattern for use of ``map``,  it should not cause side effects, either use ``isDefined`` or ``foreach``

query/src/main/scala/filodb/query/ProtoConverters.scala
```diff
@@ -222,6 +240,40 @@ object ProtoConverters {
     }
   }
 
+  implicit class IndividualQuotaFromProtoConverter(giq: GrpcMultiPartitionQueryService.IndividualQuota) {
+    def fromProto(warnDefaults: Boolean = false): IndividualQuota = {
```
amolnayak311, line 244: It's in inappropriate to add ``warnDefaults`` here. This converter should convert given proto to equivalent case class. Misusing this method will introduce bugs hard to troubleshoot. ``toProto`` and ``fromProto`` should not accept parameters and alter the behavior. . 

Have another one to convert case class to proto and write a spec similar to how it's written for others.

amolnayak311 (changes requested):
Please see changes and incorporate the changes requested. Thanks for putting this together.

amolnayak311:
Please update the description to give details if the PR and commit as per [these](https://github.com/filodb/FiloDB/blob/develop/CONTRIBUTING.md) guidelines.

core/src/main/scala/filodb.core/query/QueryContext.scala
kvpetrov: I'm only copying whatever was already in the existing properties. This PR essentially groups limits in an object and allows to have two sets of quotas/limits: (1) enforced (2) warn and allows to pass these parameters from the QueryService but it DOES NOT change the existing behavior.

kvpetrov: The warning piece is fundamental for the PR. The reason we do not enforce the limits (we have very high limits) is because it's impossible for us to enforce them. There would be a number of users whose queries will stop to work and we simply cannot let it happen (imagine alerts or recording rules), instead, we need to:
A) establish quotas that we believe are resonable on WARN level
B) find out all the users/cases where the limits are breached
C) establish SPECIFIC HIGH limits for recording rules/alerts/ some API queries
D) enforce reasonable limits for EVERYBODY but the specific cases above
E) notify the users with specific high limits to let them know that their RR/alerts will be disabled in X days if they don't take care of it.

At this stage I think it's reasonable to have more flexibility in defining what the warn and enforce levels are. The ratio is hard to come up by looking at the crystal ball. Maybe it's 5x maybe 2. This we will figure out as part of an experiment around steps A/B/C.

core/src/main/scala/filodb.core/query/QueryContext.scala
kvpetrov: this is expected, isn't it? the warning limits are smaller than enforced, ie at 2MB I warn you, at 10MB I won't let your query run.

query/src/main/scala/filodb/query/exec/BinaryJoinExec.scala
kvpetrov: pls, see the comment above, I will also update the description in the PR

core/src/main/scala/filodb.core/query/QueryContext.scala
```diff
@@ -22,26 +22,50 @@ case class PromQlQueryParams(promQl: String, startSecs: Long, stepSecs: Long, en
 
 case object UnavailablePromQlQueryParams extends TsdbQueryParams
 
+case class IndividualQuota(
```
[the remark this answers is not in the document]
kvpetrov: can we do it later, in the second PR, that PR will do nothing but renaming the object?

core/src/main/scala/filodb.core/query/QueryContext.scala
```diff
@@ -22,26 +22,50 @@ case class PromQlQueryParams(promQl: String, startSecs: Long, stepSecs: Long, en
...
                          spreadOverride: Option[SpreadProvider] = None,
                          shardOverrides: Option[Seq[Int]] = None,
                          targetSchemaProviderOverride: Option[TargetSchemaProvider] = None,
                          queryTimeoutMillis: Int = 60000, // set default to match default http-request-timeout
-                         sampleLimit: Int = 1000000,
-                         groupByCardLimit: Int = 100000,
-                         joinQueryCardLimit: Int = 100000,
-                         resultByteLimit: Long = 18000000,  // 18MB
+                         enforcedQuota: IndividualQuota = IndividualQuota(),
+                         warnQuota: IndividualQuota = IndividualQuota.apply(warnDefaults = true),
```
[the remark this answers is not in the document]
kvpetrov: pls see comment above, will update the description in the PR as well.
In short the warn most likely in the first phase will go to logs. It would be used by operators to see what the reasonable limits are.

core/src/main/scala/filodb.core/query/QueryContext.scala
```diff
@@ -22,26 +22,50 @@ case class PromQlQueryParams(promQl: String, startSecs: Long, stepSecs: Long, en
...
                          spread: Option[Int] = None,
                          spreadOverride: Option[SpreadProvider] = None,
                          shardOverrides: Option[Seq[Int]] = None,
                          targetSchemaProviderOverride: Option[TargetSchemaProvider] = None,
                          queryTimeoutMillis: Int = 60000, // set default to match default http-request-timeout
-                         sampleLimit: Int = 1000000,
-                         groupByCardLimit: Int = 100000,
-                         joinQueryCardLimit: Int = 100000,
-                         resultByteLimit: Long = 18000000,  // 18MB
+                         enforcedQuota: IndividualQuota = IndividualQuota(),
```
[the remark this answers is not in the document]
kvpetrov: I think we can move them to QueryContext too as part of a different PR, originally the parameters were int PlannerParams.

query/src/main/scala/filodb/query/ProtoConverters.scala
kvpetrov: not sure I understood this comment. queryOrigin is an option. I set queryOrigin on the proto builder if and only if option is not None. Why side effects here?

query/src/main/scala/filodb/query/ProtoConverters.scala
kvpetrov: need to talk to you about this one I guess. the problem is that it seems like protos don't have default values but scala case classes do. So, suppose you have a proto that does not have a property defined but you cannot NOT to create a corresponding scala object since its properties are not optional. So, either I define proto properties as required or?..  I noticed that most of your protos have the fields defined as optional and I believe you did it for a reason, correct?

query/src/main/scala/filodb/query/ProtoConverters.scala
```diff
@@ -180,13 +180,28 @@ object ProtoConverters {
 
   implicit class PlannerParamsToProtoConverter(pp: PlannerParams) {
     def toProto: GrpcMultiPartitionQueryService.PlannerParams = {
+      val enforcedQuotaBuilder = GrpcMultiPartitionQueryService.IndividualQuota.newBuilder()
```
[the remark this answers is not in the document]
kvpetrov: there is one defined:
Line 243.
implicit class IndividualQuotaFromProtoConverter(giq: GrpcMultiPartitionQueryService.IndividualQuota) {

core/src/main/scala/filodb.core/query/QueryContext.scala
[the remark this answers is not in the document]
amolnayak311: I leave it to you and @vishramachandran the later we do more we will have to work on preserving backward compatibility.

query/src/main/scala/filodb/query/ProtoConverters.scala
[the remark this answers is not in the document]
amolnayak311: It just converts the proto to case class (``fromProto``), we need a ``toProto`` to convert the case class to gRPC object. Anywhere we deal with converting to and from proto we should use this implicit conversions rather than copying the object building logic as we have in this PR

query/src/main/scala/filodb/query/ProtoConverters.scala
amolnayak311: Your ``map`` is setting something in an object (side effect), by design this should simply transform and give a new option transforming the original one. As I mentioned, use ``foreach`` or ``isDefined`` to set this proto field.

query/src/main/scala/filodb/query/ProtoConverters.scala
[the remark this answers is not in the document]
kvpetrov: got you, will add toProto (only fromProto is defined now)

query/src/main/scala/filodb/query/ProtoConverters.scala
amolnayak311: Correct but we never add parameters to ``toProto`` and ``fromProto``. The goal is to convert whatever the case class has (default or explicitly set values) to proto and back.

query/src/main/scala/filodb/query/ProtoConverters.scala
kvpetrov: will do

amolnayak311 (approved):
Thanks @kvpetrov for the changes. LGTM. I have retriggered the PRB which was failing.

PR #1556, opened 2023-04-12
Check for the binary join input cardinality

kvpetrov:
**Pull Request checklist**

- [] Tests for the changes have been added (for bug fixes / features) ?

Current behavior :

Currently we throw an exception if binary join input cardinality is high.

New behavior :

Throw an exception for high binary join cardinality but also add warning if we go beyond warning level. We also properly record in the logs the query and where it comes from.

### DIFF
--- a/query/src/main/scala/filodb/query/exec/BinaryJoinExec.scala
+++ b/query/src/main/scala/filodb/query/exec/BinaryJoinExec.scala
@@ -2,6 +2,7 @@ package filodb.query.exec
 
 import scala.collection.mutable
 
+import com.typesafe.scalalogging.StrictLogging
 import kamon.Kamon
 import kamon.metric.MeasurementUnit
 import monix.eval.Task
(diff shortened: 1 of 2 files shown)


query/src/main/scala/filodb/query/exec/BinaryJoinExec.scala
```diff
@@ -72,14 +73,27 @@ final case class BinaryJoinExec(queryContext: QueryContext,
...
     val taskOfResults = childResponses.map {
-      case (QueryResult(_, _, result, _, _, _), _)
-        if (
-          result.size  > queryContext.plannerParams.enforcedLimits.joinQueryCardinality &&
-          cardinality == Cardinality.OneToOne
-        ) => throw new BadQueryException(s"The join in this query has input cardinality of ${result.size} which" +
-          s" is more than limit of ${queryContext.plannerParams.enforcedLimits.joinQueryCardinality}." +
-          s" Try applying more filters or reduce time range.")
-      case (QueryResult(_, _, result, _, _, _), i) => (result, i)
+      tuple : (QueryResult, Int) => {
```
sandeep6189, line 76: Wondering if it is a good idea to separate this logic out as a function and add UTs(for the logger and exception) ?

alextheimer, line 76: Can you match each of these into a variable?
```
case (a: A, b: B) => /* do stuff */
```

alextheimer (approved):

sandeep6189 (approved):

PR #1557, opened 2023-04-13
adding warning and enforcement for ts scans/exec plan outputs

kvpetrov:
**Pull Request checklist**

**Current behavior :** 
We check if number of data scanned and produced by exec plan is within the limits given properties and defaults in the codebase. 

**New behavior :**
We add explicit check on the number of time series and allow to WARN the users that he is breaching a soft limit. We get the quotas from outside of FiloDB runtime. 

**BREAKING CHANGES**

Removal of property max-data-per-shard-query

**Other information**:

### DIFF
--- a/core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
+++ b/core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
@@ -1,14 +1,13 @@
 package filodb.core.downsample
 
-import java.util
-import java.util.concurrent.atomic.AtomicLong
-
 import scala.collection.mutable
 import scala.concurrent.{ExecutionContext, Future}
 import scala.concurrent.duration._
 
 import com.typesafe.config.Config
 import com.typesafe.scalalogging.StrictLogging
+import java.util
+import java.util.concurrent.atomic.AtomicLong
 import kamon.Kamon
 import kamon.metric.MeasurementUnit
 import kamon.tag.TagSet
@@ -21,7 +20,7 @@ import filodb.core.{DatasetRef, Types}
 import filodb.core.binaryrecord2.RecordSchema
 import filodb.core.memstore._
 import filodb.core.metadata.Schemas
-import filodb.core.query.{ColumnFilter, Filter, QuerySession}
+import filodb.core.query.{ColumnFilter, Filter, QueryContext, QuerySession}
 import filodb.core.store._
 import filodb.memory.format.{UnsafeUtils, ZeroCopyUTF8String}
 import filodb.memory.format.ZeroCopyUTF8String._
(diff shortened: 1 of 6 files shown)


core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
```diff
@@ -391,11 +390,63 @@ private val partKeyIndex = new PartKeyLuceneIndex(indexDataset, schemas.part, fa
...
+        s"Query matched $timeSeries time series, which exceeds a max warn limit of " +
+          s"${warnLimits.timeSeriesScanned} time series allowed to be queried per shard. "
+      logger.info(qContext.getQueryLogLine(msg))
+    }
+    if (estDataSize > warnLimits.timeSeriesSamplesScannedBytes) {
+      val msg =
+        s"With match of $timeSeries time series, estimate of $estDataSize bytes exceeds " +
+          s"limit of ${warnLimits.timeSeriesSamplesScannedBytes} bytes queried per shard " +
+          s"for ${schemas.apply(schemaId).name} schema. "
+      logger.info(qContext.getQueryLogLine(msg))
```
alextheimer, line 449: It's outside the scope of this PR since these are just logged, but this (and other warning messages) should say something like: "...which is approaching the enforced limit of X".

alextheimer: I think this and `ensureQueriedDataSizeWithinLimitApprox` are identical except for the range? If that's true, could you give the `require/require/if/if` logic its own method so there aren't any redundant error messages to keep track of?

query/src/main/scala/filodb/query/exec/ExecPlan.scala
```diff
@@ -211,25 +211,11 @@ trait ExecPlan extends QueryCommand {
...
         .map { srv =>
-          numResultSamples += srv.numRowsSerialized
           // fail the query instead of limiting range vectors and returning incomplete/inaccurate results
-          if (enforceSampleLimit && numResultSamples > queryContext.plannerParams.enforcedLimits.execPlanSamples)
-            throw new BadQueryException(s"This query results in more than " +
-              s"${queryContext.plannerParams.enforcedLimits.execPlanSamples} samples. " +
-              s"Try applying more filters or reduce time range.")
-
+          numResultSamples += srv.numRowsSerialized
+          checkSamplesLimit(numResultSamples)
```
alextheimer, line 216: This looks infinitely more clean/digestible 💯

alextheimer (approved):

core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
kvpetrov: Here I'm trying not to change any existing logic, just adding warnings etc. Ie if there is any functional change then it has to be in the new PR, so, it can be seen clearly.

PR #1575, opened 2023-04-26
refactor(cassandra): Mark queries as idempotent to enable retries

abhinavk96:
**Pull Request checklist**

**Current behavior :**
Idempotence is not explicitly set on any prepared statements for the queries. 

Datastax specifies that by default queries are not marked as idempotent unless overridden at the session level. 
Because of this the queries won't be retried according to the retry policy. 

Reference: https://docs.datastax.com/en/developer/java-driver/3.6/manual/idempotence/#query-idempotence
Quote from above:
> Idempotence matters for [retries](https://docs.datastax.com/en/developer/java-driver/3.6/manual/retries/) and [speculative query executions](https://docs.datastax.com/en/developer/java-driver/3.6/manual/speculative_execution/). The driver will bypass those features if the [Statement#isIdempotent()](http://docs.datastax.com/en/drivers/java/3.6/com/datastax/driver/core/Statement.html#isIdempotent--) flag is set to false, to ensure that the statement does not get executed more than once.
In most cases, you must set that flag manually.

**New behavior :**
Setting idempotent explicitly to true will enable the retry policy to act on these queries. 

**Note:**
The queries have all been deemed to be idempotent. Please review individual queries carefully to verify that assertion.


vishramachandran (dismissed):
LGTM

amolnayak311 (approved):

amolnayak311 (approved):

PR #1585, opened 2023-05-04
Removing `name` from CardinalityValue data structure to optimize the storage

sandeep6189:
**Pull Request checklist**

**Current behavior :** (link exiting issues here : In our `RocksDBCardinalityStore`, we store records of type `CardinalityValue`. This data structure stores a string which is the last value of the shardKey Seq. However, from the implementation, it doesn't seem to be used ( TODO:  check with Amol and Vish ). 

**New behavior :** Removing `name` from CardinalityValue data structure to optimize the storage

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/ratelimit/RocksDbCardinalityStore.scala
+++ b/core/src/main/scala/filodb.core/memstore/ratelimit/RocksDbCardinalityStore.scala
@@ -27,13 +27,12 @@ import filodb.memory.format.UnsafeUtils
  *   (1) only the least-significant prefix name is stored.
  *   (2) no shard ID is store
  */
-case class CardinalityValue(name: String, tsCount: Int, activeTsCount: Int,
+case class CardinalityValue(tsCount: Int, activeTsCount: Int,
                             childrenCount: Int, childrenQuota: Int)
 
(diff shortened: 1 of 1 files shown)


amolnayak311 (approved):
LGTM @vishramachandran do you recall why was this stored?

vishramachandran:
I dont recall fully, but here is my guess. At one point I think `CardinalityValue` was both stored and sent back to caller. `CardinalityRecord` I think was introduced later. @alextheimer can shed some light. I think he is the one that added `CardinalityValue`

alextheimer (approved):
`Cardinality` used to be the class we read from / wrote to RocksDB. That contained the `name` String. Later, we split `Cardinality` into two classes:
- `CardinalityValue`:  all the series counts to store in RocksDB
- `CardinalityRecord`: effectively a RocksDB `(key,CardinalityValue)` pair (FWIW I can't tell why I didn't add a `CardinalityValue` field to `CardinalityRecord`).

`CardinalityValue` only contains a `name` because `Cardinality` did, but it looks like this PR is the correct move.

sandeep6189:
@alextheimer you are right. We can simplify that aspect as well. Let me pick this up in the next PR. Closing this PR for now :)

PR #1586, opened 2023-05-04
fix(core) add RecordOutOfContainerCapacityException.

yu-shipit:
Add a more user friendly message for exceptions happen when queries involved with records too big.

**Pull Request checklist**

**New behavior :**

**BREAKING CHANGES**

Breaking changes may include:

**Other information**:

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
@@ -248,7 +248,7 @@ final class QueryActor(memStore: TimeSeriesStore,
           logger.warn(s"Query Limit Breached " +
             s"${q.queryContext.origQueryParams} ${t.getMessage}")
         case _: RecordOutOfContainerCapacityException =>
-          logger.warn(s"Query is greater than container capacity" +
+          logger.warn(s"The intermediate or final result of the Query is greater than container capacity" +
             s"${q.queryContext.origQueryParams} ${t.getMessage}")
         case e: Throwable =>
           logger.error(s"Query Error queryId=${q.queryContext.queryId} " +
--- a/coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
@@ -2,12 +2,10 @@ package filodb.coordinator
 
 import java.lang.Thread.UncaughtExceptionHandler
 import java.util.concurrent.{ForkJoinPool, ForkJoinWorkerThread}
-
 import scala.collection.mutable
 import scala.concurrent.duration.{DurationInt, FiniteDuration}
 import scala.util.{Failure, Success}
 import scala.util.control.NonFatal
-
 import akka.actor.{ActorRef, Props}
 import akka.pattern.AskTimeoutException
 import kamon.Kamon
@@ -31,6 +29,7 @@ import filodb.core.query.QueryContext
 import filodb.core.query.QueryLimitException
 import filodb.core.query.QuerySession
 import filodb.core.query.QueryStats
+import filodb.core.query.RecordOutOfContainerCapacityException
 import filodb.core.query.SerializedRangeVector
 import filodb.core.store.CorruptVectorException
 import filodb.memory.data.Shutdown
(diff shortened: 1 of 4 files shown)


core/src/main/scala/filodb.core/binaryrecord2/RecordBuilder.scala
```diff
@@ -540,7 +540,8 @@ class RecordBuilder(memFactory: MemFactory,
       val oldOffset = curRecordOffset
       if (reuseOneContainer) resetContainerPointers() else newContainer()
       logger.trace(s"Moving $recordNumBytes bytes from end of old container to new container")
-      require((containerSize - ContainerHeaderLen) > (recordNumBytes + numBytes), "Record too big for container")
+      if((containerSize - ContainerHeaderLen) <= (recordNumBytes + numBytes))
+        throw new RecordOutOfContainerCapacityException();
```
sandeep6189, line 544: Please add unit test and integration tests for this scenario

coordinator/src/main/scala/filodb.coordinator/QueryActor.scala
```diff
@@ -248,6 +247,9 @@ final class QueryActor(memStore: TimeSeriesStore,
         case _: QueryLimitException =>
           logger.warn(s"Query Limit Breached " +
             s"${q.queryContext.origQueryParams} ${t.getMessage}")
+        case _: RecordOutOfContainerCapacityException =>
+          logger.warn(s"Query is greater than container capacity" +
```
amolnayak311, line 251: Consider rephrasing "Query is greater than container capacity"

core/src/main/scala/filodb.core/binaryrecord2/RecordBuilder.scala
```diff
@@ -540,7 +540,8 @@ class RecordBuilder(memFactory: MemFactory,
       val oldOffset = curRecordOffset
       if (reuseOneContainer) resetContainerPointers() else newContainer()
       logger.trace(s"Moving $recordNumBytes bytes from end of old container to new container")
-      require((containerSize - ContainerHeaderLen) > (recordNumBytes + numBytes), "Record too big for container")
+      if((containerSize - ContainerHeaderLen) <= (recordNumBytes + numBytes))
```
amolnayak311, line 543: Update the readme as technically this is a breaking change as we change the exception type from ``IllegalArgumentException``thrown by ``require``  to ``RecordOutOfContainerCapacityException`` which extends  ``RuntimeException``

core/src/main/scala/filodb.core/query/ResultTypes.scala
```diff
@@ -177,3 +177,9 @@ class QueryLimitException(message: String, queryId: String) extends RuntimeExcep
     s"${super.getMessage}, queryId=${queryId}"
   }
 }
+
+/**
+ * Thrown when number of bytes is greater than container.
+ */
+class RecordOutOfContainerCapacityException() extends RuntimeException(
```
amolnayak311, line 184: Now that we have an exception, it would be nice to know what's the available capacity and what's the requested capacity

amolnayak311 (changes requested):

tjackpaul (changes requested):
We cannot move further with this PR since this RecordBuilder is shared across both ingestion and query layer.

core/src/main/scala/filodb.core/query/ResultTypes.scala
```diff
@@ -177,3 +177,9 @@ class QueryLimitException(message: String, queryId: String) extends RuntimeExcep
...
+  "The intermediate or final result for the query is too big. Please try to add more query filters or time range.") {}
```
tjackpaul, line 185: This is not true.

amolnayak311:
> We cannot move further with this PR since this RecordBuilder is shared across both ingestion and query layer.

@tjackpaul Why would this be a problem with this PR? This shouldn't be a new failure unless you expect ``IllegalArgumentException`` which I raised as a breaking change.

core/src/main/scala/filodb.core/query/ResultTypes.scala
amolnayak311: Agreed, the message just assumes this is used for querying.

tjackpaul:
> > We cannot move further with this PR since this RecordBuilder is shared across both ingestion and query layer.
> 
> @tjackpaul Why would this be a problem with this PR? This shouldn't be a new failure unless you expect `IllegalArgumentException` which I raised as a breaking change.

I meant with that error message, we cannot proceed further.

core/src/main/scala/filodb.core/query/ResultTypes.scala
```diff
@@ -181,5 +181,7 @@ class QueryLimitException(message: String, queryId: String) extends RuntimeExcep
 /**
  * Thrown when number of bytes is greater than container.
  */
-class RecordOutOfContainerCapacityException() extends RuntimeException(
-  "The intermediate or final result for the query is too big. Please try to add more query filters or time range.") {}
+class RecordOutOfContainerCapacityException(message: String) extends IllegalArgumentException(message) {}
+
+class QueryRecordOutOfContainerCapacityException() extends RecordOutOfContainerCapacityException(
```
amolnayak311, line 186: What's the point of having two different exceptions?

amolnayak311 (changes requested):

amolnayak311:
Additionally, please fix the checkstyle error.

PR #1592, opened 2023-05-17
Changes in CardinalityTracker to store cardinality count in Downsample cluster

sandeep6189:
**Pull Request checklist**

**Current behavior :**  CardinalityTracker support delta of cardinality counts in values of 0 or 1. All these updates is stored to RocksDB when the modifyCount is called. 

However, in downsample cluster, we have a burst write scenario, since we read all the partKeys from the index and update the cardinality count from scratch. With the existing `modifyCount` method, we are running into heavy disk write scenarios which increases the time needed to measure cardinality in downsample clusters.

**New behavior :** To mitigate the above scenario, I am adding additional data-structures in CardinalityTracker, which stores aggregates count in memory and flushes to RocksDB on a static threshold (to avoid storing all cardinality records in memory and to avoid the memory pressure).

**Test Results:** We ran a test on a kube pod, with `cpu : 1` and `memory: 2Gi` and following were the time taken to calculate cardinality - 

[With InMemoryAggregation] DocCount: 20296704 | TotalBytes: 375490632708 | TotalBytesInGB: 375.490632708 | **DurationInSecondsToCalculateCardinality: 160** | DurationSecondsScanRocksDB: 0

[WithOut Aggregation] DocCount: 20296704 | TotalBytes: 375490632708 | TotalBytesInGB: 375.490632708 | **DurationInSecondsToCalculateCardinality: 1162** | DurationSecondsScanRocksDB: 0

### DIFF
--- a/core/src/main/scala/filodb.core/binaryrecord2/RecordSchema.scala
+++ b/core/src/main/scala/filodb.core/binaryrecord2/RecordSchema.scala
@@ -263,7 +263,7 @@ final class RecordSchema(val columns: Seq[ColumnInfo],
   }
 
   /**
-    * EXPENSIVE to do at server side. Creates a stringified map with contents of this BinaryRecord.
+    * EXPENSIVE to do at server side. Creates a stringified array with contents of this BinaryRecord.
     */
   def toStringPairs(base: Any, offset: Long): Seq[(String, String)] = {
     import Column.ColumnType._
@@ -288,6 +288,47 @@ final class RecordSchema(val columns: Seq[ColumnInfo],
     }
     result
   }
+
+
+  /**
+   * Reads all the label key-value pairs from the BinaryRecord and creates a map.
+   * If multiple labels with same key --> then the last writer wins.
+   * @param base data source
+   * @param offset offset to read from
+   * @return Map of all the tags/labels. key is label-name, value is label-value
+   */
+  def toStringPairsMap(base: Any, offset: Long): mutable.Map[String, String] = {
+    import Column.ColumnType._
+    val result = mutable.Map[String, String]()
+    columnTypes.zipWithIndex.foreach {
+      case (IntColumn, i) => result(colNames(i)) = getInt(base, offset, i).toString
+      case (LongColumn, i) => result(colNames(i)) = getLong(base, offset, i).toString
+      case (DoubleColumn, i) => result(colNames(i)) = getDouble(base, offset, i).toString
+      case (StringColumn, i) => result(colNames(i)) = asJavaString(base, offset, i)
+      case (TimestampColumn, i) => result(colNames(i)) = getLong(base, offset, i).toString
+      case (MapColumn, i) => val consumer = new StringifyMapItemConsumer
+        consumeMapItems(base, offset, i, consumer)
+        result ++= consumer.stringPairs
+      case (BinaryRecordColumn, i) => result ++= brSchema(i).toStringPairs(blobBase(base, offset, i),
+        blobOffset(base, offset, i))
+        result += ("_type_" ->
+          Schemas.global.schemaName(
+            RecordSchema.schemaID(blobBase(base, offset, i),
+              blobOffset(base, offset, i))))
+      case (HistogramColumn, i) =>
+        result(colNames(i)) = bv.BinaryHistogram.BinHistogram(blobAsBuffer(base, offset, i)).toString
+    }
+    result
+  }
+
+  /**
+   * Reads all the label key-value pairs from the BinaryRecord and creates a map.
+   * If multiple labels with same key --> then the last writer wins.
+   * @param bytes data source
+   * @return Map of all the tags/labels. key is label-name, value is label-value.
+   */
+  def toStringPairsMap(bytes: Array[Byte]): mutable.Map[String, String] =
+    toStringPairsMap(bytes, UnsafeUtils.arayOffset)
 
   def colNames(base: Any, offset: Long): Seq[String] = {
     import Column.ColumnType._
--- a/core/src/main/scala/filodb.core/binaryrecord2/RecordSchema.scala
+++ b/core/src/main/scala/filodb.core/binaryrecord2/RecordSchema.scala
@@ -263,7 +263,7 @@ final class RecordSchema(val columns: Seq[ColumnInfo],
   }
 
   /**
-    * EXPENSIVE to do at server side. Creates a stringified array with contents of this BinaryRecord.
+    * EXPENSIVE to do at server side. Creates a stringified map with contents of this BinaryRecord.
     */
   def toStringPairs(base: Any, offset: Long): Seq[(String, String)] = {
     import Column.ColumnType._
@@ -288,47 +288,6 @@ final class RecordSchema(val columns: Seq[ColumnInfo],
     }
     result
   }
-
-
-  /**
-   * Reads all the label key-value pairs from the BinaryRecord and creates a map.
-   * If multiple labels with same key --> then the last writer wins.
-   * @param base data source
-   * @param offset offset to read from
-   * @return Map of all the tags/labels. key is label-name, value is label-value
-   */
-  def toStringPairsMap(base: Any, offset: Long): mutable.Map[String, String] = {
-    import Column.ColumnType._
-    val result = mutable.Map[String, String]()
-    columnTypes.zipWithIndex.foreach {
-      case (IntColumn, i) => result(colNames(i)) = getInt(base, offset, i).toString
-      case (LongColumn, i) => result(colNames(i)) = getLong(base, offset, i).toString
-      case (DoubleColumn, i) => result(colNames(i)) = getDouble(base, offset, i).toString
-      case (StringColumn, i) => result(colNames(i)) = asJavaString(base, offset, i)
-      case (TimestampColumn, i) => result(colNames(i)) = getLong(base, offset, i).toString
-      case (MapColumn, i) => val consumer = new StringifyMapItemConsumer
-        consumeMapItems(base, offset, i, consumer)
-        result ++= consumer.stringPairs
-      case (BinaryRecordColumn, i) => result ++= brSchema(i).toStringPairs(blobBase(base, offset, i),
-        blobOffset(base, offset, i))
-        result += ("_type_" ->
-          Schemas.global.schemaName(
-            RecordSchema.schemaID(blobBase(base, offset, i),
-              blobOffset(base, offset, i))))
-      case (HistogramColumn, i) =>
-        result(colNames(i)) = bv.BinaryHistogram.BinHistogram(blobAsBuffer(base, offset, i)).toString
-    }
-    result
-  }
-
-  /**
-   * Reads all the label key-value pairs from the BinaryRecord and creates a map.
-   * If multiple labels with same key --> then the last writer wins.
-   * @param bytes data source
-   * @return Map of all the tags/labels. key is label-name, value is label-value.
-   */
-  def toStringPairsMap(bytes: Array[Byte]): mutable.Map[String, String] =
-    toStringPairsMap(bytes, UnsafeUtils.arayOffset)
 
   def colNames(base: Any, offset: Long): Seq[String] = {
     import Column.ColumnType._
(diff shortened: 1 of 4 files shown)


sandeep6189:
Test this please

core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityTracker.scala
```diff
@@ -102,14 +114,104 @@ class CardinalityTracker(ref: DatasetRef,
...
+
+  /**
+   * Used to store the cardinality count for the given prefix in the downsample cluster.
+   * NOTE: In downsample cluster, tsCount == activeTsCount. So totalDelta and activeDelta is same
+   * @param prefix usually contains labels _ws_, _ns_, _metric_ and different combinations of it
+   * @param totalDelta Increase in total timeseries
+   * @param activeDelta Increase in active timeseries
+   * @param childrenDelta Increase in children count
+   */
+  private def modifyCountDS(prefix: Seq[String],
```
yu-shipit, line 194: it seems the same key would be updated multiple time. And each update requires a read before it. This can cause read amplification and compaction in rocksdb.

If the performance is a real concern, I would suggest aggregate the value for the same key before put it into rocksdb.

core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityTracker.scala
```diff
@@ -102,14 +114,104 @@ class CardinalityTracker(ref: DatasetRef,
...
+                            totalDelta: Int, activeDelta: Int, childrenDelta: Int): Unit = synchronized {
```
yu-shipit, line 195: rocksdb should be thread safe.
I think the race here is that different threads may updating the same key.
The lock granularity could be key level.

core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityTracker.scala
sandeep6189: This method is only called by `updateCardinalityCountsDS` which aggregates the count before storing in the RocksDB :)

core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityTracker.scala
sandeep6189: We are not planning on calling this from multiple threads as of now, but just to be safe we added a function scope synchronized. Can be optimized, when we plan to call this from multiple threads.

alextheimer: Just to be safe, can we add `synchronized` to all methods that directly read from / write to the `Map`? (Especially since we don't expect contention to be an issue-- we will future-proof ourselves from race conditions with very little cost).

alextheimer (dismissed):
Those perf results look awesome-- thanks @sandeep6189 for this PR.

core/src/main/scala/filodb.core/binaryrecord2/RecordSchema.scala
```diff
@@ -289,6 +289,47 @@ final class RecordSchema(val columns: Seq[ColumnInfo],
...
 
+
+  /**
+   * Reads all the label key-value pairs from the BinaryRecord and creates a map.
+   * If multiple labels with same key --> then the last writer wins.
+   * @param base data source
+   * @param offset offset to read from
+   * @return Map of all the tags/labels. key is label-name, value is label-value
+   */
+  def toStringPairsMap(base: Any, offset: Long): mutable.Map[String, String] = {
```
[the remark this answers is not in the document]
sandeep6189: I don't actually, but wanted to keep it consistent with `toStringPairs` but also agree that it is wasteful .. should I rename it to `getStringValues` and only fetch for columnTypes `StringColumn` and `MapColumn` ? ( since I only need those in downsample cardinality .. on the other hand, we use it when going over the partKeys, which only contain the tags and metric name)

core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityTracker.scala
sandeep6189: Makes sense. let me do this

core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityTracker.scala
```diff
@@ -200,5 +306,13 @@ class CardinalityTracker(ref: DatasetRef,
 
   def close(): Unit = {
     store.close()
+    cardinalityCountMapDS.clear()
```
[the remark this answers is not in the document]
sandeep6189: Thought about this, but store.close() removes the database itself, so didn't think it was necessary to flush

core/src/main/scala/filodb.core/binaryrecord2/RecordSchema.scala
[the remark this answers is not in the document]
sandeep6189: @alextheimer let me know your thoughts ?

core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityTracker.scala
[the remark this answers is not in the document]
yu-shipit: Got it. the close function is a bit misleading. This close does not only close but also delete the db.

core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityTracker.scala
```diff
@@ -37,7 +38,8 @@ class CardinalityTracker(ref: DatasetRef,
                          shardKeyLen: Int,
                          defaultChildrenQuota: Seq[Int],
                          val store: CardinalityStore,
-                         quotaExceededProtocol: QuotaExceededProtocol = NoActionQuotaProtocol) extends StrictLogging {
+                         quotaExceededProtocol: QuotaExceededProtocol = NoActionQuotaProtocol,
+                         dsCardinalityMapFlushCount: Int = 5000) extends StrictLogging {
```
amolnayak311, line 42: 1. Rename to `flushThreshold`` which can be ``Option``, if present will flush using that threshold or continue the existing path (the one used in raw)
2. We are leaking the cluster info in this class. It is ok to have flush thresholds but no references to DS or any cluster in this class, which includes repurposing the existing method to change behavior with and without flush threshold.

core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityTracker.scala
```diff
@@ -46,6 +48,16 @@ class CardinalityTracker(ref: DatasetRef,
...
 
+  /**
+   * Map used to track cardinality count in downsample cluster.
+   * WHY this is used for Downsample cardinality count only ?
+   * This is because, in downsample cluster, we read through the entire index periodically and re-calculate
+   * the cardinality count from scratch. Without an in-memory aggregation data-structure, we would be modifying
+   * the records in RocksDB too frequently, causing a slowdown because of heavy disk writes.
+   * Hence we are using this map to help us aggregate in memory and then flush to RocksDB periodically
+   */
+  private val cardinalityCountMapDS : Map[Seq[String], (Int, Int)] = Map()
```
amolnayak311, line 59: No no to mutable member variables. We can not only use immutable structures with constructs like ``fold`` and ``reduce`` but pass those as parameters to flush.

core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityTracker.scala
amolnayak311: While the performance when you tested is good, it's not guaranteed ordering of ``partKeys``  when we batch. So to @yu-shipit 's point you may see a key at least for a ws/ns after flush?  But not too worried about it at the same time,

core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityTracker.scala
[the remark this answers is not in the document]
amolnayak311: 1. ``cardinalityCountMapDS`` should not be a member variable as in previous comment
2. Even without changes, the map should always be empty, if not it means they aren't flushed.

amolnayak311 (changes requested):
Thanks for the PR and fixing the action. Left some comments

amolnayak311 (changes requested):
Thanks for the PR and fixing the action. Left some comments

core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityTracker.scala
sandeep6189: Yeah, the implementation doesn't make any assumption of ordering guarantees and same/keys prefix will most likely happen again. Chose this implementation to avoid any complexity and keep implementation plain and simple ( no sort or other transformation. The reason for threshold and guards here is to avoid runaway use of memory )

core/src/main/scala/filodb.core/binaryrecord2/RecordSchema.scala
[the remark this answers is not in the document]
sandeep6189: Initially I did that, but then I wanted to avoid two memory allocations. First creation of array and then convert to map ( not sure if Scala does optimization underneath to avoid it, but since they were two separate data-structures, I thought this would be to separate allocations )

core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityTracker.scala
```diff
@@ -102,14 +114,106 @@ class CardinalityTracker(ref: DatasetRef,
...
+          quotaExceededProtocol.quotaExceeded(ref, shard, prefix, childrenQuota)
+          throw QuotaReachedException(prefix, prefix, childrenQuota)
+        }
+
+        // store the updated parent's childrenCount
+        cardinalityCountMapDS.put(parentPrefix, updatedCountRecord)
+      }
+    }
+
+    if (cardinalityCountMapDS.size > dsCardinalityMapFlushCount) {
```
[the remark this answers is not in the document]
sandeep6189: 1. Yeah, the implementation doesn't make any assumption of ordering guarantees and same/keys prefix will most likely happen again. Chose this implementation to avoid any complexity and keep implementation plain and simple ( no sort or other transformation. The reason for threshold and guards here is to avoid runaway use of memory )

2. Yep, I was thinking of calling the flush from the invoker/caller which starts the cardinality count in DS

core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityTracker.scala
sandeep6189: Sure @amolnayak311, could we have an offline sync on this quickly ? I wanted to understand this a bit for my understanding. 

While implementation, I was going through Scala collection doc and they mention the following and I wanted to avoid creating a new collection.

"Scala collections systematically distinguish between mutable and immutable collections. A mutable collection can be updated, reduced or extended in place. This means you can change, add, or remove elements of a collection as a side effect. Immutable collections, by contrast, never change. You have still operations that simulate additions, removals, or updates, but those operations will in each case return a new collection and leave the old collection unchanged."

core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityTracker.scala
sandeep6189: Thank you :) Agree with you to avoid leaking of cluster info, so renaming the functions and variables. I do have some follow up questions, will follow up offline on those.

core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityTracker.scala
```diff
@@ -190,6 +295,9 @@ class CardinalityTracker(ref: DatasetRef,
     require(depth >= shardKeyPrefix.size,
       s"depth $depth must be at least the size of the prefix ${shardKeyPrefix.size}")
 
+    // flush any pending key value pairs to rocksDB ( useful in down-sample cardinality )
+    flushCardinalityDataToRocksDB()
```
sandeep6189, line 299: Remove it

amolnayak311 (approved):
LGTM. Thanks  for the changes @sandeep6189

PR #1594, opened 2023-05-21
feat(query): target-schema pushdown in ShardKeyRegexPlanner

alextheimer:
**Pull Request checklist**

Currently, target-schemas are not pushed-down in the `ShardKeyRegexPlanner`. This means that queries such as:
```
foo{shard-key-label=~".*"} + on(shard-key-label) foo{shard-key-label=~".*"}
```
will be materialized as e.g.:
```
E~BinaryJoinExec(binaryOp=ADD, on=List(tschemaLabel), ignoring=List()) on InProcessPlanDispatcher
-E~MultiPartitionDistConcatExec() on InProcessPlanDispatcher
  ...
-E~MultiPartitionDistConcatExec() on InProcessPlanDispatcher
  ...
```
When this happens, query limits (that would have been sufficiently high had the query been correctly pushed down) may be breached.

This PR adds logic to support the pushdown optimization in the `ShardKeyRegexPlanner`. Since it mirrored existing logic in the `SingleClusterPlanner`, a generic `getPushdownKeys` method has been created in `LogicalPlanUtils`.


coordinator/src/test/scala/filodb.coordinator/queryplanner/ShardKeyRegexPlannerSpec.scala
```diff
@@ -603,7 +603,9 @@ class ShardKeyRegexPlannerSpec extends AnyFunSpec with Matchers with ScalaFuture
       case _ => fail("Expected AggregateMapReduce for the sum operation")
     }
     mpExec.children match {
-      case plan1::plan2::Nil =>
+      case children: Seq[ExecPlan] if children.size == 2 =>
+        val plan1 = children(0)
+        val plan2 = children(1)
```
alextheimer, line 608: These did not want to work for me. I think it's because `mpExec.children` is now a `Vector` under-the-hood (unclear why that change came about).

query/src/test/scala/filodb/query/PlanValidationSpec.scala
```diff
@@ -1,24 +1,66 @@
...
+  private def removeNoise(planString: String): String = {
+    planString
+      .replaceAll("testProbe-.*]", "testActor]")
+      .replaceAll("InProcessPlanDispatcher.*\\)", "InProcessPlanDispatcher")
+  }
+
+  /**
+   * Prints the same tree for all logically-identical plans.
+   */
+  private def sortTree(planString: String): Unit = {
```
alextheimer, line 23: I got tired of tests failing due to children randomly getting swapped in `ExecPlan::printTree`. This is disabled by default, but you can use this to consistently sort the `String` trees before they're passed to `shouldEqual`.

sherali42 (approved):
thanks @alextheimer, changes look good 👍 
nitpick: can we add a comment to `canPushdown`? since it is similar logic, changes to one method should flow to another method as well. Or we can do templateMethod style here.

alextheimer:
> nitpick: can we add a comment to `canPushdown`? since it is similar logic, changes to one method should flow to another method as well. Or we can do templateMethod style here.

These felt "different enough" that combining them would overcomplicate things, but I think you're right that this is way better than before. @sherali42 can you re-approve?

sherali42 (approved):
thanks @alextheimer looks good 👍

PR #1595, opened 2023-05-22
feat(query): adding warnings to QuerySession

kvpetrov:
**Pull Request checklist**

Currently as queries get executed there is no way to convey to the user that the query runs close to the maximum capacity of FiloDB.

**New behavior :**
This PR introduces an object allowing to pass the information back to the user.

**BREAKING CHANGES**

no breaking changes

**Other information**:

### DIFF
--- a/core/src/main/scala/filodb.core/query/QueryContext.scala
+++ b/core/src/main/scala/filodb.core/query/QueryContext.scala
@@ -104,13 +104,16 @@ case class QueryWarnings(
   }
 
   override def equals(w2Compare: Any): Boolean = {
-    val w2 = w2Compare.asInstanceOf[QueryWarnings]
-    execPlanSamples.get().equals(w2.execPlanSamples.get()) &&
-    execPlanResultBytes.get().equals(w2.execPlanResultBytes.get()) &&
-    groupByCardinality.get().equals(w2.groupByCardinality.get()) &&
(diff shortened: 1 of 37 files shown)


core/src/main/scala/filodb.core/query/QueryContext.scala
```diff
@@ -48,6 +48,82 @@ object PerQueryLimits {
...
+  def updateTimeSeriesScanned(series: Int): Unit = {
+    timeSeriesScanned.updateAndGet(s => if (s<series) series else s)
+  }
+
+  def updateTimeSeriesSampleScannedBytes(bytes: Long): Unit = {
+    timeSeriesSamplesScannedBytes.updateAndGet(b => if (b<bytes) bytes else b)
+  }
+
+  override def equals(w2Compare: Any): Boolean = {
+    val w2 = w2Compare.asInstanceOf[QueryWarnings]
```
yu-shipit, line 107: should you return false if w2Compare is other type

yu-shipit (approved):
lgtm

core/src/main/scala/filodb.core/query/QueryContext.scala
kvpetrov: it would throw an exception (which I thought is perfectly fine) but checked the equals implementation in java and decided to do the same thing, ie return false

vishramachandran (approved):
LGTM.

PR #1597, opened 2023-05-26
fix(export): re-add double-quotes around label-values with commas

alextheimer:
**Pull Request checklist**

Currently, we prepare series for export by wrapping comma-containing label-values in single-quotes. This PR instead wraps them in double-quotes.

### DIFF
--- a/spark-jobs/src/main/scala/filodb/downsampler/chunk/BatchExporter.scala
+++ b/spark-jobs/src/main/scala/filodb/downsampler/chunk/BatchExporter.scala
@@ -18,7 +18,7 @@ import filodb.core.query.ColumnFilter
 import filodb.core.store.{ChunkSetInfoReader, ReadablePartition}
 import filodb.downsampler.DownsamplerContext
 import filodb.downsampler.Utils._
-import filodb.downsampler.chunk.BatchExporter.{getExportLabelValueString, DATE_REGEX_MATCHER, LABEL_REGEX_MATCHER}
+import filodb.downsampler.chunk.BatchExporter.{DATE_REGEX_MATCHER, LABEL_REGEX_MATCHER}
 import filodb.memory.format.{TypedIterator, UnsafeUtils}
 import filodb.memory.format.vectors.LongIterator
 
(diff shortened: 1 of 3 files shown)


spark-jobs/src/main/scala/filodb/downsampler/chunk/BatchExporter.scala
```diff
@@ -218,10 +209,13 @@ case class BatchExporter(downsamplerSettings: DownsamplerSettings, userStartTime
   }
 
   private def makeLabelString(labels: collection.Map[String, String]): String = {
-    val inner = labels
-      .map {case (k, v) => (k, getExportLabelValueString(v))}
-      .map {case (k, v) => s"'$k':'$v'"}
-      .mkString (",")
+    val inner = labels.map {case (k, v) =>
+      if (v.contains (",") ) {
+        String.format ("'%s\':\"%s\"", k, v)
```
yu-shipit, line 214: why the first ' is not escaped?
should it be?
"\'%s\':\"%s\""

sandeep6189 (approved):

yu-shipit (changes requested):

spark-jobs/src/main/scala/filodb/downsampler/chunk/BatchExporter.scala
alextheimer: Good catch-- I'm not sure (the escape may not be needed for the one single-quote). But I will leave this as-is for now since we know it works (we've previously merged this exact diff, but the changes were lost after a merge from `integration` to `main`).

yu-shipit: or should it be String.format ("'%s':\"%s\"", k, v) ?

alextheimer: You may be correct, but I won't touch it for now since this exact diff has worked well in the past.

yu-shipit (approved):

PR #1600, opened 2023-06-01
fix(query) gRPC exception serialization and container size fixes

amolnayak311:
**Pull Request checklist**

**Current behavior :** 

- No configuration available for gRPC max inbound message size
- Fixed 4k record container size
- Deserializing  ``QueryLimitException`` gives ``Throwable``

**New behavior :**

- Adds default max incoming message size allows configuration of this value
- Default 40k record container size for ``TopK``, ``BottomK`` and ``CountValues``. Allows overriding these values in config
- Exception type preserved in deserialization of  ``QueryLimitException``

### DIFF
--- a/core/src/main/resources/filodb-defaults.conf
+++ b/core/src/main/resources/filodb-defaults.conf
@@ -37,6 +37,10 @@ filodb {
       keep-alive-timeout-seconds = 0  # keep alive ack will timeout and connection closed  if no response is seen in this time
       # Refer to https://github.com/grpc/grpc/blob/master/doc/load-balancing.md
       load-balancing-policy = "round_robin"
+      # Maximum permissible inbound message size in gRPC, defaults to 100MB. This is more than
+      # enough from the default 4MB, ideally the query response should not be this large and the other overrides for
+      # query response size should have kicked in first to fail the query.
+      max-inbound-message-size = 104857600
   }
 
   # list of paths to dataset + ingestion config files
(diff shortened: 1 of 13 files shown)


core/src/main/resources/filodb-defaults.conf
```diff
@@ -37,6 +37,10 @@ filodb {
       keep-alive-timeout-seconds = 0  # keep alive ack will timeout and connection closed  if no response is seen in this time
       # Refer to https://github.com/grpc/grpc/blob/master/doc/load-balancing.md
       load-balancing-policy = "round_robin"
+      # Maximum permissible inbound message size in gRPC, defaults to 100MB. This is more than
+      # enough from the default 4MB, ideally the query response should not be this large and the other overrides for
+      # query response size should have kicked in first to fail the query.
+      max-inbound-message-size = 104857600
```
vishramachandran, line 43: Did we evaluate compression in GRPC?

amolnayak311: No, not yet, but most of our bytes are taken up by record container arrays and i'm not sure if I can expect to see large compression ratio on those in gRPC unless the container itself is compressed. In either case 4 MB limit might to be low especially once we consider sending multiple range vectors as part of a single streaming response message and we need a way to control this message size. The limit is in place to prevent a malicious sender from sending large messages causing receiver from going OOM.

vishramachandran: Current limit in Akka is 25MB after lz4 compression. This is for all SRVs put together though.
GRPC limits should be a balance between achieving parity and maintaining performance.

tjackpaul (approved):

sherali42 (approved):
👍 thanks @amolnayak311 for the quick fix

core/src/main/resources/filodb-defaults.conf
vishramachandran: Perhaps we need a size based batching rather than RV-count based batching

PR #1601, opened 2023-06-02
fix(query) Configurable record container sizes for topK, bottomK and countValues

amolnayak311:
**Pull Request checklist**

**Current behavior :** 
We use a fixed 4k buffer for record containers 

**New behavior :**

We change the container size 40k or any other configurable value for ``Topk``, ``BottomK`` and ``CountValue`` aggregation

### DIFF
--- a/core/src/main/resources/filodb-defaults.conf
+++ b/core/src/main/resources/filodb-defaults.conf
@@ -323,7 +323,7 @@ filodb {
     # Choices are "legacy", "antlr", and "shadow". Shadow mode uses legacy but also checks antlr for errors.
     parser = "antlr"
     container-size-overrides {
-        filodb-query-exec-localpartitionreduceaggregateexec-topbottomk = 40960
+        filodb-query-exec-aggregate-large-container                    = 65536
         filodb-query-exec-metadataexec                                 = 65536
     }
     grpc {
--- a/core/src/main/resources/filodb-defaults.conf
+++ b/core/src/main/resources/filodb-defaults.conf
@@ -324,6 +324,7 @@ filodb {
     parser = "antlr"
     container-size-overrides {
         filodb-query-exec-localpartitionreduceaggregateexec-topbottomk = 40960
+        filodb-query-exec-metadataexec                                 = 65536
     }
     grpc {
       # Override used to disable call to the multi partition calls over gRPC. Specify comma separated partition names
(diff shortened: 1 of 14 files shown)


core/src/main/resources/filodb-defaults.conf
```diff
@@ -324,6 +324,7 @@ filodb {
     parser = "antlr"
     container-size-overrides {
         filodb-query-exec-localpartitionreduceaggregateexec-topbottomk = 40960
+        filodb-query-exec-metadataexec                                 = 65536
```
vishramachandran, line 327: Use same 64k override for both ?

core/src/main/scala/filodb.core/query/QueryConfig.scala
```diff
@@ -51,7 +51,11 @@ object QueryConfig {
                                            remoteGrpcEndpoint = None,
                                            enforceResultByteLimit = false,
                                            allowPartialResultsRangeQuery = false,
-                                           allowPartialResultsMetadataQuery = true)
+                                           allowPartialResultsMetadataQuery = true,
+                                           recordContainerOverrides =
+                                             Map("filodb-query-exec-localpartitionreduceaggregateexec-topbottomk"
+                                                     -> 40960,
+                                                  "filodb-query-exec-metadataexec" -> 8192))
```
vishramachandran, line 58: Shouldn't you be using the `container-size-overrides` config ?

query/src/main/scala/filodb/query/exec/ExecPlan.scala
```diff
@@ -60,7 +60,33 @@ trait ExecPlan extends QueryCommand {
...
    */
-  def maxRecordContainerSize: Int = SerializedRangeVector.MaxContainerSize
+  def maxRecordContainerSize(cfg: QueryConfig): Int = {
+    this match {
+      // If there is a ReduceAggregateExec and aggregation is one of TopK, BottomK or CountValues,
+      // use large record container
+      case ra: ReduceAggregateExec
+                    if ra.aggrOp == AggregationOperator.TopK  ||
+                       ra.aggrOp == AggregationOperator.BottomK ||
+                       ra.aggrOp == AggregationOperator.CountValues   =>
```
vishramachandran, line 70: Quantile Aggregation also takes space due to t-digests and should be included

vishramachandran (approved):
Approved with comments

core/src/main/scala/filodb.core/query/QueryConfig.scala
amolnayak311: Actually this is used for a unit where where construction should fail and when using 64 it doesn't.

vishramachandran (approved):

PR #1603, opened 2023-06-07
feat(downsample): Measurement of downsample cardinality and  publish ws+ns cardinality

sandeep6189:
**Pull Request checklist**

**Current behavior :** We don't measure the cardinality count in DownsampleTimeSeriesShard.

**New behavior :** Enabling the measurement of cardinality count in DownsampleTimeSeriesShard.


core/src/main/resources/filodb-defaults.conf
```diff
@@ -495,6 +495,9 @@ filodb {
       # 10000 * (1KB per col * 6 cols * 3 chunks) + buffer
       shard-mem-size = 400 MB
 
+      # switch to enable/disable calculation of cardinality in downsample cluster
+      metering-enabled = true
```
sandeep6189, line 499: @amolnayak311 @sherali42 @alextheimer  new to how we manage/add configs. Wondering if we also have to add this config in helm-charts somewhere ?

sandeep6189: Also, wondering if I should start with this config as false and override for certain tsdb partitions to start with ?

core/src/main/scala/filodb.core/downsample/DownsampleCardinalityManager.scala
```diff
@@ -0,0 +1,170 @@
...
+  // to different times of day
+  private val numShardsPerNode: Int = getNumShardsPerNodeFromConfig(rawDatasetRef.dataset, filodbConfig)
+
+  /**
+   * Helper method to get the shards per node in the downsample filodb cluster. We look for the two specific
+   * config that is `min-num-nodes` ( the minimum number of nodes/pods which has to be in a cluster ) and
+   * `num-shards` (the total number of shards). num-shards/min-num-nodes gives us the required value.
+   *
+   * NOTE: This configs can be part of `dataset-configs` or `inline-dataset-configs`. Also making sure, we look
+   * for the dataset, which was used to create this instance of DownsampleTimeSeriesShard
```
alextheimer, line 37: What is the conceptual difference between these two configs? What does "inline" imply?

core/src/main/scala/filodb.core/downsample/DownsampleCardinalityManager.scala
```diff
@@ -0,0 +1,170 @@
...
+   *
+   * @return shards present per node
+   */
+  def getNumShardsPerNodeFromConfig(datasetToSearch: String, filoConfig: Config): Int = {
```
alextheimer, line 41: Is this equivalent? https://github.com/filodb/FiloDB/blob/654b7b03c9f0dea8ec81185230deb84970be162d/coordinator/src/main/scala/filodb.coordinator/ShardAssignmentStrategy.scala#L238

alextheimer: (not sure yet whether-or-not we have access to a `DatasetResourceSpec`)

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -995,7 +995,7 @@ class CardinalityCountBuilder(partSchema: PartitionSchema, cardTracker: Cardinal
         binaryValue.bytes, unsafePkOffset, partSchema.options.shardKeyColumns)
 
       // update the cardinality count by 1, since the shardKey for each document in index is unique
-      cardTracker.modifyCount(shardKey, 1, 1)
```
alextheimer, line 998: Confirming this was intended for this PR?

core/src/main/scala/filodb.core/downsample/DownsampleCardinalityManager.scala
```diff
@@ -0,0 +1,170 @@
...
+   * calculation cost of cardinality count of different shards residing in the same node at different times of day.
+   * since cardinality count in downsample cluster is essentially a rebuild from scratch, which takes few mins. We don't
+   * want all the shards to be calculating this at the same time and using up all the resources of the node.
+   *
+   * @param currentShardNum Shard Number (0 - Num-Shards)
+   * @param shardsPerNode   Number of Shards per node
+   * @param currentHour     Current hour of the day (0-23)
+   * @return True if cardinality count is to be triggered, false otherwise
+   */
+  def shouldTriggerCardinalityCount(currentShardNum: Int, shardsPerNode: Int, currentHour: Int): Boolean = {
```
alextheimer, line 104: How about a single-threaded scheduler that iterates through all shards once every flush-interval duration? That way, there's no room for edge-cases where we miss shards.

core/src/main/scala/filodb.core/downsample/DownsampleCardinalityManager.scala
```diff
@@ -0,0 +1,170 @@
...
+   * and store data in a local CardinalityStore. We then close the previous instance of CardinalityTracker and switch
+   * it with the new one we created in this call.
+   */
+  def triggerCardinalityCount(): Unit = {
+    if (downsampleStoreConfig.meteringEnabled) {
+      try {
+        val currentHour = ((System.currentTimeMillis() / 1000 / 60 / 60) % 24).toInt
+        if (shouldTriggerCardinalityCount(shardNum, numShardsPerNode, currentHour)) {
+          // get a new cardinality tracker object and re-calculate the cardinality using this
+          val newCardTracker = getNewCardTracker()
```
alextheimer, line 122: Should there be a tracker for each shard? Won't this setup only allow count data for at most one shard at any given time?

core/src/main/scala/filodb.core/downsample/DownsampleCardinalityManager.scala
sandeep6189: So currently,  filodb-default configs have two sections where a dataset's config could be present. They are `dataset-configs` and `inline-dataset-configs`. That's why we are checking for both this sections of configs

core/src/main/scala/filodb.core/downsample/DownsampleCardinalityManager.scala
sandeep6189: Yes, it is equivalent. Thanks @alextheimer for the find.  Let me see if we have access to it ^^

core/src/main/scala/filodb.core/downsample/DownsampleCardinalityManager.scala
sandeep6189: Interesting point. Taking this discussion offline to learn more about it.

core/src/main/scala/filodb.core/downsample/DownsampleCardinalityManager.scala
sandeep6189: We will have a tracker for each shard ( same for raw and DS cluster, since we are aiming to keep the setup similar to what we have in raw ( to minimize the query plan changes while fetching the data ).  

The idea here is that we don't want the physical node to be overwhelmed when all shards calculate cardinality at the same time. Hence, we are adding a check, which would only allow 1 shard to calculate its cardinality at a given point in time.

core/src/main/scala/filodb.core/downsample/DownsampleCardinalityManager.scala
```diff
@@ -0,0 +1,169 @@
...
+      tracker.setQuota(q.shardKeyPrefix, q.quota)
+    }
+    tracker
+  }
+
+  /**
+   * Closes the cardinality tracker and clean/release all the physical resources used
+   */
+  def close(): Unit = {
+    cardTracker.close()
```
amolnayak311, line 167: This won't be available if cardinality tracking is not enabled.

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
amolnayak311: Can the Builder accept the deltas? That way we can reuse the collector if needed

core/src/main/scala/filodb.core/downsample/DownsampleCardinalityManager.scala
```diff
@@ -0,0 +1,169 @@
...
+   * `num-shards` (the total number of shards). num-shards/min-num-nodes gives us the required value.
+   *
+   * NOTE: This configs can be part of `dataset-configs` or `inline-dataset-configs`. Also making sure, we look
+   * for the dataset, which was used to create this instance of DownsampleTimeSeriesShard
+   *
+   * @return shards present per node
+   */
+  def getNumShardsPerNodeFromConfig(datasetToSearch: String, filoConfig: Config): Int = {
+    // init
+    var datasetConfig: Option[Config] = None
```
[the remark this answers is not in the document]
sandeep6189: Thank You !! Updated

core/src/main/scala/filodb.core/downsample/DownsampleCardinalityManager.scala
```diff
@@ -0,0 +1,169 @@
...
+import filodb.core.memstore.PartKeyLuceneIndex
+import filodb.core.memstore.ratelimit.{CardinalityTracker, QuotaSource, RocksDbCardinalityStore}
+import filodb.core.metadata.PartitionSchema
+import filodb.core.store.StoreConfig
+import filodb.memory.format.UnsafeUtils
+
+/**
+ * Manages the calculation and storage of cardinality count in a shard of downsample cluster
+ */
+class DownsampleCardinalityManager(rawDatasetRef: DatasetRef,
```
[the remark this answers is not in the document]
sandeep6189: Ack, will do

core/src/main/scala/filodb.core/downsample/DownsampleCardinalityManager.scala
```diff
@@ -0,0 +1,169 @@
...
+   * it with the new one we created in this call.
+   */
+  def triggerCardinalityCount(): Unit = {
+    if (downsampleStoreConfig.meteringEnabled) {
+      try {
+        val currentHour = ((System.currentTimeMillis() / 1000 / 60 / 60) % 24).toInt
+        if (shouldTriggerCardinalityCount(shardNum, numShardsPerNode, currentHour)) {
+          // get a new cardinality tracker object and re-calculate the cardinality using this
+          val newCardTracker = getNewCardTracker()
+          partKeyIndex.calculateCardinality(partSchema, newCardTracker)
```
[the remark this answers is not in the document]
sandeep6189: Good point, updated the failure semantics to capture this issue and cleanup resources

amolnayak311 (approved):
LGTM, Thanks for the changes, please follow [these](https://github.com/filodb/FiloDB/blob/develop/CONTRIBUTING.md#commit-message-format) guidelines for the PR title before merge

PR #1604, opened 2023-06-07
fix(core): Negative active time series cardnality due to missed increments

vishramachandran:
Added missing increments to activeTimeSeries count.

(Cherry-pick from develop)


sandeep6189 (approved):
LGTM. Cherry pick changes which is already merged in develop

amolnayak311 (approved):

PR #1607, opened 2023-06-14
perf(coord) ensure the first SRV does not contain all record containers

amolnayak311:
**Pull Request checklist**

**Current behavior :** 
While serializing range vectors the first ``SerializedRangeVector`` (SRVs) contains a reference to all the ``RecordContainers`` and when multiple SRVs are serialized the behavior causes two problems
-  The result bytes are double counted, once as part of the first SRV and another time as part of subsequent SRVs that contain just the required record containers
- When serializing using gRPC, the message may exceed the 4 MB limit and transmits twice the required bytes over the wire. 

The issue isn't seen as part of Kryo serializer as the serialization is smart enough to preserve the references and not serialize the container twice. Also on the receiving side the deserialized values reference the same Record container not creating duplicate containers in memory unlike in gRPC based serialization.  

The streaming response uses a new RecordContainer builder for each ``RangeVector`` and thus this issue will not show up.  

**New behavior :**

- The first SRV doesn't just hold the reference to all containers but we force a snapshot of the record containers those we need  to serialize as part of first SRV
- The result bytes in the stats will show the real number of bytes we serialize (uncompressed) which would roughly be half of what we send today

### DIFF
--- a/core/src/main/scala/filodb.core/query/RangeVector.scala
+++ b/core/src/main/scala/filodb.core/query/RangeVector.scala
@@ -471,7 +471,7 @@ object SerializedRangeVector extends StrictLogging {
       // If there weren't containers before, then take all of them.  If there were, discard earlier ones, just
       // start with the most recent one we started adding to
       val containers = oldContainerOpt match {
-        case None => builder.allContainers
+        case None => builder.allContainers.toList
         case Some(firstContainer) => builder.allContainers.dropWhile(_ != firstContainer)
       }
       val srv = new SerializedRangeVector(rv.key, numRows, containers, schema, startRecordNo, rv.outputRange)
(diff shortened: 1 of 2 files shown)


sandeep6189 (dismissed):

core/src/main/scala/filodb.core/query/RangeVector.scala
```diff
@@ -471,7 +471,7 @@ object SerializedRangeVector extends StrictLogging {
       // If there weren't containers before, then take all of them.  If there were, discard earlier ones, just
       // start with the most recent one we started adding to
       val containers = oldContainerOpt match {
-        case None => builder.allContainers
+        case None => builder.allContainers.toList
```
vishramachandran, line 474: Please add detailed explanation on why this is needed

vishramachandran (approved):

core/src/main/scala/filodb.core/query/RangeVector.scala
amolnayak311: Done

vishramachandran (approved):

PR #1609, opened 2023-06-14
fix(downsample): Fixing exception when getting num-shards from config

sandeep6189:
**Pull Request checklist**

Fixing exception when getting num-shards from config. 

*Behaviour Before:* wrongly interpreted the dataset-config to be an array of Config object. Instead it is an array of string, where each string is the file path to the dataset config. Because of this mis-interpretation, we were throwing an exception.

*Behaviour After:* Fixed this misinterpretation and parsing the dataset-configs from the file. Referenced the parsing behavior from FiloSettings.scala

- [x] Tests have been added / updated

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityManager.scala
+++ b/core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityManager.scala
@@ -42,47 +42,54 @@ class CardinalityManager(datasetRef: DatasetRef,
    * @return shards present per node
    */
   def getNumShardsPerNodeFromConfig(datasetToSearch: String, filoConfig: Config): Int = {
-    val datasetConfig: Option[Config] = if (filoConfig.hasPath("dataset-configs")) {
-      filoConfig.getConfigList("dataset-configs")
-        .toArray().toList
-        .asInstanceOf[List[Config]]
-        .filter(x => x.getString("dataset") == datasetToSearch)
-        .headOption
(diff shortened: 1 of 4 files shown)


core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityManager.scala
```diff
@@ -31,6 +32,22 @@ class CardinalityManager(datasetRef: DatasetRef,
...
+   * @param filoConfig
+   * @return
+   */
+  private def getDataSetConfigs(filoConfig: Config): Seq[Config] = {
+    val datasetConfPaths = filoConfig.as[Seq[String]]("dataset-configs")
+    if (datasetConfPaths.nonEmpty) {
+      datasetConfPaths.map { d => ConfigFactory.parseFile(new java.io.File(d)) }
+    } else {
+      Seq()
+    }
```
alextheimer, line 48: `datasetConfPaths.map { d => ConfigFactory.parseFile(new java.io.File(d)) }` should be all that's needed here, I think

alextheimer (approved):
LGTM 👍

core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityManager.scala
sandeep6189: Ack, updated :)

alextheimer (approved):

PR #1611, opened 2023-06-16
fix(query) Handle NaN values as part of histogram map response

amolnayak311:
**Pull Request checklist**

**Current behavior :** 
If histogram from a Remote exec contains a NaN, the decoding fails with an error like the following

```Caused by: io.circe.DecodingFailure$$anon$2: Double: DownField(0.001),MoveRight,DownArray,DownArray,DownField(values),DownArray,DownField(result),DownField(data)```

Specifically the history shows the bucket whose value is ``NaN`` in the response.

**New behavior :**

Instead of relying on circe to decode the ``Map[String, Double]`` we have implemented a decoder that attempts to parse the ``String`` value to ``NaN``.

### DIFF
--- a/query/src/main/scala/filodb/query/PromCirceSupport.scala
+++ b/query/src/main/scala/filodb/query/PromCirceSupport.scala
@@ -1,6 +1,6 @@
 package filodb.query
 
-import io.circe.{Decoder, Encoder, HCursor, Json}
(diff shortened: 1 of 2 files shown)


vishramachandran (approved):

sherali42 (approved):
👍  thanks @amolnayak311

PR #1613, opened 2023-06-23
filodb(core) add debugging info for empty histogram.

yu-shipit:
**Pull Request checklist**

**New behavior :**

**BREAKING CHANGES**

Breaking changes may include:

**Other information**:


memory/src/main/scala/filodb.memory/format/vectors/HistogramVector.scala
```diff
@@ -578,6 +579,11 @@ class SectDeltaHistogramReader(acc2: MemoryReader, histVect: Ptr.U8)
   }
 
   override def apply(index: Int): HistogramWithBuckets = {
+    if (length <= 0) {
+      UnsafeUtils.getLong(histVect.addr)
```
amolnayak311, line 583: Why do you need this?

core/src/main/scala/filodb.core/query/PartitionTimeRangeReader.scala
```diff
@@ -33,7 +35,23 @@ final class PartitionTimeRangeReader(part: ReadablePartition,
...
+        case e : EmptyHistogramException => {
+          var message = s"${e.getMessage} infos=["
+           while (infos.hasNext) {
+             val info = infos.nextInfo
+             message +=
+               s"""${info.debugString} """
+           }
+          message += "]"
+          logger.error(s"EmptyHistogramException message ${message}")
+          throw new IllegalArgumentException(message)
```
amolnayak311, line 51: What is the current exception thrown? Do you want to throw the entire message built (possibly large) or just log and throw a concise error? 

It would be good if you give the current problem, the exception seen and the expectation to create this PR in the PR description.

amolnayak311 (changes requested):

core/src/main/scala/filodb.core/query/PartitionTimeRangeReader.scala
yu-shipit: Hi Amol,

here is the description. I updated it through the git message.

> 
> Some queries occasionally hit exceptions because of empty histogram.
> However, the same exception could not be reproduced later.
> The hunch is that the bug is caused by a race condition.
> So, adding additional debugging log to print out the chunk id chunk info and the memory dump.
>

alextheimer (dismissed):
Hopefully we'll find the bug with this info 🤞

core/src/main/scala/filodb.core/query/PartitionTimeRangeReader.scala
alextheimer: >Do you want to throw the entire message built (possibly large)

@amolnayak311 we met and discussed w/ @vishramachandran and @sherali42. The intent here is to dump everything to the logs that might be useful, but @yu-shipit  maybe we could add a `substr` call just to be doubly sure we don't log some ridiculous character count? (Full disclosure: I don't know what constitutes "reasonable." Pick something that agrees with local tests-- or maybe others have opinions about this already?)

sherali42: yup agree with @amolnayak311, no need to add debug log message to the exception. whatever we were throwing earlier `requirement failed: empty histogram` is good enough.

yu-shipit: I think we just add EmptyHistogramException to label the error message.
Thank you.

alextheimer: @sherali42 @amolnayak311  we don't need the ChunkInfo string to debug this race condition?

alextheimer: We could truncate this string too -- I'm not sure if that's a useful option, though.

amolnayak311 (approved):

alextheimer (approved):

PR #1615, opened 2023-06-26
fix(downsample): Change semantic to not throw error to avoid query exceptions when metering is not enabled

sandeep6189:
Change semantic to not throw error to avoid query exceptions when metering is not enabled

**Pull Request checklist**

**Current behavior :** We would throw query exceptions if metering is not enabled in Tsdb dowsample. This will lead to complication on the query planner side to make sure the config is followed for remote partition calls. Hence, to simplify it, we are returning empty responses if metering is not enabled with a log.

**New behavior :** Return empty response instead of throwing exception.


amolnayak311 (approved):
Thanks for the change

PR #1616, opened 2023-06-27
fix(coordinator): select random dispatcher for aggregation

alextheimer:
**Pull Request checklist**

Currently, aggregation reductions may not be evenly distribute across FiloDB nodes. Since that the dispatcher is not chosen randomly, a fixed set of nodes might perform all reductions. This PR instead selects a random dispatcher to perform each reduction.


amolnayak311 (approved):
Thanks @alextheimer for the PR. I would be interested to see how this really plays out in practice as this kicks in only when we have more than 16 children plans where we group them into `sqrt` of that number.  It might be a great catch.

PR #1618, opened 2023-06-28
feat(query): limit on actual raw data scan as opposed to current estimates

kvpetrov:
**Pull Request checklist**

**Current behavior :** 
Currently, we can limit scans on our raw data by estimating possible scan sizes, however, these estimates are not accurate which prevents us from actually using them.

**New behavior :**
Adding a limit that would act on actual data scans sizes and would stop the query from execution if the limit is breached.

**BREAKING CHANGES**

Adding a field in Protos and Akka messages.

**Other information**:


core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
```diff
@@ -348,6 +350,15 @@ class CountingChunkInfoIterator(base: ChunkInfoIterator,
     // Why two counters ?
     // 1. query stats counter to meter usage by ws/ns. This will eventually be sent as part of query result.
     dataBytesScannedCtr.addAndGet(bytesRead)
+    if (dataBytesScannedCtr.get() > maxBytesScanned) {
```
vishramachandran, line 353: Do we need to add this check for every chunk read? Or can we reduce the frequency of this check for every range vector sent out? If latter, we could move this check into ExecPlan.scala

vishramachandran: On second thoughts, lets start with this. Group-by type queries can end up scanning more before the limits kick in with the other approach.

core/src/main/scala/filodb.core/query/QueryContext.scala
```diff
@@ -111,7 +119,8 @@ case class QueryWarnings(
           groupByCardinality.get().equals(w2.groupByCardinality.get()) &&
           joinQueryCardinality.get().equals(w2.joinQueryCardinality.get()) &&
           timeSeriesSamplesScannedBytes.get().equals(w2.timeSeriesSamplesScannedBytes.get()) &&
-          timeSeriesScanned.get().equals(w2.timeSeriesScanned.get())
+          timeSeriesScanned.get().equals(w2.timeSeriesScanned.get()) &&
```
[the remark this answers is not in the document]
kvpetrov: the member variables  are atomic longs and integers and looks like AtomicIntegers with identical values are not equal to each other.
https://stackoverflow.com/questions/7567502/why-are-two-atomicintegers-never-equal

vishramachandran (approved):

alextheimer (approved):

core/src/main/scala/filodb.core/query/RangeVector.scala
```diff
@@ -296,18 +296,24 @@ final case class RawDataRangeVector(key: RangeVectorKey,
                                     partition: ReadablePartition,
                                     chunkMethod: ChunkScanMethod,
                                     columnIDs: Array[Int],
-                                    dataBytesScannedCtr: AtomicLong) extends RangeVector {
+                                    dataBytesScannedCtr: AtomicLong,
```
amolnayak311, line 299: It would be nice to have some comments here on what these Counters mean

core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
```diff
@@ -348,6 +350,15 @@ class CountingChunkInfoIterator(base: ChunkInfoIterator,
...
+      val exMessage = s"Actual raw data bytes scan of ${dataBytesScannedCtr.get()} bytes exceeds limit of " +
+        s"${maxBytesScanned} bytes queried per shard. " +
+        s"Try one or more of these: " +
+        s"(a) narrow your query filters to reduce to fewer time series matches " +
```
amolnayak311, line 357: NIT: Is it narrow timeseries matches by adding filters?

amolnayak311 (approved):
LGTM, Left some ,minor comments.

core/src/main/scala/filodb.core/query/RangeVector.scala
kvpetrov: will add the proper comment in the next iteration, otherwise the approval will disappear

core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
kvpetrov: will add the proper comment in the next iteration, otherwise the approval will disappear

PR #1624, opened 2023-07-13
fix(core) add debugging info for empty histogram.

yu-shipit:
Some queries occasionally hit exceptions because of empty histogram. However, the same exception could not be reproduced later. The hunch is that the bug is caused by a race condition. So, adding additional debugging log to print out the chunk id chunk info and the memory dump.

**Pull Request checklist**

**New behavior :**

**BREAKING CHANGES**

Breaking changes may include:

**Other information**:

### DIFF
--- a/memory/src/main/scala/filodb.memory/format/vectors/HistogramVector.scala
+++ b/memory/src/main/scala/filodb.memory/format/vectors/HistogramVector.scala
@@ -556,6 +556,7 @@ final case class HistogramCorrection(lastValue: LongHistogram, correction: LongH
 trait CounterHistogramReader extends HistogramReader with CounterVectorReader {
   def correctedValue(n: Int, meta: CorrectionMeta): HistogramWithBuckets
 }
+case class EmptyHistogramException(message: String) extends IllegalArgumentException(message)
 
 /**
  * A reader for SectDelta encoded histograms, including correction/drop functionality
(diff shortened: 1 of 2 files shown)


memory/src/main/scala/filodb.memory/format/vectors/HistogramVector.scala
```diff
@@ -578,7 +579,10 @@ class SectDeltaHistogramReader(acc2: MemoryReader, histVect: Ptr.U8)
   }
 
   override def apply(index: Int): HistogramWithBuckets = {
-    require(length > 0)
+    if (length <= 0) {
+      throw EmptyHistogramException(s"""length = $length memory=${toHexString(acc, histVect.addr)}""")
```
amolnayak311, line 583: Change this one too. It has to be "requirement failed"

amolnayak311 (changes requested):

memory/src/main/scala/filodb.memory/format/vectors/HistogramVector.scala
amolnayak311: Well, why do we need this anyways of we don't log anything, I will say please don't throw new exception, just log and keep the ``require`` as is.  We also need to ensure exceptions are serialized and deserialized over the proto. Check ``ProtoConverters`` for more details.

amolnayak311 (approved):

PR #1629, opened 2023-07-17
fix(core) fix the binary join aggregation across different partitions.

yu-shipit:
For regex shard key we need to aggregate across all nodes. InProcessPlanDispatcher is needed.

**Pull Request checklist**

**New behavior :**

**BREAKING CHANGES**

Breaking changes may include:

**Other information**:

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/queryplanner/LogicalPlanUtils.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryplanner/LogicalPlanUtils.scala
@@ -34,6 +34,7 @@ object LogicalPlanUtils extends StrictLogging {
    */
   def hasDescendantAggregate(lp: LogicalPlan): Boolean = lp match {
     case _: Aggregate                 => true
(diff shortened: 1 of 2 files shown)


coordinator/src/main/scala/filodb.coordinator/queryplanner/LogicalPlanUtils.scala
```diff
@@ -34,6 +34,7 @@ object LogicalPlanUtils extends StrictLogging {
    */
   def hasDescendantAggregate(lp: LogicalPlan): Boolean = lp match {
     case _: Aggregate                 => true
+    case bj: BinaryJoin                => bj.operator == BinaryOperator.LUnless
```
alextheimer, line 37: What about e.g. `foo{} and bar{}`? If that is pushed down and `bar` exists on only one partition, then it's possible there will be rows missing from the final result.

I think the general case should be okay:
```
case bj: BinaryJoin => true
```
(as long as there's some performance testing involved).

alextheimer (approved):
LGTM @yu-shipit -- thanks a ton for the iterative effort here 🙏

PR #1639, opened 2023-07-25
fix(query) Regex equals .* must ignore the label and match series even without the label

amolnayak311:
**Pull Request checklist**

**Current behavior :** 

Queries like ``foo{dummy=~".*"}`` are expected to match all timeseries with metric name foo even without the label dummy, in short, the filter should have same behavior as not having the filter. However, currently the matches are only made for series with the label present

**New behavior :**
 The above behavior is fixed and now the label is not considered in matching if it matches an empty string ``""``

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
+++ b/core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
@@ -822,7 +822,7 @@ class PartKeyLuceneIndex(ref: DatasetRef,
     filter match {
       case EqualsRegex(value) =>
         val regex = removeRegexAnchors(value.toString)
-        if(regex.r.pattern.matcher("").matches()) {
+        if(regex.replaceAll("\\.\\*", "") == "") {
           // Check if the given regex matches the empty string, if yes, then do not consider this label
           new MatchAllDocsQuery
         } else {
--- a/core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
+++ b/core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
@@ -817,12 +817,20 @@ class PartKeyLuceneIndex(ref: DatasetRef,
     logger.info(s"Refreshed index searchers to make reads consistent for dataset=$ref shard=$shardNum")
   }
 
+  //scalastyle:off method.length
   private def leafFilter(column: String, filter: Filter): Query = {
     filter match {
       case EqualsRegex(value) =>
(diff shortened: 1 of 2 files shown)


yu-shipit (dismissed):

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -817,12 +817,20 @@ class PartKeyLuceneIndex(ref: DatasetRef,
...
   }
 
+  //scalastyle:off method.length
   private def leafFilter(column: String, filter: Filter): Query = {
     filter match {
       case EqualsRegex(value) =>
         val regex = removeRegexAnchors(value.toString)
-        if (regex.nonEmpty) new RegexpQuery(new Term(column, regex), RegExp.NONE)
-        else leafFilter(column, NotEqualsRegex(".+"))  // value="" means the label is absent or has an empty value.
+        if(regex.r.unapplySeq("").isDefined) {
```
alextheimer, line 825: Nitpick: does `matches` work here?

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -817,12 +817,20 @@ class PartKeyLuceneIndex(ref: DatasetRef,
...
+          // Check if the given regex matches the empty string, if yes, then do not consider this label
+          val booleanQuery = new BooleanQuery.Builder
+          val allDocs = new MatchAllDocsQuery
+          booleanQuery.add(allDocs, Occur.FILTER).build()
```
alextheimer, line 829: I think we can just return a `MatchAllDocsQuery` directly (it implements `Query`, at least.)

alextheimer (dismissed):
All looks good (and thanks for the tests btw) 👍

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
amolnayak311: That's exactly what I was looking for, but its available in Scala 2.13

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
amolnayak311: Good catch, done

alextheimer (approved):

vishramachandran (approved):

PR #1641, opened 2023-08-03
perf(core): ~Two times throughput improvement for Lucene queries with enum and prefix regex filters

vishramachandran:
**Pull Request checklist**

Production profiling is showing that Lucene Regex automata is creating a hotspot in method and allocation profiling.

This PR optimizes two kinds of queries
1. Regex with enumerated values are converted to `TermInSetQuery`
2. Regex with prefix is converted to `PrefixQuery`

It also wraps Lucene queries in `ConstantScoreQuery` to prevent any scoring that may be happening.

Observed 
* 2.2x performance improvement in JMH benchmark for specific enum regex query
* 1.5x performance improvement in JMH benchmark for specific prefix regex query

Detailed Benchmark results

```
After
[info] Benchmark                                                   Mode  Cnt     Score    Error  Units
[info] PartKeyIndexBenchmark.partIdsLookupWithEnumRegexFilter     thrpt    5  7754.992 ± 31.444  ops/s
[info] PartKeyIndexBenchmark.partIdsLookupWithPrefixRegexFilters  thrpt    5  7124.124 ± 49.450  ops/s

Before
[info] Benchmark                                                   Mode  Cnt     Score    Error  Units
[info] PartKeyIndexBenchmark.partIdsLookupWithEnumRegexFilter     thrpt    5  3515.597 ± 17.021  ops/s
[info] PartKeyIndexBenchmark.partIdsLookupWithPrefixRegexFilters  thrpt    5  4661.993 ±  6.215  ops/s
```

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
+++ b/core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
@@ -828,13 +828,30 @@ class PartKeyLuceneIndex(ref: DatasetRef,
     logger.info(s"Refreshed index searchers to make reads consistent for dataset=$ref shard=$shardNum")
   }
 
-  //scalastyle:off method.length
+  val regexChars = Array('.', '?', '+', '*', '|', '{', '}', '[', ']', '(', ')', '"', '\\')
+  val regexCharsMinusPipe = (regexChars.toSet - '|').toArray
+
+  // scalastyle:off method.length
   private def leafFilter(column: String, filter: Filter): Query = {
+
+    def equalsQuery(value: String): Query = {
+      if (value.nonEmpty) new TermQuery(new Term(column, value))
+      else leafFilter(column, NotEqualsRegex(".+")) // value="" means the label is absent or has an empty value.
+    }
+
     filter match {
       case EqualsRegex(value) =>
         val regex = removeRegexAnchors(value.toString)
-        if (regex.replaceAll("\\.\\*", "").nonEmpty) new RegexpQuery(new Term(column, regex), RegExp.NONE)
-        else leafFilter(column, NotEqualsRegex(".+")) // value="" means the label is absent or has an empty value.
+        val regexCharArray = regex.toCharArray
+        if (regexCharArray.forall(c => !regexChars.contains(c))) { // if all regex chars absent
+          // treat like Equals
+          equalsQuery(regex)
+        } else if (regexCharArray.forall(c => !regexCharsMinusPipe.contains(c))) { // if pipe is only regex char present
+          new TermInSetQuery(column, regex.split('|').map(t => new BytesRef(t)): _*)
+        } else if (regex.replaceAll("\\.\\*", "").nonEmpty) {
+          new RegexpQuery(new Term(column, regex), RegExp.NONE)
+        }
+        else leafFilter(column, NotEqualsRegex(".+"))  // value="" means the label is absent or has an empty value.
 
       case NotEqualsRegex(value) =>
         val term = new Term(column, removeRegexAnchors(value.toString))
@@ -843,9 +860,10 @@ class PartKeyLuceneIndex(ref: DatasetRef,
         booleanQuery.add(allDocs, Occur.FILTER)
         booleanQuery.add(new RegexpQuery(term, RegExp.NONE), Occur.MUST_NOT)
         booleanQuery.build()
+
       case Equals(value) =>
-        if (value.toString.nonEmpty) new TermQuery(new Term(column, value.toString))
-        else leafFilter(column, NotEqualsRegex(".+"))  // value="" means the label is absent or has an empty value.
+        equalsQuery(value.toString)
+
       case NotEquals(value) =>
         val str = value.toString
         val term = new Term(column, str)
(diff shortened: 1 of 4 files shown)


core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -817,16 +816,40 @@ class PartKeyLuceneIndex(ref: DatasetRef,
...
+    }
+
     filter match {
       case EqualsRegex(value) =>
         val regex = removeRegexAnchors(value.toString)
-        if (regex.replaceAll("\\.\\*", "") == "")
+        if (regex == "") {
+          // if label=~"" then match empty string or label not present condition too
+          leafFilter(column, NotEqualsRegex(".+"))
+        } else if (regex.replaceAll("\\.\\*", "") == "") {
```
yu-shipit, line 836: do we also need to consider cases like ".{0,10}"

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -817,16 +816,40 @@ class PartKeyLuceneIndex(ref: DatasetRef,
...
+          // if label=~".*" then match all docs since promQL matches .* with absent label too
           new MatchAllDocsQuery
-        else if (regex.nonEmpty)
+        } else if (regex.forall(c => !regexChars.contains(c))) {
+          // if all regex special chars absent, then treat like Equals
+          equalsQuery(regex)
+        } else if (regex.forall(c => !regexCharsMinusPipe.contains(c))) {
+          // if pipe is only regex special char present, then convert to IN query
+          new TermInSetQuery(column, regex.split('|').map(t => new BytesRef(t)): _*)
+        } else if (regex.endsWith(".*") && regex.length > 2 &&
```
yu-shipit, line 845: should we also handle cases like ".*.*"

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -817,16 +816,40 @@ class PartKeyLuceneIndex(ref: DatasetRef,
...
+
+    def equalsQuery(value: String): Query = {
+      if (value.nonEmpty) new TermQuery(new Term(column, value))
+      else leafFilter(column, NotEqualsRegex(".+")) // value="" means the label is absent or has an empty value.
+    }
+
     filter match {
       case EqualsRegex(value) =>
         val regex = removeRegexAnchors(value.toString)
-        if (regex.replaceAll("\\.\\*", "") == "")
```
alextheimer, line 825: Previously, `""` would get caught by this `if` condition and `MatchAllDocsQuery` would be used. Now, `NotEqualsRegex(".+")` is used. Is that intentional?

Also, I don't think this is correct:
```
          // if label=~"" then match empty string or label not present condition too
           leafFilter(column, NotEqualsRegex(".+"))
```
-- since that regex would not match any empty strings.

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
vishramachandran: `..` would not be a prefix query. So it should not match for this condition

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
vishramachandran: No. This one should match `.*` like regexes that match all docs only.

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
vishramachandran: @amolnayak311  's PRs earlier had some regressions. This PR reorders the ifs and restores behavior for `label=~""` to behave prior to his change. So this begs the question. Was @yu-shipit the right behavior ?

Lets discuss in person.

yu-shipit: ~"" is converted to NotEqualsRegex(".+") to handle non-existing label. 
EqualsRegex(".*")  requires that the label exists.

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
yu-shipit: sorry the \* was not escaped.
I mean we can drop all the .\* on the right side.

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
alextheimer: I've tested `label=~""` in a Prometheus server, and it looks like the behavior there matches this `NotEqualsRegex(".+")` filter.

alextheimer (approved):

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
vishramachandran: Although possible, it unlikely to get any or too many of these cases. I am leaning towards keeping code simple to maintain. The fallback always works functionally

amolnayak311 (approved):

amolnayak311:
LGTM Thanks for the optimization and fix.

PR #1644, opened 2023-08-13
perf(query): reduce regex query fanout

alextheimer:
**Pull Request checklist**

Currently, the `ShardKeyRegexPlanner` asks its inner planner to create an `ExecPlan` for each one of a `LogicalPlan`'s shard keys. This is inefficient; if 100 shard keys are matched but all data lies on only two partitions, it would be much more efficient to instead ask the inner planner to materialize one plan per partition.

This PR materializes `ExecPlan`s in this more-efficient manner. The `ShardKeyRegexPlanner` now passes `LogicalPlans` with original regex filters to lower-level planners. The `MultiPartitionPlanner` now has infrastructure added to resolve shard-keys and determine the partitions to be queried; it now materializes plans per-partition (instead of per-shard-key). Again, regex is passed to lower-level planners; the `SingleClusterPlanner` now also has the infrastructure to materialize regex-filtered plans.


vishramachandran (commented):
@alextheimer  Query Stats need to be keyed with "multiple-ns" for the namespace field for NS-Regex queries within partition.

alextheimer:
@vishramachandran it looks like the latest `develop` has this covered (I've rebased the PR):
- In `TimeSeriesShard`: https://github.com/filodb/FiloDB/blob/6cb5433b202239fd1ac4baaf201468532444e8a2/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala#L1936
- In `DownsampledTimeSeriesShard`: https://github.com/filodb/FiloDB/blob/6cb5433b202239fd1ac4baaf201468532444e8a2/core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala#L355

`QueryStats` are also updated in e.g. label-value iterators, but I'll leave those keys alone since they aren't related to PromQL queries.

prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
```diff
@@ -241,9 +241,14 @@ sealed trait Vector extends Expression {
         case NotRegexMatch   => require(labelValue.length <= Parser.REGEX_MAX_LEN,
                                          s"Regular expression filters should be <= ${Parser.REGEX_MAX_LEN} characters")
                                 ColumnFilter(labelMatch.label, query.Filter.NotEqualsRegex(labelValue))
-        case RegexMatch      => require(labelValue.length <= Parser.REGEX_MAX_LEN,
-                                         s"Regular expression filters should be <= ${Parser.REGEX_MAX_LEN} characters")
-                                ColumnFilter(labelMatch.label, query.Filter.EqualsRegex(labelValue))
+        case RegexMatch      =>
+          // Relax the length limit only for matchers that contain at most the "|" special character.
+          if (QueryUtils.containsUnescapedNonPipeRegexChars(labelValue)) {
```
alextheimer, line 246: Double-check equal with Lucene functionality.

coordinator/src/main/scala/filodb.coordinator/queryplanner/LogicalPlanUtils.scala
```diff
@@ -422,7 +423,15 @@ object LogicalPlanUtils extends StrictLogging {
       .map(_.asInstanceOf[RawSeries]).flatMap{ rs =>
       val shardKeyFilters = getShardKeyFilters(rs)
       val interval = LogicalPlanUtils.getSpanningIntervalSelector(rs)
-      shardKeyFilters.map{ shardKey =>
+      val resolvedFilters = shardKeyFilters.flatMap{ filters =>
+        val updFilters = filters.map { filter => filter.filter match {
+          case EqualsRegex(values: String) if !QueryUtils.containsUnescapedNonPipeRegexChars(values) =>
+            values.split('|').map(value => ColumnFilter(filter.column, Equals(value))).toSeq
```
alextheimer, line 429: Make work for escaped pipes

prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
alextheimer: No longer applies.

coordinator/src/test/scala/filodb.coordinator/queryplanner/MultiPartitionPlannerSpec.scala
```diff
@@ -923,94 +923,82 @@ class MultiPartitionPlannerSpec extends AnyFunSpec with Matchers with PlanValida
...
-    val promQlQueryParams = PromQlQueryParams("test{job = \"app\"}", startSeconds, step, endSeconds)
-
-    val execPlan = engine.materialize(lp, QueryContext(origQueryParams = promQlQueryParams, plannerParams =
-      PlannerParams(processMultiPartition = true)))
-    val stitchRvsExec = execPlan.asInstanceOf[StitchRvsExec]
-    stitchRvsExec.children.size shouldEqual (2)
-    stitchRvsExec.children(0).isInstanceOf[PromQlRemoteExec] shouldEqual (true)
-    stitchRvsExec.children(1).isInstanceOf[PromQlRemoteExec] shouldEqual (true)
-
+    execPlan.isInstanceOf[PromQlRemoteExec] shouldEqual (true)
```
amolnayak311, line 931: Why should it generate just one? The previous one was generating two children,  both for different partition endpoints.

coordinator/src/test/scala/filodb.coordinator/queryplanner/MultiPartitionPlannerSpec.scala
```diff
@@ -923,94 +923,82 @@ class MultiPartitionPlannerSpec extends AnyFunSpec with Matchers with PlanValida
...
-    queryParams2.endSecs shouldEqual endSeconds
-    queryParams2.stepSecs shouldEqual step
-    remoteExec2.queryContext.plannerParams.processFailure shouldEqual true
-    remoteExec2.queryContext.plannerParams.processMultiPartition shouldEqual false
-    remoteExec2.queryEndpoint shouldEqual "remote-url"
-
+    remoteExec.queryEndpoint shouldEqual "remote-url2"
   }
 
+//  // TODO: what is this the purpose of this test? The lookback is 300s, and the assignment
```
amolnayak311, line 944: Why comment out the test?

coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
```diff
@@ -96,7 +98,12 @@ class MultiPartitionPlanner(partitionLocationProvider: PartitionLocationProvider
     } else logicalPlan match {
       case mqp: MetadataQueryPlan             => materializeMetadataQueryPlan(mqp, qContext).plans.head
       case lp: TsCardinalities                => materializeTsCardinalities(lp, qContext).plans.head
-      case _                                  => walkLogicalPlanTree(logicalPlan, qContext).plans.head
+      case _                                  =>
+        val result = walkLogicalPlanTree(logicalPlan, qContext)
```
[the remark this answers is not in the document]
alextheimer: Good point -- this was probably needed for some previous PR iteration, but tests pass without it, now. I will drop this change since the extra logic isn't necessary.

coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
```diff
@@ -389,64 +332,127 @@ class MultiPartitionPlanner(partitionLocationProvider: PartitionLocationProvider
...
       case sqw: SubqueryWithWindowing => super.materializeSubqueryWithWindowing(qContext, sqw)
       case bj: BinaryJoin => materializeMultiPartitionBinaryJoinNoSplitLeaf(bj, qContext)
       case sv: ScalarVectorBinaryOperation => super.materializeScalarVectorBinOp(qContext, sv)
       case aif: ApplyInstantFunction => super.materializeApplyInstantFunction(qContext, aif)
       case svdp: ScalarVaryingDoublePlan => super.materializeScalarPlan(qContext, svdp)
       case aaf: ApplyAbsentFunction => super.materializeAbsentFunction(qContext, aaf)
+      case ps: PeriodicSeries => materializeNonSplitPeriodicAndRawSeries(ps, qContext)
+      case rcm: RawChunkMeta => materializeNonSplitPeriodicAndRawSeries(rcm, qContext)
+      case rs: RawSeries => materializeNonSplitPeriodicAndRawSeries(rs, qContext)
       case x => throw new IllegalArgumentException(s"unhandled type: ${x.getClass}")
```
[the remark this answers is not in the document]
alextheimer: I wish I knew this sooner, it bugged me that `match`es never threw compile-time errors:
> It is, however, possible to do this for sealed traits whose subclasses are all also at least sealed or final (thus effectively "turning off" the capability of loading new subclasses at runtime). In that case, Scala will optimize the pattern match to a more efficient representation, and you will get a warning for a non-exhaustive pattern match.

From [here](https://stackoverflow.com/a/65489322).

I will add a line for each subclass 👍

coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
```diff
@@ -389,64 +332,127 @@ class MultiPartitionPlanner(partitionLocationProvider: PartitionLocationProvider
...
    *   contains leaves that individually span partitions.
    */
   private def materializePlanHandleSplitLeaf(logicalPlan: LogicalPlan,
                                              qContext: QueryContext): PlanResult = {
     val qParams = qContext.origQueryParams.asInstanceOf[PromQlQueryParams]
-    val hasMultiPartitionLeaves = LogicalPlan.findLeafLogicalPlans(logicalPlan)
-                                             .exists(getPartitions(_, qParams).size > 1)
+    // Create one plan per RawSeries/shard-key pair, then resolve its partitions.
+    // If any resides on more than one partition, the leaf is "split".
+    val hasMultiPartitionLeaves =
```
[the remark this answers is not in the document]
alextheimer: This is confusing because it should be renamed to `hasSplitPartitionLeaves`. "Multi-partition" implied "split" when shard-keys were filtered only by equality.

Detailed explanation for completeness: The old logic was
```
    val hasMultiPartitionLeaves = LogicalPlan.findLeafLogicalPlans(logicalPlan)
                                             .exists(getPartitions(_, qParams).size > 1)
```
Suppose some plan has regex shard-key filters. If the resolved keys reside in multiple partitions, the `getPartitions.size` step might return some value `> 1`. The old logic would interpret that as a partition split even though none might exist.

coordinator/src/test/scala/filodb.coordinator/queryplanner/MultiPartitionPlannerSpec.scala
alextheimer: Nope, they were generated for the same endpoints:
```
    val remoteExec = stitchRvsExec.children(0).asInstanceOf[PromQlRemoteExec]
    ...
    remoteExec.queryEndpoint shouldEqual "remote-url"

    val remoteExec2 = stitchRvsExec.children(1).asInstanceOf[PromQlRemoteExec]
    ...
    remoteExec2.queryEndpoint shouldEqual "remote-url"
```
This test used to generate two children because one child plan was materialized per shard-key.

coordinator/src/test/scala/filodb.coordinator/queryplanner/MultiPartitionPlannerSpec.scala
alextheimer: I was hoping others would know what this test was for-- it looks like we used to expect partition-split plans would always return at least one data-point per partition, regardless of lookback/offset. That doesn't jive with the way we handle partition-splits elsewhere in the planners, so I think this test should be removed.

The reason it passed prior to this PR is because the MPP sent `PeriodicSeries` and `RawSeries` through the old `materializePeriodicAndRawSeries` logic. That always materialized at least one plan per partition.

`materializePeriodicAndRawSeries` was built to materialize plans with equality-only filters, so that would not work for this PR. Instead, plans are:
- if split, passed to the regex-handling `materializeSplitLeafPlan`, which accounts correctly for lookback.
- if not split, forwarded as-is to any partitions that own the query data.

coordinator/src/test/scala/filodb.coordinator/queryplanner/MultiPartitionPlannerSpec.scala
alextheimer: Some additional comments: this test needed an update because -- now that all split plans (plans with at least one shard-key that resides in >1 partitions) are materialized via `materializePlanHandleSplitLeaf`-- an empty result is returned when all timestamp evaluations require data from >1 partitions.

That would happen here:
- This is an instant query, so only one timestamp is evaluated at `t=1000s`.
- The lookback is `100s`.
- The partition split happens at `950s`.

I've pushed the start of the second partition back far enough to fit the entire `100s` lookback, and now we should only expect one `RemoteExec` to be returned.

amolnayak311: Got it, so the method would return ``EmptyResultExec`` if left unchanged? If yes why not keep that test and additionally test your case too which goes to just one partition for a timestamp >= 1050?

coordinator/src/test/scala/filodb.coordinator/queryplanner/ShardKeyRegexPlannerSpec.scala
```diff
@@ -72,48 +78,33 @@ class ShardKeyRegexPlannerSpec extends AnyFunSpec with Matchers with ScalaFuture
...
     val execPlan = engine.materialize(lp, QueryContext(origQueryParams = promQlQueryParams))
-    execPlan.isInstanceOf[MultiPartitionDistConcatExec] shouldEqual(true)
-    execPlan.children(0).children.head.isInstanceOf[MultiSchemaPartitionsExec]
-    execPlan.children(1).children.head.asInstanceOf[MultiSchemaPartitionsExec].filters.
-      contains(ColumnFilter("_ns_", Equals("App-1"))) shouldEqual(true)
-    execPlan.children(0).children.head.asInstanceOf[MultiSchemaPartitionsExec].filters.
-      contains(ColumnFilter("_ns_", Equals("App-2"))) shouldEqual(true)
-  }
-
-  it("should generate Exec plan for implicit ws query") {
```
amolnayak311, line 92: @alextheimer if this test doesn't break your work leave it there. There are some changes in ``ShardKeyRegexPlanner`` that this test asserts

alextheimer: [discussed offline]

coordinator/src/test/scala/filodb.coordinator/queryplanner/MultiPartitionPlannerSpec.scala
alextheimer: Good point -- I will re-add the `EmptyResultExec` test.

coordinator/src/main/scala/filodb.coordinator/queryplanner/QueryUtils.scala
```diff
@@ -0,0 +1,35 @@
+package filodb.coordinator.queryplanner
+
+import scala.collection.mutable
+
+/**
+ * Storage for miscellaneous utility functions.
+ */
+object QueryUtils {
```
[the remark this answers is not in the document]
alextheimer: Much cleaner -- I've added almost exactly this (but with SeqExtractors: https://stackoverflow.com/a/19147469)

PR #1653, opened 2023-08-18
feat(core): support for per dataset cassandra keyspace overrides

sherali42:
**Pull Request checklist**

As of now there is a stipulation that all the datasets that are configured in a FiloDB cluster needs to use one Keyspace for raw (actively ingesting) data. With this change, keyspaces can be configured per dataset with the exception of MetaStore (checkpoints).
This is especially useful to merge two clusters for better resource utilization without the need for data migration. With the caveat that checkpoint records need to be migrated from source cluster to target cluster.


amolnayak311 (approved):
LGTM. QQ, Would you need to initialize in ``PerShardCardinalityBuster`` too?

yu-shipit (approved):

PR #1662, opened 2023-08-28
maint(core): upgrade to Lucene 9.7.0

alextheimer:
**Pull Request checklist**

Lucene 9.7.0 [requires document field value types to be consistent](https://github.com/apache/lucene/blame/fb8183332bc5fa74a8eb36105be1026ae7aa0a8e/lucene/core/src/java/org/apache/lucene/index/IndexingChain.java#L563). Currently, `__partId__` may have either numeric and/or string values; this causes Lucene to throw exceptions.

This PR splits `__partId__` into two separate fields: `__partIdNumeric__` and `__partIdString__`. Now, field values are consistently-typed (and compatible with Lucene 9.7.0).


core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -646,7 +648,7 @@ class PartKeyLuceneIndex(ref: DatasetRef,
       s"with startTime=$startTime endTime=$endTime into dataset=$ref shard=$shardNum")
     val doc = luceneDocument.get()
     val docToAdd = doc.facetsConfig.build(doc.document)
-    val term = new Term(PART_ID, documentId)
+    val term = new Term(PART_ID_STRING, documentId)
```
vishramachandran, line 651: Unrelated to your PR, but can you please improve by adding docs on why documentId is assigned to partId ? @amolnayak311  can add more here, but this should be called only from DownsampleTimeSeriesShard where there is no concept of partId. Hence documentId is used.

vishramachandran (approved):
LGTM with some minor nitpicks.

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
amolnayak311: > but this should be called only from DownsampleTimeSeriesShard where there is no concept of partId.

Hi @vishramachandran regarding above, the upsert is called from any ``*TimeSeriesShard``, however the default parameter ``documentId`` is only overridden from ``DownSampleTimeSeriesShard`` as you mentioned. Initially it was assumed to be the string representation of partId to uniquely identify the document to upsert, however, in downsample index, we still need a unique identifier that wasn't a numeric value (which was ephemeral) hence the method was amended to optionally accept that unique identifier.

amolnayak311 (dismissed):
LGTM, left some comments to @vishramachandran response and we may choose to change the name but overall even the current PR works just fine.

amolnayak311:
Thanks again @alextheimer for putting this PR together to upgrade the much needed latest version :)

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
alextheimer: I like the `PART_ID` prefix because both are really just partition identifiers by different names. Either way works for me, though -- I will change it if there's a strong preference for `DOCUMENT_ID`.

amolnayak311 (approved):
LGTM Thanks

PR #1665, opened 2023-09-09
filodb(core) include the start time for range functions such as rate.

yu-shipit:
**Pull Request checklist**

**New behavior :**
 
The result of a range vector function is based on the data in a range. Previous the first millisecond of data was excluded from the range, now it is included. Let's say, we have a  ts counter { [0s, 1.0], [5s, 6], [10s, 16].
```
Previously, rate(count)[10s] at timestamp 10 just has two data points [5s, 6] and [10s, 16] it may result in a result (16 - 6) /5 =2.

Now, the [0s, 1.0] is included so, the rate would be (16 - 1) / 10 = 1.5
```
After this change, Filodb will be compliant to Prometheus.
**BREAKING CHANGES**

Breaking changes may include:

**Other information**:

### DIFF
--- a/query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
+++ b/query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
@@ -155,16 +155,17 @@ final case class PeriodicSamplesMapper(startMs: Long,
    * pronounced if [1i] notation was used and step == lookback.
    */
   private def extendLookback(rv: RangeVector, window: Long): Long = {
+    window
     // TODO There is a code path is used by Histogram bucket extraction path where
     // underlying vector need not be a RawDataRangeVector. For those cases, we may
     // not able to reliably extend lookback.
     // Much more thought and work needed - so punting the bug
-    val pubInt = rv match {
-      case rvrd: RawDataRangeVector if (functionId.exists(_.onCumulCounter) && stepMultipleNotationUsed)
-          => rvrd.publishInterval.getOrElse(0L)
-      case _ => 0L
-    }
-    window + pubInt
+//    val pubInt = rv match {
+//      case rvrd: RawDataRangeVector if (functionId.exists(_.onCumulCounter) && stepMultipleNotationUsed)
+//          => rvrd.publishInterval.getOrElse(0L)
+//      case _ => 0L
+//    }
+//    window + pubInt
   }
 
   // Transform source double or long to double schema
@@ -343,7 +344,7 @@ class SlidingWindowIterator(raw: RangeVectorCursor,
   override def hasNext: Boolean = curWindowEnd <= end
   override def next(): TransientRow = {
     val curWindowStart = curWindowEnd - window
-    // current window is: (curWindowStart, curWindowEnd]. Excludes start, includes end.
+    // current window is: [curWindowStart, curWindowEnd]. Includes start, end.
 
     // Add elements to window until end of current window has reached
     while (rows.hasNext && rows.head.timestamp <= curWindowEnd) {
@@ -377,7 +378,7 @@ class SlidingWindowIterator(raw: RangeVectorCursor,
     */
   private def shouldAddCurToWindow(curWindowStart: Long, cur: TransientRow): Boolean = {
     // cur is inside current window
-    cur.timestamp > curWindowStart
+    cur.timestamp >= curWindowStart
   }
 
   /**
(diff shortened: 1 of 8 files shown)


query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
```diff
@@ -155,16 +155,17 @@ final case class PeriodicSamplesMapper(startMs: Long,
    * pronounced if [1i] notation was used and step == lookback.
    */
   private def extendLookback(rv: RangeVector, window: Long): Long = {
```
yu-shipit, line 157: Since we include the start time, we do not need to extend the lookback window anymore.
Here is an example, the window is 15s and no need to extend the window to 15001ms
`[2023-09-12 15:47:51,816] INFO  query-sched-1 default f.query.Query$ [] - functionId=Some(Rate) stepMultipleNotationUsed=false window=15000 rv=RawDataRangeVector(/shard:3/b2[schema=prom-counter  _metric_=my_counter,tags={_ns_: Test-data-1694557937_p1, _ws_: aci-telemetry, code: 817, instance: inst-4, mode: nice, service: service-2, version: 3.3.13}] [grp16],TimeSeriesPartition(shard=3,partId=112){b2[schema=prom-counter  _metric_=my_counter,tags={_ns_: Test-data-1694557937_p1, _ws_: aci-telemetry, code: 817, instance: inst-4, mode: nice, service: service-2, version: 3.3.13}]},TimeRangeChunkScan(1694557922000,1694557937000),[I@65c327b0,0,200000000,dcd2633e-6b01-4d89-8d73-124b18cacdfd)
`

alextheimer:
@yu-shipit Could you include a quick summary of the PR in the header comment?

query/src/test/scala/filodb/query/exec/PeriodicSamplesMapperSpec.scala
```diff
@@ -98,9 +98,9 @@ class PeriodicSamplesMapperSpec extends AnyFunSpec with Matchers with ScalaFutur
     val rv = timeValueRVPk(samples, partKey)
 
     val expectedResults = List(
-      500000L -> 125d,
-      900000L ->100d,
-      1300000 -> 400d
+      500000L -> 100d,
+      900000L -> 80d,
+      1300000 -> 320d
```
alextheimer, line 103: Do these changes mean we should expect queries before/after this PR to produce different results? Are the results in this PR more correct?

query/src/test/scala/filodb/query/exec/MultiSchemaPartitionsExecSpec.scala
```diff
@@ -239,9 +239,9 @@ class MultiSchemaPartitionsExecSpec extends AnyFunSpec with Matchers with ScalaF
       ColumnFilter("job", Filter.Equals("myCoolService".utf8)))
     val execPlan = MultiSchemaPartitionsExec(QueryContext(), dummyDispatcher,
       dsRef, 0, filters, AllChunkScan, "_metric_")
-    val start = now - (numRawSamples-100) * reportingInterval
+    val start = now - (numRawSamples-100) * reportingInterval + 1
```
alextheimer, line 242: Why `+1` here?

alextheimer (commented):
Most changes look good-- just a couple questions. Thanks for the tedious work here @yu-shipit 🙏

query/src/test/scala/filodb/query/exec/PeriodicSamplesMapperSpec.scala
yu-shipit: Yes for some queries.
Supposed that we query (rate(foo[3s])) the following data points.
```
foo
1000ms  -> 1
2000ms ->  2
3000ms -> 4

```
Previously the window is [1001ms, 3000ms], which contains only 2 data points-- 2, 4.
It uses 2 and 4 for calculation, and results in (4-2) / (3-2) = 2 

However the new window is [1000ms, 3000ms], which contains 3 data points-- 1, 2, 4.
It uses 1 and 4 for calculation, and results in (4-1) / (3-1) = 1.5


 We will have one more samples for the

query/src/test/scala/filodb/query/exec/MultiSchemaPartitionsExecSpec.scala
yu-shipit: This make the start time same as the previous one.

Previous the start time has a 1ms offset

alextheimer (approved):

PR #1666, opened 2023-09-11
perf(card): Adding config support for DS card flushCount and perf logs for cardinality calculation time

sandeep6189:
**Pull Request checklist**

**Current behavior :** We issue a RocksDB write for each each iteration of partKey while traversing through lucene index during DS cardinality store calculation. For large indexes (10-12 billions), this could take up to 2 hours.   

**New behavior :** Reducing the number of RocksDB write by in-memory aggregation up to a configured `flushCount` after which we issue bulk RocksDB writes.

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityManager.scala
+++ b/core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityManager.scala
@@ -32,6 +32,15 @@ class CardinalityManager(datasetRef: DatasetRef,
   // physical resources for duplicate calculation
   private var isCardTriggered: Boolean = false
 
+  // number of partKeys to aggregate in memory at any given time, before we write the cardinality information to
+  // rocksDB. This is done to reduce the time taken to calculate cardinality of partitions with high number of
+  // TS by reducing the number writes issued to RocksDB ( From our profiling, high number of RocksDB writes lead to
+  // performance bottlenecks)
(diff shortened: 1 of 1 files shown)


core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityManager.scala
```diff
@@ -143,34 +149,7 @@ class CardinalityManager(datasetRef: DatasetRef,
...
-            } catch {
-              case ex: Exception =>
-                // Very unlikely scenario, but can happen if the disk call fails.
-                logger.error(s"[CardinalityManager]Error closing card tracker! shardNum=$shardNum", ex)
-                // setting cardTracker to None in this case, since the exception happened on the close. We
-                // can't rely on the card store. The next trigger should re-build the card store and card tracker
-                cardTracker = None
-            }
-          }
+          createNewCardinalityTrackerAndCalculate(indexRefreshCount)
```
alextheimer, line 152: This method is an exact copy-paste of the above deleted block?

alextheimer (approved):

core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityManager.scala
sandeep6189: Yes, it is a copy paste. Separated the function to simplify the concern of each function ( and make sure function doesn't grow beyond 50 lines )

PR #1669, opened 2023-09-14
fix(query) SetOperator LOR are not compliant with prometheus

amolnayak311:
**Pull Request checklist**

**Current behavior :** 

The prometheus ``OR`` operation for a range query currently looks for existence of join keys in the range and then does not give results for RHS even if the LHS is not available for entire time range in the query range. This behavior is not same as prometheus

Consider ``sum(foo{}) OR vector(1)`` for range ``now-7d`` to ``now``. Suppose ``sum(foo{})`` is only available from  ``now-7d`` to ``now - 1d``, we must see ``vector(1)`` from  ``now-1d`` to ``now``.  Currently. because we check the existence of a join key in the entire range, we don't see ``vector(1)`` in the results for the missing time range. 

Additionally, all binary joins considered absence of ``on`` and assumed that to be empty list. This isn't the right assumption as not providing``on`` and ``on()`` both would give ``on`` as empty list. This PR fixes this by making ``on`` an option where its ``None`` if on isn't specified and ``Some`` if its specified. An example is ``foo{} OR vector(1)`` was treated same as ``foo{} OR on() vector(1)`` before the fix. 

**New behavior :**

Above incompatibility is fixed. 

**Additional notes**

Similar regression in behavior for ``AND`` and ``UNLESS`` too might be present and would need more investigation and a fix on similar lines


query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
```diff
@@ -187,59 +188,175 @@ final case class SetOperatorExec(queryContext: QueryContext,
...
+              cur
+            }
+
+            override def close(): Unit = {
+              rows.close()
+            }
+          }
+          IteratorBackedRangeVector(x.key, rowsCursor, x.outputRange)
+        })
+        // This order is important, when lhs Rvs are executed they set a bitmap which is used when iterating RHS
```
yu-shipit, line 292: This part of code is same as the logic for the left. can we have a common function for it?

query/src/main/scala/filodb/query/exec/BinaryJoinExec.scala
```diff
@@ -15,6 +15,25 @@ import filodb.query._
...
+  val metricColumn: String
+
+  require(on == None || ignoring == Nil, "Cannot specify both 'on' and 'ignoring' clause")
+  require(on.map(!_.contains(metricColumn)).getOrElse(true), "On cannot contain metric name")
+
+  lazy val onLabels: Option[Set[Utf8Str]] = on.map(_.map(Utf8Str(_)).toSet)
+  val ignoringLabels = ignoring.map(Utf8Str(_)).toSet
+  val ignoringLabelsForJoin = ignoringLabels + metricColumn.utf8
+
+  def joinKeys(rvk: RangeVectorKey): Map[Utf8Str, Utf8Str] =
```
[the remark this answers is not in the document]
amolnayak311: I'm not sure I follow the question, is the onLabel ``None``? In that case the join key will contain **all** the labels except the ``ignoringLabels``.

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
amolnayak311: Its not the same, the lhs is **setting** a bit and rhs is **checking** for the set bit and returning a NaN or a value. Am I missing something?

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
```diff
@@ -134,6 +128,13 @@ final case class SetOperatorExec(queryContext: QueryContext,
           if (resVal.key.labelValues == rv.key.labelValues) {
             rhsMap.put(jk, StitchRvsExec.stitch(rv, resVal, outputRvRange) )
           } else {
+            // TODO: Is it ok to overwrite the rhs of multiple RVs match the joinKey
+            //  Consider the case where two Range vectors match the join key and one of them has a non NaN value while
+            //  while other one has a NaN value for a given time step and the one with a non NaN value is the final RV
+            //  in the map. When we run this logic
+            //    val res = if (rhsRow.getDouble(1).isNaN) Double.NaN else lhsRow.getDouble(1) the rhs is not NaN and
```
alextheimer, line 135: Looks like we shouldn't care whether-or-not the RHS value is `NaN`: http://demo.robustperception.io:9090/graph?g0.expr=vector(0)%20and%20vector(Nan)&g0.tab=1&g0.stacked=0&g0.range_input=1h

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
```diff
@@ -187,59 +188,175 @@ final case class SetOperatorExec(queryContext: QueryContext,
...
+          var idx = 0
+          val rows = x.rows()
+          val rowsCursor: RangeVectorCursor = new RangeVectorCursor {
+            val cur = new TransientRow()
+
+            override def hasNext: Boolean = rows.hasNext
+
+            override def next(): RowReader = {
+              val reader = rows.next()
+              // TODO: What is the expected behavior for Histograms?
```
alextheimer, line 275: It looks like the previous implementation was agnostic to the row type. Is it enough to just add a `case` here for histograms and set the value accordingly?

alextheimer (commented):
Looks good @amolnayak311-- just a couple comments. I'll review the histogram changes once those are added, too.

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
```diff
@@ -187,59 +188,175 @@ final case class SetOperatorExec(queryContext: QueryContext,
...
+              val value = reader.getDouble(1)
+              val res = if (value.isNaN) {
```
alextheimer, line 247: Can you use `Double.isNan` here, instead to prevent the extra object creation?

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
```diff
@@ -187,59 +188,175 @@ final case class SetOperatorExec(queryContext: QueryContext,
...
+                Double.NaN
+              } else {
+                bitSet += idx
```
alextheimer, line 250: Since we only set the bit when the value is not NaN, the RHS value will be returned. But Prometheus would return the NaN: http://demo.robustperception.io:9090/graph?g0.expr=vector(Nan)%20or%20sum(up)&g0.tab=1&g0.stacked=0&g0.range_input=1h

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
amolnayak311: Good catch but unfortunately we can't distinguish between ``vector(Nan)`` and value not being there which itself is ``NaN``

query/src/main/scala/filodb/query/exec/BinaryJoinExec.scala
```diff
@@ -15,6 +15,25 @@ import filodb.query._
...
+  val on: Option[Seq[String]]
+  val ignoring: Seq[String]   // TODO: Should this be an option too?
+  val metricColumn: String
+
+  require(on == None || ignoring == Nil, "Cannot specify both 'on' and 'ignoring' clause")
+  require(on.map(!_.contains(metricColumn)).getOrElse(true), "On cannot contain metric name")
+
+  lazy val onLabels: Option[Set[Utf8Str]] = on.map(_.map(Utf8Str(_)).toSet)
+  val ignoringLabels = ignoring.map(Utf8Str(_)).toSet
+  val ignoringLabelsForJoin = ignoringLabels + metricColumn.utf8
```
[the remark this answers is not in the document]
amolnayak311: I'm not sure I get follow the concern. Shouldn't it be adding them up?

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
```diff
@@ -187,59 +188,175 @@ final case class SetOperatorExec(queryContext: QueryContext,
...
+
+    case class OrSetOpIteratorWrapper(lhs: Seq[RangeVector], rhs: Seq[RangeVector]) {
+      // Has a side effect where LHS are consumed first and a bit is set to indicate the time instant where there
+      // is at least one non Nan value. the rhs values are then iterated over and emitted only where the given timestep
+      // has no value in any of the LHS timeseries
+      assert(outputRvRange.isDefined, "Expected to see a non null outputRvRange")
+      val (start, end, step) = (outputRvRange.get.startMs, outputRvRange.get.endMs, outputRvRange.get.stepMs)
+      // 64 bits in one word (long), 1 bit per step
+      assert(start == end || step > 0, "Expected to see a non 0 step size when start and end are not same")
+      val wordsRequired = if (start == end) 1 else ((((end - start) / step) >> 6) + 1).toInt
```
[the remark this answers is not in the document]
amolnayak311: I'm not  sure any optimization is done behind the scene.

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
amolnayak311: Previous behavior was also incorrect. :). This behavior is consistent with AND. While I don't necessarily say this is a good thing to do. Im not sure how to deal with histograms yet and hence left a TODO

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
amolnayak311: Can't distinguish NaN for a missing value vs a legitimate value. I guess supporting this if required will need more changes.

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
alextheimer: We can leave it as long as any `or` query with a histogram operand can fail (since it wouldn't have explicitly failed before). I can't think of any example that isn't a nonsense query, though.

query/src/main/scala/filodb/query/exec/BinaryJoinExec.scala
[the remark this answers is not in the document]
alextheimer: Huh-- I had no idea metric names were dropped during joins: http://demo.robustperception.io:9090/graph?g0.expr=node_network_up%7Bjob%3D%22node%22%7D%20%2B%20label_replace(up%7Bjob%3D%22node%22%7D%2C%22device%22%2C%22eth0%22%2C%22%22%2C%22%22)&g0.tab=1&g0.stacked=0&g0.range_input=1h

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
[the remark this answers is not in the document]
alextheimer: Super nitpick: please update this if you see this comment before the merge (I don't see this in the updates since the last review, but it's not worth calling out / blocking the merge).

alextheimer (approved):
LGTM as long as this is okay: https://github.com/filodb/FiloDB/pull/1669#discussion_r1351321877

alextheimer:
If possible, please add a quick blurb for our future selves about how this same solution can (possibly) be applied to `AND` and `UNLESS`.

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
amolnayak311: I'll assume it is ok. We can look at any issues that may come up and fix accordingly. 
CC @vishramachandran

PR #1281, opened 2021-11-04
feat(query): add promql @ modifier

alextheimer:
**Pull Request checklist**

- [x] The commit(s) message(s) follows the contribution [guidelines](CONTRIBUTING.md) ?
- [ ] Tests for the changes have been added (for bug fixes / features) ?
- [ ] Docs have been added / updated (for bug fixes / features) ?

**New behavior :**
Adds support for queries such as:
```
http_requests_total @ 1609746000
http_requests_total[5m] @ 1609746000
http_requests_total[5m:30s] @ 1609746000
```
where the `@` modifier simply aligns a query's evaluation time with the `@` timestamp.

Per the PromQL spec, `@` can be used in conjunction with `offset`, as in:
```
http_requests_total @ 1609746000 offset 5m
http_requests_total offset 5m @ 1609746000
```
where the relative order of the modifiers does not matter.

See [here](https://prometheus.io/docs/prometheus/latest/querying/basics/#modifier) for more information.

Important notes:
- `start()` and `end()` are not currently supported
- From prometheus.io: "This modifier is disabled by default since it breaks the invariant that PromQL does not look ahead of the evaluation time for samples. It can be enabled by setting --enable-feature=promql-at-modifier flag." This flag is not currently supported, and this PR would enable `@` by default.

PR #1678, opened 2023-09-29
feat(query) Implement @modifier.

yu-shipit:
1. Override range param based on @modifier timestamp for PeriodicSamplesMapper.
2. Add RepeatTransformer and RepeatValueVector to populate data for all timestamps.
3. Add GRPC functions for RepeatValueVector.

**Pull Request checklist**

**New behavior :**
Will support `@modifier`.
The problem is described in [this doc](https://docs.google.com/document/d/1uSbD3T2beM-iX4-Hp7V074bzBRiRNlqUdcWP6JTDQSs/edit).
The [prometheus PR](https://github.com/prometheus/prometheus/pull/8121).
Filodb will support advance usage such as
```
rate(process_cpu_seconds_total[1m]) 
  and
topk(7, rate(process_cpu_seconds_total[1h] @ 1234))

```

**Other information**:


yu-shipit:
Thanks for @alextheimer PR.
https://github.com/filodb/FiloDB/pull/1281

coordinator/src/main/scala/filodb.coordinator/queryplanner/DefaultPlanner.scala
```diff
@@ -338,17 +342,24 @@ trait  DefaultPlanner {
...
-        Some(InternalRangeFunction.lpToInternalFunc(RangeFunctionId.Last)),
-        qContext,
-        stepMultipleNotationUsed = false,
-        Seq(),
-        offsetMs,
-        rawSource = false
-      ))
+     val realStartMs = sqww.atMs.getOrElse(sqww.startMs)
+     val realEndMs = sqww.atMs.getOrElse(sqww.endMs)
+     val realStep = sqww.atMs.map(_ => 0L).getOrElse(sqww.stepMs)
```
alextheimer, line 347: Why is `atMs` used to get the step?

coordinator/src/main/scala/filodb.coordinator/queryplanner/DefaultPlanner.scala
```diff
@@ -338,17 +342,24 @@ trait  DefaultPlanner {
...
+
+     innerExecPlan.plans.foreach(plan => {
+       plan.addRangeVectorTransformer(PeriodicSamplesMapper(
+         realStartMs, realStep, realEndMs,
```
alextheimer, line 351: Should the periodic mapper be passed the `real` time parameters or the actual ones? It gives the correct timestamps when the `real` parameters are passed?

coordinator/src/main/scala/filodb.coordinator/queryplanner/DefaultPlanner.scala
```diff
@@ -338,17 +342,24 @@ trait  DefaultPlanner {
...
+         window,
+         Some(InternalRangeFunction.lpToInternalFunc(RangeFunctionId.Last)),
+         qContext,
+         stepMultipleNotationUsed = false,
+         Seq(),
+         offsetMs,
+         rawSource = false
+       ))
+       sqww.atMs.map(_ => plan.addRangeVectorTransformer(RepeatTransformer(sqww.startMs, sqww.stepMs, sqww.endMs,
```
alextheimer, line 360: Similar question to the above: shouldn't the `RepeatTransformer` be passed the `real` time params?

alextheimer (commented):
Thanks for the PR @yu-shipit -- just a couple of questions. Also, please add some tests that show the `ExecPlans` for simple queries and typical `@` use-cases.

coordinator/src/main/scala/filodb.coordinator/queryplanner/DefaultPlanner.scala
yu-shipit: the value of atMs is not used for step. when atMs is provided step should be 0 to turn the query an instant query.

coordinator/src/main/scala/filodb.coordinator/queryplanner/DefaultPlanner.scala
yu-shipit: The `RepeatTransformer` should take the original parameters.
Say we have range query` [1, 10, 100] @ 50`
We will have the following plan.
```
--RepeatTransformer(1, 10, 100) => populate the value returned by PeriodicMapper  forever every step
   --PeriodicMapper(50,0,50) => return only one value at timestamp 50
```

yu-shipit: sorry. I think maybe the term real is confusing.
the real time is the time with modifier applied.
Maybe I should change realStartMs to something like dataStartMs.

coordinator/src/main/scala/filodb.coordinator/queryplanner/LongTimeRangePlanner.scala
```diff
@@ -53,6 +53,8 @@ import filodb.query.exec._
     val earliestRawTime = earliestRawTimestampFn
     val offsetMillis = LogicalPlanUtils.getOffsetMillis(periodicSeriesPlan)
     val (maxOffset, minOffset) = (offsetMillis.max, offsetMillis.min)
+    require(periodicSeriesPlan.validateAtModifier(earliestRawTime),
+      s"$periodicSeriesPlan @modifier and query range should be all greater or less than $earliestRawTime")
 
     val lookbackMs = LogicalPlanUtils.getLookBackMillis(periodicSeriesPlan).max
     val startWithOffsetMs = periodicSeriesPlan.startMs - maxOffset
```
alextheimer, line 60: Would it be straightforward/correct to use these values to determine whether-or-not the query needs to be rejected? e.g:
```
startWithOffset - lookbackMs, all_at_timestamps >= earliestRawTime -> OK
endWithOffset, all_at_timestamps < earliestRawTime -> OK
else -> REJECT
```

So, the only additional method we'd need is something like `getAllAtTimestampMillis` that just walks the plan and aggregates a list of at-modifier timestamps from the leaves.

(Please sanity-check that this is consistent with the below if-else chain if you do explore this option.)

coordinator/src/test/scala/filodb.coordinator/queryplanner/PlannerHierarchySpec.scala
```diff
@@ -1837,47 +1837,67 @@ class PlannerHierarchySpec extends AnyFunSpec with Matchers with PlanValidationS
     validatePlan(execPlan, expectedPlan)
   }
 
-  it("Modifier should have PeriodicSamplesMapper wrapped by RepeatTransformer") {
+  it("Modifier should have PeriodicSamplesMapper wrapped by RepeatTransformer and all data in raw cluster") {
```
alextheimer, line 1840: Could you add a test case for each possible scenario?
- query routed entirely to raw (already exists)
- query routed entirely to DS (already exists)
- query routed to both (should reject)
- query start/end are in raw, but lookback extends into DS (should reject)
- query start/end are in raw, but offset pushes it entirely into DS (ok)
- [any other important scenarios I'm forgetting]

alextheimer (changes requested):

coordinator/src/main/scala/filodb.coordinator/queryplanner/LongTimeRangePlanner.scala
alextheimer: Is there a reason the `getDataSource`/`hasAtModifier` implementation is preferable?

query/src/main/scala/filodb/query/LogicalPlan.scala
```diff
@@ -71,23 +71,8 @@ sealed trait PeriodicSeriesPlan extends LogicalPlan {
...
-      1
-    } else {
-      2
-    }
-  }
-
-  def hasAtModifier: Boolean = false
-
+  // The array of all at modifier
+  def atModifierTimestampsWithOffset: Seq[Long] = Seq()
```
alextheimer, line 75: I see in the `LongTimeRangePlanner` that all `RawSeries` plans (and a couple others) are sent directly to the raw cluster regardless of timestamps. Should that be addressed in this PR?

alextheimer: This default implementation might bite us later when a new `PeriodicSeries` is added -- can you do one of either:
- Remove the default implementation.
- Create a `match` function in some utility class that stores this logic for all classes (it's easy to safely give a default implementation for a bunch of classes that way). Something like this:
```
        case lc: LabelCardinality          => materializeLabelCardinalityPlan(lc, qContext)
        ...
        case ts: TsCardinalities           => materializeTSCardinalityPlan(qContext, ts)
        case _: LabelValues |
             _: ApplyLimitFunction |
             _: SeriesKeysByFilters |
             _: ApplyInstantFunctionRaw |
             _: RawSeries |
             _: LabelNames                 => <insert default implementation here>
```

coordinator/src/main/scala/filodb.coordinator/queryplanner/LongTimeRangePlanner.scala
```diff
@@ -53,13 +53,26 @@ import filodb.query.exec._
...
+      // should be in raw cluster.
+      atModifierTimestampsWithOffset.forall(at => at - lookbackMs >= earliestRawTime)
+    } else if (endWithOffsetMs - lookbackMs < earliestRawTime) {
+      // should be in down sample cluster.
+      atModifierTimestampsWithOffset.forall(at => at - lookbackMs < earliestRawTime)
+    } else {
+      atModifierTimestampsWithOffset.isEmpty
+    }
+    require(isAtModifierValid, s"@modifier $atModifierTimestampsWithOffset is not supported because it queries data" +
+      s" from both downsample and raw cluster. Please adjust the start and end time if you want to use @modifier.")
```
alextheimer, line 74: This is pretty good, I think. Can you say "because it queries raw and downsampled data" (let's not mention "cluster" -- that might be confusing to anyone unfamiliar with filodb)?

alextheimer: > Please adjust the start and end time

Should we give all options?:
> please adjust the query range, offsets, selector durations, and @modifier timestamps

Or maybe more generally:
> please adjust query parameters

query/src/main/scala/filodb/query/LogicalPlan.scala
yu-shipit: can we apply @modifier for RawSeries?

alextheimer: Is something like this valid promql?
```
foo{label="value"}[1h] @ 12345
```

FWIW it would be more appropriate to just adjust the query range. Maybe since the logic currently sends all pure RawSeries queries to the raw cluster, we should just reject any with an @modifier in the DS cluster? And we could include a message like "please adjust the time-range"?

yu-shipit: `foo{label="value"}[1h] ` always goes to the raw cluster.
As discussed, we can fix this later.

coordinator/src/main/scala/filodb.coordinator/queryplanner/LongTimeRangePlanner.scala
yu-shipit: not needed.

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
```diff
@@ -119,16 +119,17 @@ class SingleClusterPlanner(val dataset: Dataset,
       //For binary join LHS & RHS should have same times
       val boundParams = periodicSeriesPlan.get.head match {
         case p: PeriodicSeries => (p.startMs, p.stepMs, WindowConstants.staleDataLookbackMillis,
-          p.offsetMs.getOrElse(0L), p.endMs)
-        case w: PeriodicSeriesWithWindowing => (w.startMs, w.stepMs, w.window, w.offsetMs.getOrElse(0L), w.endMs)
+          p.offsetMs.getOrElse(0L), p.atMs.getOrElse(0), p.endMs)
+        case w: PeriodicSeriesWithWindowing => (w.startMs, w.stepMs, w.window, w.offsetMs.getOrElse(0L),
+          w.atMs.getOrElse(0L), w.endMs)
```
alextheimer, line 124: Double-checking: no need to update these start/step/end values like we do for subqueries?

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
```diff
@@ -119,16 +119,17 @@ class SingleClusterPlanner(val dataset: Dataset,
...
         case _  => throw new UnsupportedOperationException(s"Invalid plan: ${periodicSeriesPlan.get.head}")
       }
 
       val newStartMs = boundToStartTimeToEarliestRetained(boundParams._1, boundParams._2, boundParams._3,
         boundParams._4)
-      if (newStartMs <= boundParams._5) { // if there is an overlap between query and retention ranges
+      if (newStartMs <= boundParams._6) { // if there is an overlap between query and retention ranges
         if (newStartMs != boundParams._1)
-          Some(LogicalPlanUtils.copyLogicalPlanWithUpdatedTimeRange(logicalPlan, TimeRange(newStartMs, boundParams._5)))
+          Some(LogicalPlanUtils.copyLogicalPlanWithUpdatedTimeRange(logicalPlan, TimeRange(newStartMs, boundParams._6)))
```
alextheimer, line 132: Outside the scope of this PR, but it's probably safe to say it's time for those `boundParams` to be given their own `case class` 😆

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
```diff
@@ -583,23 +584,35 @@ class SingleClusterPlanner(val dataset: Dataset,
     val execRangeFn = if (logicalPlanWithoutBucket.function == RangeFunctionId.AbsentOverTime) Last
                       else InternalRangeFunction.lpToInternalFunc(logicalPlanWithoutBucket.function)
 
+    val realScanStartMs = lp.atMs.getOrElse(logicalPlanWithoutBucket.startMs)
+    val realScanEndMs = lp.atMs.getOrElse(logicalPlanWithoutBucket.endMs)
+    val realScanStepMs: Long = lp.atMs.map(_ => 0L).getOrElse(lp.stepMs)
```
alextheimer, line 589: Same as some other comments: should we keep the original duration between start/end?

core/src/main/scala/filodb.core/query/RangeVector.scala
```diff
@@ -195,6 +195,103 @@ final case class ScalarVaryingDouble(val timeValueMap: Map[Long, Double],
...
+
+  val recordSchema: RecordSchema = schema
+
+  // There is potential for optimization.
+  // The parent transformer does not need to iterate all rows.
+  // It can transform one data because data at all steps are identical. It just need to return a RepeatValueVector.
+  override def rows(): RangeVectorCursor = {
+    import NoCloseCursor._
+    // If rowReader is empty, iterate nothing.
+    val it = Iterator.from(0, rowReader.map(_ => stepMs.toInt).getOrElse(0))
```
alextheimer, line 230: Should this be `getOrElse(1)` to prevent an infinite loop?

query/src/main/scala/filodb/query/LogicalPlan.scala
```diff
@@ -70,6 +70,9 @@ sealed trait PeriodicSeriesPlan extends LogicalPlan {
   def endMs: Long
 
   def replacePeriodicSeriesFilters(filters: Seq[ColumnFilter]): PeriodicSeriesPlan
+
+  // The array of all at modifier
+  def atModifierTimestampsWithOffset: Seq[Long] = Seq()
```
alextheimer, line 75: Double-checking: https://github.com/filodb/FiloDB/pull/1678#discussion_r1386119976

Can this default implementation be removed?

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
yu-shipit: Yeah. we have one more argument, which change the number to 6.

coordinator/src/main/scala/filodb.coordinator/queryplanner/DefaultPlanner.scala
```diff
@@ -338,17 +342,24 @@ trait  DefaultPlanner {
...
-        Some(InternalRangeFunction.lpToInternalFunc(RangeFunctionId.Last)),
-        qContext,
-        stepMultipleNotationUsed = false,
-        Seq(),
-        offsetMs,
-        rawSource = false
-      ))
+     val realScanStartMs = sqww.atMs.getOrElse(sqww.startMs)
+     val realScanEndMs = sqww.atMs.getOrElse(sqww.endMs)
+     val realScanStep = sqww.atMs.map(_ => 0L).getOrElse(sqww.stepMs)
```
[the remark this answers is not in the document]
yu-shipit: Somehow an instant query requires 0 to be step .

query/src/main/scala/filodb/query/LogicalPlan.scala
yu-shipit: image
A few classes have to implement this functions to remove the default one

alextheimer (approved):

PR #1680, opened 2023-10-09
fix(cluster): Fixes for clusterV2 in kubernetes setup 

sandeep6189:
**Pull Request checklist**

**Current behavior :** There were two issues in the ClusterV2 setup. They are:
i.  Incorrect resolution of hostnames, resulting in no communication between pods
ii. ShardHealthStats reporting assigned and unassigned status incorrectly

**New behavior :** Fixes for above issues is added

### DIFF
--- a/coordinator/src/main/scala/filodb/coordinator/v2/FiloDbClusterDiscovery.scala
+++ b/coordinator/src/main/scala/filodb/coordinator/v2/FiloDbClusterDiscovery.scala
@@ -72,16 +72,10 @@ class FiloDbClusterDiscovery(settings: FilodbSettings,
 
   lazy private val hostNames = {
     require(settings.minNumNodes.isDefined, "[ClusterV2] Minimum Number of Nodes config not provided")
-    if (settings.k8sHostFormat.isDefined && settings.akkaPort.isDefined) {
+    if (settings.k8sHostFormat.isDefined) {
       // This is used in kubernetes setup. We read the host format config and resolve the FQDN using env variables
       val hosts = (0 until settings.minNumNodes.get)
-        .map(i => {
-          val currentHostname = String.format(
-            settings.k8sHostFormat.get,
-            i.toString
-          )
-          s"${currentHostname}:${settings.akkaPort.get}"
-        })
+        .map(i => String.format(settings.k8sHostFormat.get, i.toString))
       logger.info(s"[ClusterV2] hosts to communicate: " + hosts)
       hosts.sorted
     } else if (settings.hostList.isDefined) {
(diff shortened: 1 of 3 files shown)


coordinator/src/main/scala/filodb/coordinator/v2/FiloDbClusterDiscovery.scala
```diff
@@ -71,12 +71,22 @@ class FiloDbClusterDiscovery(settings: FilodbSettings,
...
-    require(settings.minNumNodes.isDefined, "Minimum Number of Nodes config not provided")
+    require(settings.minNumNodes.isDefined, "[ClusterV2] Minimum Number of Nodes config not provided")
     if (settings.k8sHostFormat.isDefined) {
-      (0 until settings.minNumNodes.get).map(i => String.format(settings.k8sHostFormat.get, i.toString))
+      // This is used in kubernetes setup. We read the host format config and resolve the IP address from each host
+      val hosts = (0 until settings.minNumNodes.get)
+        .map(i => {
+          val resolvedIp = InetAddress.getByName(String.format(settings.k8sHostFormat.get, i.toString))
+          // return resolvedIpAddress and port combination for akka communication
+          s"${resolvedIp.getHostAddress()}:${settings.akkaPort}"
```
vishramachandran, line 81: This is a lazy val and will be initialized only once. Can the IP address of hosts change for any reason when any pod is bounced due to a crash etc? If yes, then this change will not capture the new IP address.

sandeep6189: discussed offline on this. We are now using hostname for akka actor registration and getting actor ref to communicate between the pods.

coordinator/src/main/scala/filodb/coordinator/v2/FiloDbClusterDiscovery.scala
```diff
@@ -71,12 +71,23 @@ class FiloDbClusterDiscovery(settings: FilodbSettings,
   def shardsForLocalhost(numShards: Int): Seq[Int] = shardsForOrdinal(ordinalOfLocalhost, numShards)
 
   lazy private val hostNames = {
-    require(settings.minNumNodes.isDefined, "Minimum Number of Nodes config not provided")
-    if (settings.k8sHostFormat.isDefined) {
-      (0 until settings.minNumNodes.get).map(i => String.format(settings.k8sHostFormat.get, i.toString))
+    require(settings.minNumNodes.isDefined, "[ClusterV2] Minimum Number of Nodes config not provided")
+    if (settings.k8sHostFormat.isDefined && settings.akkaPort.isDefined) {
```
vishramachandran, line 75: The k8sHostFormat config value itself can contain `${akka.remote.netty.tcp.port}` in it - it will be resolved. So code for adding`akkaPort` is not needed. But the current approach is also okay if you want to keep it.

vishramachandran (approved):
Looks great!

vishramachandran (approved):

coordinator/src/main/scala/filodb/coordinator/v2/FiloDbClusterDiscovery.scala
sandeep6189: Ack, Agree with you. Updated the config to simplify this.

PR #1681, opened 2023-10-16
feat(prometheus): relax length limit for pipe-only EqualsRegex filters

alextheimer:
**Pull Request checklist**

Currently, regex filter lengths are limited to 1000 chars. However, given that a [recent optimization](https://github.com/filodb/FiloDB/pull/1641) converts EqualsRegex pipe-only (i.e. `|`-only) filters to Lucene `IN` queries, this limit can now be relaxed specifically for these queries.

This PR relaxes the length-limit for regex filters that contain _at most_ the pipe special regex character. For example, the query...
```
foo{label1=~"baz|bar|bat|...|zab", label2=~"hello", label3=~"foo|far[0-9]", label4!~"hello"}
```
...will relax the length limit only for `label1` and `label2`.

### DIFF
--- a/prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
+++ b/prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
@@ -1,7 +1,8 @@
 package filodb.prometheus.ast
 
 import scala.util.Try
-import filodb.core.{GlobalConfig, query}
+
+import filodb.core.{query, GlobalConfig}
 import filodb.core.query.{ColumnFilter, RangeParams}
 import filodb.prometheus.Utils
 import filodb.prometheus.parse.Parser
--- a/prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
+++ b/prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
@@ -1,9 +1,9 @@
 package filodb.prometheus.ast
 
 import scala.util.Try
-
-import filodb.core.{query, GlobalConfig}
+import filodb.core.{GlobalConfig, query}
 import filodb.core.query.{ColumnFilter, RangeParams}
+import filodb.prometheus.Utils
 import filodb.prometheus.parse.Parser
 import filodb.query._
 
@@ -265,9 +265,14 @@ sealed trait Vector extends Expression {
         case NotRegexMatch   => require(labelValue.length <= Parser.REGEX_MAX_LEN,
                                          s"Regular expression filters should be <= ${Parser.REGEX_MAX_LEN} characters")
                                 ColumnFilter(labelMatch.label, query.Filter.NotEqualsRegex(labelValue))
-        case RegexMatch      => require(labelValue.length <= Parser.REGEX_MAX_LEN,
-                                         s"Regular expression filters should be <= ${Parser.REGEX_MAX_LEN} characters")
-                                ColumnFilter(labelMatch.label, query.Filter.EqualsRegex(labelValue))
+        case RegexMatch      =>
+          // Relax the length limit only for matchers that contain at most the "|" special character.
+          if (!Utils.isPipeOnlyRegex(labelValue)) {
+            require(labelValue.length <= Parser.REGEX_MAX_LEN,
+              s"Regular expression filters should be <= ${Parser.REGEX_MAX_LEN} characters " +
+                s"when non-`|` special characters are used.")
+          }
+          ColumnFilter(labelMatch.label, query.Filter.EqualsRegex(labelValue))
         case NotEqual(false) => ColumnFilter(labelMatch.label, query.Filter.NotEquals(labelValue))
         case other: Any      => throw new IllegalArgumentException(s"Unknown match operator $other")
       }
--- a/prometheus/src/main/scala/filodb/prometheus/parse/Parser.scala
+++ b/prometheus/src/main/scala/filodb/prometheus/parse/Parser.scala
@@ -1,9 +1,9 @@
 package filodb.prometheus.parse
 
 import com.typesafe.scalalogging.StrictLogging
-
 import filodb.core.GlobalConfig
 import filodb.core.query.{ColumnFilter, Filter, QueryConfig}
+import filodb.prometheus.Utils
 import filodb.prometheus.ast._
 import filodb.query.{LabelValues, LogicalPlan}
 
@@ -138,9 +138,14 @@ object Parser extends StrictLogging {
             case NotRegexMatch => require(l.value.length <= REGEX_MAX_LEN,
                                    s"Regular expression filters should be <= $REGEX_MAX_LEN characters")
                                   ColumnFilter(l.label, Filter.NotEqualsRegex(l.value))
-            case RegexMatch =>  require(l.value.length <= REGEX_MAX_LEN,
-                                   s"Regular expression filters should be <= $REGEX_MAX_LEN characters")
-                                ColumnFilter(l.label, Filter.EqualsRegex(l.value))
+            case RegexMatch    =>
+              // Relax the length limit only for matchers that contain at most the "|" special character.
+              if (!Utils.isPipeOnlyRegex(l.value)) {
+                require(l.value.length <= Parser.REGEX_MAX_LEN,
+                  s"Regular expression filters should be <= ${Parser.REGEX_MAX_LEN} characters " +
+                    s"when non-`|` special characters are used.")
+              }
+              ColumnFilter(l.label, Filter.EqualsRegex(l.value))
             case NotEqual(false) => ColumnFilter(l.label, Filter.NotEquals(l.value))
             case other: Any => throw new IllegalArgumentException(s"Unknown match operator $other")
           }
--- a/prometheus/src/main/scala/filodb/prometheus/parse/Parser.scala
+++ b/prometheus/src/main/scala/filodb/prometheus/parse/Parser.scala
@@ -1,6 +1,7 @@
 package filodb.prometheus.parse
 
 import com.typesafe.scalalogging.StrictLogging
+
 import filodb.core.GlobalConfig
 import filodb.core.query.{ColumnFilter, Filter, QueryConfig}
 import filodb.prometheus.Utils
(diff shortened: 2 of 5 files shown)


prometheus/src/main/scala/filodb/prometheus/query/PrometheusModel.scala
```diff
@@ -7,15 +7,16 @@ import filodb.core.binaryrecord2.{BinaryRecordRowReader, StringifyMapItemConsume
...
 import filodb.core.query.{Result => _, _}
+import filodb.prometheus.Utils
 import filodb.prometheus.parse.Parser.REGEX_MAX_LEN
 import filodb.query.{QueryResult => FiloQueryResult, _}
 import filodb.query.AggregationOperator.Avg
 import filodb.query.exec.{ExecPlan, HistToPromSeriesMapper}
 
 object PrometheusModel {
   import com.softwaremill.quicklens._
-  val conf = GlobalConfig.defaultsFromUrl
```
alextheimer, line 17: Changed so the resolved config is given (instead of just the defaults).

core/src/main/scala/filodb.core/query/QueryUtils.scala
```diff
@@ -0,0 +1,25 @@
+package filodb.core.query
+
+/**
+ * Storage for utility functions.
+ */
+object QueryUtils {
```
alextheimer, line 6: The functions here keep the logic consistent between Lucene and the Parser.

core/src/main/scala/filodb.core/query/QueryUtils.scala
```diff
@@ -0,0 +1,25 @@
...
+    str.exists(regexChars.contains(_))
+  }
+
+  /**
+   * Returns true iff the argument string contains no special regex
+   *   characters except the pipe character ('|').
+   * True is also returned when the string contains no pipes.
+   */
+  def isPipeOnlyRegex(str: String): Boolean = {
+    str.forall(!regexCharsMinusPipe.contains(_))
```
alextheimer, line 23: I've (more-or-less) copy-pasted the Lucene logic here to keep the diff simple/safe. To correctly handle escaped characters, we'd need something 10x less pretty:
```
  /**
   * Returns the set of unescaped special regex chars in the argument string.
   * Special chars are: . ? + * | { } [ ] ( ) " \
   */
  def getUnescapedSpecialRegexChars(str: String): Set[Char] = {
    // Match special chars preceded by any count of backslash pairs and either
    //   some non-backslash character or the beginning of a line.
    val regex = "(?<=(^|[^\\\\]))(\\\\\\\\)*([.?+*|{}\\[\\]()\"\\\\])".r
    regex.findAllMatchIn(str)
      .map(_.group(3)) // get the special char -- third capture group
      .map(_(0)) // convert the string to a char
      .toSet
  }
```

core/src/main/scala/filodb.core/query/QueryUtils.scala
```diff
@@ -0,0 +1,25 @@
...
+ */
+object QueryUtils {
+  private val regexChars = Array('.', '?', '+', '*', '|', '{', '}', '[', ']', '(', ')', '"', '\\')
+  private val regexCharsMinusPipe = (regexChars.toSet - '|').toArray
+
+  /**
+   * Returns true iff the argument string contains any special regex chars.
+   */
+  def containsRegexChars(str: String): Boolean = {
+    str.exists(regexChars.contains(_))
```
amolnayak311, line 14: Why ``exists`` when the original implementation was ``forall``?

amolnayak311 (dismissed):
@alextheimer do you know the max number of IN we can add in Lucene or any performance implications of having a large number of values? Today we might be protected by the max length of regex check but if we disable those for pipe only regex, do we have a clear understanding of what will/will not work?

I will let @vishramachandran comment on this too.

alextheimer:
@amolnayak311 I've added a relaxed-but-enforced limit, as well.

core/src/main/scala/filodb.core/query/QueryUtils.scala
alextheimer: Seemed more scala-thonic 🤷 
I can change it back if there's some weird difference between the two and/or the consensus is that `forall` is more readable?

amolnayak311: they aren't same , ``exists`` returns true of at least one is true and ``forall`` requires all to be ``true``

```
scala> val l = List(1, 2, 3, 4)
l: List[Int] = List(1, 2, 3, 4)

scala> l.exists(_ > 2)
res5: Boolean = true

scala> l.forall(_ > 2)
res6: Boolean = false

```

amolnayak311: I see the difference after speaking offline, totally missed the `!`, I agree the refactor is more readable than original .

prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
```diff
@@ -265,9 +267,23 @@ sealed trait Vector extends Expression {
         case NotRegexMatch   => require(labelValue.length <= Parser.REGEX_MAX_LEN,
                                          s"Regular expression filters should be <= ${Parser.REGEX_MAX_LEN} characters")
                                 ColumnFilter(labelMatch.label, query.Filter.NotEqualsRegex(labelValue))
-        case RegexMatch      => require(labelValue.length <= Parser.REGEX_MAX_LEN,
-                                         s"Regular expression filters should be <= ${Parser.REGEX_MAX_LEN} characters")
-                                ColumnFilter(labelMatch.label, query.Filter.EqualsRegex(labelValue))
+        case RegexMatch      =>
+          // Relax the length limit only for matchers that contain at most the "|" special character.
+          val shouldRelax = queryConfig.getBoolean("should-relax-pipe-only-equals-regex-limit") &&
```
sandeep6189, line 272: wondering if we can consider merging the two flags to a single flag for simplicity ? We can get the desired functionality just by `relaxed-pipe-only-equals-regex-limit` flag

```
val pipeOnlyCharLimit : Option[int] = 
if queryConfig.hasPath("relaxed-pipe-only-equals-regex-limit") {
 queryConfig.getInt("relaxed-pipe-only-equals-regex-limit");
}
else {
  None
}

val shouldRelax = QueryUtils.isPipeOnlyRegex && pipeOnlyCharLimit != None);
```

prometheus/src/main/scala/filodb/prometheus/parse/Parser.scala
```diff
@@ -138,9 +140,23 @@ object Parser extends StrictLogging {
             case NotRegexMatch => require(l.value.length <= REGEX_MAX_LEN,
                                    s"Regular expression filters should be <= $REGEX_MAX_LEN characters")
                                   ColumnFilter(l.label, Filter.NotEqualsRegex(l.value))
-            case RegexMatch =>  require(l.value.length <= REGEX_MAX_LEN,
-                                   s"Regular expression filters should be <= $REGEX_MAX_LEN characters")
-                                ColumnFilter(l.label, Filter.EqualsRegex(l.value))
+            case RegexMatch    =>
+              // Relax the length limit only for matchers that contain at most the "|" special character.
+              val shouldRelax = queryConf.getBoolean("should-relax-pipe-only-equals-regex-limit") &&
```
sandeep6189, line 145: Same as above ^

prometheus/src/main/scala/filodb/prometheus/query/PrometheusModel.scala
```diff
@@ -42,9 +42,22 @@ object PrometheusModel {
           case MatchType.EQUAL => Filter.Equals(m.getValue)
           case MatchType.NOT_EQUAL => Filter.NotEquals(m.getValue)
           case MatchType.REGEX_MATCH =>
-                            require(m.getValue.length <= REGEX_MAX_LEN, s"Regular expression filters should " +
-                              s"be <= ${REGEX_MAX_LEN} characters")
-                            Filter.EqualsRegex(m.getValue)
+            // Relax the length limit only for matchers that contain at most the "|" special character.
+            val shouldRelax = queryConfig.getBoolean("should-relax-pipe-only-equals-regex-limit") &&
```
sandeep6189, line 46: Same as above ^

core/src/main/scala/filodb.core/query/QueryUtils.scala
```diff
@@ -0,0 +1,25 @@
+package filodb.core.query
```
sandeep6189, line 1: Kudos on moving the utility functions to a different file :) Much appreciated and thanks a tonne :)

core/src/main/scala/filodb.core/query/QueryUtils.scala
```diff
@@ -0,0 +1,25 @@
...
+ * Storage for utility functions.
+ */
+object QueryUtils {
+  private val regexChars = Array('.', '?', '+', '*', '|', '{', '}', '[', ']', '(', ')', '"', '\\')
+  private val regexCharsMinusPipe = (regexChars.toSet - '|').toArray
+
+  /**
+   * Returns true iff the argument string contains any special regex chars.
+   */
+  def containsRegexChars(str: String): Boolean = {
```
sandeep6189, line 13: Wondering if it is a good idea to have a unit test just for the `containsRegexChars` and `isPipeOnlyRegex`. ( Please feel free to discard if we are already testing this path in unit tests )

prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
alextheimer: That is way better (especially since there are already so many configs) -- I will combine these.

core/src/main/scala/filodb.core/query/QueryUtils.scala
alextheimer: Good call -- I'll add a couple specifically for these methods.

sandeep6189: Thanks a tonne :)

sandeep6189 (approved):

PR #1684, opened 2023-11-01
hack for GRPC execution plans

kvpetrov:
**Pull Request checklist**

**New behavior :**

**BREAKING CHANGES**

Breaking changes may include:

**Other information**:

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/ActorPlanDispatcher.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/ActorPlanDispatcher.scala
@@ -1,16 +1,14 @@
 package filodb.coordinator
 
 import java.util.concurrent.TimeUnit
-
 import scala.concurrent.duration.FiniteDuration
-
 import akka.actor.ActorRef
-import akka.pattern.{ask, AskTimeoutException}
+import akka.pattern.{AskTimeoutException, ask}
 import akka.util.Timeout
+import filodb.coordinator.client.QueryCommands.ProtoExecPlan
 import monix.eval.Task
 import monix.execution.Scheduler
 import monix.reactive.Observable
-
 import filodb.core.QueryTimeoutException
 import filodb.core.query.{QueryStats, QueryWarnings, ResultSchema}
 import filodb.core.store.ChunkSource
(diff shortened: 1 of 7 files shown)


coordinator/src/main/scala/filodb/coordinator/ProtoConverters.scala
```diff
@@ -0,0 +1,376 @@
...
+import java.util.concurrent.TimeUnit
+import scala.collection.JavaConverters
+import scala.concurrent.Await
+import scala.concurrent.duration.Duration
+
+// scalastyle:off number.of.methods
+// scalastyle:off number.of.types
+// scalastyle:off file.size.limit
+// scalastyle:off method.length
+object ProtoConverters {
```
amolnayak311, line 21: Why not use the existing ``ProtoConverters``, especially when the ExecPlans are in query? Please use the same  object and add your converters there. Also make sure you tests are written for each and every converter.

coordinator/src/main/scala/filodb.coordinator/ActorPlanDispatcher.scala
```diff
@@ -44,7 +42,15 @@ case class ActorPlanDispatcher(target: ActorRef, clusterName: String) extends Pl
           emptyPartialResult
         })
       } else {
-        val fut = (target ? plan.execPlan) (t).map {
+        // HACK
```
amolnayak311, line 45: Is this really a hack?

query/src/main/scala/filodb/query/ProtoConverters.scala
```diff
@@ -3,13 +3,10 @@ package filodb.query
 
 import java.util.concurrent.TimeoutException
 import java.util.concurrent.atomic.{AtomicInteger, AtomicLong}
-
 import scala.collection.JavaConverters._
```
amolnayak311, line 6: These empty lines are needed for style checks, add them back.

amolnayak311 (changes requested):

coordinator/src/main/scala/filodb/coordinator/ProtoConverters.scala
kvpetrov: dispatchers are in the coordinator, otherwise, I would have kept them in the original ProtoConverters, in fact I started writing the logic in the original ProtoConverters.

coordinator/src/main/scala/filodb.coordinator/ActorPlanDispatcher.scala
kvpetrov: yes, this is a hack, not meant to be committed, this PR is just a POC that shows how proto exec plans can works end to end and solicit the feedback. Here, for example the question is how would we want to proceed with the configuration switch that would enable proto serialization for the exec plans. I think the most preferable option is to pass it as a query parameter, smth like protoExec=true.

query/src/main/scala/filodb/query/ProtoConverters.scala
kvpetrov: sorry, this is just a typo

PR #1689, opened 2023-12-01
fix(query) Corner case for Set Operator OR with on()

amolnayak311:
**Pull Request checklist**

**Current and New behavior :** 

SetOperator OR fixed in [PR](https://github.com/filodb/FiloDB/pull/1669) has a corner case specifically when using ``OR on``. This since we use a Map which doesn't follow the order of the keys, RHS Rvs were getting executed prior to all LHS rvs get executed. Since the correctness of result depends on the order or execution since we use a bitmap to track all time instants where some LHS RV has results, there were some time instants where RHS RV returns results when it should have returned ``NaN``. This PR fixes this edge case.


kvpetrov (approved):
LGTM

PR #1692, opened 2023-12-07
misc(query): limit partKey/partId lookup results early

sherali42:
**Pull Request checklist**

- Stop collecting results once limit is hit. This will avoid unnecessary cpu cycles and heap usage

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
+++ b/core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
@@ -1190,11 +1190,10 @@ class PartIdCollector(limit: Int) extends SimpleCollector {
   }
 
   override def collect(doc: Int): Unit = {
-    if (partIdDv.advanceExact(doc)) {
+    if (result.length >= limit) {
+      throw new CollectionTerminatedException
+    } else if (partIdDv.advanceExact(doc)) {
       result += partIdDv.longValue().toInt
-      if (result.length >= limit) {
-        throw new CollectionTerminatedException
-      }
     } else {
       throw new IllegalStateException("This shouldn't happen since every document should have a partIdDv")
     }
@@ -1240,14 +1239,13 @@ class PartKeyRecordCollector(limit: Int) extends SimpleCollector {
   }
 
   override def collect(doc: Int): Unit = {
-    if (partKeyDv.advanceExact(doc) && startTimeDv.advanceExact(doc) && endTimeDv.advanceExact(doc)) {
+    if (records.size >= limit) {
+      throw new CollectionTerminatedException
+    } else if (partKeyDv.advanceExact(doc) && startTimeDv.advanceExact(doc) && endTimeDv.advanceExact(doc)) {
       val pkBytesRef = partKeyDv.binaryValue()
       // Gotcha! make copy of array because lucene reuses bytesRef for next result
       val pkBytes = util.Arrays.copyOfRange(pkBytesRef.bytes, pkBytesRef.offset, pkBytesRef.offset + pkBytesRef.length)
       records += PartKeyLuceneIndexRecord(pkBytes, startTimeDv.longValue(), endTimeDv.longValue())
-      if (records.size >= limit) {
-        throw new CollectionTerminatedException
-      }
     } else {
       throw new IllegalStateException("This shouldn't happen since every document should have partIdDv and startTimeDv")
     }
--- a/core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
+++ b/core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
@@ -903,7 +903,7 @@ class PartKeyLuceneIndex(ref: DatasetRef,
                          startTime: Long,
                          endTime: Long,
                          limit: Int = Int.MaxValue): debox.Buffer[Int] = {
-    val collector = new PartIdCollector(limit) // passing zero for unlimited results
+    val collector = new PartIdCollector(limit)
     searchFromFilters(columnFilters, startTime, endTime, collector)
     collector.result
   }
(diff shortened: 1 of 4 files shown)


alextheimer (dismissed):
Excellent catch @sherali42 @amolnayak311

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -901,8 +901,9 @@ class PartKeyLuceneIndex(ref: DatasetRef,
   //scalastyle:on method.length
   def partIdsFromFilters(columnFilters: Seq[ColumnFilter],
                          startTime: Long,
-                         endTime: Long): debox.Buffer[Int] = {
-    val collector = new PartIdCollector() // passing zero for unlimited results
+                         endTime: Long,
+                         limit: Int = Int.MaxValue): debox.Buffer[Int] = {
+    val collector = new PartIdCollector(limit) // passing zero for unlimited results
```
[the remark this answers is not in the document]
sherali42: nice catch @alextheimer copy paste error, fixed it.

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -1190,6 +1192,9 @@ class PartIdCollector extends SimpleCollector {
   override def collect(doc: Int): Unit = {
     if (partIdDv.advanceExact(doc)) {
       result += partIdDv.longValue().toInt
+      if (result.length >= limit) {
+        throw new CollectionTerminatedException
```
sandeep6189, line 1196: So instead of long-running queries ( resulting in taking away resources for large durations > n min ), we are failing fast here ?

sandeep6189 (dismissed):
Thanks a bunch @sherali42 @amolnayak311 . This will definitely help protect the system with runaway queries

core/src/test/scala/filodb.core/memstore/PartKeyLuceneIndexSpec.scala
```diff
@@ -65,10 +65,18 @@ class PartKeyLuceneIndexSpec extends AnyFunSpec with Matchers with BeforeAndAfte
...
 
+    // return only 4 partIds - empty filter
+    val partNumsLimit = keyIndex.partIdsFromFilters(Nil, start, end, 4)
+    partNumsLimit shouldEqual debox.Buffer(0, 1, 2, 3)
+
     val filter2 = ColumnFilter("Actor2Code", Equals("GOV".utf8))
     val partNums2 = keyIndex.partIdsFromFilters(Seq(filter2), start, end)
     partNums2 shouldEqual debox.Buffer(7, 8, 9)
 
+    // return only 4 partIds - empty filter
```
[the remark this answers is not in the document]
sherali42: 👍 fixed, thanks for catching

sandeep6189 (approved):

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -1190,6 +1192,9 @@ class PartIdCollector extends SimpleCollector {
   override def collect(doc: Int): Unit = {
     if (partIdDv.advanceExact(doc)) {
       result += partIdDv.longValue().toInt
+      if (result.length >= limit) {
```
amolnayak311, line 1195: I think we should move these checks right when the ``collect`` call is made.  While the document isn't very clear it will terminate the further collection  on the segment but that means there will be one record collected every other segment before we stop collecting in that segment so we will essentially collect ``limit + (segments - 1)`` results. \

amolnayak311 (changes requested):

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -1176,7 +1178,7 @@ class TopKPartIdsCollector(limit: Int) extends Collector with StrictLogging {
   }
 }
 
-class PartIdCollector extends SimpleCollector {
+class PartIdCollector(limit: Int) extends SimpleCollector {
   val result: debox.Buffer[Int] = debox.Buffer.empty[Int]
```
amolnayak311, line 1182: Now that we know the limit not sure how pre-initializing to size limit would play. That is using ``ofSize`` instead of ``empty``. With default size of 8 is almost always guaranteed to re allocate the buffer.

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -1220,7 +1224,7 @@ class PartIdStartTimeCollector extends SimpleCollector {
   }
 }
 
-class PartKeyRecordCollector extends SimpleCollector {
+class PartKeyRecordCollector(limit: Int) extends SimpleCollector {
   val records = new ArrayBuffer[PartKeyLuceneIndexRecord]
```
amolnayak311, line 1228: Same as before for pre-initialing with size to avoid reallocating space.

amolnayak311 (approved):
LGTM, left a couple of comments. Thanks for the PR @sherali42

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
sherali42: for now leaving it as is

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
sherali42: its hard to fix the size, will defer it to later PRs

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -901,8 +901,9 @@ class PartKeyLuceneIndex(ref: DatasetRef,
   //scalastyle:on method.length
   def partIdsFromFilters(columnFilters: Seq[ColumnFilter],
                          startTime: Long,
-                         endTime: Long): debox.Buffer[Int] = {
-    val collector = new PartIdCollector() // passing zero for unlimited results
+                         endTime: Long,
```
vishramachandran, line 904: Can you remove the limit from this  and all the partId collectors ? This specific method is used in regular PromQL queries and it should not silently limit. The data could be aggregated

vishramachandran: Never mind - the default value for that param I see is Int.MaxValue, and PromQL API path is not passing in any value.

PR #1696, opened 2023-12-14
fix(cardinality): Records with overflow doesn't pass dataset + Aggregation issue in TsCardReduceExec

sandeep6189:
**Current behavior :**

PR fixes two issues:

1. Cardinality records with `_overflow_` prefix doesn't pass dataset as expected by the QueryService in V2 APIs. 

2. Correctness issue: Incorrect aggregation when the hashmap reaches the map size. As of today, we would skip the keys already present in the map and always aggregate the overflow bucket.

**Pull Request checklist**

**New behavior :**
The fixes and unit tests are added for the above issues.

### DIFF
--- a/query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
+++ b/query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
@@ -20,6 +20,7 @@ import filodb.memory.format._
 import filodb.memory.format.ZeroCopyUTF8String._
 import filodb.query._
 import filodb.query.Query.qLogger
+import filodb.query.exec.TsCardExec.MAX_RESULT_SIZE
 
 trait MetadataDistConcatExec extends NonLeafExecPlan {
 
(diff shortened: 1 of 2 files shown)


query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
```diff
@@ -106,8 +119,17 @@ final case class TsCardReduceExec(queryContext: QueryContext,
...
                                  firstSchema: Task[ResultSchema],
                                  querySession: QuerySession): Observable[RangeVector] = {
-    val taskOfResults = childResponses.flatMap(res => Observable.fromIterable(res._1.result))
-      .foldLeftL(new mutable.HashMap[ZeroCopyUTF8String, CardCounts])(mapFold)
+
+    // capture the result schema of responses
+    var rs = ResultSchema.empty
+    val flatMapData = childResponses.flatMap(res => {
+      if (rs == ResultSchema.empty) {
+        rs = res._1.resultSchema
```
yu-shipit, line 127: do all child responses have the same schema?

kvpetrov (approved):
LGTM

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
sandeep6189: Yep, for the same version, each of them will have same schema

query/src/test/scala/filodb/query/exec/MetadataExecSpec.scala
```diff
@@ -434,4 +435,64 @@ class MetadataExecSpec extends AnyFunSpec with Matchers with ScalaFutures with B
       }
     }
   }
+
+  it ("should add overflow group") {
```
alextheimer, line 439: A nontrivial test for a relatively simple fix -- the effort here is much appreciated

alextheimer (approved):
Thanks for this ultra-quick fix @sandeep6189

PR #1701, opened 2024-01-18
fix(query) Fixes for split partition queries

amolnayak311:
**Pull Request checklist**

**Current behavior :** 

The query processing assumes the partition split happens exactly at a given point in time, however, in reality there is a delay around that time and this period of uncertainty needs to be considered while computing the partition assignments

**New behavior :**

Fixes the issue by adding this period of uncertainty around the partition split


coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
```diff
@@ -550,15 +557,12 @@ class MultiPartitionPlanner(partitionLocationProvider: PartitionLocationProvider
...
-      // "distinct" in case this is a BinaryJoin
-      val partitions = getPartitions(logicalPlan, qParams).distinct.sortBy(_.timeRange.startMs)
-      require(!partitions.isEmpty, s"Partition assignments is not expected to be empty for query ${qParams.promQl}")
-      val timeRange = TimeRange(1000 * qParams.startSecs, 1000 * qParams.endSecs)
-      val stepMsOpt = if (qParams.startSecs == qParams.endSecs) None else Some(1000 * qParams.stepSecs)
-      getAssignmentQueryRanges(partitions, timeRange,
+    val timeRange = TimeRange(1000 * qParams.startSecs, 1000 * qParams.endSecs)
+    val stepMsOpt = if (qParams.startSecs == qParams.endSecs) None else Some(1000 * qParams.stepSecs)
+    val partitions = getPartitions(logicalPlan, qParams).distinct.sortBy(_.timeRange.startMs)
+    require(partitions.nonEmpty, s"Partition assignments is not expected to be empty for query ${qParams.promQl}")
```
alextheimer, line 563: Confirming: this won't affect, say, scalar queries?

coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
```diff
@@ -657,7 +670,36 @@ class MultiPartitionPlanner(partitionLocationProvider: PartitionLocationProvider
...
+      // period 1 - 1:10 can not be served from one partition alone and needs to be computed on query service. Here
+      // we will handle this case and add the missing execPlan if needed
+
+      if (partitions.length > 1 && queryConfig.routingConfig.supportRemoteRawExport) {
+        // Here we check if the assignment ranges cover the entire query duration
+        val (_, lastTimeRange) = assignmentRanges.last
+        if (lastTimeRange.endMs < timeRange.endMs) {
+          // this means we need to add the missing time range to the end to execute the bit on Query service
+          val (gapStartTimeMs, gapEndTimeMs) = stepMsOpt match {
+            case Some(step)   =>   (snapToStep(lastTimeRange.endMs + 1, step, timeRange.startMs),
```
alextheimer, line 686: Why don't these timestamps need the same `* 1000 / 1000 + 1` treatment as before?

coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
```diff
@@ -657,7 +670,36 @@ class MultiPartitionPlanner(partitionLocationProvider: PartitionLocationProvider
             (Some(next), ep)
         }
       }
-      execPlans
+      // Edge case with namespace across multiple partitions includes a case where the a part of time range
+      // of the original params fall in the new partition but that duration is less than the lookback. For example,
+      // Consider the timesplit happening at 1:00, the lookback of the query is 20 mins and the original query params
+      // have end date of 1:10. Then the period of 1 - 1:20 will not have any results, this is due to that fact the
+      // period 1 - 1:10 can not be served from one partition alone and needs to be computed on query service. Here
```
alextheimer, line 677: > this is due to that fact the period 1 - 1:10 can not be served from one partition alone and needs to be computed on query service.

Should we callout the `getQueryTimeRanges` method? Really that just needs to be updated to drop the old "empty result at partition splits" logic.

coordinator/src/test/scala/filodb.coordinator/StreamingResultsExecSpec.scala
```diff
@@ -153,7 +153,7 @@ class StreamingResultsExecSpec extends AnyFunSpec with Matchers with ScalaFuture
     resp(2).asInstanceOf[StreamQueryResultFooter].queryStats.getTimeSeriesScannedCounter().get() shouldEqual 2
   }
 
-  it("should execute Aggregation Exec Plans from Memstore in result streaming mode using actor plan dispatcher") {
+  ignore("should execute Aggregation Exec Plans from Memstore in result streaming mode using actor plan dispatcher") {
```
alextheimer, line 156: This is intentional?

alextheimer (dismissed):
Approved (with a couple questions)

coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
amolnayak311: Because we are doing snapToStep now which will take care to align the start exactly at the time where it needs to start.

coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
amolnayak311: Great point, this forced me to try few cases which failed with this. I pushed a fix for that. As far as the above line is concerned we are good.

```
Caused by: java.lang.IllegalArgumentException: requirement failed
	at scala.Predef$.require(Predef.scala:268)
	at filodb.query.BinaryJoin.<init>(LogicalPlan.scala:462)
	at filodb.query.BinaryJoin.copy(LogicalPlan.scala:460)
	at filodb.coordinator.queryplanner.PlannerUtil$.rewritePlanWithRemoteRawExport(DefaultPlanner.scala:666)
```

alextheimer (approved):

PR #1707, opened 2024-01-31
fix(query): metric name ColumnFilter should replace `_bucket` only when present in suffix

sandeep6189:
**Pull Request checklist**

**Current behavior :** The query plan updates the `_metric_` ColumnFilter to replace the `_bucket` string, even when not present in suffix. This results in data not being found, since the metric name column filter is wrong.

**New behavior :** Fixed the erroneous behavior by adding a check to only replace when `_bucket` is present in the suffix.


coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
```diff
@@ -620,15 +621,21 @@ class SingleClusterPlanner(val dataset: Dataset,
...
-             Right(lp.right.get.copy(series = rawSeriesLp.copy(filters = filtersWithoutBucket)))
-          (nameFilter, leFilter, newLp)
+          // the convention for histogram bucket queries is to have the "_bucket" string in the suffix
+          if (!nameFilter.get.endsWith("_bucket")) {
+            (nameFilter, leFilter, lp)
+          }
+          else {
+            val filtersWithoutBucket = rawSeriesLp.filters.filterNot(_.column.equals(PromMetricLabel)).
+              filterNot(_.column == "le") :+ ColumnFilter(PromMetricLabel,
+              Equals(nameFilter.get.replace("_bucket", "")))
```
yu-shipit, line 631: Do we only replace the _bucket in the end?
consider this case
`time_bucket_bucket  => time_bucket`

sandeep6189: @yu-shipit excellent point :) I wanted to keep the PR as simple as I can, without breaking any existing scenarios. So I just added an additional check before the existing replace. 

But you make an excellent point. This is definitely a good case. 

 @alextheimer @amolnayak311 @sherali42 should we tackle this case that @yu-shipit pointed out ? Any other scenario that we should think for ?

sandeep6189: I went ahead and added a helper function to do this. Please let me know if there is any concern here.

alextheimer (approved):
LGTM -- thanks for both debugging and resolving this @sandeep6189 🙏

sandeep6189:
test this please

yu-shipit (approved):

PR #1709, opened 2024-02-01
perf(query) Option to disable Lucene caching

amolnayak311:
**Pull Request checklist**

**Current behavior :** 

When we profile a FIloDB instance with a large index and which sees a lot of repeating queries, we see the following flame graph (as taken by async profiler)

Screenshot 2024-02-02 at 10 00 10 AM

We  see almost 80% stack traces are related to index searches and about 50% of them are related to caching.  Also [this](https://www.mail-archive.com/java-user@lucene.apache.org/msg51649.html) archive mentions disabling caching which may or may not work in our case   (also they mention disabling cache in a different way than what this PR does).  The idea is to have that knob to let us disable the caching and then profile again.

**New behavior :**

We now support flag based enable/disable caching. By default the caching is enabled (existing default)

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
+++ b/core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
@@ -133,7 +133,8 @@ class PartKeyLuceneIndex(ref: DatasetRef,
                          retentionMillis: Long, // only used to calculate fallback startTime
                          diskLocation: Option[File] = None,
                          val lifecycleManager: Option[IndexMetadataStore] = None,
-                         useMemoryMappedImpl: Boolean = true
+                         useMemoryMappedImpl: Boolean = true,
+                         disableIndexCaching: Boolean = false
                         ) extends StrictLogging {
 
   import PartKeyLuceneIndex._
@@ -240,7 +241,26 @@ class PartKeyLuceneIndex(ref: DatasetRef,
   private val utf8ToStrCache = concurrentCache[UTF8Str, String](PartKeyLuceneIndex.MAX_STR_INTERN_ENTRIES)
 
   //scalastyle:off
-  private val searcherManager = new SearcherManager(indexWriter, null)
+  private val searcherManager =
+    if (disableIndexCaching) {
+      new SearcherManager(indexWriter,
+        new SearcherFactory() {
+          override def newSearcher(reader: IndexReader, previousReader: IndexReader): IndexSearcher = {
+            val indexSearcher = super.newSearcher(reader, previousReader)
+            indexSearcher.setQueryCache(null)
+            indexSearcher.setQueryCachingPolicy(new QueryCachingPolicy() {
+              override def onUse(query: Query): Unit = {
+
+              }
+
+              override def shouldCache(query: Query): Boolean = false
+            })
+            indexSearcher
+          }
+        })
+    } else {
+      new SearcherManager(indexWriter, null)
+    }
   //scalastyle:on
 
   //start this thread to flush the segments and refresh the searcher every specific time period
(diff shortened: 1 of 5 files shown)


core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -240,7 +241,26 @@ class PartKeyLuceneIndex(ref: DatasetRef,
...
 
   //scalastyle:off
-  private val searcherManager = new SearcherManager(indexWriter, null)
+  private val searcherManager =
+    if (disableIndexCaching) {
+      new SearcherManager(indexWriter,
+        new SearcherFactory() {
+          override def newSearcher(reader: IndexReader, previousReader: IndexReader): IndexSearcher = {
+            val indexSearcher = super.newSearcher(reader, previousReader)
+            indexSearcher.setQueryCache(null)
```
sandeep6189, line 250: do we need to still set the cache to null if shouldCache returns false ?

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -240,7 +241,26 @@ class PartKeyLuceneIndex(ref: DatasetRef,
...
+            indexSearcher.setQueryCachingPolicy(new QueryCachingPolicy() {
+              override def onUse(query: Query): Unit = {
```
sandeep6189, line 252: quick question: what does onUse do ?

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -284,6 +284,8 @@ class TimeSeriesShard(val ref: DatasetRef,
                            filodbConfig.getBoolean("memstore.index-faceting-enabled-shard-key-labels")
   private val indexFacetingEnabledAllLabels = filodbConfig.getBoolean("memstore.index-faceting-enabled-for-all-labels")
   private val numParallelFlushes = filodbConfig.getInt("memstore.flush-task-parallelism")
+  private val disableIndexCaching = filodbConfig.getBoolean("memstore.disable-index-caching")
```
sandeep6189, line 287: do we also need this in downsample time series shard?

sandeep6189 (approved):

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
whizkido: Describe why it is detrimental too

alextheimer (approved):
Approved, but +1 to all of @sandeep6189's questions (let's defer the DS config to a separate PR if that's a less-straightforward change, but the consistency would be nice to have).

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
amolnayak311: Nou sure its needed yet but not ruling out the possibility

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
amolnayak311: Good question @sandeep6189 Initially I set the caching policy and disable caching anything but later found setting cache to null also disables cache (at least from unit tests) and also eliminates the check for whether to cache for not. However, I left both changes in. 

Hello @whizkido good to see your comments  :)

When we profile a FIloDB instance with a large index and which sees a lot of repeating queries, we see the following flame graph (as taken by async profiler)

Screenshot 2024-02-02 at 10 00 10 AM

We  see almost 80% stack traces are related to index searches and about 50% of the are related to caching.  Also [this](https://www.mail-archive.com/java-user@lucene.apache.org/msg51649.html) archive mentions disabling caching which may or may not work in our case   (also they mention disabling cache in a different way than what this PR does).  The idea is to have that knob to let us disable the caching and then profile again. It might be worse but we can measure and check.

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
amolnayak311: Taken from Lucene documentation

> Callback that is called every time that a cached filter is used. This is typically useful if the policy wants to track usage statistics in order to make decisions.

PR #1712, opened 2024-02-09
feat(histogram): Adding support for min max for prom and delta histograms

sandeep6189:
**Pull Request checklist**

**New behavior :** Adding support for min max for prom and delta histograms

### DIFF
--- a/core/src/main/resources/filodb-defaults.conf
+++ b/core/src/main/resources/filodb-defaults.conf
@@ -183,6 +183,33 @@ filodb {
       downsample-schema = "delta-histogram"
     }
 
+    prom-histogram-min-max {
+      columns = ["timestamp:ts",
+                 "sum:double:detectDrops=true",
+                 "count:double:detectDrops=true",
+                 "min:double:detectDrops=true",
+                 "max:double:detectDrops=true",
+                 "h:hist:counter=true"]
+      value-column = "h"
+      downsamplers = ["tTime(0)", "dLast(1)", "dLast(2)", "dLast(3)", "dLast(4)", "hLast(5)"]
+      # Downsample periods are determined by counter dips of the count column
+      downsample-period-marker = "counter(2)"
+      downsample-schema = "prom-histogram-min-max"
+    }
+
+    delta-histogram-min-max {
+      columns = ["timestamp:ts",
+        "sum:double:{detectDrops=false,delta=true}",
+        "count:double:{detectDrops=false,delta=true}",
+        "min:double:{detectDrops=false,delta=true}",
+        "max:double:{detectDrops=false,delta=true}",
+        "h:hist:{counter=false,delta=true}"]
+      value-column = "h"
+      downsamplers = ["tTime(0)", "dSum(1)", "dSum(2)", , "dSum(3)", "dSum(4)", "hSum(5)"]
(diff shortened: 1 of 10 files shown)


core/src/main/resources/filodb-defaults.conf
```diff
@@ -183,6 +183,33 @@ filodb {
...
+      # Downsample periods are determined by counter dips of the count column
+      downsample-period-marker = "counter(2)"
+      downsample-schema = "prom-histogram-min-max"
+    }
+
+    delta-histogram-min-max {
+      columns = ["timestamp:ts",
+        "sum:double:{detectDrops=false,delta=true}",
+        "count:double:{detectDrops=false,delta=true}",
+        "min:double:{detectDrops=false,delta=true}",
```
amolnayak311, line 204: Based on [this](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#histogram) page the min and max fields are optional. For delta this especially means the value for min field can now be negative unlike sum, count and max.
Also, how should we interpret the word “optional”.  Is it guaranteed that once min and max are published they will always be published on each interval (even if same value or delta of 0) or the lack of this field be treated as delta of 0 ( I’m assuming this might be done to minimize data sent over the wire from agents?) in our case these are different schemas and we don’t want to fail queues with multiple schemas found error.

sandeep6189: Good question @amolnayak311 . From what I understand, we are adding a different schema altogether, which *needs* to have min-max provided by the agent (we are creating a new type/proto altogether to avoid this confusion) and the gateway would then consistently create the required schema, avoiding this issue altogether.

@sherali42

core/src/main/resources/filodb-defaults.conf
```diff
@@ -183,6 +183,33 @@ filodb {
       downsample-schema = "delta-histogram"
     }
 
+    prom-histogram-min-max {
```
amolnayak311, line 186: Do you want to call it prom anymore or something like otel? Technically Prometheus histograms dont have min max and prom stands for Prometheus.

sandeep6189: I like the idea of calling it `otel_histogram`.

core/src/main/scala/filodb.core/metadata/Dataset.scala
```diff
@@ -115,7 +115,7 @@ object DatasetOptions {
                                       metricColumn = "__name__",
                                       // defaults that work well for Prometheus
                                       ignoreShardKeyColumnSuffixes =
-                                        Map("__name__" -> Seq("_bucket", "_count", "_sum")),
+                                        Map("__name__" -> Seq("_bucket", "_count", "_sum", "_min", "_max")),
```
sherali42, line 118: @sandeep6189 this is not needed, we can revert this change

sandeep6189: Ack, reverting

core/src/main/resources/filodb-defaults.conf
sandeep6189: updated to Otel-histogram

sherali42 (approved):
thanks @sandeep6189 for adding support for min/max histograms

PR #1715, opened 2024-02-14
fix(query): histogram queries with PeriodSeriesWithWindowing logical plan don't apply le label filter

sandeep6189:
**Pull Request checklist**

**Current behavior :**  Histogram bucket queries with PeriodSeriesWithWindowing logical plan (queries such as sum(rate) etc. use this logical plan) don't apply the InstantVectorFunctionMapper transformer correctly for le label filter. 

**New behavior :** Adding the required transformer to correctly pick up the required bucket given in the query.

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
@@ -618,13 +618,16 @@ class SingleClusterPlanner(val dataset: Dataset,
     }
   }
 
+  // scalastyle:off method.length
   override private[queryplanner] def materializePeriodicSeriesWithWindowing(qContext: QueryContext,
                                                      lp: PeriodicSeriesWithWindowing,
                                                      forceInProcess: Boolean): PlanResult = {
 
-    val logicalPlanWithoutBucket = if (queryConfig.translatePromToFilodbHistogram) {
-       removeBucket(Right(lp))._3.right.get
-    } else lp
+    val (nameFilter: Option[String], leFilter: Option[String], logicalPlanWithoutBucket: PeriodicSeriesWithWindowing) =
+      if (queryConfig.translatePromToFilodbHistogram) {
+       val result = removeBucket(Right(lp))
+        (result._1, result._2, result._3.right.get)
+    } else (None, None, lp)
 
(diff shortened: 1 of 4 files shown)


coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
```diff
@@ -585,6 +588,15 @@ class SingleClusterPlanner(val dataset: Dataset,
       realScanStepMs, realScanEndMs, window, Some(execRangeFn), qContext,
       logicalPlanWithoutBucket.stepMultipleNotationUsed,
       paramsExec, logicalPlanWithoutBucket.offsetMs, rawSource)))
+
+    // Add the le filter transformer to select the required bucket
+    if (nameFilter.isDefined && nameFilter.head.endsWith("_bucket") && leFilter.isDefined) {
```
amolnayak311, line 593: ``nameFilter.exists(_.endsWith("_bucket"))`` is more readable. Pattern match is even better

```
(nameFilter, leFilter) match {

   case Some(filter), Some(le) if filter.endsWith("_bucket") => // Your logic here, use variable le and filter directly, no need to use head/get on options

case _  =>  //NOP

}
```

coordinator/src/test/scala/filodb.coordinator/queryplanner/SingleClusterPlannerSpec.scala
```diff
@@ -2202,6 +2202,26 @@ class SingleClusterPlannerSpec extends AnyFunSpec with Matchers with ScalaFuture
       isInstanceOf[StaticFuncArgs] shouldEqual(true)
   }
 
+  it("should add le label filter correctly for histogram query with PeriodicSeriesWithWindowing logical plan") {
```
amolnayak311, line 2205: Can you assert in the same style as we do in ``PlannerHierarchySpec``? Personally I prefer that as one can see the actual plan visually.

amolnayak311 (changes requested):

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
sandeep6189: Sounds good @amolnayak311 !! Updating this ^

amolnayak311 (approved):
Thanks @sandeep6189

PR #1722, opened 2024-02-27
fix(cluster): Adding metric to track when actor ref is not getting resolved

sandeep6189:
**Pull Request checklist**

**New behavior :** Adding a counter to track the occurrences of actor refs not getting resolved.

### DIFF
--- a/coordinator/src/main/scala/filodb/coordinator/v2/FiloDbClusterDiscovery.scala
+++ b/coordinator/src/main/scala/filodb/coordinator/v2/FiloDbClusterDiscovery.scala
@@ -8,7 +8,6 @@ import akka.pattern.ask
 import akka.util.Timeout
 import com.typesafe.scalalogging.StrictLogging
 import kamon.Kamon
-import kamon.tag.TagSet
 import monix.execution.{CancelableFuture, Scheduler}
 import monix.reactive.Observable
 import scala.collection.mutable
@@ -25,11 +24,13 @@ class FiloDbClusterDiscovery(settings: FilodbSettings,
 
   private val discoveryJobs = mutable.Map[DatasetRef, CancelableFuture[Unit]]()
 
-  // default tags
-  val tags = Map("source" -> "FiloDbClusterDiscovery")
-  val actorResolvedFailedCounter = Kamon.counter("actor-resolve-failed").withTags(TagSet.from(tags))
-  val clusterDiscoveryCounter = Kamon.counter("filodb-cluster-discovery").withTags(TagSet.from(tags))
-  val unassignedShardsGauge = Kamon.gauge("v2-unassigned-shards").withTags(TagSet.from(tags))
+
+  // Metric to track actor resolve failures
+  val actorResolvedFailedCounter = Kamon.counter("actor-resolve-failed")
+  // Metric to track cluster discovery runs
+  val clusterDiscoveryCounter = Kamon.counter("filodb-cluster-discovery")
+  // Metric to track if we have unassigned shards on a given pod
+  val unassignedShardsGauge = Kamon.gauge("v2-unassigned-shards")
 
   lazy val ordinalOfLocalhost: Int = {
     if (settings.localhostOrdinal.isDefined) settings.localhostOrdinal.get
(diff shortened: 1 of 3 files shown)


coordinator/src/main/scala/filodb/coordinator/v2/FiloDbClusterDiscovery.scala
```diff
@@ -24,6 +25,10 @@ class FiloDbClusterDiscovery(settings: FilodbSettings,
 
   private val discoveryJobs = mutable.Map[DatasetRef, CancelableFuture[Unit]]()
 
+  // default tags
+  val tags = Map("source" -> "FiloDbClusterDiscovery")
```
sherali42, line 29: why do we need this static tag?

sherali42 (approved):

coordinator/src/main/scala/filodb/coordinator/v2/FiloDbClusterDiscovery.scala
```diff
@@ -99,7 +107,7 @@ class FiloDbClusterDiscovery(settings: FilodbSettings,
     val empty = CurrentShardSnapshot(ref, new ShardMapper(numShards))
     def fut = (nca ? GetShardMapScatter(ref)) (t).asInstanceOf[Future[CurrentShardSnapshot]]
     Observable.fromFuture(fut).onErrorHandle { e =>
-      logger.error(s"Saw exception on askShardSnapshot: $e")
```
alextheimer, line 102: This will log even if clustering V1 is enabled?

alextheimer: Could you put the exception as the second argument to `logger.error`? I see it wasn't there before, but that will make stack traces more conventional.

coordinator/src/main/scala/filodb/coordinator/v2/FiloDbClusterDiscovery.scala
```diff
@@ -110,7 +118,18 @@ class FiloDbClusterDiscovery(settings: FilodbSettings,
...
-      ncaRef <- Observable.fromFuture(nca.resolveOne(settings.ResolveActorTimeout).recover{case _=> ActorRef.noSender})
+      ncaRef <- Observable.fromFuture(nca.resolveOne(settings.ResolveActorTimeout)
+        .recover {
+          // recovering with ActorRef.noSender
+          case e =>
+            // log the exception we got while trying to resolve and emit metric
+            logger.error(s"[ClusterV2] Actor Resolve Failed ! actor: ${nca.toString()}", e)
+            actorResolvedFailedCounter
+              .withTag("dataset", ref.dataset)
+              .withTag("actor", nca.toString())
```
alextheimer, line 129: Could these metrics suddenly explode in cardinality during a network outage?

coordinator/src/main/scala/filodb/coordinator/v2/FiloDbClusterDiscovery.scala
```diff
@@ -122,13 +141,20 @@ class FiloDbClusterDiscovery(settings: FilodbSettings,
   private val datasetToMapper = new ConcurrentHashMap[DatasetRef, ShardMapper]()
   def registerDatasetForDiscovery(dataset: DatasetRef, numShards: Int): Unit = {
     require(failureDetectionInterval.toMillis > 5000, "failure detection interval should be > 5s")
-    logger.info(s"Starting discovery pipeline for $dataset")
+    logger.info(s"[ClusterV2] Starting discovery pipeline for $dataset")
```
alextheimer, line 144: Same as above: these will log even during V1 clustering?

coordinator/src/main/scala/filodb/coordinator/v2/FiloDbClusterDiscovery.scala
sandeep6189: We use this class only for V2. So it won't be logged.

Surely, updating the exception log

coordinator/src/main/scala/filodb/coordinator/v2/FiloDbClusterDiscovery.scala
sandeep6189: Thinking, so the actor and actor path should be same here for V2. since we are still using hostnames. But I agree that, we will have to update it, if we go the ip route

sandeep6189: Sync'd offline. Only used for V2. we only log this for failure scenarios and worst cardinality is 21k

coordinator/src/main/scala/filodb/coordinator/v2/FiloDbClusterDiscovery.scala
sandeep6189: Nope, this is exclusive to V2

alextheimer (approved):

PR #1723, opened 2024-02-29
fix @modifier for multiple partition time series.

yu-shipit:
**Pull Request checklist**

**New behavior :**

**BREAKING CHANGES**

Breaking changes may include:

**Other information**:

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/queryplanner/LogicalPlanParser.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryplanner/LogicalPlanParser.scala
@@ -16,6 +16,7 @@ object QueryConstants {
   val ClosingCurlyBraces = "}"
   val Space = " "
   val Offset = "offset"
+  val At = "@"
   val OpeningRoundBracket = "("
   val ClosingRoundBracket = ")"
   val OpeningSquareBracket = "["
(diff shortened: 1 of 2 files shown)


coordinator/src/main/scala/filodb.coordinator/queryplanner/LogicalPlanParser.scala
```diff
@@ -64,7 +65,8 @@ object LogicalPlanParser {
 
   private def periodicSeriesToQuery(periodicSeries: PeriodicSeries): String = {
     // Queries like sum(foo) should not have window even though stale lookback is present
-    s"${rawSeriesLikeToQuery(periodicSeries.rawSeries, false)}"
+    val at = periodicSeries.atMs.fold("")(atMs => s" $At${atMs / 1000}")
```
alextheimer, line 68: What do you think about using `map` and `orElse` here? `fold` implies `atMs` is a sequence, I think.

alextheimer (approved):
Thanks @yu-shipit for the ultra-quick PR -- especially these tests as well 🙏 

As discussed offline:
- `RawSeries` don't contain an @mod attribute, so we can't just add the @mod string from `rawSeriesLikeToQuery`.
- `RawSeries` are the only case where an @mod timestamp will be baked into the plan range.

PR #1735, opened 2024-03-05
misc(query): feature flag to exclude start time for range functions

sherali42:
**Pull Request checklist**

- feature flag to disable [starttime inclusiveness in range functions](https://github.com/filodb/FiloDB/pull/1665)

### DIFF
--- a/core/src/main/resources/filodb-defaults.conf
+++ b/core/src/main/resources/filodb-defaults.conf
@@ -318,6 +318,9 @@ filodb {
 
   query {
 
+    # feature flag to enable start time inclusive in window lookback
+    inclusive_range = true
+
     # feature flag for query result streaming
     streaming-query-results-enabled = false
 
--- a/core/src/main/resources/filodb-defaults.conf
+++ b/core/src/main/resources/filodb-defaults.conf
@@ -319,7 +319,7 @@ filodb {
   query {
 
     # feature flag to enable start time inclusive in window lookback
-    inclusive_range = true
+    inclusive-range = true
 
     # feature flag for query result streaming
     streaming-query-results-enabled = false
--- a/core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
+++ b/core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
@@ -456,8 +456,8 @@ extends Iterator[ChunkSetInfoReader] {
     // advance window pointers and reset read index
     if (curWindowEnd == -1L) {
       curWindowEnd = start
-      val windowStart = if (isInclusiveRange) window else window - 1
-      curWindowStart = start - Math.max(windowStart, 0)  // window cannot be below 0, ie start should never be > end
+      val windowDuration = if (isInclusiveRange) window else window - 1
+      curWindowStart = start - Math.max(windowDuration, 0)  // window cannot be below 0, ie start should never be > end
     } else {
       curWindowEnd += step
       curWindowStart += step
--- a/core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
+++ b/core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
@@ -435,7 +435,8 @@ class WindowedChunkIterator(rv: RawDataRangeVector, start: Long, step: Long, end
                             var curWindowEnd: Long = -1L,
                             var curWindowStart: Long = -1L,
                             private var readIndex: Int = 0,
-                            windowInfos: Buffer[ChunkSetInfoReader] = Buffer.empty[ChunkSetInfoReader])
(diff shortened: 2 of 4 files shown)


yu-shipit (approved):
lgtm.

core/src/main/resources/filodb-defaults.conf
```diff
@@ -318,6 +318,9 @@ filodb {
 
   query {
 
+    # feature flag to enable start time inclusive in window lookback
+    inclusive_range = true
```
alextheimer, line 322: I think all of our filodb configs are `kebab-case`:
```suggestion
    inclusive-range = true
```

core/src/main/scala/filodb.core/store/ChunkSetInfo.scala
```diff
@@ -455,7 +456,8 @@ extends Iterator[ChunkSetInfoReader] {
     // advance window pointers and reset read index
     if (curWindowEnd == -1L) {
       curWindowEnd = start
-      curWindowStart = start - Math.max(window, 0)  // window cannot be below 0, ie start should never be > end
+      val windowStart = if (isInclusiveRange) window else window - 1
```
alextheimer, line 459: What do you think about this, instead?
```suggestion
      val windowDuration = if (isInclusiveRange) window else window - 1
```

query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
```diff
@@ -390,7 +397,8 @@ class SlidingWindowIterator(raw: RangeVectorCursor,
     * @param curWindowStart start time of the current window
     */
   private def shouldRemoveWindowHead(curWindowStart: Long): Boolean = {
-    (!windowQueue.isEmpty) && windowQueue.head.timestamp < curWindowStart
+    val windowStart = if (FiloQueryConfig.isInclusiveRange) curWindowStart else curWindowStart - 1
```
alextheimer, line 400: Should this be `+1`? Same as on line 387?

query/src/main/scala/filodb/query/exec/rangefn/RateFunctions.scala
```diff
@@ -329,8 +340,14 @@ class RateOverDeltaChunkedFunctionD extends ChunkedDoubleRangeFunction {
                                 endRowNum: Int): Unit =
     sumFunc.addTimeDoubleChunks(doubleVectAcc, doubleVect, doubleReader, startRowNum, endRowNum)
 
-  override def apply(windowStart: Long, windowEnd: Long, sampleToEmit: TransientRow): Unit =
-    sampleToEmit.setValues(windowEnd, sumFunc.sum / (windowEnd - (windowStart)) * 1000)
+  override def apply(windowStart: Long, windowEnd: Long, sampleToEmit: TransientRow): Unit = {
+    val curWindowStart = if (FiloQueryConfig.isInclusiveRange)
+                            windowStart
+                         else
+                            windowStart - 1
```
alextheimer, line 347: Shouldn't we make the window smaller if the range is not inclusive?

query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
```diff
@@ -378,7 +384,8 @@ class SlidingWindowIterator(raw: RangeVectorCursor,
     */
   private def shouldAddCurToWindow(curWindowStart: Long, cur: TransientRow): Boolean = {
     // cur is inside current window
-    cur.timestamp >= curWindowStart
+    val windowStart = if (FiloQueryConfig.isInclusiveRange) curWindowStart else curWindowStart + 1
+    cur.timestamp >= windowStart
```
alextheimer, line 388: Shouldn't we only need to change the initial start timestamp? Then there's no need to update this (or any other) logic?

The assumption I'm making is that this is always true:
```
curWindowEnd - curWindowStart == adjustedWindowDuration
```

sherali42: fix is to revert the new start-inclusive behavior if config is set to false

query/src/main/scala/filodb/query/exec/rangefn/RateFunctions.scala
sherali42: fix is to revert the new start-inclusive behavior if config is set to false

query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
sherali42: since we check if an element is falling out of the window, this should work

alextheimer (approved):
LGTM -- thanks for the discussion/changes 🙏

PR #1745, opened 2024-03-21
misc(iceberg-export): remove assert and allow None label-values as some labels may not exist in all timeseries

nikitag55:
Current behavior :

Some labels as configured in label-column-mapping (to add as more columns in the table) may not exist in all the timeseries. So, the assert statement can fail the export job if the label may not exist in the timeseries and hence, label-value can be None.

Currently, `none` strings are appearing in the Iceberg Table when some column has no values. 

New behavior :

- Remove the assert statement such that, if the label doesn't exist in the timeseries. 
- Remove dynamic setting of `none` string in favor of Iceberg columns can be of only NULL Type or NOT NULL Type.
- When there is no value for a label, write it as `None` from spark to the Iceberg Table column, iceberg will automatically treat it as a `null` value and will show empty values in the table's column. 
- Remove assert and handle key not found in Map gracefully.

### DIFF
--- a/spark-jobs/src/main/scala/filodb/downsampler/chunk/BatchExporter.scala
+++ b/spark-jobs/src/main/scala/filodb/downsampler/chunk/BatchExporter.scala
@@ -160,12 +160,12 @@ case class BatchExporter(downsamplerSettings: DownsamplerSettings, userStartTime
     val dataSeq = new mutable.ArrayBuffer[Any](exportTableConfig.tableSchema.fields.length)
     // append all dynamic column values
     exportTableConfig.labelColumnMapping.foreach { pair =>
-      // scalastyle: off null
+      // scalastyle:off null
       val result = exportData.labels.get(pair._1) match {
         case Some(labelValue) => labelValue
         case None => null
       }
-      // scalastyle: on null
+      // scalastyle:on null
       dataSeq.append(result)
     }
     // append all fixed column values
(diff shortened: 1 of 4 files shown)


alextheimer (approved):

spark-jobs/src/main/scala/filodb/downsampler/chunk/BatchExporter.scala
```diff
@@ -160,8 +160,11 @@ case class BatchExporter(downsamplerSettings: DownsamplerSettings, userStartTime
     val dataSeq = new mutable.ArrayBuffer[Any](exportTableConfig.tableSchema.fields.length)
     // append all dynamic column values
     exportTableConfig.labelColumnMapping.foreach { pair =>
-      val labelValue = exportData.labels.get(pair._1)
-      dataSeq.append(labelValue.getOrElse(downsamplerSettings.exportDefaultDynamicColumnValue))
+      val result = exportData.labels.get(pair._1) match {
+        case Some(labelValue) => labelValue
+        case None => None
+      }
```
alextheimer, line 166:
```suggestion
      // scalastyle: off null
      val result = exportData.labels.get(pair._1) match {
        case Some(labelValue) => labelValue
        case None => null
      }
      // scalastyle: on null
```

spark-jobs/src/main/scala/filodb/downsampler/chunk/BatchExporter.scala
```diff
@@ -160,8 +160,13 @@ case class BatchExporter(downsamplerSettings: DownsamplerSettings, userStartTime
...
     // append all dynamic column values
     exportTableConfig.labelColumnMapping.foreach { pair =>
-      val labelValue = exportData.labels.get(pair._1)
-      dataSeq.append(labelValue.getOrElse(downsamplerSettings.exportDefaultDynamicColumnValue))
+      // scalastyle: off null
+      val result = exportData.labels.get(pair._1) match {
+        case Some(labelValue) => labelValue
+        case None => null
+      }
+      // scalastyle: on null
```
alextheimer, line 168:
```suggestion
      // scalastyle:off null
      val result = exportData.labels.get(pair._1) match {
        case Some(labelValue) => labelValue
        case None => null
      }
      // scalastyle:on null
```

alextheimer (approved):

PR #1749, opened 2024-03-26
fix(query) Range Vector Transformers were not applied for ScalarBinaryOperationExec

amolnayak311:
**Pull Request checklist**

**Current behavior :**

Queries such as ``time() - 60 * 60`` generate plans like

```
T~ScalarOperationMapper(operator=SUB, scalarOnLhs=true)
-FA1~TimeFuncArgs(RangeParams(1711402246,0,1711402246))
-E~ScalarBinaryOperationExec(params = RangeParams(1711402246,0,1711402246), operator = MUL, lhs = Left(60.0), rhs = Left(60.0)) on InProcessPlanDispatcher(QueryConfig(...)))
```

However, the ``RangeVectorTransformer`` never gets applied during evaluating.  This result this is simply executing ``60*60`` which gives ``3600``

**New behavior :**

Now we execute the transformers which should give the tight result for ``time() - 60*60``

### DIFF
--- a/query/src/main/scala/filodb/query/exec/ScalarBinaryOperationExec.scala
+++ b/query/src/main/scala/filodb/query/exec/ScalarBinaryOperationExec.scala
@@ -1,7 +1,9 @@
 package filodb.query.exec
 
+import kamon.Kamon
 import monix.eval.Task
 import monix.execution.Scheduler
+import monix.reactive.Observable
 
 import filodb.core.DatasetRef
 import filodb.core.metadata.Column.ColumnType
@@ -9,6 +11,9 @@ import filodb.core.query._
 import filodb.core.store.ChunkSource
 import filodb.query.{BinaryOperator, QueryResponse, QueryResult}
 import filodb.query.exec.binaryOp.BinaryOperatorFunction
+
+
+
 
 /**
   * Exec Plans for scalar binary operations which can execute locally without being dispatched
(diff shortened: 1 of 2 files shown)


query/src/main/scala/filodb/query/exec/ScalarBinaryOperationExec.scala
```diff
@@ -58,12 +63,21 @@ case class ScalarBinaryOperationExec(queryContext: QueryContext,
                        querySession: QuerySession)
                       (implicit sched: Scheduler): Task[QueryResponse] = {
     val rangeVectors : Seq[RangeVector] = Seq(ScalarFixedDouble(params, evaluate))
-    Task.eval { // not async
-      QueryResult(
-        queryContext.queryId, resultSchema, rangeVectors, QueryStats(), QueryWarnings(),
-        querySession.resultCouldBePartial,
-        querySession.partialResultsReason
-      )
+    Kamon.runWithSpan(Kamon.currentSpan(), false) {
```
alextheimer, line 66: Why override `execute` here at all? All scalars do this, I guess, so `ScalarRangeVectors` can be returned?

alextheimer (approved):
PR LGTM -- thanks for this fix 🙏

query/src/main/scala/filodb/query/exec/ScalarBinaryOperationExec.scala
amolnayak311: Yes

PR #1753, opened 2024-03-28
fix(Cardinality): Adding filodb partition check for cardinality queries 

sandeep6189:
**Pull Request checklist**

**Current behavior :**  We don't have any validation check for the cardinality queries. 

**New behavior :** 
1. Adding a filodb partition check for TsCardinalities queries.
2. Passing filodb-partition in QueryContext for Metering queries

### DIFF
--- a/core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesStore.scala
+++ b/core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesStore.scala
@@ -39,6 +39,7 @@ extends TimeSeriesStore with StrictLogging {
 
   private val datasets = new HashMap[DatasetRef, NonBlockingHashMapLong[DownsampledTimeSeriesShard]]
   private val quotaSources = new HashMap[DatasetRef, ConfigQuotaSource]
+  private val FILODB_PARTITION = filodbConfig.getString("partition")
 
   val stats = new ChunkSourceStats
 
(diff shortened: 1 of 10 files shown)


core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesStore.scala
```diff
@@ -73,8 +74,12 @@ extends TimeSeriesStore with StrictLogging {
     }
   }
 
-  def scanTsCardinalities(ref: DatasetRef, shards: Seq[Int],
+  def scanTsCardinalities(queryContext: QueryContext, ref: DatasetRef, shards: Seq[Int],
                           shardKeyPrefix: Seq[String], depth: Int): Seq[CardinalityRecord] = {
+    // adding an additional check to verify if the partition is same as mentioned in query context
+    require(isCorrectPartitionForCardinalityQuery(queryContext, FILODB_PARTITION),
```
yu-shipit, line 80: It seems `queryContext` is only used for this newly added check.
Should this check happen before this function?

core/src/main/scala/filodb.core/store/ChunkSource.scala
```diff
@@ -198,7 +198,26 @@ trait ChunkSource extends RawChunkSource with StrictLogging {
...
+
+  /**
+   * Additional check to ensure the query is for the given filodb partition
+   *
+   * @param queryContext QueryContext containing traceInfo of the originating partition
+   * @return
+   */
+  def isCorrectPartitionForCardinalityQuery(queryContext: QueryContext, filodbPartition: String): Boolean = {
+    if (
+      queryContext.traceInfo.contains(FILODB_PARTITION_KEY) &&
```
yu-shipit, line 211: So the only invalid case is that we find a `trace info` does not match the `filodbPartition` from the `partition key`?

core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesStore.scala
sandeep6189: Yes. As of now, query context is not being used in TsCardinalities.  Thought of multiple ways, but querycontext was easier to pass without impacting the existing code flow.

We need this check when it is being executed on the filodb ( and also the component should have access to the config ). Before this, we are still in logical / exec plan realm.

core/src/main/scala/filodb.core/store/ChunkSource.scala
sandeep6189: Yes, that is the only invalid case. This is done to make sure the existing api flows in which TsCardinalities is queried is not impacted.

yu-shipit (approved):

PR #1756, opened 2024-04-10
fix(cassandra): Marking cassandra reads as idempotent

sandeep6189:
**Pull Request checklist**

**Current behavior :** Read queries is not retried by default since it is not marked as idempotent.

**New behavior :** Marking read queries as idempotent for better retry handling.


alextheimer (approved):
Docs for reference: https://docs.datastax.com/en/developer/java-driver/3.6/manual/retries/#retries-and-idempotence

Feels weird calling a read "idempotent," but it looks like you're right that _any_ query error will prevent a retry.

PR #1757, opened 2024-04-10
fix(query): Reverting the min max suffix replacement

sandeep6189:
**Pull Request checklist**

**Current behavior :** Assertion error happening in cases where
1.  metric name has a suffix `_min`  or `_max` AND 
2. query have lots of label filters which causes some `TimeSeriesShard` returning `no` data AND
3. have another metric without the suffix.

**New behavior :** We are reverting the query change added with the support for min/max histograms to fix the bug in the short term.


sherali42 (approved):
thanks @sandeep6189 for taking this

PR #1759, opened 2024-04-11
fix(query) misleading message when label filter is too long.

yu-shipit:
**Pull Request checklist**

**Current behavior :** 

Error code does not indicate that the regex is too long.
`valid_query: Cannot parse […]`

**New behavior :**
Error code will indicate that the regex is too long.
The error will contain meaningful message
`Regular expression filters should be <=`

**BREAKING CHANGES**

Breaking changes may include:

**Other information**:

### DIFF
--- a/prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
+++ b/prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
@@ -21,6 +21,8 @@ object WindowConstants {
                                   getDuration("stale-sample-after").toMillis
 }
 
+class RegexLengthLimitException(msg: String) extends IllegalArgumentException(msg)
+
 sealed trait JoinMatching {
   def labels: Seq[String]
 }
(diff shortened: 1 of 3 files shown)


prometheus/src/main/scala/filodb/prometheus/ast/Vectors.scala
```diff
@@ -21,6 +21,8 @@ object WindowConstants {
                                   getDuration("stale-sample-after").toMillis
 }
 
+class RegexLengthLimitException(msg: String) extends IllegalArgumentException(msg)
```
sandeep6189, line 24: Nice +1

prometheus/src/main/scala/filodb/prometheus/parse/AntlrParser.scala
```diff
@@ -64,13 +64,14 @@ object AntlrParser extends StrictLogging {
         return expr
       }
     } catch {
-      case e: ParseCancellationException => {}
-      case exc: Throwable => exc.printStackTrace
+      case e: ParseCancellationException => println(e.getMessage)
```
sandeep6189, line 67: is println required ?

sandeep6189 (dismissed):

prometheus/src/main/scala/filodb/prometheus/parse/AntlrParser.scala
yu-shipit: remove the println

sandeep6189 (approved):

PR #1763, opened 2024-04-26
feat(cluster): Enhancing liveness check to include cassandra and kafka connectivity at startup

sandeep6189:
**Pull Request checklist**

**Current behavior :** The liveness check just returned `UP` response without any consideration.

**New behavior :** Liveness endpoint checks for connectivity to kafka and cassandra i.e. the shards to be in recovery or active state.

### DIFF
--- a/http/src/main/scala/filodb/http/HealthRoute.scala
+++ b/http/src/main/scala/filodb/http/HealthRoute.scala
@@ -1,7 +1,5 @@
 package filodb.http
 
-import scala.concurrent.duration.DurationInt
-
 import akka.actor.ActorRef
 import akka.http.scaladsl.model.StatusCodes
 import akka.http.scaladsl.server.Directives._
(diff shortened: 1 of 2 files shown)


http/src/main/scala/filodb/http/HealthRoute.scala
```diff
@@ -27,6 +25,21 @@ class HealthRoute(coordinatorActor: ActorRef, v2ClusterEnabled: Boolean, setting
...
                                          ShardStatusAssigned.getClass)
 
+  def checkIfAllShardsLiveClusterV1(shardEvents: collection.Map[DatasetRef, Seq[ShardEvent]]): Boolean = {
+    shardEvents.values.flatten.forall {
+      case IngestionStarted(_, _, _) | RecoveryStarted(_, _, _, _) | RecoveryInProgress(_, _, _, _) => true
+      case _ => false
+    }
+  }
+
+  def checkIfAllShardsLiveClusterV2(shardHealthSeq: Seq[DatasetShardHealth]): Boolean = {
```
alextheimer, line 35: If there are `N` shards in the cluster, is it possible that the V2 coordinator node returns fewer than `N` statuses? I.e. if a single host is unresponsive?

alextheimer (approved):
Approved (with one question -- if that is not a problem, feel free to merge)

http/src/main/scala/filodb/http/HealthRoute.scala
sandeep6189: This ask call on the coordinator queries only the given instance/host which is queried. It will not coordinate with other hosts. Hence, we will only get the local shard health information. 

Saying that, I have added additional check to make sure that all the local shards responses is collected before we make a decision for liveness.

http/src/main/scala/filodb/http/HealthRoute.scala
```diff
@@ -42,17 +55,43 @@ class HealthRoute(coordinatorActor: ActorRef, v2ClusterEnabled: Boolean, setting
...
+    // connections are initialized correctly on startup. This will help us in detecting any connection issues to
+    // kafka and cassandra on startup.
+    // NOTE: We don't wait for bootstrap to finish or normal ingestion to start or shards to be active. As soon as
+    // ingestion or recovery is started, we consider the filodb pods to be live
     path ("__liveness") {
       get {
-        complete(httpLiveness("UP"))
+        if (v2ClusterEnabled) {
+          onSuccess(asyncAsk(coordinatorActor, LocalShardsHealthRequest, settings.queryAskTimeout)) {
+            case m: Seq[DatasetShardHealth]@unchecked =>
```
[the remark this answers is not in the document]
sandeep6189: Yes, I agree that it is not ideal. Continuing with the given format for now and will add a tracking work item to see if we can remove this.

alextheimer (approved):
The additional tests are much appreciated, as well 🙏

PR #1771, opened 2024-05-14
fix(QueryStats): return namespace regex value as given instead of using "multiple" string

sandeep6189:
**Pull Request checklist**

**Current behavior :** We send a string "multiple" when passing a regex for a shard key label/column filter. This doesn't give us visibility of of the actual value passed in the query stats.

**New behavior :** Added a match case which allows to return the column-filter regex values.


yu-shipit (approved):
lgtm

amolnayak311 (approved):

PR #1788, opened 2024-05-30
feat(histograms): Query support for max-min histograms

sandeep6189:
**Pull Request checklist**

**Current behavior :** functions like `histogram_quantile` didn't consider using  `max` and `min` columns while calculating the quantile value. As a result, the approximation was extrapolated using bucket tops, which is not a representative value.
 
**New behavior :** Using the `max` and `min`  column values while calculating quantiles for more accurate results.

**BREAKING CHANGES**

Breaking changes may include:
- Extended the max-histograms query planning support to include the min value as well. This should not break the existing max-histograms functions like `histogram_max`


core/src/main/scala/filodb.core/query/TransientRow.scala
```diff
@@ -93,13 +93,13 @@ class TransientHistRow(var timestamp: Long = 0L,
 }
 
 // 0: Timestamp, 1: Histogram, 2: Max/Double
-final class TransientHistMaxRow(var max: Double = 0.0) extends TransientHistRow() {
+final class TransientHistMaxMinRow(var max: Double = 0.0, var min: Double = 0.0) extends TransientHistRow() {
   override def setDouble(columnNo: Int, valu: Double): Unit =
-    if (columnNo == 2) max = valu else throw new IllegalArgumentException()
+    if (columnNo == 2) max = valu else if (columnNo == 3) min = valu else throw new IllegalArgumentException()
```
alextheimer, line 98: Ack'd the column numbers weren't originally there, but could you add `columnNo` to these exception messages?

memory/src/main/scala/filodb.memory/format/vectors/Histogram.scala
```diff
@@ -316,17 +320,18 @@ final case class MaxHistogram(innerHist: MutableHistogram, max: Double) extends
...
-      if (b == 0 && bucketTop(0) <= 0) return bucketTop(0)
+      if (bucketNum == 0 && bucketTop(0) <= 0) return bucketTop(0)
       else {
         // interpolate quantile within le bucket
-        var (bucketStart, bucketEnd, count) = (0d, Math.min(bucketTop(b), max), bucketValue(b))
-        if (b > 0) {
-          bucketStart = bucketTop(b-1)
-          count -= bucketValue(b-1)
-          rank -= bucketValue(b-1)
+        var (bucketStart, bucketEnd, count) = (Math.max(0d, min),
```
alextheimer, line 329: Not a chance I completely understand what's going on here, but off the top of my head: what if `bucketNum = 0`, `le = 0.001`, and `min = 99999999`? Wouldn't the interpolated value be huge in that case?

query/src/main/scala/filodb/query/exec/SelectRawPartitionsExec.scala
```diff
@@ -13,15 +13,31 @@ import filodb.query.Query
...
+
+  // Returns Some(colID) the ID of a "min" column if one of given colIDs is a histogram.
+  def histMinColumn(schema: Schema, colIDs: Seq[Types.ColumnId]): Option[Int] = {
+    colIDs.find { id => schema.data.columns(id).columnType == HistogramColumn }
+      .flatMap { histColID =>
+        schema.data.columns.find(c => c.name == "min" && c.columnType == DoubleColumn).map(_.id)
+      }
+  }
+
+  def histMaxMinColumns(schema: Schema, colIDs: Seq[Types.ColumnId]): Option[Int] = {
```
alextheimer, line 36: What is this used for?

query/src/main/scala/filodb/query/exec/rangefn/InstantFunction.scala
```diff
@@ -362,18 +364,33 @@ final case class HistogramQuantileImpl() extends HistToDoubleIFunction {
...
     * @param scalarParams - a single value between 0 and 1, the quantile to calculate.
     */
-  final def apply(hist: Histogram, max: Double, scalarParams: Seq[Double]): Double = {
+  final def apply(hist: Histogram, max: Double, min: Double = Double.NaN, scalarParams: Seq[Double]): Double = {
     require(scalarParams.length == 1, "Quantile (between 0 and 1) required for histogram quantile")
     val maxHist = hist match {
-      case h: MutableHistogram => MaxHistogram(h, max)
-      case other: Histogram    => MaxHistogram(MutableHistogram(other), max)
+      case h: MutableHistogram => MaxMinHistogram(h, max)
+      case other: Histogram    => MaxMinHistogram(MutableHistogram(other), max)
```
alextheimer, line 393: Should `min` be passed as an argument here, too, if it's not `Double.NaN`?

query/src/main/scala/filodb/query/exec/SelectRawPartitionsExec.scala
```diff
@@ -13,15 +13,31 @@ import filodb.query.Query
...
+    colIDs.find { id => schema.data.columns(id).columnType == HistogramColumn }
+      .flatMap { histColID =>
+        schema.data.columns.find(c => (c.name == "min" || c.name == "max") && c.columnType == DoubleColumn).map(_.id)
```
yu-shipit, line 39: (c.name == "min" || c.name == "max") should this be &&?

core/src/main/resources/filodb-defaults.conf
```diff
@@ -304,7 +304,7 @@ filodb {
   }
 
   # # of shards each application/metric is spread out to = 2^spread
-  spread-default = 1
+  spread-default = 2
```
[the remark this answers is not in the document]
sandeep6189: nope, was using it for testing. removing

memory/src/main/scala/filodb.memory/format/vectors/Histogram.scala
sandeep6189: This would be a case of bad instrumentation from user then. (since they can't expect to configure bucket that low and emit min with such a huge value) The idea is that we need to improve the accuracy of quantile calculation by extrapolating between [min, max] range

sandeep6189: Some context : the min and max is also published by the user as part of histogram ( this are not query time inputs )

query/src/main/scala/filodb/query/exec/SelectRawPartitionsExec.scala
```diff
@@ -44,18 +60,18 @@ object SelectRawPartitionsExec extends  {
...
   }
 
-  def newXFormersForHistMax(schema: Schema,
-                            colIDs: Seq[Types.ColumnId],
-                            transformers: Seq[RangeVectorTransformer]): Seq[RangeVectorTransformer] = {
-    histMaxColumn(schema, colIDs).map { maxColID =>
+  def newXFormersForHistMaxMin(schema: Schema,
+                               colIDs: Seq[Types.ColumnId],
+                               transformers: Seq[RangeVectorTransformer]): Seq[RangeVectorTransformer] = {
+    histMaxMinColumns(schema, colIDs).map { colId =>
```
sherali42, line 70: can this be replaced with `histMaxColumn`, dont think col_id is used any where?

query/src/main/scala/filodb/query/exec/rangefn/InstantFunction.scala
sherali42: +1, good catch @alextheimer. i think in this case, since `HistogramMaxQuantileImpl` is invoked only when historgram schema as only `max` column and not `min`. may be make `min` as default 0 instead of NaN and pass it along?

sherali42 (approved):
PR looks good to me @sandeep6189, thanks for taking care of all the campatibility scenarios.

query/src/main/scala/filodb/query/exec/SelectRawPartitionsExec.scala
sandeep6189: updated this function to make sure we return just one option[int] value. we use it as a signal to replace the transformers.

grpc/src/main/protobuf/query_service.proto
```diff
@@ -635,7 +635,7 @@ enum InternalRangeFunction {
   LAST_OVER_TIME        = 21;
   AVG_WITH_SUM_AND_COUNT_OVER_TIME = 22;
   SUM_AND_MAX_OVER_TIME = 23;
-  LAST_SAMPLE_HIST_MAX  = 24;
+  LAST_SAMPLE_HIST_MAX_MIN  = 24;
```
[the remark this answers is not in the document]
sandeep6189: This is function id used exclusively for Hubble for gcdb. we are planning to test it more with Hubble team.

core/src/test/scala/filodb.core/TestData.scala
```diff
@@ -393,20 +393,21 @@ object MachineMetricsData {
...
+  def histMaxMin(histStream: Stream[Seq[Any]]): Stream[Seq[Any]] =
     histStream.map { row =>
       val hist = row(3).asInstanceOf[bv.LongHistogram]
       // Set max to a fixed ratio of the "last bucket" top value, ie the last bucket with an actual increase
       val highestBucketVal = hist.bucketValue(hist.numBuckets - 1)
       val lastBucketNum = ((hist.numBuckets - 2) to 0 by -1).filter { b => hist.bucketValue(b) == highestBucketVal }
                             .lastOption.getOrElse(hist.numBuckets - 1)
       val max = hist.bucketTop(lastBucketNum) * 0.8
-      ((row take 3) :+ max) ++ (row drop 3)
+      val min = hist.bucketTop(lastBucketNum) * 0.1
```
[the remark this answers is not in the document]
sandeep6189: For unit test, I wanted to simulate the data in a way to keep the max as higher value and min as the lower value

query/src/main/scala/filodb/query/exec/SelectRawPartitionsExec.scala
sandeep6189: @sherali42 @alextheimer Good question. It is a signal used in SelectRawPartittionsExec to replace the transformers for max-min histograms.

query/src/main/scala/filodb/query/exec/rangefn/InstantFunction.scala
sandeep6189: Need default for min as  Double.NaN for the HistMaxMinAggregator ( if not, then it forcibly brings down the min to 0 even if the min is set to a higher number like 10 or 20 )

let me pass 0 for HistogramMaxQuantile function

gateway/src/main/scala/filodb/timeseries/TestTimeseriesProducer.scala
```diff
@@ -221,13 +222,15 @@ object TestTimeseriesProducer extends StrictLogging {
 
     val instanceBase = System.currentTimeMillis
     var prevTimestamp = startTime
+    var timestamp = startTime
     Stream.from(0).map { n =>
-      val instance = n % numTimeSeries + instanceBase
-      val dc = instance & oneBitMask
-      val partition = (instance >> 1) & twoBitMask
+      val instance = 1// n % numTimeSeries + instanceBase
```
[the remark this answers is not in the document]
sandeep6189: Unintentional changes

gateway/src/main/scala/filodb/timeseries/TestTimeseriesProducer.scala
```diff
@@ -204,13 +205,13 @@ object TestTimeseriesProducer extends StrictLogging {
    * the cardinality of time series for testing purposes.
    */
   def genHistogramData(startTime: Long, numTimeSeries: Int = 16, histSchema: Schema): Stream[InputRecord] = {
-    val numBuckets = 10
+    val numBuckets = 3
```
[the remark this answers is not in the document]
sandeep6189: Not intentional, removing

grpc/src/main/protobuf/query_service.proto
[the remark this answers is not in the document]
sandeep6189: talked to  @kvpetrov about this. So we aren't using this protos in production as of now, so we are safe with this change.

query/src/main/scala/filodb/query/exec/SelectRawPartitionsExec.scala
sandeep6189: That is a good point. I added it to make sure the underlying schema must have both max and min ( since the transformers are changed to have both )

memory/src/main/scala/filodb.memory/format/vectors/Histogram.scala
alextheimer: > since they can't expect to configure bucket that low and emit min with such a huge value

What about a bucketing schema like this?:
```
le = 1, 2, 4, 8 , 16
```
And the histogram accounts for a single value: `2`. Doesn't that cause the same scenario? I.e.:
```
bucketNum = 0 -> le = 1
min = 2 -> min > le
```

alextheimer: I see -- it looks like you're right that this state:
> bucketNum = 0, le = 0.001, and min = 99999999

would never happen.

query/src/main/scala/filodb/query/exec/SelectRawPartitionsExec.scala
alextheimer: Ah I see -- how do you feel about returning a `Boolean` to indicate whether-or-not it has min/max columns?

sandeep6189: Surely 👍 Updated it and added more unit tests as well

alextheimer (approved):

PR #1792, opened 2024-06-07
(misc-iceberg-export): Remove sortWithinPartitions to avoid spill(memory) and spill(disk)

nikitag55:
Current behavior :

In case of large volume of data to be exported, with sortWithinPartitions() is causing spill(disk) & spill(memory).

New behavior :

Removing this operation will not do sorting within partitions and will not cause memory & disk spill.


alextheimer (approved):
Thanks for the updates @nikitag55 -- LGTM

PR #1795, opened 2024-06-10
fix(query) Add aggregate presenter to remote query when target schema definition is available

amolnayak311:
**Pull Request checklist**

**Current behavior :** 

When an aggregate is pushed remotely, aggregate presenter is skipped in certain cases where the query service expects to see the result of the aggregate coming from multiple shards on the remote. However, when target schema is defined the aggregate is guaranteed to coming from one shard not requiring any additional presenter to be applied on the Query service. In other words, the entire aggregation needs to be sent to the remote partition (and the shard in that partition). The current logic does not consider the presence of target schema and thus erroneously does not apply the aggregate presenter. 

**New behavior :**

Logic changed to apply the presenter if target schema is defined.


alextheimer (approved):
LGTM -- thanks for taking care of this @amolnayak311. This should be the only spot we need the change; the SCP is another candidate, but I see `*noPushdown` variants of methods (e.g. `materializeAggregateNoPushdown`) are explicitly called by wrapper methods there when target-schemas are applied.

PR #1804, opened 2024-06-22
feat(core): target-schema / data-export regex filters

alextheimer:
**Pull Request checklist**

Currently:
- Data-export configs do not support regex filters.
- Target-schema configs have no out-of-the-box support for regex filters.

This PR adds a new `ColumnFilterMap` type and makes use of it as part of:
- a new `TargetSchemaProvider` class; provides target-schema configs.
- `DownsamplerMain`; provides data-export configs.

From the Javadoc:
```
  /**
   * Maps sets of {@link ColumnFilter} "keys" to element "values".
   *
   * Convenient for use-cases that require e.g. configuration of rules against
   *   label-value filters (filter->rule pairs). The {@link #get} method
   *   accepts a set of label-value pairs, identifies any matching filter sets,
   *   then returns the corresponding element.
   *
   * For example, consider the (filters,element) pairs:
   *     - ({l1="v1",l2=~"v2.*"}, elt1)
   *     - ({l1="v1",l2="v3"},    elt2)
   *     - ({l2=~"v4.*"},         elt3)
   *
   * The following {label:value} sets would map to these elements:
   *     - {l1=v1,l2=v2,l3=v3} -> elt1
   *     - {l1=v1}             -> (None)
   *     - {l2=v4A}            -> elt3
   *
   * A label-value map "matches" a set of filters when:
   *     - all filter labels are present in the map
   *     - every filter is matched by label in the map.
   ...
```

Now, both target-schemas and data-export configs have the infrastructure to support regex filters. Data-export configs are now written as:
```
data-export {
  key-labels = ["l1", "l2"]
  groups = [
    {
      key = ["l1=\"v1\"", "l2=~\"v2.*\""]  <-- only this field is updated
      table = "my_cool_table"
      ...
    }
  ]
}
```


core/src/main/scala/filodb.core/Utils.scala
```diff
@@ -38,4 +43,225 @@ object Utils extends StrictLogging {
...
+   *                           can be a non-Equals filter.
+   */
+  class ColumnFilterMap[T](filterElementPairs: Iterable[(Iterable[ColumnFilter], T)]){
+
+    // The main idea:
+    // - Require that filter sets added map "keys" include at most one non-Equals filter.
+    // - Use tries to match get() argument label-values pairs with Equals filters efficiently
+    //   (see get() for more details).
+    // - The trie get() matches Equals filters; store each element with its one final
+    //   (possibly non-Equals) filter as values in the trie. Iterate these filters as appropriate.
```
alextheimer, line 155: This trie-based implementation is more complicated than originally expected, and it isn't particularly efficient.
An alternative is implemented [here](https://github.com/filodb/FiloDB/blob/517af29d61555f054eff3a170cdae538a1ade032/core/src/main/scala/filodb.core/TargetSchemaProvider.scala#L99). Its `get` runtime complexity grows logarithmically with the total count of mappings, and it might actually be easier to reason about.

There is a drawback: it doesn't currently support regex filters, but it can be updated to support prefix-only `EqualsRegex` filters (e.g. `foo=~"bar-.*"`).

core/src/main/scala/filodb.core/Utils.scala
```diff
@@ -38,4 +43,161 @@ object Utils extends StrictLogging {
...
+   *
+   * NOTE: This implementation is intended to favor simplicity over efficiency.
+   *   Other implementations should be considered if the {@link #get} runtime
+   *   (described in the method's javadoc) is insufficient.
+   *
+   * @tparam T the type of elements to store/return.
+   * @param filterElementPairs (filter-set,element) pairs; at most one of filter-set
+   *                           can be a non-Equals filter.
+   */
+  class ColumnFilterMap[T](filterElementPairs: Iterable[(Iterable[ColumnFilter], T)]){
```
alextheimer, line 78: This implementation is more complicated than originally expected, and it isn't particularly efficient.
An alternative is implemented [here](https://github.com/filodb/FiloDB/blob/517af29d61555f054eff3a170cdae538a1ade032/core/src/main/scala/filodb.core/TargetSchemaProvider.scala#L99). Its `get` runtime complexity grows logarithmically with the total count of mappings, and it might actually be easier to reason about.

There is a drawback: it doesn't currently support regex filters, but it can be updated to support prefix-only `EqualsRegex` filters (e.g. `foo=~"bar-.*"`).

core/src/main/scala/filodb/core/utils/ColumnFilterMap.scala
```diff
@@ -0,0 +1,164 @@
...
+ *
+ * NOTE: This implementation is intended to favor simplicity over efficiency.
+ *   Other implementations should be considered if the {@link #get} runtime
+ *   (described in the method's javadoc) is insufficient.
+ *
+ * @tparam T the type of elements to store/return.
+ * @param filterElementPairs (filter-set,element) pairs; at most one of filter-set
+ *                           can be a non-Equals filter.
+ */
+class ColumnFilterMap[T](filterElementPairs: Iterable[(Iterable[ColumnFilter], T)]){
```
alextheimer, line 40: This implementation is more complicated than originally expected, and it isn't particularly efficient (especially when different labels are used between sets and/or regex filters are common).

An alternative is implemented [here](https://github.com/filodb/FiloDB/blob/517af29d61555f054eff3a170cdae538a1ade032/core/src/main/scala/filodb.core/TargetSchemaProvider.scala#L99). Its `get` runtime complexity grows logarithmically with the total count of mappings, and it might actually be easier to reason about.

There is a drawback: it doesn't currently support regex filters, but it can be updated to support prefix-only `EqualsRegex` filters (e.g. `foo=~"bar-.*"`).

core/src/main/scala/filodb/core/utils/ColumnFilterMap.scala
```diff
@@ -0,0 +1,164 @@
...
+ *   then returns the corresponding element.
+ *
+ * For example, consider the (filters,element) pairs:
+ *     - ({l1="v1",l2=~"v2.*"}, elt1)
+ *     - ({l1="v1",l2="v3"},    elt2)
+ *     - ({l2=~"v4.*"},         elt3)
+ *
+ * The following {label:value} sets would map to these elements:
+ *     - {l1=v1,l2=v2,l3=v3} -> elt1
+ *     - {l1=v1}             -> (None)
```
yu-shipit, line 23: I am not clear about this case.
should this reuslt be `elt1, elt2`  instead of
```- ({l1="v1",l2="v3"},    elt2)```

File: core/src/main/scala/filodb/core/utils/ColumnFilterMap.scala [hunk 5]
```diff
@@ -0,0 +1,164 @@
...
+    sortedFiltersToElements.foreach { case (filters, element) =>
+      filters.last.filter match {
+        case Filter.Equals(value) => equalsFilterMap(value.toString) = element
+        case _ => otherFilters.append((filters.last, element))
+      }
+    }
+    val filterToElementMap = FilterToElementMap(equalsFilterMap.toMap, otherFilters)
+    labelSeqToMap
+      .getOrElseUpdate(labelNames, new mutable.HashMap[Seq[String], FilterToElementMap]())
+      .put(prefixValues, filterToElementMap)
```
yu-shipit, line 127: If I understand correctly,
```
filters = (l1=v1, l2=v2, l3=~"v3.*") 
will be converted to
Map([l1,l2,l3] => Map([v1, v2] => FilterToElementMap(filters)))
```
Do we have to store every filter into FilterToElementMap or just the last filter?

core/src/main/scala/filodb/core/utils/ColumnFilterMap.scala
```diff
@@ -0,0 +1,164 @@
...
+   * Maps individual filters to elements.
+   * Equals filters are stored as a Map for efficiency.
+   */
+  protected case class FilterToElementMap(equalsFilters: Map[String, T],
+                                          otherFilters: Iterable[(ColumnFilter, T)]) {
+    /**
+     * Returns an occupied Option iff at least one filter is matched.
+     * The Equals map is checked first before other filters are iterated.
+     */
+    def get(string: String): Option[T] = {
```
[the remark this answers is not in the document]
alextheimer: I'll keep this as `string` because `pattern` sort of implies it's used to match something. The string is processed through each of the filters (actual `ColumnFilter`s or the map of "`Equals`" filters) until a filter matches the string.

I'll update the docs so it's more clear that the string is "getting matched" and not doing the matching itself.

core/src/main/scala/filodb/core/utils/ColumnFilterMap.scala
alextheimer: Super close -- the last line should be:
```
// only the last filter is added to the map
Map([l1,l2,l3] => Map([v1, v2] => FilterToElementMap(filters.last)))
```

I'll steal this example, expand on it a bit, and add it to the docs.

yu-shipit: That makes sense. 
I have a question regarding this case. 
```
filters = ((l1=v1),  (l1=v2))
```
In this case, the filters will have the same key in the map?

alextheimer: Yep, that's right 👍 
Snippet from some new docs:
```
  // As an example, consider a ColumnFilterMap composed of these entries:
...
  //     {l1="v1", l2="v2a"} -> F
  //     {l1="v1", l2="v2b"} -> G
  //     {l1="v1a", l2="v2a"} -> H
...
  //   The nested maps would be structured as:
...
  //     (l1, l2) -> {
  //         (v1) -> [l2="v2a":F, l2="v2b":G]
  //         (v1a) -> [l2="v2a":H]
  //     }
```

yu-shipit: Got it thank you for the explanation.

core/src/main/scala/filodb/core/utils/ColumnFilterMap.scala
```diff
@@ -0,0 +1,197 @@
...
+  //   the inner hash-map of example (3) above would be keyed on (labelD-value, labelE-value) and
+  //   store (labelC-filter, element) pairs.
+  // - When a set of labels-value pairs is processed through the ColumnFilterMap, the values are
+  //   used to access each inner map until (a) the inner map contains a corresponding value, and
+  //   (b) the final filter matches the final label-value. When a match is found, the filter's paired
+  //   element is returned.
+  //
+  // As an example, consider a ColumnFilterMap composed of these entries:
+  //
+  //     {l1="v1", l2="v2", l3="v3a"} -> A
```
sandeep6189, line 63: noob question: what are the `element` that we are referencing  here - A, B, C etc ? ( I am confused since the data structure already stores label name and label values )

sandeep6189: Seeing from below, `ExportTableConfig` and `Seq(TargetSchemaChange)` looks like the element stored essentially ?

sandeep6189 (approved):

sandeep6189:
This PR is a great example on the best practices for code documentation and unit tests. Kudos @alextheimer  💯

core/src/main/scala/filodb/core/utils/ColumnFilterMap.scala
```diff
@@ -0,0 +1,197 @@
+package filodb.core.utils
+
+import scala.collection.mutable
+import scala.collection.mutable.ArrayBuffer
+
+import filodb.core.query.{ColumnFilter, Filter}
```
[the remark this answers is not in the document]
alextheimer: Mostly: that's where the target-schema stuff lives. But (a) `ColumnFilterMap`s aren't exclusively for use with target-schemas, and (b) I think you're right a move to `filodb.core.query` would be more conventional -- updated ✅

core/src/main/scala/filodb/core/utils/ColumnFilterMap.scala
```diff
@@ -0,0 +1,197 @@
...
+ *   (described in the method's javadoc) is insufficient.
+ *
+ * @tparam T the type of elements to store/return.
+ * @param filterElementPairs (filter-set,element) pairs; at most one of filter-set
+ *                           can be a non-Equals filter.
+ */
+class ColumnFilterMap[T](filterElementPairs: Iterable[(Iterable[ColumnFilter], T)]){
+
+  // The main idea:
+  // - Require that filter sets include at most one non-Equals filter.
```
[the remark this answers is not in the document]
alextheimer: It was originally a simplification to support only one regex label. But I'm glad you left this comment -- I've revisited the idea, and I think it's actually simplified the logic further. Let me know what you think (I know this PR is a nontrivial time commitment-- I can save the new changes for a separate PR, as well).

core/src/main/scala/filodb/core/utils/ColumnFilterMap.scala
alextheimer: That's right -- those two will be the most-commonly stored values.

core/src/main/scala/filodb/core/utils/ColumnFilterMap.scala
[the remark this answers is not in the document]
alextheimer: Most important bit is here: https://github.com/filodb/FiloDB/blob/4c3c5a2afc5c538fb2ce488235e95fedbcd406be/core/src/main/scala/filodb.core/query/ColumnFilterMap.scala#L37

spark-jobs/src/test/scala/filodb/downsampler/DownsamplerMainSpec.scala
```diff
@@ -490,10 +492,178 @@ class DownsamplerMainSpec extends AnyFunSpec with Matchers with BeforeAndAfterAl
...
+          |          partition-by-columns = []
+          |          rules = []
+          |        }
+          |      ]
+          |    }
+          |""".stripMargin
+      ).withFallback(conf)
+      val settings = new DownsamplerSettings(testConf)
+      val rdd = sparkCtx.parallelize(Seq(
+        Row("my_ws1", "my_ns1"),
```
yu-shipit, line 649: Do we need consider case 

```
Row("my_ws1"),
and
Row("my_ws1", "my_ns1", "another label"),
```

yu-shipit (commented):
lgtm.

core/src/test/scala/filodb.core/query/ColumnFilterSpec.scala
```diff
@@ -0,0 +1,126 @@
...
+      cfMap.get(Map()) shouldEqual None
+    }
+
+    {
+      // single equals filter
+      val entries = Seq(
+        (Seq(equals("a", "a")), "foo")
+      )
+      val cfMap = new ColumnFilterMap[String](entries)
+      cfMap.get(Map("a" -> "a")).get shouldEqual "foo"
```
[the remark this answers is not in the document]
alextheimer: Good call -- I've added a few more tests.

spark-jobs/src/test/scala/filodb/downsampler/DownsamplerMainSpec.scala
alextheimer: Nope -- there will always be as many columns as the labels configured as the "key" above.

yu-shipit (approved):

sandeep6189 (approved):
looks good :) Thanks @alextheimer

PR #1807, opened 2024-06-27
feat(histograms): Supporting back compatibility of queries which spans across prom histograms and otel histograms for same metric

sandeep6189:
**Pull Request checklist**

**Current behavior :** back compatibility of queries spanning across prom-histograms and otel histograms with same metric name.

**New behavior :** supporting this behavior.

**BREAKING CHANGES**

- Changed storage schema of `otel-cumulative-histogram` and `otel-delta-histogram`. Moved histogram to column 3 from column 5.

### DIFF
--- a/core/src/main/resources/filodb-defaults.conf
+++ b/core/src/main/resources/filodb-defaults.conf
@@ -189,11 +189,11 @@ filodb {
       columns = ["timestamp:ts",
                  "sum:double:detectDrops=true",
                  "count:double:detectDrops=true",
+                 "h:hist:counter=true",
                  "min:double:detectDrops=true",
-                 "max:double:detectDrops=true",
-                 "h:hist:counter=true"]
+                 "max:double:detectDrops=true"]
       value-column = "h"
-      downsamplers = ["tTime(0)", "dLast(1)", "dLast(2)", "dMin(3)", "dMax(4)", "hLast(5)"]
+      downsamplers = ["tTime(0)", "dLast(1)", "dLast(2)", "hLast(3)", "dMin(4)", "dMax(5)"]
       # Downsample periods are determined by counter dips of the count column
       downsample-period-marker = "counter(2)"
       downsample-schema = "otel-cumulative-histogram"
@@ -203,11 +203,11 @@ filodb {
       columns = ["timestamp:ts",
         "sum:double:{detectDrops=false,delta=true}",
         "count:double:{detectDrops=false,delta=true}",
+        "h:hist:{counter=false,delta=true}",
         "min:double:{detectDrops=false,delta=true}",
-        "max:double:{detectDrops=false,delta=true}",
-        "h:hist:{counter=false,delta=true}"]
+        "max:double:{detectDrops=false,delta=true}"]
       value-column = "h"
-      downsamplers = ["tTime(0)", "dSum(1)", "dSum(2)", "dMin(3)", "dMax(4)", "hSum(5)"]
+      downsamplers = ["tTime(0)", "dSum(1)", "dSum(2)", "hSum(3)", "dMin(4)", "dMax(5)"]
       downsample-period-marker = "time(0)"
       downsample-schema = "otel-delta-histogram"
     }
--- a/core/src/main/resources/filodb-defaults.conf
+++ b/core/src/main/resources/filodb-defaults.conf
@@ -396,6 +396,8 @@ filodb {
       partitions-deny-list = ""
     }
 
+    # Config to disable workspaces where we don't want max min columns to be selected for back compatibility
+    workspaces-disabled-max-min = ["disabled_ws"]
 
     routing {
       enable-remote-raw-exports = false
--- a/core/src/main/resources/filodb-defaults.conf
+++ b/core/src/main/resources/filodb-defaults.conf
@@ -397,6 +397,7 @@ filodb {
     }
 
     # Config to disable workspaces where we don't want max min columns to be selected for back compatibility
+    # If unspecified, all workspaces are disabled by default.
     workspaces-disabled-max-min = ["disabled_ws"]
 
     routing {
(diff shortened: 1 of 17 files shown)


core/src/main/scala/filodb.core/store/ReadablePartition.scala
```diff
@@ -129,5 +129,14 @@ trait ReadablePartition extends FiloPartition {
                           countInfoIterator: CountingChunkInfoIterator): RangeVectorCursor =
     new PartitionTimeRangeReader(this, method.startTime, method.endTime, countInfoIterator, columnIDs)
 
+  final def isBackCompatibleHistograms(schemaNameToCheck : String, schemaHashToCheck : Int) : Boolean = {
+    val sortedSchemas = Seq(schema.name, schemaNameToCheck).sortBy(_.length)
+    val ret = if ( (sortedSchemas(0) == "prom-histogram") && (sortedSchemas(1) == "otel-cumulative-histogram") ) true
+    else if ( (sortedSchemas(0) == "delta-histogram") && (sortedSchemas(1) == "otel-delta-histogram") ) true
+    else if (schema.schemaHash == schemaHashToCheck) true
```
sandeep6189, line 136: move this to first step

alextheimer (approved):
LGTM -- many thanks @sandeep6189 for the additional tests

query/src/main/scala/filodb/query/exec/MultiSchemaPartitionsExec.scala
```diff
@@ -56,13 +57,22 @@ final case class MultiSchemaPartitionsExec(queryContext: QueryContext,
     (lookupRes, Some(columnName))
   }
 
+  def isMaxMinEnabledForWorkspace(ws: Option[String]) : Boolean = {
```
alextheimer, line 60: Why does this accept an `Option`?

query/src/main/scala/filodb/query/exec/MultiSchemaPartitionsExec.scala
```diff
@@ -56,13 +57,22 @@ final case class MultiSchemaPartitionsExec(queryContext: QueryContext,
...
+    ws.isDefined match {
+      case true => (GlobalConfig.workspacesDisabledForMaxMin.isDefined) &&
+        (!GlobalConfig.workspacesDisabledForMaxMin.get.contains(ws.get))
```
alextheimer, line 63: This will say "max-min is disabled for this workspace" if `GlobalConfig.workspacesDisabledForMaxMin` is not defined. Is that intentional?

query/src/main/scala/filodb/query/exec/MultiSchemaPartitionsExec.scala
```diff
@@ -56,13 +57,22 @@ final case class MultiSchemaPartitionsExec(queryContext: QueryContext,
...
+  }
+
   // scalastyle:off method.length
   private def finalizePlan(source: ChunkSource,
                            querySession: QuerySession): SelectRawPartitionsExec = {
     val partMethod = FilteredPartitionScan(ShardSplit(shard), filters)
     Kamon.currentSpan().mark("filtered-partition-scan")
     var lookupRes = source.lookupPartitions(dataset, partMethod, chunkMethod, querySession)
     val metricName = filters.find(_.column == metricColumn).map(_.filter.valuesStrings.head.toString)
+    val ws = filters.find(_.column == TsCardinalities.LABEL_WORKSPACE).map(_.filter.valuesStrings.head.toString)
```
alextheimer, line 75: Could you add one additional condition: that the `ws` filter is an instance of `Equals`?

query/src/main/scala/filodb/query/exec/MultiSchemaPartitionsExec.scala
sandeep6189:
```
  private def finalizePlan(source: ChunkSource,
                           querySession: QuerySession): SelectRawPartitionsExec = {
    val partMethod = FilteredPartitionScan(ShardSplit(shard), filters)
    Kamon.currentSpan().mark("filtered-partition-scan")
    var lookupRes = source.lookupPartitions(dataset, partMethod, chunkMethod, querySession)
    val metricName = filters.find(_.column == metricColumn).map(_.filter.valuesStrings.head.toString)
    val ws = filters.find(_.column == TsCardinalities.LABEL_WORKSPACE).map(_.filter.valuesStrings.head.toString)
    var newColName = colName
```

This is because, when we find the string from filters, the search by default gives us an Option[String]. Even metricName is Option[String].

query/src/main/scala/filodb/query/exec/MultiSchemaPartitionsExec.scala
sandeep6189: Yes, that is intentional. will give the reasoning offline

query/src/main/scala/filodb/query/exec/MultiSchemaPartitionsExec.scala
sandeep6189: Sure. Updated

query/src/main/scala/filodb/query/exec/MultiSchemaPartitionsExec.scala
alextheimer: Understood 👍

core/src/main/resources/filodb-defaults.conf
```diff
@@ -396,6 +396,8 @@ filodb {
       partitions-deny-list = ""
     }
 
+    # Config to disable workspaces where we don't want max min columns to be selected for back compatibility
+    workspaces-disabled-max-min = ["disabled_ws"]
```
alextheimer, line 400:
```suggestion
    # Config to disable workspaces where we don't want max min columns to be selected for back compatibility.
    # If unspecified, all workspaces are disabled by default.
    workspaces-disabled-max-min = ["disabled_ws"]
```

alextheimer (dismissed):

alextheimer (approved):

PR #1809, opened 2024-07-06
fix(downsample): apply correct export rules

alextheimer:
**Pull Request checklist**

Currently, the `ColumnFilterMap` used in `BatchExporter::getRuleIfShouldExport` will choose an export config arbitrarily from multiple key filter matches. For example, if the map contains:
```
{l1=foo.*} -> config1
{l1=fooA.*} -> config2
```
Then this series would technically match both keys:
```
{l1=fooABC}
```
-- and either `config1` or `config2` may be returned.

Since both contain their own `allow`/`block` rules, it's possible the series should be exported according to `config1` but no export rule is returned because it is blocked by `config2`.

Since anyways an `RDD` with _all_ rows is iterated once per table, it is both less confusing and more correct to consider exclusively the rules for one table per iteration. This PR adds a table config parameter to `getRuleIfShouldExport`, and it consolidates all row filtering logic.

### DIFF
--- a/spark-jobs/src/main/scala/filodb/downsampler/chunk/BatchExporter.scala
+++ b/spark-jobs/src/main/scala/filodb/downsampler/chunk/BatchExporter.scala
@@ -109,7 +109,9 @@ case class BatchExporter(downsamplerSettings: DownsamplerSettings, userStartTime
   /**
    * Returns the Spark Rows to be exported.
    */
-  def getExportRows(readablePartitions: Seq[ReadablePartition], exportTableConfig: ExportTableConfig): Iterator[Row] = {
+  def getExportRows(readablePartitions: Seq[ReadablePartition],
+                    keyFilters: Seq[ColumnFilter],
+                    exportTableConfig: ExportTableConfig): Iterator[Row] = {
     readablePartitions
       .iterator
       .map { part =>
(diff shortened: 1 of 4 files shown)


sandeep6189 (approved):

spark-jobs/src/main/scala/filodb/downsampler/chunk/BatchExporter.scala
```diff
@@ -230,18 +232,21 @@ case class BatchExporter(downsamplerSettings: DownsamplerSettings, userStartTime
...
-        }
+  def getRuleIfShouldExport(partKeyMap: collection.Map[String, String],
+                            keyFilters: Seq[ColumnFilter],
+                            exportTableConfig: ExportTableConfig): Option[ExportRule] = {
+    if (!matchAllFilters(keyFilters, partKeyMap)) {
+      return None
+    }
+    exportTableConfig.exportRules.takeWhile { rule =>
+      // step through rules while we still haven't matched a "block" filter
+      !rule.blockFilterGroups.exists { filterGroup =>
```
sandeep6189, line 243: nitpick: I am guessing we already have a unit test where we have two/three matching rules and one of them is not selected for block-filters ?

alextheimer: Added ✅

srivik (approved):

PR #1810, opened 2024-07-06
perf(query): add FastColumnFilterMap

alextheimer:
**Pull Request checklist**

Adds `FastColumnFilterMap`. Snippet from the javadoc:
```
 * Bare-bones but (in some cases) ultra-fast ColumnFilterMap.
 * At most two label-values are read during a get() call:
 *   - One to be matched by equality.
 *   - One to be matched by any arbitrary filter.
 * Useful for performance-critical use-cases where this minimal functionality is sufficient:
 *   at most two labels are filtered, and only one supports non-equals filters.
```

Local tests demonstrate that-- although it is significantly more constrained--  `FastColumnFilterMap`  is about 4x faster than `DefaultColumnFilterMap` in these limited-filtering scenarios.

### DIFF
--- a/core/src/test/scala/filodb.core/query/ColumnFilterMapSpec.scala
+++ b/core/src/test/scala/filodb.core/query/ColumnFilterMapSpec.scala
@@ -13,10 +13,10 @@ class ColumnFilterMapSpec extends AnyFunSpec with Matchers {
     ColumnFilter(column, Filter.EqualsRegex(value))
   }
 
-  it("should store/retrieve ColumnFilterMap elements correctly") {
+  it("DefaultColumnFilterMap should store/retrieve elements correctly") {
     {
       // empty map
-      val cfMap = new ColumnFilterMap[String](Nil)
+      val cfMap = new DefaultColumnFilterMap[String](Nil)
       cfMap.get(Map("foo" -> "bar")) shouldEqual None
       cfMap.get(Map("foo" -> "bar", "baz" -> "bat")) shouldEqual None
       cfMap.get(Map()) shouldEqual None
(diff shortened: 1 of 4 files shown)


core/src/test/scala/filodb.core/query/ColumnFilterMapSpec.scala
```diff
@@ -127,4 +127,97 @@ class ColumnFilterMapSpec extends AnyFunSpec with Matchers {
       cfMap.get(Map("b" -> "bbb123", "c" -> "ccc123")).get shouldEqual "4r"
     }
   }
+
```
yu-shipit, line 130: Can you add a test case ?
`regex("c", "c.*cc.*")), "4r"),`

yu-shipit (commented):
Just curious if how faster this would be than the non-optimized one?

core/src/test/scala/filodb.core/query/ColumnFilterMapSpec.scala
alextheimer: Added tests to both `Default` and `Fast` maps ✅

alextheimer:
@yu-shipit I've updated my "10x" claim in the header -- it is closer to 4x more performant with some more-comprehensive tests.

I tested this by loading two `ColumnFilterMap` implementations with the same real-world mappings. Then I wrote a `getSeries` method that returned
- A matching series 50% of the time.
- Of the matching 50%:
  - 50% matched regex filters.
  - 50% matched equality filters.

For each implementation, I called `getSeries` and `columnFilterMap.get(...)` 1M times and recorded the total time each took to complete. A typical result (in milliseconds):
```
DefaultColumnFilterMap: 2199
FastColumnFilterMap:    611
```

yu-shipit (approved):
lgtm. Thank you for improving this.

PR #1817, opened 2024-07-23
feat(core): Add open/close support for Tantivy Index

rfairfax:
Commit 1 of adding Tantivy index support.  This PR is broken down into a small subset of the overall Tantivy logic to allow for easier partial review.  The index code is not usable end to end until all parts are committed.

This adds the basic Rust project skeleton and supports opening, applying schema, and closing the index. Many methods are unimplemented and will be added in follow up PRs.  End to end testing is not available in this PR as the index test suite requires ingestion and query support for verification.

**Pull Request checklist**

**New behavior :**

Part 1 of Tantivy index support.  The functionality is still incomplete and not ready for usage yet.


alextheimer (approved):
LGTM @rfairfax -- only one minor comment from me. Massive thanks for laying all of this groundwork 🙏

core/src/rust/src/index.rs
```diff
@@ -0,0 +1,129 @@
...
+    errors::{JavaException, JavaResult},
+    exec::jni_exec,
+    jnienv::JNIEnvExt,
+    state::{
+        field_constants::{self, facet_field_name, LABEL_LIST},
+        IndexHandle,
+    },
+};
+
+pub const WRITER_MEM_BUDGET: usize = 50 * 1024 * 1024;
```
amolnayak311, line 27: Suppose we run this in a container, I believe this comes from the total - heap mem? Do you think there is value in making this configurable or this is a good enough statically configured value?

core/src/rust/src/index.rs
```diff
@@ -0,0 +1,129 @@
...
+        return Err(JavaException::new_runtime_exception(
+            "More than one map field specified",
+        ));
+    }
+
+    let default_field = if len == 1 {
+        let name = env.get_object_array_element(map_fields, 0)?.into();
+        let name = env.get_rust_string(&name)?;
+
+        let field = builder.add_json_field(
```
amolnayak311, line 108: We wont need this to be ``FAST`` field?

amolnayak311 (approved):
Thanks @rfairfax LGTM, left few comments/questions

core/src/rust/Cargo.lock
```diff
@@ -0,0 +1,1266 @@
+# This file is automatically @generated by Cargo.
```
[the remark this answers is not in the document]
rfairfax: You want to include this to ensure that on every build you use the same dependency versions.  Without this file Cargo automatically pulls in latest versions, which could lead to unexpected build breaks or differences in behavior depending on build servers.

core/src/rust/src/index.rs
rfairfax: My general thought is that this is already smaller than Lucene, so I'm not concerned about it causing new issues.  If we think there's legit reason to change to a config value it's easy enough to do, but my gut says to do that once we have a need to change the value to avoid config explosion.

core/src/rust/src/index.rs
rfairfax: Yes we do.  I have this set to fast in a future PR but I'll port that back to this.

core/src/rust/src/state.rs
```diff
@@ -0,0 +1,92 @@
...
+}
+
+const PART_KEY_INDEX_RAW_CLASS: &str = "filodb/core/memstore/PartKeyIndexRaw";
+
+pub mod field_constants {
+    pub fn facet_field_name(name: &str) -> String {
+        format!("{}{}", FACET_FIELD_PREFIX, name)
+    }
+
+    pub const PART_ID: &str = "__partIdDv__";
```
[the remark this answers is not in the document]
rfairfax: I'm actually going to add partIdField back - we end up needing it for some later scenarios so it can't be safely removed yet without a larger refactor in the Scala level APIs.

partIdDv is an i64 field - unless I'm misunderstanding what you mean by "is a string"?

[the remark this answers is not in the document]
alextheimer: Ah duh you're right about that second bit -- thanks for the documentation here 🙏

alextheimer (approved):

amolnayak311 (approved):

PR #1819, opened 2024-07-25
feat(core): Now metadata queries support _type_ filter

vishramachandran:
**Pull Request checklist**

**Current behavior :** 
Metadata queries do not support filtering based on metric type using `_type_` filter

**New behavior :**
Added support.
Note that only new documents of persisted downsample index will have the _type_ field. Index needs to be rebuilt if full support is needed. Can be skipped if metadata queries don't hit downsample index.

### DIFF
--- a/core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
+++ b/core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
@@ -174,7 +174,8 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
       schemas.part.binSchema.toStringPairs(partKey.base, partKey.offset).map(pair => {
         pair._1.utf8 -> pair._2.utf8
       }).toMap ++
-        Map("_type_".utf8 -> Schemas.global.schemaName(RecordSchema.schemaID(partKey.base, partKey.offset)).utf8)
+        Map(Schemas.TypeLabel.utf8 ->
+          Schemas.global.schemaName(RecordSchema.schemaID(partKey.base, partKey.offset)).utf8)
     }
   }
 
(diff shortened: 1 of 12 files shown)


rfairfax (approved):

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
```diff
@@ -678,6 +678,9 @@ class PartKeyLuceneIndex(ref: DatasetRef,
     // If configured and enabled, Multi-column facets will be created on "partition-schema" columns
     createMultiColumnFacets(partKeyOnHeapBytes, partKeyBytesRefOffset)
 
+    val schemaName = Schemas.global.schemaName(RecordSchema.schemaID(partKeyOnHeapBytes, UnsafeUtils.arayOffset))
+    luceneDocument.get().addField(Schemas.TypeLabel, schemaName)
```
amolnayak311, line 682: Nit pick, this is same as ``addIndexedField`` and while looking up  that the method in the code when someone is trying to understand the setup, this gets missed.  If this line is changed to ``addIndexedField`` that would be helpful.

amolnayak311 (commented):
Thanks for the changes @vishramachandran I have one question and one change if you can push. Other than that looks good to me,.

core/src/main/scala/filodb.core/memstore/PartKeyLuceneIndex.scala
vishramachandran: sure

core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
```diff
@@ -174,7 +174,8 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
       schemas.part.binSchema.toStringPairs(partKey.base, partKey.offset).map(pair => {
         pair._1.utf8 -> pair._2.utf8
       }).toMap ++
-        Map("_type_".utf8 -> Schemas.global.schemaName(RecordSchema.schemaID(partKey.base, partKey.offset)).utf8)
```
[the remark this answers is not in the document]
vishramachandran: No. It is only an indexed field and we would not see the `_type_` field in promQl results just because of that.
Type field is added into the result by extracting label-value pairs from partKey and not from indexed fields. They should not be connected

amolnayak311 (approved):

PR #1822, opened 2024-07-30
fix(core) send the right promql to remote in HighAvailabilityPlanner.

yu-shipit:
**Pull Request checklist**

**Current behavior : 
For queries such as `1 - sum(foo)`, when local shards are unhealthy, the `HighAvailabilityPlanner` will create a remote plan  `1 - sum(foo)` that runs in its healthy buddy cluster. This, however, cause that the query behaves equivalent to `1 - Remote(1 - sum(foo))`.

**New behavior :**
The same query will be materialized something like` 1 - Remote(sum(foo))` to produce the correct result.

**BREAKING CHANGES**

Breaking changes may include:

**Other information**:

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,14 @@ brew bottle --skip-relocation kafka
 brew reinstall `ls kafka*bottle*`
 ```
 
+Newer versions of ZooKeeper start an admin HTTP server on port 8080, which conflicts with the FiloDB servers.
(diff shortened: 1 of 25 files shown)


sandeep6189 (approved):

sherali42 (approved):
thanks @yu-shipit @sandeep6189 for the quick fix 👍

PR #1835, opened 2024-08-14
feat(core): Add Rust side of Tantivy query support

rfairfax:
This contains the Rust code needed to support all query patterns for FiloDB.  This is part 3 of the series and is not usuable end to end until additional PRs complete.

This PR splits the logic into a common tantivy library, where collectors and query extensions are implemented, and the filodb library, where specific JVM bridge methods and business logic exists.

The majority of this PR are custom collectors and caching to better match the performance of Lucene in the FiloDB use cases.  The glue logic to connect these collectors to JVM methods are for the most part very straight forward - parse a query, run it, serialize results.

**Pull Request checklist**

n/a

**New behavior :**

Adds query methods for Tantivy index.  The Scala side of this will be in the next PR to glue it all together.


core/src/rust/filodb_core/src/state.rs
```diff
@@ -0,0 +1,144 @@
+//! State objects shared with Java
```
rfairfax, line 1: The git diff is a bit confused with this file.  It is mostly a move as is, with the execute_cachable_query methods being new.

core/src/rust/tantivy_utils/src/query/cache.rs
```diff
@@ -0,0 +1,245 @@
...
+{
+    fn equivalent(&self, key: &(SegmentId, QueryType)) -> bool {
+        self.0 == key.0 && *self.1 == key.1
+    }
+}
+
+// Tuning parameters for query cache
+const QUERY_CACHE_MAX_SIZE_BYTES: u64 = 50_000_000;
+// Rough estimate of bitset size - 250k docs
+const QUERY_CACHE_AVG_ITEM_SIZE: u64 = 31250;
```
alextheimer, line 68: Can these be made configurable?

alextheimer (commented):
Thanks for the PR @rfairfax (and especially for the huge count of tests, as well). Let me know when you've taken a look at these comments 🙏

core/src/rust/tantivy_utils/src/collectors/column_cache.rs
```diff
@@ -0,0 +1,266 @@
...
+
+/// Opening DynamicColumn instances requires parsing some headers
+/// and other items that while fast, can add up if you're doing this
+/// thousands of times a second.  Since columns for a given segment
+/// are immutable once created caching this parsed data is safe
+/// and cheap and can result in major speedups on things like
+/// point queries.
+#[derive(Clone)]
+pub struct ColumnCache {
+    cache: Arc<Cache<(SegmentId, String), DynamicColumn>>,
```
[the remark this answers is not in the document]
rfairfax: Arc does not have a lock - it's atomically reference counted.  If I change it to Box then cloning the ColumnCache struct, which is required for the collectors to hold it without having to mangle their lifetimes via a reference, becomes a complete copy of the cache.  By keeping the inner reference an Arc the clone operation copies a pointer + increments a ref count (2 words), which makes this highly efficient and sharable without a lock like a Mutex or RwLock.

core/src/rust/filodb_core/src/reader.rs
```diff
@@ -0,0 +1,558 @@
...
+            env.new_object_array(results.len() as i32, "java/lang/String", JObject::null())?;
+        for (i, item) in results.drain().enumerate() {
+            env.set_object_array_element(&java_ret, i as i32, env.new_string(item)?)?;
+        }
+
+        Ok(java_ret.into_raw())
+    })
+}
+
+const MAX_TERMS_TO_ITERATE: usize = 10_000;
```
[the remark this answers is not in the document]
rfairfax: Will add a comment as to where this came from

core/src/rust/filodb_core/src/query/cachable_query.rs
```diff
@@ -0,0 +1,265 @@
...
+use tantivy_utils::field_constants;
+
+use super::parse_query;
+
+/// A query that can potentially be cached
+///
+/// We can't just hold a reference to Tantivy's Query object because
+/// they don't implement Hash/Equals so they can't be a key
+#[derive(Debug, Hash, PartialEq, Eq, PartialOrd, Ord, Clone)]
+pub enum CachableQuery {
```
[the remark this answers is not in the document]
rfairfax: Sure.  This is an artifact of an earlier version where I cached everything, before realizing on simple queries it wasn't worth it.  Will clean up the naming.

core/src/rust/filodb_core/src/query.rs
```diff
@@ -0,0 +1,983 @@
...
+    input: &'a [u8],
+    schema: &Schema,
+    default_field: Option<Field>,
+) -> IResult<&'a [u8], Box<dyn Query>, ParserError> {
+    let (input, column) = parse_string(input)?;
+    let (input, text) = parse_string(input)?;
+
+    let query = query_with_field_and_value(schema, default_field, &column, |field, prefix| {
+        Ok(Box::new(TermQuery::new(
+            Term::from_field_text(field, &value_with_prefix(prefix, &text)),
```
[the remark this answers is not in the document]
rfairfax: Will add comments explaining this - but the simple version is for JSON (map) fields, the encoded value is `<dynamic field name>\0s<value>`.  We need to focus our input query to adjust to the expected prefix.

core/src/rust/filodb_core/src/query.rs
```diff
@@ -0,0 +1,983 @@
...
+    /// Query must match
+    Must = 1,
+    /// Query must not match
+    MustNot = 2,
+}
+
+pub fn parse_query<'a>(
+    input: &'a [u8],
+    schema: &Schema,
+    default_field: Option<Field>,
```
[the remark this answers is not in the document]
rfairfax: Will add a comment, but this is used to search the map field when the input name isn't one of the top level index names.  This makes us behave like Lucene, which has dynamic schema, even though Tantivy has fixed schema.

core/src/rust/filodb_core/src/profile.rs
```diff
@@ -0,0 +1,73 @@
...
+pub extern "system" fn Java_filodb_core_memstore_TantivyNativeMethods_00024_dumpCacheStats(
+    mut env: JNIEnv,
+    _class: JClass,
+    handle: jlong,
+) -> jstring {
+    jni_exec(&mut env, |env| {
+        let index = IndexHandle::get_ref_from_handle(handle);
+
+        let (column_hits, column_misses) = index.column_cache.stats();
+        let (query_hits, query_misses) = index.query_cache_stats();
```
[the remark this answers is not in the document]
rfairfax: The library I'm using doesn't appear to expose any kind of stats reset method.  However, it doesn't matter too much at the moment because the only place that is using this is the JMH tests and they drop the whole JVM after every test anyway.

core/src/rust/tantivy_utils/src/query/cache.rs
rfairfax: Sure, will wire this up (and the others called out)

core/src/rust/filodb_core/src/query/cachable_query.rs
```diff
@@ -0,0 +1,265 @@
...
+        }
+    }
+}
+
+#[derive(Clone, Default)]
+pub struct CachableQueryWeighter;
+
+impl Weighter<(SegmentId, CachableQuery), Arc<BitSet>> for CachableQueryWeighter {
+    fn weight(&self, key: &(SegmentId, CachableQuery), val: &Arc<BitSet>) -> u64 {
+        let bitset_size = ((val.max_value() as usize + 63) / 64) * 8;
```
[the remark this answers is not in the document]
rfairfax: Unless I'm missing something I don't see a capacity on BitSet.  They do have this calculation on a private method, but we can't get at it.

core/src/rust/tantivy_utils/src/collectors/limited_collector/unlimited_collector.rs
```diff
@@ -0,0 +1,111 @@
+//! Helper to wrap collectors as unlimited
```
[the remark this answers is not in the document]
rfairfax: Now I remember why I did this - FacetCollector's segment collector is in a private module, so I can't add a trait to it.  I need this adapter pattern to adapt the inbox Tantivy collectors without changing their code. 

I'll add a comment to that effect

core/src/rust/filodb_core/src/query.rs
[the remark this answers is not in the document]
alextheimer: Ah makes sense -- thank you 🙏

core/src/rust/tantivy_utils/src/collectors/column_cache.rs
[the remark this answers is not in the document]
alextheimer: Ah -- I misunderstood some docs and somehow figured `Arcs` protect more than just the reference count. Definitely keep it as-is 👍

alextheimer (dismissed):
Awesome stuff @rfairfax -- thanks for fielding all of that feedback. PR LGTM 👍

core/src/rust/filodb_core/src/profile.rs
```diff
@@ -0,0 +1,112 @@
...
+
+        let output = format!(
```
amolnayak311, line 27: While it's ok to log, the value will determine the changes we will make to ``column-cache-count``. Does it make sense to have a gauge for this as a metric (need not be part of this PR)?

core/src/rust/filodb_core/src/query.rs
```diff
@@ -0,0 +1,997 @@
...
+) -> IResult<&'a [u8], Box<dyn Query>, ParserError> {
+    let (input, column) = parse_string(input)?;
+    let (input, text) = parse_string(input)?;
+
+    let query = query_with_field_and_value(schema, default_field, &column, |field, prefix| {
+        Ok(Box::new(TermQuery::new(
+            Term::from_field_text(field, &value_with_prefix(prefix, &text)),
```
amolnayak311, line 215: Wouldn't just giving the field that contains the map suffice? I found this prefix was needed only for `RegexQuery`

core/src/rust/filodb_core/src/query.rs
amolnayak311, line 1: Should be renamed to ``parsers.rs``/``query_parsers.rs``? If I was browsing the code I would expect to see the query logic in here.

core/src/rust/filodb_core/src/query/filodb_query.rs
```diff
@@ -0,0 +1,265 @@
...
+                let field = schema.get_field(field_constants::PART_KEY)?;
+                let term = Term::from_field_bytes(field, part_key);
+                let query = TermQuery::new(term, IndexRecordOption::Basic);
+
+                Ok(Box::new(query))
+            }
+            FiloDBQuery::ByPartIds(part_ids) => {
+                let part_id_field = schema.get_field(field_constants::PART_ID)?;
+
+                let mut terms = vec![];
```
amolnayak311, line 73: Nit pick, can we initialize with capacity since we know the number of part_ids?

core/src/rust/filodb_core/src/query/filodb_query.rs
```diff
@@ -0,0 +1,265 @@
...
+            FiloDBQuery::All => Ok(Box::new(AllQuery)),
+            FiloDBQuery::ByPartId(part_id) => {
+                let part_id_field = schema.get_field(field_constants::PART_ID)?;
+                let term = Term::from_field_i64(part_id_field, *part_id as i64);
+
+                let query = TermQuery::new(term, IndexRecordOption::Basic);
+
+                Ok(Box::new(query))
+            }
+            FiloDBQuery::ByEndTime(ended_before) => {
```
amolnayak311, line 92: Nit pick: I know this is how its named in the Lucent implementation, but can the variable be ended_at? ``before`` makes one believe its end Exclusive.

core/src/rust/filodb_core/src/query/filodb_query.rs
```diff
@@ -0,0 +1,265 @@
...
+                Ok(Box::new(query))
+            }
+        }
+    }
+}
+
+#[derive(Clone, Default)]
+pub struct CachableQueryWeighter;
+
+impl Weighter<(SegmentId, FiloDBQuery), Arc<BitSet>> for CachableQueryWeighter {
```
amolnayak311, line 108: Some comments on what this does with quick_cache will be helpful. There are some Tantivy internals like BitSet used here and some explanation on how it's used and the intuition of how this weight is calculated will be helpful.

core/src/rust/filodb_core/src/reader.rs
amolnayak311, line 1: This is what I was expecting ``query.rs`` to be :smile: but given it reads the index,`` reader.rs`` makes sense.

core/src/rust/filodb_core/src/reader.rs
```diff
@@ -0,0 +1,560 @@
...
+
+use crate::{
+    errors::{JavaException, JavaResult},
+    exec::jni_exec,
+    jnienv::JNIEnvExt,
+    query::filodb_query::FiloDBQuery,
+    state::IndexHandle,
+};
+
+const TERM_INFO_CLASS: &str = "filodb/core/memstore/TermInfo";
```
amolnayak311, line 32: Though I don't expect to, but if someone refactors and changes the packaging.  Would there be a  compilation failure or there will be a is a runtime exception when running using Tantivy?

core/src/rust/tantivy_utils/src/collectors/limited_collector.rs
```diff
@@ -0,0 +1,90 @@
...
+        &self,
+        weight: &dyn Weight,
+        segment_ord: u32,
+        reader: &SegmentReader,
+        limiter: &mut LimitCounter,
+    ) -> Result<<Self::Child as SegmentCollector>::Fruit, TantivyError> {
+        let mut segment_collector = self.for_segment(segment_ord, reader)?;
+        let mut scorer = weight.scorer(reader, 1.0)?;
+
+        match (reader.alive_bitset(), self.requires_scoring()) {
```
amolnayak311, line 54: Can we have some comments on what this logic does?

core/src/rust/tantivy_utils/src/query/shared_doc_set.rs
```diff
@@ -0,0 +1,176 @@
+//! Low memcpy sharable docset
```
amolnayak311, line 1: We can include this one to review as well.

core/src/rust/tantivy_utils/src/query/prefix_query.rs
```diff
@@ -0,0 +1,96 @@
+//! Query that does a prefix match
```
amolnayak311, line 1: Perhaps you,  @alextheimer and I can sync up and understand more about this piece of code.

core/src/rust/tantivy_utils/src/query/bitset_weight.rs
```diff
@@ -0,0 +1,90 @@
+//! Weight adapter for a cached bitset
```
amolnayak311, line 1: I guess we can syncup offline for understand more about these tantivy internals.

amolnayak311 (approved):
Looks great, left some comments

core/src/rust/filodb_core/src/query.rs
rfairfax: The term query on a JSON field won't match if it doesn't have the correct prefix.

core/src/rust/filodb_core/src/reader.rs
rfairfax: It would be a runtime exception for ClassNameNotFound or something similar, caught via unit tests.

core/src/rust/filodb_core/src/profile.rs
rfairfax: Let me think about this for a future PR.  In practice with this cache I'm not sure we'll ever tune it - 1000 entries is not much RAM and we don't have that many actual columns to cache / segments even as we grow.

core/src/rust/filodb_core/src/query/filodb_query.rs
rfairfax: Yep, good call

core/src/rust/filodb_core/src/query/filodb_query.rs
rfairfax: Updated

core/src/rust/tantivy_utils/src/collectors/limited_collector.rs
rfairfax: Added

core/src/rust/filodb_core/src/profile.rs
rfairfax: I thought about this some more - what I'd like to do is a follow up PR that adds the following metrics:

* Cache hit rates for both caches
* Ingestion commit duration

I'll make sure these are in before we merge to develop.

amolnayak311 (approved):

alextheimer (approved):

PR #1842, opened 2024-08-22
feat(core): Do not index any existing _type_ field since it is a reserved field

vishramachandran:
**Pull Request checklist**

Addition of `_type_` field to index is now configurable for each cluster, false by default for now.
Also, if the part-key already has the `_type_` field we don't index that since it is a reserved field that we populate.


rfairfax:
We will need to merge this logic into the tantivy branch and adjust the tantivy code to honor it - let me know once this is in develop and I'll handle that.

sherali42 (approved):
👍  thanks @vishramachandran

PR #1843, opened 2024-08-26
feat(core): Add Scala side of Tantivy query support

rfairfax:
This contains the Scala side of query support.  This completes the basic end to end flow, so it also includes regression and performance tests.

There will be some small cleanup PRs after this for things like additional metrics, but with this PR the Tantivy index can be successfully activated via config.

**Pull Request checklist**

**New behavior :**

Adds query logic to call into Rust code and tests

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/PartKeyIndex.scala
+++ b/core/src/main/scala/filodb.core/memstore/PartKeyIndex.scala
@@ -313,11 +313,13 @@ abstract class PartKeyIndexRaw(ref: DatasetRef,
    */
   def labelNamesEfficient(colFilters: Seq[ColumnFilter], startTime: Long, endTime: Long): Seq[String]
 
+  protected final val LABEL_NAMES_AND_VALUES_DEFAULT_LIMIT = 100
+
   /**
    * Use faceting to get field/index values given a column filter and time range
    */
   def labelValuesEfficient(colFilters: Seq[ColumnFilter], startTime: Long, endTime: Long,
-                           colName: String, limit: Int = 100): Seq[String]
+                           colName: String, limit: Int = LABEL_NAMES_AND_VALUES_DEFAULT_LIMIT): Seq[String]
 
   /**
    * Add new part key to index
--- a/core/src/main/scala/filodb.core/memstore/PartKeyIndex.scala
+++ b/core/src/main/scala/filodb.core/memstore/PartKeyIndex.scala
@@ -2,6 +2,7 @@ package filodb.core.memstore
 
 import java.io.File
 import java.nio.charset.StandardCharsets
+import java.nio.file.Path
 import java.util.regex.Pattern
 
 import scala.collection.immutable.HashSet
@@ -69,13 +70,25 @@ abstract class PartKeyIndexRaw(ref: DatasetRef,
     .withTag("dataset", ref.dataset)
     .withTag("shard", shardNum)
 
+  protected val labelValuesQueryLatency = Kamon.histogram("index-label-values-query-latency",
+      MeasurementUnit.time.nanoseconds)
+    .withTag("dataset", ref.dataset)
+    .withTag("shard", shardNum)
+
   protected val queryIndexLookupLatency = Kamon.histogram("index-partition-lookup-latency",
       MeasurementUnit.time.nanoseconds)
     .withTag("dataset", ref.dataset)
     .withTag("shard", shardNum)
 
-  protected val indexDiskLocation = diskLocation.map(new File(_, ref.dataset + File.separator + shardNum))
+  protected val partIdFromPartKeyLookupLatency = Kamon.histogram("index-ingestion-partId-lookup-latency",
+      MeasurementUnit.time.nanoseconds)
+    .withTag("dataset", ref.dataset)
+    .withTag("shard", shardNum)
+
+  private val _indexDiskLocation = diskLocation.map(new File(_, ref.dataset + File.separator + shardNum))
     .getOrElse(createTempDir(ref, shardNum)).toPath
+
+  def indexDiskLocation: Path = _indexDiskLocation
 
   // If index rebuild is triggered or the state is Building, simply clean up the index directory and start
   // index rebuild
@@ -269,6 +282,11 @@ abstract class PartKeyIndexRaw(ref: DatasetRef,
    * Number of documents in flushed index, excludes tombstones for deletes
    */
   def indexNumEntries: Long
+
+  /**
+   * Memory used by index in mmap segments, mostly for index data
+   */
+  def indexMmapBytes: Long
 
   /**
    * Closes the index for read by other clients. Check for implementation if commit would be done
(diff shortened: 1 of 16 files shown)


core/src/main/scala/filodb.core/memstore/PartKeyTantivyIndex.scala
```diff
@@ -272,12 +398,134 @@ object ByteBufferEncodingUtils {
...
   }
 }
 
+object TantivyQueryBuilder {
+  private val bufferLocal = new ThreadLocal[ArrayBuffer[Byte]]() {
+    override def initialValue(): ArrayBuffer[Byte] = new ArrayBuffer[Byte](4096)
+  }
+}
+
+class TantivyQueryBuilder extends PartKeyQueryBuilder with StrictLogging {
```
rfairfax, line 413: This matches the decoding logic seen in the previous Rust PR

core/src/test/scala/filodb.core/memstore/PartKeyIndexRawSpec.scala
```diff
@@ -0,0 +1,926 @@
...
+                               facetEnabledAllLabels: Boolean,
+                               facetEnabledShardKeyLabels: Boolean,
+                               shardNum: Int,
+                               retentionMillis: Long,
+                               diskLocation: Option[File] = None,
+                               lifecycleManager: Option[IndexMetadataStore] = None): PartKeyIndexRaw
+
+  // scalastyle:off method.length
+  // scalastyle:off cyclomatic.complexity
+  def commonPartKeyTests(keyIndex: PartKeyIndexRaw, partBuilder: RecordBuilder): Unit = {
```
rfairfax, line 40: All of these tests are moved as is from the existing Lucene class.  The only changes where to make the asserts accept results in any order as Tantivy doesn't guarantee order in returns.

No new test logic was added in this class.

core/src/test/scala/filodb.core/memstore/PartKeyLuceneIndexSpec.scala
```diff
@@ -51,83 +45,19 @@ class PartKeyLuceneIndexSpec extends AnyFunSpec with Matchers with BeforeAndAfte
     }
   }
 
-  it("should add part keys and parse filters correctly") {
```
rfairfax, line 54: All deletes here are just moves to the new abstract base class

project/plugins.sbt
```diff
@@ -1,8 +1,8 @@
 addSbtPlugin("org.scalastyle" %% "scalastyle-sbt-plugin" % "1.0.0")
 
-addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.14.5")
+addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.14.9")
```
rfairfax, line 3: This was needed because the test refactors broke the assembly step - they got tripped up on valid JVM 11 byte code.  Turns out the first version of this with working JVM 11 support was 0.14.9 and we were just getting lucky it wasn't breaking before.

project/plugins.sbt
```diff
@@ -1,8 +1,8 @@
...
 
-addSbtPlugin("pl.project13.scala" % "sbt-jmh" % "0.3.7")
+addSbtPlugin("pl.project13.scala" % "sbt-jmh" % "0.4.7")
```
rfairfax, line 5: 0.4.7 adds significant new profiling support

core/src/main/scala/filodb.core/memstore/PartKeyTantivyIndex.scala
```diff
@@ -272,12 +398,134 @@ object ByteBufferEncodingUtils {
...
+
+  override protected def visitRangeQuery(column: String, start: Long, end: Long, occur: PartKeyQueryOccur): Unit = {
+    writeOccur(occur)
+
+    // Type byte, col len, col, start, end
+    buffer += LONG_RANGE_TYPE_BYTE
+    writeString(column)
+    writeLong(start)
+    writeLong(end)
+  }
```
alextheimer, line 520: Where is this invoked? And should this have a unit test added, as well?

alextheimer (commented):
Awesome stuff as usual @rfairfax -- just a couple nitpick comments, but otherwise LGTM.

Massive thanks for all the benchmark work as well 🙏

core/src/main/scala/filodb.core/memstore/PartKeyTantivyIndex.scala
rfairfax: This is invoked in the base class as part of visitQueryWithStartAndEnd.  Today, Tantivy won't get any calls to this as we made a change to send the start/end time as explicit parameters for optimization vs part of the query.  However, I think the right thing is to keep this implemented should we change that logic in the future.

You're correct that I'm missing a test for this case.  Will add one.

core/src/main/scala/filodb.core/memstore/PartKeyTantivyIndex.scala
```diff
@@ -107,20 +130,40 @@ class PartKeyTantivyIndex(ref: DatasetRef,
...
+    results.toSeq
   }
 
   override def labelNamesEfficient(colFilters: Seq[ColumnFilter], startTime: Long, endTime: Long): Seq[String] = {
-    ???
+    val start = System.nanoTime()
+    val queryBuilder = new TantivyQueryBuilder()
+    val query = queryBuilder.buildQuery(colFilters)
+
+    val results = TantivyNativeMethods.labelNames(indexHandle, query, 100, startTime, endTime)
```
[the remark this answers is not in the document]
rfairfax: I'll add a constant and adjust it here and in Lucene - the goal is to be consistent in return sizes between the indexes.

core/src/main/scala/filodb.core/memstore/PartKeyIndex.scala
```diff
@@ -69,14 +70,26 @@ abstract class PartKeyIndexRaw(ref: DatasetRef,
...
-  protected val indexDiskLocation = diskLocation.map(new File(_, ref.dataset + File.separator + shardNum))
+  protected val partIdFromPartKeyLookupLatency = Kamon.histogram("index-ingestion-partId-lookup-latency",
+      MeasurementUnit.time.nanoseconds)
+    .withTag("dataset", ref.dataset)
+    .withTag("shard", shardNum)
+
+  private val _indexDiskLocation = diskLocation.map(new File(_, ref.dataset + File.separator + shardNum))
     .getOrElse(createTempDir(ref, shardNum)).toPath
 
+  def indexDiskLocation: Path = _indexDiskLocation
```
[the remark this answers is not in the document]
rfairfax: I need to save it to a member val because `createTempDir` being called every time would create a new folder on each ask of the location.

I have a wrapping method so that code like the test code can access this without going to the variable directly.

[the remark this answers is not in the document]
alextheimer: Ah -- understood

core/src/main/scala/filodb.core/memstore/PartKeyTantivyIndex.scala
rfairfax: Added a test

core/src/main/scala/filodb.core/memstore/PartKeyTantivyIndex.scala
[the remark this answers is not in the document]
rfairfax: Updated

alextheimer (dismissed):
Thanks for the changes @rfairfax -- PR LGTM 👍

core/src/rust/filodb_core/src/index.rs
```diff
@@ -56,9 +57,22 @@ pub extern "system" fn Java_filodb_core_memstore_TantivyNativeMethods_00024_newI
             .open_or_create(directory.clone())?;
 
         let writer = index.writer::<TantivyDocument>(WRITER_MEM_BUDGET)?;
+
+        let mut merge_policy = LogMergePolicy::default();
+        merge_policy.set_del_docs_ratio_before_merge(0.1);
```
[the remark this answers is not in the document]
amolnayak311: I don't want to cause a config overload but is .1 a good number across to be hardcoded even in cases where there is high churn? Would frequent segment merges cause cpu spikes (Generally lesser segments will help with query performance)?

amolnayak311 (dismissed):
Thanks a ton @rfairfax for this PR. Just left one question on config for delete ratio.

core/src/rust/filodb_core/src/index.rs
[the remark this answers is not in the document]
rfairfax: I'm not sure we ever need to change this per instance so it feels like it's not something I'd put in a config file, though it can be converted to a config in the future.

This number only indicates the lower bound on when it is considered for merges.  It does not guarantee a merge at 10%, just that it will be compared to other files for merge candidates.  In general it only merges when there are many files that benefit from it.

Their logic is logarithmic for merging, so it tends to avoid constant merging.  In general I found the defaults besides this value were very sane and they recommend changing the delete ratio from 0 but don't provide a sensible default due to back compat.

My vote is to keep this at this point, monitor perf, and adjust later if needed.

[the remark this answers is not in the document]
amolnayak311: Thanks @rfairfax for the quick changes

amolnayak311 (approved):

alextheimer (approved):

PR #1846, opened 2024-09-07
fix(query): negative rate/increase due to NaN chunk

sherali42:
**Pull Request checklist**

This change is to handle a case where a time series ends with a NaN value. Currently this results into a new chunk/row with a single NaN value. 
As we default NaN to zero value during rate/increase window calculation, this sets highestValue to be zero. This is resulting into negative rate/increase values, since rate = highestValue-lowestValue.
This fix is to short `addTimeChunks` operation and return without any effect on highestValue calculation.

### DIFF
--- a/query/src/main/scala/filodb/query/exec/rangefn/RateFunctions.scala
+++ b/query/src/main/scala/filodb/query/exec/rangefn/RateFunctions.scala
@@ -173,6 +173,10 @@ abstract class ChunkedRateFunctionBase extends CounterChunkedRangeFunction[Trans
                     startRowNum: Int, endRowNum: Int,
                     startTime: Long, endTime: Long): Unit = {
     val dblReader = reader.asDoubleReader
+    // if the chunk is a single row NaN value, then return. Prometheus end of time series marker.
+    // This is to make sure we don't set highestValue to zero. avoids negative rate/increase values.
+    if (startRowNum == 0 && endRowNum == 0 && dblReader.apply(acc, vector, startRowNum).isNaN)
+      return
     if (startTime < lowestTime || endTime > highestTime) {
       numSamples += endRowNum - startRowNum + 1
       if (startTime < lowestTime) {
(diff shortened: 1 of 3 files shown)


query/src/main/scala/filodb/query/exec/rangefn/RateFunctions.scala
```diff
@@ -173,6 +173,10 @@ abstract class ChunkedRateFunctionBase extends CounterChunkedRangeFunction[Trans
                     startRowNum: Int, endRowNum: Int,
                     startTime: Long, endTime: Long): Unit = {
     val dblReader = reader.asDoubleReader
+    // if the chunk is a single row NaN value, then return. Prometheus end of time series marker.
+    // This is to make sure we don't set highestValue to zero. avoids negative rate/increase values.
+    if (startRowNum == 0 && endRowNum == 0 && dblReader.apply(acc, vector, startRowNum).isNaN)
```
yu-shipit, line 178: Just curious if check should be
 `if (startRowNum == endRowNum && dblReader.apply(acc, vector, startRowNum).isNaN)`

sherali42: for this issue to manifest, it has to be single row NaN chunk. We could just test for `endRowNum == 0` this implies startRowNum is 0.

sandeep6189: Thanks @sherali42 . So basically, we are not updating the highestValue in this condition ? 
Should we also add a unit test for this edge case ?

sandeep6189 (approved):

query/src/main/scala/filodb/query/exec/rangefn/RateFunctions.scala
sherali42: @sandeep6189 added a test case, PTAL

sandeep6189 (approved):
Thank you @sherali42 for taking this up 👍

PR #1874, opened 2024-10-29
perf(query) Memoize the part of the logical plan tree traversal for reduced memory allocation and faster planning

amolnayak311:
**Pull Request checklist**

**Current behavior :** 

Though there is no expected change in the behavior for the end users,  the memory allocation for  methods ``getPushdownShards``, ``isTargetSchemaChanging`` and  ``QueryUtils.makeAllKeyValueCombos`` is high. This high memory allocation can be reduced by memoizing the responses which are repeatedly called by recursing the logical plan tree

With no caching enabled the JMH tests for materialize give following results

```
[info] Do not assume the numbers tell you what you want them to tell.
[info] Benchmark                                        Mode  Cnt   Score    Error  Units
[info] PlannerBenchmark.benchmarkMaterializePlan        avgt    5  20.819 ± 13.145  ms/op
```

Screenshot 2024-10-29 at 3 03 08 PM

Screenshot 2024-10-29 at 3 02 34 PM

As seen above, ``getPushdownShards``, ``isTargetSchemaChanging``  show up as two hotspots for memory allocation profiling. 

**New behavior :**

After memoizing the responses, the hotspots are no longer visible. The planner's materialize is 2-3X improvement in materialize times and then two calls don't show up in allocation hotspots anymore. 

```
[info] Do not assume the numbers tell you what you want them to tell.
[info] Benchmark                                        Mode  Cnt  Score   Error  Units
[info] PlannerBenchmark.benchmarkMaterializePlan        avgt    5  7.642 ± 3.868  ms/op
```

Screenshot 2024-10-29 at 3 18 34 PM

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
@@ -4,6 +4,7 @@ import scala.collection.mutable.ArrayBuffer
 import scala.concurrent.duration._
 
 import akka.actor.ActorRef
+import com.github.benmanes.caffeine.cache.{Cache, Caffeine}
 import com.typesafe.scalalogging.StrictLogging
 import kamon.Kamon
 
@@ -65,6 +66,36 @@ class SingleClusterPlanner(val dataset: Dataset,
   private val shardColumns = dsOptions.shardKeyColumns.sorted
   private val dsRef = dataset.ref
 
+  private val shardPushdownCache: Option[Cache[(LogicalPlan, Option[Seq[Int]]), Option[Set[Int]]]] =
+      if (queryConfig.cachingConfig.singleClusterPlannerCachingEnabled) {
+        Some(
+          Caffeine.newBuilder()
+          .maximumSize(queryConfig.cachingConfig.singleClusterPlannerCachingSize)
+          .recordStats()
+          .build()
+        )
+      } else {
+        None
+      }
+
+
+  private val tSchemaChangingCache: Option[Cache[(Seq[ColumnFilter], Long, Long), Some[Boolean]]] =
+    if (queryConfig.cachingConfig.singleClusterPlannerCachingEnabled) {
+      Some(
+        Caffeine.newBuilder()
+          .maximumSize(queryConfig.cachingConfig.singleClusterPlannerCachingSize)
+          .recordStats()
+          .build()
+      )
+    } else {
+      None
+    }
+
+  private[queryplanner] def invalidateCaches(): Unit = {
+    shardPushdownCache.foreach(_.invalidateAll())
+    tSchemaChangingCache.foreach(_.invalidateAll())
+  }
+
   // failed failover counter captures failovers which are not possible because at least one shard
   // is down both on the primary and DR clusters, the query will get executed only when the
   // partial results are acceptable otherwise an exception is thrown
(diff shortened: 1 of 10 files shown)


coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
```diff
@@ -106,6 +133,25 @@ class SingleClusterPlanner(val dataset: Dataset,
...
+   *   (2) changes between the argument timestamps.
+   */
+  def isTargetSchemaChanging(shardKeyFilters: Seq[ColumnFilter],
+                             startMs: Long, endMs: Long,
+                             qContext: QueryContext): Boolean = {
+    tSchemaChangingCache match {
+      case Some(cache) =>
+        cache.get((shardKeyFilters, startMs, endMs), _ => {
+          Some(isTargetSchemaChangingInner(shardKeyFilters, startMs, endMs, qContext))
+        }).getOrElse(true)
```
alextheimer, line 149: When might that `getOrElse` be needed?

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
```diff
@@ -65,6 +66,36 @@ class SingleClusterPlanner(val dataset: Dataset,
...
+          .maximumSize(queryConfig.cachingConfig.singleClusterPlannerCachingSize)
+          .recordStats()
+          .build()
+        )
+      } else {
+        None
+      }
+
+
+  private val tSchemaChangingCache: Option[Cache[(Seq[ColumnFilter], Long, Long), Some[Boolean]]] =
```
alextheimer, line 82: Instead of `Some[Boolean]`, would just `Boolean` work here?

coordinator/src/test/scala/filodb.coordinator/queryplanner/SingleClusterPlannerSpec.scala
```diff
@@ -386,6 +387,7 @@ class SingleClusterPlannerSpec extends AnyFunSpec with Matchers with ScalaFuture
   // Target-Schema start
 
   it("should stitch results when target-schema changes during query range") {
+    engine.invalidateCaches()
```
alextheimer, line 390: You can add `implements BeforeAndAfterEach` to the test class then `override def beforeEach(): Unit` so this runs before every test.

core/src/main/scala/filodb.core/query/QueryUtils.scala
```diff
@@ -1,16 +1,25 @@
...
 import scala.collection.mutable
 import scala.collection.mutable.ArrayBuffer
 
 /**
  * Storage for utility functions.
  */
 object QueryUtils {
-  val REGEX_CHARS = Array('.', '?', '+', '*', '|', '{', '}', '[', ']', '(', ')', '"', '\\')
+  val REGEX_CHARS: Array[Char] = Array('.', '?', '+', '*', '|', '{', '}', '[', ']', '(', ')', '"', '\\')
+  private val COMBO_CACHE_SIZE = 2048
```
alextheimer, line 12: Should this be configurable, too?
Good catch btw -- I wonder if this lit up in your flame plots, too 🤔

alextheimer: > and  QueryUtils.makeAllKeyValueCombos is high
Missed this bit 👏

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
amolnayak311: It doesn't compile., I will have to either use ``java.lang.Boolean`` or ``Some[scala.Boolean]``. Caffeine is a java library and it has the following signature 

`public final class Caffeine<K extends Object, V extends Object> `

coordinator/src/main/scala/filodb.coordinator/queryplanner/SingleClusterPlanner.scala
amolnayak311: Because its a ``Some[scala.Boolean]`` and the signature returns a ``scala.Boolean``

coordinator/src/test/scala/filodb.coordinator/queryplanner/SingleClusterPlannerSpec.scala
amolnayak311: Good point,  let me do that.

core/src/main/scala/filodb.core/query/QueryUtils.scala
amolnayak311: It's an Object and a util and doesn't have a config yet, passing it a config might be too much. I'll let it go as a constant for now and then we can extract a config if needed? I am open to have a different value.

coordinator/src/test/scala/filodb.coordinator/queryplanner/SingleClusterPlannerSpec.scala
amolnayak311: Pushed changes

alextheimer (approved):

PR #1875, opened 2024-10-31
fix(coordinator): update LogicalPlanParser to correctly handle scalar functions in queries

Tanner-Meng:
**Pull Request checklist**

**Current Behavior:**  
Scalar functions are excluded from queries on the `implicit-ws` endpoint but work as expected on the `global` endpoint.

**New Behavior:**  
Scalar functions are now correctly included in queries on the `implicit-ws` endpoint by updating LogicalPlanParser, consistent with `global`.


coordinator/src/test/scala/filodb.coordinator/queryplanner/LogicalPlanParserSpec.scala
```diff
@@ -148,6 +148,51 @@ class LogicalPlanParserSpec extends AnyFunSpec with Matchers {
     res shouldEqual "sum(rate(http_requests_total{job=\"app\"}[300s:60s] @100))"
   }
 
+  it("should correctly generate queries when LogicalPlan contains scalar() function") {
```
yu-shipit, line 151: can you add tests for query without any binary operators?

I mean no operators like `+-*/><`

for example
```
scalar(avg_over_time(http_requests_total[1h]))
```

yu-shipit (dismissed):

coordinator/src/test/scala/filodb.coordinator/queryplanner/LogicalPlanParserSpec.scala
Tanner-Meng: Understood.
I think binary operators are rarely used but could we keep other mathematical operators like `+-*/`?

Tanner-Meng: I've removed binary operators like `<>`

yu-shipit: oh < and > are valid operators.
I mean you had test cases for with binary operators.
And you also need to test the one without binary operators.

Since you added those. It looks good to me.

yu-shipit (approved):

PR #1882, opened 2024-11-07
fix(core): Error downsampled query if lookback is short when querying rate/increase

vishramachandran:
**Pull Request checklist**

Rate requires 2 samples in lookback window. Instead of returning empty result for rate in downsampled dataset (such queries with small window and smaller publish interval are encountered frequently in downsampled dataset), ~~we extend the lookback to two downsampled resolutions~~  we error the query out when lookback is insufficient. This fix is imperfect, but serves as a workaround until a better solution is delivered. Status quo affects customers by producing wrong query results and leads to thinking that there is data loss, or even worse, implicit trust that the query results are right.

### DIFF
--- a/query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
+++ b/query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
@@ -87,11 +87,11 @@ final case class PeriodicSamplesMapper(startMs: Long,
           val rdrv = rv.asInstanceOf[RawDataRangeVector]
           val chunkedHRangeFunc = rangeFuncGen().asChunkedH
           val minResolutionMs = rdrv.minResolutionMs
-          if (chunkedHRangeFunc.isInstanceOf[CounterChunkedRangeFunction[_]] && windowLength < minResolutionMs)
+          if (chunkedHRangeFunc.isInstanceOf[CounterChunkedRangeFunction[_]] && windowLength < minResolutionMs * 2)
             throw new IllegalArgumentException(s"Minimum resolution of data for this time range is " +
               s"${minResolutionMs}ms. However, a lookback of ${windowLength}ms was chosen. This will not " +
-              s"yield intended results for rate/increase functions since each lookback window contains " +
-              s"lesser than 2 samples. Increase lookback to more than ${minResolutionMs}ms")
+              s"yield intended results for rate/increase functions since each lookback window can contain " +
+              s"lesser than 2 samples. Increase lookback to more than ${minResolutionMs * 2}ms")
           val windowPlusPubInt = extendLookback(rv, windowLength)
           IteratorBackedRangeVector(rv.key,
             new ChunkedWindowIteratorH(rdrv, startWithOffset, adjustedStep, endWithOffset,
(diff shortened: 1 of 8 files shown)


query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
```diff
@@ -165,23 +166,32 @@ final case class PeriodicSamplesMapper(startMs: Long,
...
-//      case rvrd: RawDataRangeVector if (functionId.exists(_.onCumulCounter) && stepMultipleNotationUsed)
-//          => rvrd.publishInterval.getOrElse(0L)
-//      case _ => 0L
-//    }
-//    window + pubInt
+  private[exec] def extendLookback(rv: RangeVector, window: Long): Long = {
+    if (functionId.contains(InternalRangeFunction.Rate)) { // only extend for rate function
+      val pubInt = rv match {
+        case rvrd: RawDataRangeVector =>
+          if (rvrd.partition.schema == Schemas.promCounter ||
```
yu-shipit, line 180: should we add a function in schema?
isCumulative()?
In the future if we have more schema type people just have to implement this function.

vishramachandran: Good suggestion. Making this change

query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
```diff
@@ -87,32 +87,33 @@ final case class PeriodicSamplesMapper(startMs: Long,
...
+          val extendedWindow = extendLookback(rdrv, windowLength)
+          if (chunkedDRangeFunc.isInstanceOf[CounterChunkedRangeFunction[_]] &&
+              extendedWindow < rdrv.minResolutionMs * 2)
             throw new IllegalArgumentException(s"Minimum resolution of data for this time range is " +
               s"${minResolutionMs}ms. However, a lookback of ${windowLength}ms was chosen. This will not " +
-              s"yield intended results for rate/increase functions since each lookback window contains " +
-              s"lesser than 2 samples. Increase lookback to more than ${minResolutionMs}ms")
-          val windowPlusPubInt = extendLookback(rv, windowLength)
+              s"yield intended results for rate/increase functions since each lookback window can contain " +
+              s"lesser than 2 samples. Increase lookback to more than ${minResolutionMs * 2}ms")
```
amolnayak311, line 113: This is 4x of `rdrv.minResolutionMs`, the logic and the message in the two cases are different. I guess what we want is not to change ``minResolutionMs`` but introduce something like ``minExpected:Lookback = minResolutionMs * 2`` and use that in the condition and the exception

amolnayak311 (changes requested):

query/src/main/scala/filodb/query/exec/PeriodicSamplesMapper.scala
vishramachandran: Fixed

amolnayak311 (approved):

amolnayak311 (approved):

PR #1887, opened 2024-11-13
fix(query): Fix histogram query rewrite to exclude summaries

vishramachandran:
**Pull Request checklist**

Today we ingest summaries are three or more time series. For example, for a summary named `foo`: 
* One or more quantiles can be reported as time series with metric name `foo`. Today, this is ingested as a gauge in FiloDB.
* We have a sum reported at `foo_sum`. This is a counter.
* We have a `foo_count`. This is also a counter.
Details [in the original prom doc](https://prometheus.io/docs/concepts/metric_types/#summary)

Problem:
When `foo_count` is queried, we lookup to see if metric name `foo` exists. If it exists, we simply rewrite the query to `foo::count` which ends up querying the `count` column on `foo`. This is for backward compatibility on prometheus histograms.

In the case of summaries, the `count` column does not exist on `foo` and it leads to query failure.

Fix:
Rewrite the query only if the second lookup exists, and *additionally has a column called `count`*. This additional check will be true for histograms only.

Testing:
Existing unit tests in `MultiSchemaPartitionsExecSpec` cover this case.

### DIFF
--- a/query/src/main/scala/filodb/query/exec/MultiSchemaPartitionsExec.scala
+++ b/query/src/main/scala/filodb/query/exec/MultiSchemaPartitionsExec.scala
@@ -4,6 +4,7 @@ import kamon.Kamon
 import monix.execution.Scheduler
 
 import filodb.core.{DatasetRef, GlobalConfig}
+import filodb.core.memstore.PartLookupResult
 import filodb.core.metadata.Schemas
 import filodb.core.query.{ColumnFilter, QueryConfig, QueryContext, QuerySession, QueryWarnings}
 import filodb.core.query.Filter.Equals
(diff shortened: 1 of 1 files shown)


query/src/main/scala/filodb/query/exec/MultiSchemaPartitionsExec.scala
```diff
@@ -53,7 +56,15 @@ final case class MultiSchemaPartitionsExec(queryContext: QueryContext,
     // clear stats since previous call to lookupPartitions set the stat with metric name that has suffix
     querySession.queryStats.clear()
     val lookupRes = source.lookupPartitions(dataset, partMethod, chunkMethod, querySession)
-    (lookupRes, Some(columnName))
+    // Accept only if lookupResult schema has a data column with name as the requested columnName.
+    // This is true generally for histogram queries where we rewrite the query to remove _sum, _count suffix.
+    // Check is needed since summaries are not columnized, and we don't want the rewrite to happen for summaries.
+    // If column does not exist, return original lookup result and don't change the query
+    val schemas = source.schemas(datasetRef).get
+    if (lookupRes.firstSchemaId.exists(s => schemas.apply(s).data.columns.exists(_.name == columnName)))
```
sandeep6189, line 64: Not a blocker since we already do two full lookups: 

Curious if we can avoid the second "full" scan  (only scan for 1 or two ts) ? This will allow for the check before scanning the index for all entries in the 2nd lookup ?

sandeep6189 (approved):
thanks for taking this up @vishramachandran

query/src/main/scala/filodb/query/exec/MultiSchemaPartitionsExec.scala
vishramachandran: Fair suggestion. This requires a more involved change, best done in a different PR on a need basis.

PR #1910, opened 2024-12-10
feat(query): delete validation for "split" queries contain binary joins with offsets

Boyuan-Yu:
**Pull Request checklist**

**Current behavior :** 

- It will throw this exception because of validation.
 > invalid_query: This query contains selectors that individually read data from multiple partitions. This is likely because a selector's data was migrated between partitions. These "split" queries cannot contain binary joins with offsets.

- maxRemoteRawExportTimeRange is 3 days

**New behavior :**

- It handles the "split" query correctly.
- Increased the maxRemoteRawExportTimeRange to 31 days so that remote raw export is supported for longer days.


Boyuan-Yu:
test this please

Boyuan-Yu:
test this please

coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
```diff
@@ -603,7 +579,6 @@ class MultiPartitionPlanner(val partitionLocationProvider: PartitionLocationProv
                                        qContext: QueryContext): PlanResult = {
     // TODO: Reassess this validate, we should also support binary joins in split leaf as long as they are within
     //  the limits of max range of data exported
-    validateSplitLeafPlan(logicalPlan)
```
alextheimer, line 606: Can you remove the above comments as well?

core/src/main/scala/filodb.core/query/QueryConfig.scala
```diff
@@ -83,7 +83,7 @@ object QueryConfig {
 }
 case class RoutingConfig(
                           supportRemoteRawExport: Boolean                = false,
-                          maxRemoteRawExportTimeRange: FiniteDuration    = 3 days,
+                          maxRemoteRawExportTimeRange: FiniteDuration    = 31 days,
```
alextheimer, line 86: Leave this default as-is (better to configure this externally because it's unclear yet what new problems the change from 3d -> 31d might uncover).

coordinator/src/test/scala/filodb.coordinator/queryplanner/PlannerHierarchySpec.scala
```diff
@@ -2818,14 +2789,18 @@ class PlannerHierarchySpec extends AnyFunSpec with Matchers with PlanValidationS
...
-          plannerParams = PlannerParams(processMultiPartition = true)))
-      }
-    }
+    val lp = Parser.queryRangeToLogicalPlan(query, TimeStepParams(startSec, stepSec, endSec))
+    val promQlQueryParams = PromQlQueryParams(query, startSec, stepSec, endSec)
+    val execPlan = engine.materialize(lp,
+      QueryContext(origQueryParams = promQlQueryParams,
+        plannerParams = PlannerParams(processMultiPartition = true))
+    )
+    val expectedPlan =
```
alextheimer, line 2798: Could you try a query like:
```
sum(rate(test{_ws_ = "demo", _ns_ = "localNs"}[1h]))
+
sum(rate(test{_ws_ = "demo", _ns_ = "localNs"}[1h] offset 30m))
```

The left/right of the split are stitched correctly together, but we're looking for one additional stitch, as well, since some datapoints cannot be computed without values from both sides of the split.

core/src/main/scala/filodb.core/query/QueryConfig.scala
Boyuan-Yu: You are right. In Splunk, I've seen lots of query blocked by this validation, so we should leave as it is.

alextheimer (approved):

PR #1920, opened 2024-12-23
fix(query): Raw data export with StitchRvsExec correctly sets the `rawSource` value in PeriodicSamplesMapper RVT

sandeep6189:
**Pull Request checklist**

- [x] The commit(s) message(s) follows the contribution [guidelines](CONTRIBUTING.md) ?
- [x] Tests for the changes have been added (for bug fixes / features) ?
- [x] Docs have been added / updated (for bug fixes / features) ?

**Current behavior :**  Bug: `rawSource` in `PeriodicSamplesMapper` RVT, is incorrectly set to false leading to failed histogram queries during remote raw export

**New behavior :** rawSource is set to true for the  `PeriodicSamplesMapper` RVT allowing for queries to go through.

PR #1933, opened 2025-01-16
fix(query): (Revert): Raw data export with StitchRvsExec correctly sets the `rawSource` value in PeriodicSamplesMapper RVT"

sandeep6189:
Reverts filodb/FiloDB#1920. This is needed to fix the integ tests and PRB.


alextheimer (approved):
Could you add a quick one-liner to the header about why this is needed?

PR #1943, opened 2025-01-31
fix(memstore): account for clock drift during latency measurement

alextheimer:
**Pull Request checklist**

"Negative" ingestion latencies i.e. where a `RecordContainer`'s timestamp occurs after the current time) might occur as a result of clock skew.

Currently, negative ingestion latencies will cause Kamon histograms to throw exceptions.

This PR adds one additional metric `negativeIngestionPipelineLatency` to record negative pipeline latencies (as their absolute values). Additionally, values recorded into the two usual ingestion latency metrics are now wrapped inside `Math.max(0, value)` calls.

### DIFF
--- a/core/src/main/scala/filodb.core/RateLimiter.scala
+++ b/core/src/main/scala/filodb.core/RateLimiter.scala
@@ -21,7 +21,7 @@ class RateLimiter(period: Duration) {
    */
   def attempt(): Boolean = {
     val nowMillis = System.currentTimeMillis()
-    if (System.currentTimeMillis() - lastSuccessMillis > period.toMillis) {
+    if (nowMillis - lastSuccessMillis > period.toMillis) {
       lastSuccessMillis = nowMillis
       return true
     }
--- a/core/src/main/scala/filodb.core/RateLimiter.scala
+++ b/core/src/main/scala/filodb.core/RateLimiter.scala
@@ -1,0 +1,30 @@
+package filodb.core
+
+import scala.concurrent.duration.Duration
+
+/**
+ * Rate-limiter utility.
+ * @param period a "successful" attempt will be indicated only after a full
+ *               period has elapsed since the previous success.
+ */
+class RateLimiter(period: Duration) {
+  var lastSuccessMillis = 0L;
+
+  /**
+   * Returns true to indicate an attempt was "successful", else it was "failed".
+   * Successes are returned only after a full period has elapsed since the previous success.
+   *
+   * NOTE: this operation is not thread-safe, but if at least one concurrent invocation is
+   *   successful, then one of the successful timestamps will be recorded internally as the
+   *   most-recent success. In practice, this means async calls may succeed in bursts (which
+   *   may be acceptable in some use-cases).
+   */
+  def attempt(): Boolean = {
+    val nowMillis = System.currentTimeMillis()
+    if (System.currentTimeMillis() - lastSuccessMillis > period.toMillis) {
+      lastSuccessMillis = nowMillis
+      return true
+    }
+    false
+  }
+}
--- a/core/src/main/scala/filodb.core/RateLimiter.scala
+++ b/core/src/main/scala/filodb.core/RateLimiter.scala
@@ -8,7 +8,7 @@ import scala.concurrent.duration.Duration
  *               period has elapsed since the previous success.
  */
 class RateLimiter(period: Duration) {
-  var lastSuccessMillis = 0L;
+  private var lastSuccessMillis = 0L;
 
   /**
    * Returns true to indicate an attempt was "successful", else it was "failed".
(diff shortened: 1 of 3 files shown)


core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -1301,7 +1307,18 @@ class TimeSeriesShard(val ref: DatasetRef,
...
+
+    // Account for clock skew, otherwise exceptions are thrown when negative latencies are recorded.
+    val ingestionPipelineLatency = currentTime - container.timestamp
+    shardStats.ingestionPipelineLatency.record(Math.max(0, ingestionPipelineLatency))
+    // It would be helpful to know if latencies are frequently negative;
+    //   record their absolute values into a separate histogram.
+    // We'll continue to count floored-to-zero values in the above metric; this will
+    //   keep e.g. its rate() as-expected when latencies are negligibly negative.
+    if (ingestionPipelineLatency < 0) {
+      shardStats.negativeIngestionPipelineLatency.record(-ingestionPipelineLatency)
```
srivik, line 1319: Do you think adding a log message when negative latencies are detected would be helpful with troubleshooting?

srivik (approved):

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -1301,7 +1307,18 @@ class TimeSeriesShard(val ref: DatasetRef,
     }
 
     val currentTime = System.currentTimeMillis()
-    shardStats.ingestionPipelineLatency.record(currentTime - container.timestamp)
+
+    // Account for clock skew, otherwise exceptions are thrown when negative latencies are recorded.
```
sandeep6189, line 1311: Interesting problem (even more difficult to solve). I am guessing the currentTimeMillis is epoch time ?
Good idea to track it with metric. This is probably a task for kube team to arrest the drift, but we might need to provide evidence for the same. Should we log some instances of currentTime as well as container.timestamp to have concrete evidence for them ?

amolnayak311: It might be an expected behavior with the clocks getting synced periodically. I am assuming this dpwsry happen very frequently.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
amolnayak311: Without additional details in the logs, I am not sure what could be done especially if this isn't very frequent.

amolnayak311 (approved):

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
alextheimer: I avoided logs before because if, for some reason-- not necessarily clock skew--  all containers show a negative ingestion latency, there might be some significant overhead involved to log everything. The separate metric was added mostly as a lightweight confirmation that the problem exists, and it would expose the scale of the problem as well (How negative? How frequent?).

I've now added a `RateLimiter` utility that allows logs once every 10s (hard-coded, but I can make that configurable if preferred. Maybe `rate-limited-logs-period`?). It is not thread-safe, but it is good enough here (see the PR for details). The other options are:
- `synchronize` on the `RateLimiter` or track the most-recent `RateLimiter.attempt()` success internally as an `AtomicLong` (but both will introduce some contention on the ingestion path).
- Use a `ThreadLocal[RateLimiter]`, which feels like overkill.

core/src/main/scala/filodb.core/RateLimiter.scala
```diff
@@ -0,0 +1,30 @@
...
+   * Successes are returned only after a full period has elapsed since the previous success.
+   *
+   * NOTE: this operation is not thread-safe, but if at least one concurrent invocation is
+   *   successful, then one of the successful timestamps will be recorded internally as the
+   *   most-recent success. In practice, this means async calls may succeed in bursts (which
+   *   may be acceptable in some use-cases).
+   */
+  def attempt(): Boolean = {
+    val nowMillis = System.currentTimeMillis()
+    if (System.currentTimeMillis() - lastSuccessMillis > period.toMillis) {
```
sandeep6189, line 24: `nowMillis - lastSuccessMillis` ?

We can avoid an additional sys call here.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
sandeep6189: log volume is a genuine concern. Thank you for addressing this. We can be more conservation and even 15-30s should be enough I feel ( since we just want evidence of it happening, and the frequency of logging is a lesser concern ).

sandeep6189 (approved):

core/src/main/scala/filodb.core/RateLimiter.scala
alextheimer: Whoops, that slipped through -- good catch

sandeep6189 (approved):

PR #1944, opened 2025-02-04
feat(query) Range function to identify outliers using MAD

vishramachandran:
**Pull Request checklist**

New range function `last_over_time_is_mad_outlier()` which classifies the last data point in the range window as an anomaly  or not based on other points in the window.

This is especially useful in subqueries. This query executes a subquery and calculates outliers in request rate for a service. The outlier points are returned as part of the results.
```
last_over_time_is_mad_outlier(3, sum(rate(http_req_total{...}[5m]))[1d:5m])
```


amolnayak311:
Thanks @vishramachandran for the PR, QQ, is it possible to use `mad_over_time`  to achieve this? The challenge is MAD is not the only approach and we will end up implementing multiple` last_over_time_is_<>_outlier` if others need to be supported.

sherali42 (approved):

PR #1948, opened 2025-02-07
fix(memstore): fix flaky rate-limiter test

alextheimer:
**Pull Request checklist**

New `RateLimiter` tests are flaky. Two possible resons:
- `System.currentTimeMillis()` is [random enough](https://stackoverflow.com/a/2979239) that the tests fail.
- One of the test’s `Thread.sleep()`s was too short prior to a validation.

This PR addresses both possible problems (and will be merged only after 5 consecutive PRB successes).

### DIFF
--- a/core/src/main/scala/filodb.core/RateLimiter.scala
+++ b/core/src/main/scala/filodb.core/RateLimiter.scala
@@ -20,7 +20,7 @@ class RateLimiter(period: Duration) {
    *   may be acceptable in some use-cases).
    */
   def attempt(): Boolean = {
-    val nowMillis = System.currentTimeMillis()
+    val nowMillis = Math.max(System.currentTimeMillis(), lastSuccessMillis)
     if (nowMillis - lastSuccessMillis > period.toMillis) {
       lastSuccessMillis = nowMillis
       return true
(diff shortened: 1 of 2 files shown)


alextheimer:
PRB run 1 successful.

alextheimer:
PRB run 2 successful.

alextheimer:
PRB run 3 successful.

core/src/main/scala/filodb.core/RateLimiter.scala
```diff
@@ -20,7 +20,8 @@ class RateLimiter(period: Duration) {
    *   may be acceptable in some use-cases).
    */
   def attempt(): Boolean = {
-    val nowMillis = System.currentTimeMillis()
+    // Wrapping in a max() because System.currentTimeMillis() is not monotonic.
+    val nowMillis = Math.max(System.currentTimeMillis(), lastSuccessMillis)
```
amolnayak311, line 24: Use ```System.nanoTime```

amolnayak311 (changes requested):

core/src/main/scala/filodb.core/RateLimiter.scala
alextheimer: I generally avoid that because I know there is a performance cost, but I've made the change and removed the `max` (assumption is that the cost difference will be negligible in practice).

alextheimer:
PR run 5 successful.

amolnayak311 (approved):

PR #1964, opened 2025-03-06
perf(query) Optimize RepeatValueVector to not store the key and the first item in the data eagerly

amolnayak311:
**Pull Request checklist**

**Current behavior :** 

``RepeatValueVector``  calls ``next`` on the underlying transformer and stores the value of the first item as part of it. This might be unnecessary and the call is deferred to the point when ``rows`` is invoked. Currently both the key and the first item of the underlying  RV (as ``TransientRow``) is stored in the instance of RepeatValueVector.

**New behavior :**

We do not invoke next or keys on the wrapped RangeVector until the rows are requested.

**BREAKING CHANGES :**

`RepeatValueVector` no longer implements `SerializableRangeVector` This means , any deserialization of this vector will arrive over the wire as a SerializableRangeVector (which would have consumed ``RepeatValueVector`` on the producer side ) and the not as ``RepeatValueVector`` itself.  The change in signature of ``RepeatValueVector`` which now accepts a ``RangeVector`` which may or mayn't be serializable, requires us to consume this range vector on the producer side and send as a ``SerializedRangeVector``.

### DIFF
--- a/core/src/main/scala/filodb.core/query/RangeVector.scala
+++ b/core/src/main/scala/filodb.core/query/RangeVector.scala
@@ -209,18 +209,22 @@ final case class RangeParams(startSecs: Long, stepSecs: Long, endSecs: Long)
  * @param rowReader the row read that provide the value.
  * @param schema the schema.
  */
-final class RepeatValueVector(rangeVectorKey: RangeVectorKey,
+final case class RepeatValueVector(rv: RangeVector,
                               startMs: Long, stepMs: Long, endMs: Long,
-                              rowReader: Option[RowReader],
                               schema: RecordSchema) extends SerializableRangeVector with StrictLogging {
   override def outputRange: Option[RvRange] = Some(RvRange(startMs, stepMs, endMs))
   override val numRows: Option[Int] = Some((endMs - startMs) / math.max(1, stepMs) + 1).map(_.toInt)
+
+  lazy val key: RangeVectorKey = rv.key
 
   // This will now be used only during serialization of this vector in protos
   // This is room for optimization here as we cant instantiate anything under 2048 bytes which is the MinContainerSize
   def containers: Seq[RecordContainer] = {
     val builder = new RecordBuilder(MemFactory.onHeapFactory, RecordBuilder.MinContainerSize)
-    rowReader.map(builder.addFromReader(_, schema, 0))
+    val rvCursor = rows()
+    if (rvCursor.hasNext) {
+      builder.addFromReader(rvCursor.next(), schema, 0)
+    }
     builder.allContainers.toList
   }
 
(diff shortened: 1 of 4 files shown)


query/src/main/scala/filodb/query/exec/RangeVectorTransformer.scala
```diff
@@ -340,11 +340,9 @@ final case class RepeatTransformer(startMs: Long, stepMs: Long, endMs: Long, exe
             limit: Int,
             sourceSchema: ResultSchema,
             paramResponse: Seq[Observable[ScalarRangeVector]]): Observable[RepeatValueVector] = {
+    val rs = new RecordSchema(sourceSchema.columns, brSchema = sourceSchema.brSchemas)
```
amolnayak311, line 343: Instantiated once to avoid repeated instantiation of this same times as the number of range vectors.

core/src/main/scala/filodb.core/query/RangeVector.scala
```diff
@@ -266,47 +271,28 @@ final class RepeatValueVector(rangeVectorKey: RangeVectorKey,
...
-object RepeatValueVector extends StrictLogging {
-  import filodb.core._
-  def apply(rv: RangeVector,
-            startMs: Long, stepMs: Long, endMs: Long,
-            schema: RecordSchema,
-            execPlan: String,
-            queryStats: QueryStats): RepeatValueVector = {
-    val startNs = Utils.currentThreadCpuTimeNanos
-    try {
-      ChunkMap.validateNoSharedLocks(execPlan)
```
yu-shipit, line 298: should we keep this lock?

yu-shipit (dismissed):

query/src/main/scala/filodb/query/exec/RangeVectorTransformer.scala
```diff
@@ -340,11 +340,9 @@ final case class RepeatTransformer(startMs: Long, stepMs: Long, endMs: Long, exe
...
     source.map { rv =>
-        RepeatValueVector(rv, startMs, stepMs, endMs,
-          new RecordSchema(sourceSchema.columns, brSchema = sourceSchema.brSchemas),
-          execPlan,
-          querySession.queryStats)
```
sandeep6189, line 347: quick question: Previously we were tracking the time taken to create the `RepeatValueVector` object using the `queryStats.CpuNanosCounter`. Wondering if we still need to do it or it will be tracked under the caller itself ?

core/src/main/scala/filodb.core/query/RangeVector.scala
sandeep6189: @yu-shipit curious on why this lock was needed ? was it for iterating over the RV ?

sandeep6189 (approved):
Awesome optimization @amolnayak311 !!

query/src/main/scala/filodb/query/exec/RangeVectorTransformer.scala
amolnayak311: That was done when we iterated though the RV to get the first item, now that we just instantiate I removed that bit.

core/src/main/scala/filodb.core/query/RangeVector.scala
amolnayak311: Its needed for SerializableRangeVectors but I am not sure if this vector still needs to be a SerializableRangeVector?

yu-shipit (commented):
Thank you for the deep investigation and fix.
I also learned a lot from this.

yu-shipit (approved):
Thank you for the deep investigation and fix.
I also learned a lot from this.

PR #1969, opened 2025-03-12
feat(config): Enable type field indexing and query circuit breaker by default

vishramachandran:
**Pull Request checklist**

Type field indexing is tested in production, so enabling by default.
Also enabling query circuit breaker to protect system when under distress.

### DIFF
--- a/core/src/main/resources/filodb-defaults.conf
+++ b/core/src/main/resources/filodb-defaults.conf
@@ -448,7 +448,7 @@ filodb {
 
     circuit-breaker {
       # query circuit breaker enabled or not
-      enabled = false
+      enabled = true
       # open the circuit when number of failures reaches this count
       open-when-num-failures = 5
       # move to half-open after this time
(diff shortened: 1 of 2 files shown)


whizkido:
This would nice. do we have measurement for how many times the circuit breaker is invoked.

core/src/main/resources/filodb-defaults.conf
```diff
@@ -448,7 +448,7 @@ filodb {
 
     circuit-breaker {
       # query circuit breaker enabled or not
-      enabled = false
+      enabled = true
```
amolnayak311, line 451: We don't have data to see enable this by default, let's not enable this feature.

amolnayak311 (changes requested):

core/src/main/resources/filodb-defaults.conf
vishramachandran: What data do we need? And why can't we collect this data in a perf test during this release ?

core/src/main/resources/filodb-defaults.conf
```diff
@@ -858,7 +858,7 @@ filodb {
     disable-index-caching = false
 
     # Whether to add the _type_ label to all time series for the purpose of filtering
-    type-field-indexing-enabled = false
+    type-field-indexing-enabled = true
```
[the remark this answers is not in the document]
vishramachandran: It will override the value sent with our schema string. `_type_` is a reserved label by convention.

vishramachandran:
Closing since the team is not ready to do this now.

PR #1970, opened 2025-03-12
perf(query) Remove unnecessary serialization if the execution is local

amolnayak311:
**Pull Request checklist**

**Current behavior :** 

The makeResult converts the ``RangeVector``  to a ``SerializableRangeVector`` if it's not one. This prevents us from benefiting from target schema optimization where both LHS and RHS are on the same shard.  We end up serializing the range vector even if we don't have to

**New behavior :**

We check the ``dispatcher`` in the `` ExecPlan`` and if it is a local, we don't convert the ``RangeVector`` to ``SerializedRangeVector``


query/src/main/scala/filodb/query/exec/ExecPlan.scala
```diff
@@ -458,17 +461,22 @@ trait ExecPlan extends QueryCommand {
...
+                  numResultSamples += srv.numRowsSerialized
+                  checkSamplesLimit(numResultSamples, querySession.warnings)
+                  val srvBytes = srv.estimatedSerializedBytes
+                  resultSize += srvBytes
+                  querySession.queryStats.getResultBytesCounter(Nil).addAndGet(srvBytes)
+                  checkResultBytes(resultSize, querySession.queryConfig, querySession.warnings)
+                  srv
+            case rv: RangeVector           => rv
+          }
+          .filter {
```
srivik, line 476: Should filtering be moved to .map ?

amolnayak311: You mean before `.map`?

amolnayak311: That won't work as we need to also have another case

`case srv: SerializedRangeVector if srv.numRowsSerialized <= 0`

as `.map` will give one rv for every rv it sees and can't filter.  To combine, we might have to do something like a `fold`

query/src/main/scala/filodb/query/exec/ExecPlan.scala
```diff
@@ -458,17 +467,22 @@ trait ExecPlan extends QueryCommand {
...
-              checkResultBytes(resultSize, querySession.queryConfig, querySession.warnings)
-              srv
+                  numResultSamples += srv.numRowsSerialized
+                  checkSamplesLimit(numResultSamples, querySession.warnings)
+                  val srvBytes = srv.estimatedSerializedBytes
+                  resultSize += srvBytes
+                  querySession.queryStats.getResultBytesCounter(Nil).addAndGet(srvBytes)
+                  checkResultBytes(resultSize, querySession.queryConfig, querySession.warnings)
+                  srv
+            case rv: RangeVector           => rv
```
alextheimer, line 480: Thinking out loud: does this mean a query like this might not breach any sample-based limits?
```
sum by (_node_) (
    sum_over_time(foo{...}[1000y])
)
```

That would probably be OK if the `dataBytesScanned` limit were tripped, instead, but It looks like that is enforced only in `MultiSchemaPartitionsExec`'s override of `checkResultBytes`-- which Is run only here in this SRV-specific case.

Not 100% sure, but it looks like some obscenely-long-range queries might not trip any limits as long as e.g. series cardinality is low throughout the query range.

amolnayak311: ``dataBytesScanned`` is unaffected. The ``resultBytes`` (bytes serialized) and  ``numResultSamples`` limits won't be breached. IMO, both these are relevant if we move data around, but with target schema, this move is not happening and thus penalizing the query for breaching these limits is not right. The ``dataBytesScanned`` however since unaffected by this change might still be violated.

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
```diff
@@ -106,16 +106,23 @@ final case class SetOperatorExec(queryContext: QueryContext,
...
+                       rhsSchema: ResultSchema, querySession: QuerySession): Iterator[RangeVector] = {
+
+    lazy val builder = SerializedRangeVector.newBuilder()
+    lazy val recSchema = SerializedRangeVector.toSchema(rhsSchema.columns, rhsSchema.brSchemas)
+    lazy val execPlanName = queryWithPlanName(queryContext)
+    val mappedRhsRvs = rhsRvs.map {
+      case srv: SerializedRangeVector  => srv
+      case rv: RangeVector               => SerializedRangeVector.apply(rv, builder,
+                                            recSchema, execPlanName, querySession.queryStats)
+    }
```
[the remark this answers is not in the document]
amolnayak311: Thanks for the second pair of eyes.

query/src/test/scala/filodb/query/exec/ExecPlanSpec.scala
```diff
@@ -216,4 +256,140 @@ class ExecPlanSpec extends AnyFunSpec with Matchers with ScalaFutures {
...
+      case error: QueryError =>
+        assert(false, "expected QueryResult got " + error)
+    }
+
+    // Same call with preventRangeVectorSerialization=true should not give a SerializedRangeVector
+    val plan1 = makeFixedLeafExecPlan(Seq(rv), schema)
+    plan1.execute(memStore, querySession.copy(preventRangeVectorSerialization=true)).runToFuture.futureValue match {
+      case res: QueryResult  =>
+        val rv = res.result.head
+        // By default we expect a SerializableRangeVector
```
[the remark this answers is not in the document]
amolnayak311: Good catch.

query/src/main/scala/filodb/query/exec/ExecPlan.scala
```diff
@@ -679,9 +693,16 @@ abstract class NonLeafExecPlan extends ExecPlan {
     // across threads. Note that task/observable will not run on the thread where span is present since
     // kamon uses thread-locals.
     // Dont finish span since this code didnt create it
+    // Prevent serialization if one of the two conditions happen
+    // 1. The querySession already has the preventRangeVectorSerialization set to true
+    // 2. If the dispatcher is an inProcessDispatcher and the config has enabled this feature
+    val preventRangeVectorSerialization = qSession.preventRangeVectorSerialization ||
+                                          (dispatcher.isLocalCall
+                                              && qSession.queryConfig.enableLocalDispatch)
```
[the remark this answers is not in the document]
amolnayak311: Thanks made the change

alextheimer (approved):

PR #1972, opened 2025-03-14
perf(memstore) TimeRangeChunkScan optimization when querying data in current ingesting chunk

amolnayak311:
**Pull Request checklist**

**Current behavior :** 

When ``TimeRangeChunkScan`` is provided, all chunkInfos are iterated and the relevant ``ChunkSetInfos`` are filtered that contains the data for the given range. The change will avoid this scan and first check if the requested data is available in the currentInfo and return that one chunk of interest. 

**New behavior :**

Just return the ``currentInfo`` if thats where all the requested data range lies.

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
+++ b/core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
@@ -336,9 +336,19 @@ extends ChunkMap(initMapSize) with ReadablePartition {
   def infos(method: ChunkScanMethod): ChunkInfoIterator = method match {
     case AllChunkScan        => allInfos
     case InMemoryChunkScan   => allInfos
-    case r: TimeRangeChunkScan => allInfos.filter { ic =>
-                                    ic.intersection(r.startTime, r.endTime).isDefined
-                                  }
+    case r: TimeRangeChunkScan => if (currentInfo != nullInfo
+                                      && r.startTime <= r.endTime
+                                      && currentInfo.startTime <= r.startTime) {
+                                     try {
+                                        new OneChunkInfo(currentInfo)
+                                     } catch {
+                                        case e: Throwable => chunkmapReleaseShared(); throw e;
+                                     }
+                                 } else {
+                                    allInfos.filter { ic =>
+                                      ic.intersection(r.startTime, r.endTime).isDefined
+                                    }
+                                }
     case WriteBufferChunkScan => if (currentInfo == nullInfo) ChunkInfoIterator.empty
                                 else {
                                   // Return a single element iterator which holds a shared lock.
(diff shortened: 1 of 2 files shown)


core/src/test/scala/filodb.core/memstore/TimeSeriesPartitionSpec.scala
```diff
@@ -656,4 +657,23 @@ class TimeSeriesPartitionSpec extends MemFactoryCleanupTest with ScalaFutures {
...
+    val data = singleSeriesReaders().take(11)
+    val minData = data.map(_.getDouble(1))
+    val initTS = data(0).getLong(0)
+    data.take(10).zipWithIndex.foreach { case (r, i) => part.ingest(0, r, ingestBlockHolder,
+      flushIntervalMillis = flushIntervalMillis,
+      acceptDuplicateSamples = acceptDuplicateSamples,
+      createChunkAtFlushBoundary = timeAlignedChunksEnabled)
+    }
+    val infos = part.infos(TimeRangeChunkScan(initTS, initTS + 100000000))
+    assert(!infos.isInstanceOf[FilteredChunkInfoIterator])
```
alextheimer, line 672: Might be worth asserting that it's a `OneChunkInfo`

alextheimer (dismissed):
LGTM -- just one non-blocking comment 👍

core/src/test/scala/filodb.core/memstore/TimeSeriesPartitionSpec.scala
amolnayak311: I wanted to but it's a private class and I will have to change the visibility so left it there. I also thought of having  string search based assertion on the class name's toString but thought that wasn't the best way.

srivik (dismissed):

core/src/main/scala/filodb.core/memstore/TimeSeriesPartition.scala
```diff
@@ -336,9 +336,21 @@ extends ChunkMap(initMapSize) with ReadablePartition {
...
-                                  }
+    case r: TimeRangeChunkScan =>
+                                     if (currentInfo != nullInfo
+                                      && r.startTime <= r.endTime
+                                      && currentInfo.startTime <= r.startTime) {
+                                     try {
+                                        chunkmapAcquireShared()
+                                        new OneChunkInfo(currentInfo)
+                                     } catch {
+                                        case e: Throwable => chunkmapReleaseShared(); throw e;
```
yu-shipit, line 347: should this be in the finally clause?
Because acquire and release do not match in most cases.

amolnayak311: No, Whoever get the iterator closes it. Its kind of unintuitive as not abstracted but thats how it is unfortunately. We have a similar code. in ``ChunkMap.scala``

```
 final def chunkmapIterate(): ElementIterator = {
    new LazyElementIterator(() => {
      chunkmapAcquireShared()
      try {
        new MapIterator(first, first + size)
      } catch {
        case e: Throwable => chunkmapReleaseShared(); throw e;
      }
    })
  }
```

sherali42 (approved):

PR #1978, opened 2025-03-24
perf(query) Use the last chunk and avoid scan if the data is availlable in the latest ingesting chunk

amolnayak311:
**Pull Request checklist**

When TimeRangeChunkScan is provided, all chunkInfos are iterated and the relevant ChunkSetInfos are filtered that contains the data for the given range. The change will avoid this scan and first check if the requested data is available in the currentInfo and return that one chunk of interest.

New behavior :

Just return the currentInfo if thats where all the requested data range lies.

### DIFF
--- a/memory/src/main/scala/filodb.memory/data/ChunkMap.scala
+++ b/memory/src/main/scala/filodb.memory/data/ChunkMap.scala
@@ -220,6 +220,19 @@ abstract class ChunkMap(var capacity: Int) {
     arrayGet(realIndex(first + size - 1))
   }
 
+
+  final def chunkmapDoGetLastIterator(): ElementIterator = {
+    new LazyElementIterator(() => {
+      chunkmapAcquireShared()
+      try {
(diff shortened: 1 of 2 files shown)


memory/src/main/scala/filodb.memory/data/ChunkMap.scala
```diff
@@ -220,6 +220,19 @@ abstract class ChunkMap(var capacity: Int) {
     arrayGet(realIndex(first + size - 1))
   }
 
+
+  final def chunkmapDoGetLastIterator(): ElementIterator = {
```
vishramachandran, line 224: Can you add a unit test in `ChunkMapTest`  that exercises the code and validates locks ?

amolnayak311: Done, the code path is getting used in the new test and the fact that `chunkmapFree` at the end which acquires exclusive lock works validated that no locks are help.

vishramachandran (approved):

PR #1990, opened 2025-04-10
test(core): switching to part keys v2 table and minor serialization test

kvpetrov:
**Pull Request checklist**

- [ X ] The commit(s) message(s) follows the contribution [guidelines](CONTRIBUTING.md) ?
- [ X ] Tests for the changes have been added (for bug fixes / features) ?

Unit tests are run with the legacy style of partkeys C* table

**New behavior :**
Unit tests are run with the parkeys C* table v2

**BREAKING CHANGES**
n/a

### DIFF
--- a/core/src/test/resources/application_test.conf
+++ b/core/src/test/resources/application_test.conf
@@ -48,7 +48,7 @@ filodb {
     create-tables-enabled = true
     num-token-range-splits-for-scans = 2
     index-scan-parallelism-per-shard = 2
-    part-keys-v2-table-enabled = false
+    part-keys-v2-table-enabled = true
     pk-v2-table-num-buckets = 100
   }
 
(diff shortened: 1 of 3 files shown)


sandeep6189 (dismissed):

core/src/test/resources/application_test.conf
```diff
@@ -48,7 +48,7 @@ filodb {
     create-tables-enabled = true
     num-token-range-splits-for-scans = 2
     index-scan-parallelism-per-shard = 2
-    part-keys-v2-table-enabled = false
+    part-keys-v2-table-enabled = true
```
alextheimer, line 51: Should the corresponding [config](https://github.com/filodb/FiloDB/blob/1cc39d9baef00610f9e72aad03e86f0c967e17d0/core/src/main/resources/filodb-defaults.conf#L548) in filo-defaults also be updated since all tests going forward will use v2 tables?

kvpetrov: I don't want to change the defaults, the reason is that in that case the production clusters will pick up the change but it should be applied only to NEW clusers. In other words, the partitions that are already built will stay on the good old table structure as we don't have the way to transfer the data.

alextheimer (approved):

PR #2012, opened 2025-05-21
feat(query): update unless Operator to compute each row in RV

Boyuan-Yu:
**Pull Request checklist**

**Current behavior :** 
The unless operator only evaluated the join key and ignored per-row timestamps.
Example:
`mymetric{} unless (mymetric{} offset 1m)  `
This would return an empty result because the keys match, regardless of timestamp differences.

**New behavior :**
The unless operator now considers timestamps for each row during evaluation.
Example:
`mymetric{} unless (mymetric{} offset 1m) ` 
This will return data points from the initial 1 minute where mymetric has no corresponding matches 1 minute in the past.

**Additional Fixes**
Fixed a utility function `assertListEquals` used in unit tests that previously gave false positives.

### DIFF
--- a/query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
+++ b/query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
@@ -73,7 +73,8 @@ final case class SetOperatorExec(queryContext: QueryContext,
           case LAND => val rhsSchema = if (rhsResp.nonEmpty) rhsResp.head._1 else ResultSchema.empty
             setOpAnd(lhsRvs, rhsRvs, rhsSchema, querySession)
           case LOR => setOpOr(lhsRvs, rhsRvs)
-          case LUnless => setOpUnless(lhsRvs, rhsRvs)
+          case LUnless => val rhsSchema = if (rhsResp.nonEmpty) rhsResp.head._1 else ResultSchema.empty
+            setOpUnless(lhsRvs, rhsRvs, rhsSchema, querySession)
           case _ => throw new IllegalArgumentException("requirement failed: Only and, or and unless are supported ")
         }
         // check for timeout after dealing with metadata, before dealing with numbers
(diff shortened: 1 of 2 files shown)


query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
```diff
@@ -375,31 +376,107 @@ final case class SetOperatorExec(queryContext: QueryContext,
...
+    }
+
+    // 3) Stitch shards of RHS into a single RV per join-key
+    val rhsMap  = new mutable.HashMap[Map[Utf8Str, Utf8Str], RangeVector]()
+    mappedRhs.foreach { rv =>
       val jk = joinKeys(rv.key)
-      rhsKeysSet += jk
+      if (!isEmpty(rv, rhsSchema)) {
+        if (rhsMap.contains(jk)) {
+          val resVal = rhsMap(jk)
```
alextheimer, line 405: What do you think about `joinKeyRv`?

alextheimer, line 405: Since these join keys could be huge, it might be better to do one single map access (i.e. instead of a hash/equals once for `contains` and once for `rhsMap(jk)`).

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
```diff
@@ -375,31 +376,107 @@ final case class SetOperatorExec(queryContext: QueryContext,
...
+        // append or update into the ArrayBuffer for this join‐key
+        val buf   = result.getOrElse(jk, ArrayBuffer.empty[RangeVector])
+        val outRv = IteratorBackedRangeVector(lhs.key, cursor, period)
+        if (index >= 0) buf.update(index, outRv) else buf.append(outRv)
+        result.put(jk, buf)
+
+      } else {
+        // no RHS for this key ⇒ emit LHS (and stitch duplicates)
+        val buf = result.getOrElseUpdate(jk, ArrayBuffer.empty[RangeVector])
+        buf.indexWhere(r => r.key.labelValues == lhs.key.labelValues) match {
```
alextheimer, line 466: When will an LHS series already exist in the buffer?

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
```diff
@@ -375,31 +376,107 @@ final case class SetOperatorExec(queryContext: QueryContext,
...
+    val period = lhsRvs.headOption.flatMap(_.outputRange)
+
+    // 5) For each LHS, either emit unchanged (if no RHS) or do the “unless”‐zip
+    lhsRvs.foreach { lhs =>
+      val jk = joinKeys(lhs.key)
+
+      if (rhsMap.contains(jk)) {
+        var index = -1
+        val lhsStitched =
+          if (result.contains(jk)) {
```
alextheimer, line 426: When will the result contain LHS join keys? Er, more precisely: when will the result ever contain something other than keys as-is in from the LHS?

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
```diff
@@ -375,31 +376,107 @@ final case class SetOperatorExec(queryContext: QueryContext,
...
+          override def close(): Unit = {
+            lhsRows.close()
+            rhsRows.close()
           }
-          result(lhs.key.labelValues) = stitchedOrNewValue
         }
-        accumulated
+
+        // append or update into the ArrayBuffer for this join‐key
+        val buf   = result.getOrElse(jk, ArrayBuffer.empty[RangeVector])
```
alextheimer, line 458: Should this access `lhs.key`, instead?

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
```diff
@@ -375,31 +376,107 @@ final case class SetOperatorExec(queryContext: QueryContext,
...
+
+    // 3) Stitch shards of RHS into a single RV per join-key
+    val rhsMap  = new mutable.HashMap[Map[Utf8Str, Utf8Str], RangeVector]()
+    mappedRhs.foreach { rv =>
       val jk = joinKeys(rv.key)
-      rhsKeysSet += jk
+      if (!isEmpty(rv, rhsSchema)) {
+        if (rhsMap.contains(jk)) {
+          val resVal = rhsMap(jk)
+          if (resVal.key.labelValues == rv.key.labelValues) {
```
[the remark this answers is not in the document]
Boyuan-Yu: This is the same piece of code with `setOpAnd`, should we fix that in this PR as well? - according to comments there:

> TODO: isEmpty is expensive and iterates all the rows, we need a performant implementation, we may very well not  keep this check and it will do less iterations in case where we dont have several empty Range vectors
When spread changes, we need to account for multiple Range Vectors with same key coming from different shards Each of these range vectors would contain data for different time ranges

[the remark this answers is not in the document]
alextheimer: I see. So, if it has a join key that might be shared by other series, we do the stitch.

We won't update the others, but let's set the standard in this PR (and see if any of the `unless` tests break): just use `if (rhsMap.isPresent())` here because that makes the logic easier to understand.

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
```diff
@@ -375,31 +376,107 @@ final case class SetOperatorExec(queryContext: QueryContext,
...
+            rhsMap.put(jk, StitchRvsExec.stitch(rv, resVal, outputRvRange))
+          } else {
+            rhsMap.put(jk, rv)
```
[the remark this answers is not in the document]
alextheimer: In the `unless` case: let's consider a join key as "present" at a timestamp if any matching series has _any_ (`NaN` or otherwise) value at that timestamp. That is [consistent with prometheus](https://demo.promlabs.com/query?g0.expr=up+unless+%28up+%2B+NaN%29&g0.show_tree=0&g0.tab=table&g0.range_input=1h&g0.res_type=auto&g0.res_density=medium&g0.display_mode=lines&g0.show_exemplars=0).

We'll skip the `else` and always `stitch`. What do you think?

One open question I can think of is: Will we ever run into something like this, where each series has the same join key?
```
series 1: ---
series 2:       ---
series 3:    ---
```
or this?
```
series 1: -  -  -
series 2:  -  -  -
series 3:   -  -  -
```
If yes: will those be stitched correctly?

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
alextheimer: OK, I see. `result` is the map of join keys to matching `RangeVector`s.
The reason it's needed is for a fast lookup of a matching series (if one exists) so we can `stitch`? I.e. the reason `result` is needed is because we need to handle possibly repeat series labels (not join-key labels)?

If that's the purpose: should we consider something like a hash-map? Worst-case scenario: all LHS keys have the same join key. Would that mean there's a linear scan of all series every time we invoke `resVal.indexWhere` to add a `RangeVector` to `result`?

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
```diff
@@ -375,31 +376,107 @@ final case class SetOperatorExec(queryContext: QueryContext,
...
+        val lhsStitched =
+          if (result.contains(jk)) {
+            val resVal = result(jk)
+            index = resVal.indexWhere(r => r.key.labelValues == lhs.key.labelValues)
+            if (index >= 0) {
+              StitchRvsExec.stitch(lhs, resVal(index), outputRvRange)
+            } else lhs
+          } else lhs
+
+        // build an “unless” cursor: NaN where RHS had any sample
```
alextheimer, line 434: Is `NaN` right? I'm having a hard time proving this one way or another in [Prometheus](https://demo.promlabs.com/query?g0.expr=up&g0.show_tree=0&g0.tab=graph&g0.range_input=1h&g0.res_type=auto&g0.res_density=medium&g0.display_mode=lines&g0.show_exemplars=0).

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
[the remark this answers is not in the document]
Boyuan-Yu: We tested and the `stitch` works as expected. It can handle both cases above.

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
Boyuan-Yu: The `NaN` align with other binary operators results and get cleaned through the plan.

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
Boyuan-Yu: Updated to use a hash-map to store the result

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
Boyuan-Yu: When spread changes, we need to account for multiple Range Vectors with same key coming from different shards
Each of these range vectors would contain data for different time ranges, therefore we stitch them.

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
alextheimer: Sync'd offline -- this should indeed be `jk`

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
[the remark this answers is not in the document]
alextheimer: Discussed offline: we will trust the comment for now and handle the spread-change case.

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
```diff
@@ -413,23 +413,26 @@ final case class SetOperatorExec(queryContext: QueryContext,
...
+        val innerMap = result.getOrElse(jk, mutable.Map.empty)
+
+        // Stitch any duplicate LHS series by full labelValues
+        val prevRvOpt = innerMap.get(lhs.key.labelValues)
+        val lhsStitched = prevRvOpt match {
+          case Some(prevRv) =>
+            StitchRvsExec.stitch(lhs, prevRv, outputRvRange)
+          case None =>
+            lhs
+        }
```
alextheimer: Nice cleanup 👍

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
alextheimer: Bumping this comment (not a blocker)

alextheimer (approved):

query/src/main/scala/filodb/query/exec/SetOperatorExec.scala
Boyuan-Yu: Updated

alextheimer (approved):

PR #2013, opened 2025-05-23
fix(query): update AbsentFunctionMapper to return NaN for rows not absent

Boyuan-Yu:
**Pull Request checklist**

**Current behavior:**
AbsentFunctionMapper omits any input rows whose output is present, so downstream consumers get a "key not found" when they try to look them up.
Example
```
AbsentFunctionMapper((1, 10), (2, 5), (3, NaN), (4, NaN))
  // yields: ((3, 1), (4, 1))
  
mymetric{}
*
scalar(absent_over_time(mymetric{}[2m] offset 2m))
 // gives key not found error
```

**New behavior:**
AbsentFunctionMapper will explicitly emit NaN for any row that isn’t “absent,” avoiding missing-key errors.
Example
```
AbsentFunctionMapper((1, 10), (2, 5), (3, NaN), (4, NaN))
  // yields: ((1, NaN), (2, NaN), (3, 1), (4, 1))
```

### DIFF
--- a/query/src/test/scala/filodb/query/exec/rangefn/AbsentFunctionSpec.scala
+++ b/query/src/test/scala/filodb/query/exec/rangefn/AbsentFunctionSpec.scala
@@ -93,7 +93,7 @@ class AbsentFunctionSpec extends AnyFunSpec with Matchers with ScalaFutures with
     rows shouldEqual expectedRows
   }
 
-  it("should not generate range vector when sample is present") {
+  it("should generate range vector with NaN as value when sample is present") {
     val columnFilter = Seq(ColumnFilter("host", Equals("host1")), ColumnFilter("instance", Equals("instance1")))
     val absentFunctionMapper = exec.AbsentFunctionMapper(columnFilter, RangeParams(1, 20, 1), "metric")
     val resultObs = absentFunctionMapper(Observable.fromIterable(testSample), querySession, 1000, resultSchema, Nil)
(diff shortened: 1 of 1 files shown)


query/src/test/scala/filodb/query/exec/rangefn/AbsentFunctionSpec.scala
```diff
@@ -134,15 +138,20 @@ class AbsentFunctionSpec extends AnyFunSpec with Matchers with ScalaFutures with
...
     val keys = result.map(_.key.labelValues)
     val rows = result.flatMap(_.rows.map(x => (x.getLong(0), x.getDouble(1))).toList)
     keys.head shouldEqual expectedKeys
-    rows shouldEqual expectedRows
+    rows.zip(expectedRows).foreach {
+      case ((tsActual, vActual), (tsExp, vExp)) =>
+        tsActual shouldEqual tsExp
+        if (vExp.isNaN)    { vActual.isNaN shouldBe true }
+        else               { vActual shouldEqual vExp }
+    }
```
alextheimer, line 154: Ignoring `zip`, isn't this the same as an `equals`? Foreach element:
1. If `exp=NaN`, make sure `act=NaN`
2. else make sure `exp=act`

In other words: "make sure each `exp=act`".

Could this diff be undone?

query/src/test/scala/filodb/query/exec/rangefn/AbsentFunctionSpec.scala
```diff
@@ -100,7 +100,11 @@ class AbsentFunctionSpec extends AnyFunSpec with Matchers with ScalaFutures with
     val result = resultObs.toListL.runToFuture.futureValue
     val keys = result.map(_.key.labelValues)
     val rows = result.flatMap(_.rows.map(_.getDouble(1)).toList)
-    rows.isEmpty shouldEqual true
```
alextheimer, line 103: Could you update the name of this test? (Awkwardly adding the comment here because I can't apply it above)

alextheimer (commented):
Could you include in the PR header a sample query that might give a "key not found" error?

query/src/test/scala/filodb/query/exec/rangefn/AbsentFunctionSpec.scala
alextheimer: Discussed offline -- explanation from @Boyuan-Yu:
> The logic is equivalent to a normal equal. Except for the NaN case where equal fails to compare them because Double.NaN == Double.NaN is false (which is Non-intuitive).

PR #2030, opened 2025-07-13
feat(query): Raw to aggregated query translation Round 2

vishramachandran:
**Pull Request checklist**

test case refactor, more test cases, rule versions with active/inactive status, more comments and docs


sandeep6189 (approved):

sandeep6189:
Only two considerations:

1. Emit counter to track when HQE query translation was successful. ( If you want, we can do it in QS but need to send a true/false flag )

2.  Consider removing `.*` regex filters before trying to match tags in the Include/Exclude rule.

PR #2035, opened 2025-07-29
fix(query) Add support to disable remote stitch by tenants

amolnayak311:
**Pull Request checklist**

**Current behavior :** 
Enabling and disabling stick is a global config with no way to disable at tenant level if the global config is enabled. 

**New behavior :**
Add capabilities to disable the feature per tenant

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
@@ -286,7 +286,7 @@ class MultiPartitionPlanner(val partitionLocationProvider: PartitionLocationProv
       case raw: RawSeries                  =>
                                               val params = qContext.origQueryParams.asInstanceOf[PromQlQueryParams]
                                               if(getPartitions(raw, params).tail.nonEmpty
-                                                && queryConfig.routingConfig.supportRemoteRawExport)
+                                                && supportRemoteRawExport(raw))
                                                 this.walkLogicalPlanTree(
                                                   raw.copy(supportsRemoteDataCall = true),
                                                   qContext,
@@ -313,7 +313,7 @@ class MultiPartitionPlanner(val partitionLocationProvider: PartitionLocationProv
               val values = QueryUtils.splitAtUnescapedPipes(re.value.toString)
               values.foreach(valueSet.add)
             case _ => throw new IllegalArgumentException(
-              s"""shard keys must be filtered by equality or "|"-only regex. filter=${colFilter}""")
+              s"""shard keys must be filtered by equality or "|"-only regex. filter=$colFilter""")
           }
           acc
         }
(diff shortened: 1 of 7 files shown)


coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
```diff
@@ -972,6 +973,20 @@ class MultiPartitionPlanner(val partitionLocationProvider: PartitionLocationProv
     PlanResult(Seq(resPlan))
   }
   //scalastyle:on method.length
+
+  def supportRemoteRawExport(lp: LogicalPlan): Boolean = {
+    val matchingColumnNameForDisabledTenants = queryConfig.routingConfig.stitchDisabledTenantColumn
```
alextheimer, line 978: Maybe something like `tenantIdentifierColumn`? Then below could be e.g. `tenantIdentifiers`.

coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
```diff
@@ -972,6 +973,20 @@ class MultiPartitionPlanner(val partitionLocationProvider: PartitionLocationProv
...
+    val matchingColumnValues = lp.planColumnFilters().filter(_ match {
+      case ColumnFilter(name, Equals(_: String)) if name == matchingColumnNameForDisabledTenants         => true
+      case _                                                                                             => false
+    }).map(f => f.filter.valuesStrings.head.asInstanceOf[String]).toSet
+    // Check if any value in this set is present in disabled tenants
+    val isDisabledTenant = matchingColumnValues
+                            .intersect(queryConfig.routingConfig.tenantsWithDisabledRemoteStitch).nonEmpty
+    queryConfig.routingConfig.supportRemoteRawExport && !isDisabledTenant
```
alextheimer, line 986: Could you use `isExportEnabledTenant`? (Just to get rid of that double-negative)

query/src/main/scala/filodb/query/LogicalPlan.scala
```diff
@@ -22,6 +22,33 @@ sealed trait LogicalPlan {
     */
   def isTimeSplittable: Boolean = true
 
+  // scalastyle:off cyclomatic.complexity
+  def planColumnFilters(): Seq[ColumnFilter] = this match {
```
alextheimer, line 26: Does the existing `LogicalPlan::getRawSeriesFilters` work? Could raw data for metadata plans also need to be exported?

alextheimer (approved):
Approved with non-blocking comments ✅ 
Appreciate the comprehensive tests 🙏

coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
amolnayak311: Do you mean just change the variables and do this?

```
val isExportEnabledTenant = !matchingColumnValues
                            .intersect(queryConfig.routingConfig.tenantsWithDisabledRemoteStitch).nonEmpty
    queryConfig.routingConfig.supportRemoteRawExport && isExportEnabledTenant
```
?

query/src/main/scala/filodb/query/LogicalPlan.scala
amolnayak311: Good question, I thought it might work but then noticed cases like ``ApplyLimitFunction`` which itself has its filters apart from the filters of its child plan.

coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
amolnayak311: I know the name is awkward but ``tenantIdentifierColumn`` doesn't tell column is used for.

PR #2047, opened 2025-08-18
fix(query) Support for histogram raw data exports and stitch

amolnayak311:
**Pull Request checklist**

**Current behavior :** 
When a stitch needs to happen and multiple partitions export raw data, the underlying `RangeFunction` implementation only supported `TransientRow`. This causes stitch on histogram to fail with` scala.NotImplementedError: an implementation is missing`

**New behavior :**
The `RangeFunction` is now made generic to support both variants, `TransientRow` and `TransientHistRow`.  The PR has added matching functiontionality similar to its chunked counterpart in this PR. 
```
def histChunkedFunction(schema: ResultSchema,
                          func: Option[InternalRangeFunction],
                          funcParams: Seq[FuncArgs] = Nil): RangeFunctionGenerator = func match {
    case None                 => () => new LastSampleChunkedFunctionH
    case Some(LastSampleHistMaxMin) => require(schema.columns(2).name == "max" && schema.columns(3).name == "min")
                                 () => new LastSampleChunkedFunctionHMax(schema.colIDs(2), schema.colIDs(3))
    case Some(SumAndMaxOverTime) => require(schema.columns(2).name == "max")
                                 () => new SumAndMaxOverTimeFuncHD(schema.colIDs(2))
    case Some(RateAndMinMaxOverTime) => require(schema.columns(2).name == "max" && schema.columns(3).name == "min")
                                 () => new RateAndMinMaxOverTimeFuncHD(schema.colIDs(2), schema.colIDs(3))
    case Some(Last)           => () => new LastSampleChunkedFunctionH
    case Some(SumOverTime)    => () => new SumOverTimeChunkedFunctionH
    case Some(Rate) if schema.columns(1).isCumulative
                              => () => new HistRateFunction
    case Some(Increase) if schema.columns(1).isCumulative
                              => () => new HistIncreaseFunction
    case Some(Rate)           => () => new RateOverDeltaChunkedFunctionH
    case Some(Increase)       => () => new SumOverTimeChunkedFunctionH
    case _                    => ??? //TODO enumerate all possible cases
  }
```

### DIFF
--- a/query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
+++ b/query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
@@ -128,13 +128,13 @@ class MaxOverTimeChunkedFunctionL(var max: Long = Long.MinValue) extends Chunked
  *   * Add (t4, 3): deque = [(t3, 15), (t4, 3)] (3 < 15, so keep both)
  *   * Add (t5, 12): deque = [(t3, 15), (t5, 12)] (12 > 3, remove 3; 12 < 15, keep 15)
  *   * Head value is always maximum: 15
- * 
+ *
  * - For min tracking with same values:
  *   * The deque would maintain smallest values, with head containing minimum
- * 
+ *
  * Time complexity: O(1) amortized for offer(), O(log n) worst case for individual operations
  * Space complexity: O(k) where k is the effective window size (typically much smaller than total values)
- * 
+ *
  * @param ordering Determines whether to track max (Ordering.Double) or min (Ordering.Double.reverse)
  */
 private case class MaxMinTracker(ordering: Ordering[Double]) {


core/src/main/scala/filodb.core/query/TransientRow.scala
```diff
@@ -105,17 +109,42 @@ class TransientHistRow(var timestamp: Long = 0L,
   def getBlobNumBytes(columnNo: Int): Int = throw new IllegalArgumentException()
 
   override def toString: String = s"TransientHistRow(t=$timestamp, v=$value)"
+
+  override def copyFrom(r: RowReader): Unit = r match {
+    case h: TransientHistRow   if h.value.isInstanceOf[LongHistogram]
+                                                          => timestamp = h.timestamp
+                                                             val lh = h.value.asInstanceOf[LongHistogram]
+                                                             value = LongHistogram(lh.buckets, lh.values.clone())
```
alextheimer, line 117: Does `copy` work here, too (instead of `clone`)?

core/src/main/scala/filodb.core/query/TransientRow.scala
```diff
@@ -105,17 +109,42 @@ class TransientHistRow(var timestamp: Long = 0L,
...
   override def setDouble(columnNo: Int, valu: Double): Unit =
     if (columnNo == 2) max = valu else if (columnNo == 3) min = valu
     else throw new IllegalArgumentException(s"ColumnId: ${columnNo} is invalid")
   override def getDouble(columnNo: Int): Double =
     if (columnNo == 2) max else if (columnNo == 3) min else throw new IllegalArgumentException()
 
   override def toString: String = s"TransientHistMaxMinRow(t=$timestamp, h=$value, max=$max, min=$min)"
+
+  override def copyFrom(r: RowReader): Unit = r match {
+    // TODO: Change the copy for this to handle BinaryRowReader
```
alextheimer: Leftover `TODO`?

core/src/main/scala/filodb.core/query/TransientRow.scala
```diff
@@ -105,17 +109,42 @@ class TransientHistRow(var timestamp: Long = 0L,
...
+                                                             value = LongHistogram(lh.buckets, lh.values.clone())
+    case h: TransientHistRow   if h.value.isInstanceOf[MutableHistogram]
+                                                          => timestamp = h.timestamp
+                                                             value = h.value.asInstanceOf[MutableHistogram].copy()
+    case b: BinaryRecordRowReader
+        if b.schema.columnTypes(1) == ColumnType.HistogramColumn   =>
+                                                           timestamp = b.getLong(0)
+                                                           value = b.getHistogram(1).asInstanceOf[HistogramWithBuckets]
+    case _                                                         =>
+                                                           throw new IllegalArgumentException("Unknown Row reader")
```
alextheimer, line 126: Could you include the reader's class name in the message? (Same goes for other `Unknown Row reader` exceptions, too.)

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
```diff
@@ -112,8 +112,167 @@ class MaxOverTimeChunkedFunctionL(var max: Long = Long.MinValue) extends Chunked
...
 }
 
-class SumOverTimeFunction(var sum: Double = Double.NaN, var count: Int = 0) extends RangeFunction {
-  override def addedToWindow(row: TransientRow, window: Window): Unit = {
+private case class MaxMinTracker(ordering: Ordering[Double]) {
+  private val values: util.ArrayDeque[(Long, Double)] = new util.ArrayDeque[(Long, Double)]()
+
+  def offer(ts: Long, value: Double): Boolean = {
+    if (!value.isNaN) {
+      while(!values.isEmpty && ordering.compare(values.peekLast()._2, value) < 0) {
```
alextheimer, line 145: Could you document that `Ordering.Double` will give a "max" tracker, and `Ordering.Double.reverse` will give a "min" tracker?

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
```diff
@@ -112,8 +112,167 @@ class MaxOverTimeChunkedFunctionL(var max: Long = Long.MinValue) extends Chunked
...
+    } else {
+      false
+    }
+  }
+
+  def removeValuesOlderThan(ts: Long): Unit = {
+    while(!values.isEmpty && values.peekFirst()._1 <= ts) {
+      values.removeFirst()
+    }
+  }
```
alextheimer, line 183: I didn't believe this at first 🤯
Here's a proof it works in case it's worth pasting there:
```
// This method would behave incorrectly in one scenario:
//   when a value exists closer to the front of the queue
//   than another with an older timestamp. In that case,
//   some older elements would not be removed.
// 
// Suppose this is a "max" tracker, and suppose
//   such a pair of values exists.
// Given that the newer value is closer to the front of
//   the queue, it has a value at least as large as the
//   older value.
// But it can't be equal to the older value; elements
//   are only removed during offer() while the order
//   is violated. In that case, the newer timestamp
//   would be closer to the end of the queue.
// And it cannot be greater than the older value;
//   the smaller value would have been removed
//   during offer().
// So, the supposed pair of elements cannot exist.
```

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
```diff
@@ -112,8 +112,167 @@ class MaxOverTimeChunkedFunctionL(var max: Long = Long.MinValue) extends Chunked
...
+      }
+      count += 1
+    }
+  }
+
+  override def removedFromWindow(row: TransientHistRow, window: Window[TransientHistRow]): Unit = {
+    // TODO: Very expensive, is there a better way? Can we subtract assuming the counter correction is already done?
+    val histValue = row.value
+    if (histValue.numBuckets > 0) {
+      // Since histogram subtraction is complex and not typically supported,
```
alextheimer, line 282: Could you give an example of what would break if we did the subtraction here?

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
```diff
@@ -112,8 +112,71 @@ class MaxOverTimeChunkedFunctionL(var max: Long = Long.MinValue) extends Chunked
...
+      // Recalculate sum from all remaining items in window
+      for (i <- 0 until window.size) {
+        val windowHist = window(i).value
+        if (windowHist.numBuckets > 0) {
+          sum match {
+            case hist if hist.numBuckets == 0 => sum = bv.MutableHistogram(windowHist)
+            case hist: bv.MutableHistogram    => hist.add(windowHist)
+          }
+          count += 1
+        }
```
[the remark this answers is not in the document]
alextheimer: The bot straight-up ignored your comments above 😆

query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
```diff
@@ -397,6 +403,56 @@ object RangeFunction {
...
+    case Some(AvgOverTime)                        => () => new AvgOverDeltaFunctionH()
+    case Some(StdDevOverTime)                     => throw new NotImplementedError(notImplemented("StdDevOverTime"))
+    case Some(StdVarOverTime)                     => throw new NotImplementedError(notImplemented("StdVarOverTime"))
+    case Some(Changes)                            => throw new NotImplementedError(notImplemented("Changes"))
+    case Some(QuantileOverTime)                   => throw new NotImplementedError(notImplemented("QuantileOverTime"))
+    case Some(MedianAbsoluteDeviationOverTime)    =>
+                                       throw new NotImplementedError(notImplemented("MedianAbsoluteDeviationOverTime"))
+    case Some(LastOverTimeIsMadOutlier)        =>
+                                       throw new NotImplementedError(notImplemented("LastOverTimeIsMadOutlier"))
+    case _                                      => ??? //TODO enumerate all possible cases
```
alextheimer: Leftover `TODO`?

query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
```diff
@@ -397,6 +403,56 @@ object RangeFunction {
...
+    // Since cumulative is already handled previously, this is for delta temporality
+    case Some(Rate)                               => () => new RateOverDeltaFunctionH()
+    case Some(Increase)                           => () => new SumOverTimeFunctionH() // Sum of deltas over time
+    case Some(Delta)                              => () => DeltaFunctionH
+    case Some(Irate) if schema.columns(1).isCumulative
+                                                  => () => IRateFunctionH
+    case Some(Idelta)                             => () => IDeltaFunctionH
+    case Some(Irate)                              => () => IRatePeriodicFunctionH
+    case Some(LastSampleHistMaxMin) => require(schema.columns(2).name == "max" && schema.columns(3).name == "min")
+      () => new LastSampleFunctionH(true)
```
alextheimer: [Nitpick] Could you use named parameters here?

query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
```diff
@@ -436,12 +485,38 @@ object RangeFunction {
...
+            queryConfig: QueryConfig): Unit = {
+
+    for (i <- (window.size - 1) to 0 by -1) {
+      val row = window(i)
+      // TODO: Is this ok? The only implementation that's not HistogramWithBuckets is PromRateHistogram
+      //  That's private in HistogramQuantileMapper, also is one bucket not NaN a good check?
+      val hist = row.getHistogram(1).asInstanceOf[HistogramWithBuckets]
+      if (hist.numBuckets > 0 && !hist.bucketValue(0).isNaN ) {
+        sampleToEmit.setValues(endTimestamp, hist)
+        return
```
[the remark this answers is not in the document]
alextheimer: Confirming: I'm also unsure about the `TODO`.

query/src/main/scala/filodb/query/exec/rangefn/RangeInstantFunctions.scala
```diff
@@ -68,10 +72,55 @@ object RangeInstantFunctions {
...
+      require(window.head.timestamp >= startTimestamp, "Possible internal error, found samples < startTimestamp")
+      require(window.last.timestamp <= endTimestamp, "Possible internal error, found samples > endTimestamp")
+
+      val lastSample = window.last.value
+      val prevSampleRow = window(window.size - 2)
+      val prevSample = prevSampleRow.value
+
+      // Check if histograms have compatible bucket schemes
+      if (prevSample.buckets == lastSample.buckets && lastSample.numBuckets > 0) {
+        val resultArray = new Array[Double](lastSample.numBuckets)
```
alextheimer: [Nitpick] Could you call this `resultBucketValues`?

query/src/main/scala/filodb/query/exec/rangefn/RangeInstantFunctions.scala
```diff
@@ -111,15 +160,49 @@ object IRateFunction extends RangeFunction {
...
+  def addedToWindow(row: TransientHistRow, window: Window[TransientHistRow]): Unit = {}
+  def removedFromWindow(row: TransientHistRow, window: Window[TransientHistRow]): Unit = {}
+
+  def apply(startTimestamp: Long,
+            endTimestamp: Long,
+            window: Window[TransientHistRow],
+            sampleToEmit: TransientHistRow,
+            queryConfig: QueryConfig): Unit = {
+    val result = RangeInstantFunctions.instantValueH(startTimestamp,
+      endTimestamp, window, true)
```
alextheimer: [Nitpick] Could you use named parameters here?

query/src/main/scala/filodb/query/exec/rangefn/RangeInstantFunctions.scala
```diff
@@ -111,15 +160,49 @@ object IRateFunction extends RangeFunction {
...
+  def addedToWindow(row: TransientHistRow, window: Window[TransientHistRow]): Unit = {}
+  def removedFromWindow(row: TransientHistRow, window: Window[TransientHistRow]): Unit = {}
+
+  def apply(startTimestamp: Long,
+            endTimestamp: Long,
+            window: Window[TransientHistRow],
+            sampleToEmit: TransientHistRow,
+            queryConfig: QueryConfig): Unit = {
+    val result = RangeInstantFunctions.instantValueH(startTimestamp,
+      endTimestamp, window, false)
```
alextheimer: [Nitpick] Could you use named parameters here?

query/src/main/scala/filodb/query/exec/rangefn/RateFunctions.scala
```diff
@@ -112,14 +145,31 @@ object RateFunction extends RangeFunction {
...
 
   def apply(startTimestamp: Long,
             endTimestamp: Long,
-            window: Window,
+            window: Window[TransientHistRow],
+            sampleToEmit: TransientHistRow,
+            queryConfig: QueryConfig): Unit = {
+    val result = RateFunctions.extrapolatedRateH(startTimestamp,
+      endTimestamp, window, true, true)
+    sampleToEmit.setValues(endTimestamp, result) // TODO need to use a NA instead of NaN
```
alextheimer, line 161: What does this `TODO` mean?

query/src/main/scala/filodb/query/exec/rangefn/RateFunctions.scala
```diff
@@ -415,4 +490,38 @@ class RateOverDeltaChunkedFunctionH(var h: bv.MutableHistogram = bv.Histogram.em
...
+  def addedToWindow(row: TransientHistRow, window: Window[TransientHistRow]): Unit = {}
+  def removedFromWindow(row: TransientHistRow, window: Window[TransientHistRow]): Unit = {}
+
+  def apply(startTimestamp: Long,
+            endTimestamp: Long,
+            window: Window[TransientHistRow],
+            sampleToEmit: TransientHistRow,
+            queryConfig: QueryConfig): Unit = {
+    val result = RateFunctions.extrapolatedRateH(startTimestamp,
+      endTimestamp, window, false, false)
```
alextheimer: [Nitpick] Could you use named parameters here?

alextheimer (commented):
Finished a first-pass review of all non-test files. Mostly LGTM-- the only scary-looking thing is the full window iteration during `SumOverTimeFunctionH::removedFromWindow`. Let's talk about that one in-person (I don't 100% understand why plain subtraction doesn't work).

query/src/test/scala/filodb/query/exec/BufferableCounterCorrectionIteratorHSpec.scala
```diff
@@ -0,0 +1,226 @@
...
+import filodb.core.query.TransientHistRow
+import filodb.memory.format.vectors.{CustomBuckets, HistogramWithBuckets, MutableHistogram}
+import org.scalatest.funspec.AnyFunSpec
+import org.scalatest.matchers.should.Matchers
+
+class BufferableCounterCorrectionIteratorHSpec extends AnyFunSpec with Matchers {
+
+  // Helper method to create histogram with bucket values
+  private def createHistogram(bucketValues: Array[Double]): HistogramWithBuckets = {
+    // TODO: How would GeometricHistogram and ExpHistogram Work?
```
alextheimer, line 12: Leftover `TODO`?

core/src/main/scala/filodb.core/query/TransientRow.scala
amolnayak311: No, ``copy`` doesn't exist and needs``copyToXX``, any concern with clone?

core/src/main/scala/filodb.core/query/TransientRow.scala
amolnayak311: Do you mean log the schema? No point logging the class name of row reader as it will be ``BinaryRecordRowReader``.

core/src/main/scala/filodb.core/query/TransientRow.scala
amolnayak311: Thanks for catching this

query/src/main/scala/filodb/query/exec/rangefn/RateFunctions.scala
amolnayak311: Ah never mind will remove it

core/src/main/scala/filodb.core/query/TransientRow.scala
alextheimer: I didn't know the difference prior to this PR, but found this: https://stackoverflow.com/a/31013156
Looks like `copy` is a more-efficient `memcpy` equivalent, but it could be true that `clone` is implemented the same. Anyways if `copy` doesn't exist, feel free to mark this resolved.

core/src/main/scala/filodb.core/query/TransientRow.scala
alextheimer: Couldn't it be any arbitrary `RowReader` type?

core/src/main/scala/filodb.core/query/TransientRow.scala
amolnayak311: Yeah, ``copy`` gives a compilation error. Marking this as resolved

core/src/main/scala/filodb.core/query/TransientRow.scala
amolnayak311: Let me log the class name,

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
amolnayak311: > Ordering.Double will give a "max" tracker

Not sure that would be an accurate statement, technically ``Ordering`` is a [comparator](https://www.scala-lang.org/api/3.x/scala/math/Ordering.html)

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
amolnayak311: Let me think through and put some comments with an example. The found myself reading the above comment multiple times to understand what it tries to say :D

query/src/test/scala/filodb/query/exec/BufferableCounterCorrectionIteratorHSpec.scala
amolnayak311: No,  I don't know how to create and test a ExpHistogram just wanted to leave a TODO if any one reviewing has an idea.

amolnayak311:
> Finished a first-pass review of all non-test files. Mostly LGTM-- the only scary-looking thing is the full window iteration during SumOverTimeFunctionH::removedFromWindow. Let's talk about that one in-person (I don't 100% understand why plain subtraction doesn't work).

@alextheimer i dont know who would it behave with counter correction, its not something I can test in any integration tests to be certain it will work or not.

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
amolnayak311: Done, updated.

core/src/main/scala/filodb.core/query/TransientRow.scala
amolnayak311: Done

query/src/main/scala/filodb/query/exec/rangefn/AggrOverTimeFunctions.scala
alextheimer: I just mean that a `MaxMinTracker` constructed with an `Ordering.Double` will keep track of the greatest values (ala line 141).

PR #2054, opened 2025-08-29
Added Kafka consumer to ingest data to Prometheus

MichaelBragilevsky:
**Pull Request checklist**

**Current behavior :**
Currently gateway produces data to Kafka

**New behavior :**
Added new Kafka consumer TestTimeseriesPrometheusConsumer to ingest data to Prometheus by using pushgateway

**Limitations :**
Supports only counter/gauge metric types

### DIFF
--- a/dev-gateway.sh
+++ b/dev-gateway.sh
@@ -6,3 +6,6 @@ args=${@:-"conf/timeseries-dev-source.conf"}
 java -Dconfig.file=conf/timeseries-filodb-server.conf  \
      -Dkamon.prometheus.embedded-server.port=9097  \
      -cp gateway/target/scala-2.12/gateway-* filodb.gateway.GatewayServer $args &
+
+# Starts a Prometheus Kafka Consumer for timeseries data
+java -cp gateway/target/scala-2.12/gateway-* filodb.timeseries.TestTimeseriesPrometheusConsumer $args &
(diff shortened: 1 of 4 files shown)


dev-gateway.sh
```diff
@@ -6,3 +6,6 @@ args=${@:-"conf/timeseries-dev-source.conf"}
 java -Dconfig.file=conf/timeseries-filodb-server.conf  \
      -Dkamon.prometheus.embedded-server.port=9097  \
      -cp gateway/target/scala-2.12/gateway-* filodb.gateway.GatewayServer $args &
+
+# Starts a Prometheus Kafka Consumer for timeseries data
+java -cp gateway/target/scala-2.12/gateway-* filodb.timeseries.TestTimeseriesPrometheusConsumer $args &
```
sandeep6189: can we add a config to enable local prometheus ingestion ? ( to make sure existing automation goes through without any issue )

sandeep6189:
@MichaelBragilevsky How do we ensure we are writing same data to FiloDB and Prometheus ? Should we use the same metric generator, but different mapping functions, each of which write to FiloDB and Prometheus ?

MichaelBragilevsky:
> @MichaelBragilevsky How do we ensure we are writing same data to FiloDB and Prometheus ? Should we use the same metric generator, but different mapping functions, each of which write to FiloDB and Prometheus ?

We ensure we are writing the same data to FiloDB and Prometheus by using one source of data, that is Kafka

dev-gateway.sh
MichaelBragilevsky: moved the start of the consumer to the mosaic-query-tools debugger script, as this is more related to the integration tests and the consumer should be run explicitly as part of the dev/test phase.

sandeep6189 (approved):

PR #2058, opened 2025-09-03
Split concatenated namespaces

bhavanap1131-cmd:
Split pipe-concatenated namespace regex (e.g., _ns_=~"nsA|nsB|nsC") using QueryUtils.splitAtUnescapedPipes:

Single value ->  Equals("nsA")

Multiple values ->  In(Set("nsA","nsB","nsC"))

Per-namespace attribution: when In(...) is present, build a metric group-by per namespace and increment counters accordingly (timeSeriesScanned, cpuNanos; preserves existing behavior elsewhere).

### DIFF
--- a/core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
+++ b/core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
@@ -81,6 +81,7 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
   private val indexTtlMs = downsampleTtls.last.toMillis
   private val clusterType = filodbConfig.getString("cluster-type")
   private val deploymentPartitionName = filodbConfig.getString("deployment-partition-name")
+  private val NamespaceCol = "namespace"
 
   private val downsampleStoreConfig = StoreConfig(filodbConfig.getConfig("downsampler.downsample-store-config"))
   private val typeFieldIndexingEnabled = filodbConfig.getBoolean("memstore.type-field-indexing-enabled")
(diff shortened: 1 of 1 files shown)


core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
```diff
@@ -372,6 +378,87 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
...
+      val shardKeys = schemas.part.options.shardKeyColumns
+
+      nsValues.foreach { v =>
+        // replace namespace=In(...) with namespace=Equals(v) for per-NS attribution
+        val perNsFilters = normalizedFilters.updated(nsIdx, ColumnFilter(NamespaceCol, Filter.Equals(v)))
+        val startNs = Utils.currentThreadCpuTimeNanos
+
+        // per-namespace lookup purely for accounting
+        val recsForNs = partKeyIndex.partKeyRecordsFromFilters(
+          perNsFilters, chunkMethod.startTime, chunkMethod.endTime)
```
alextheimer, line 426: This correctly gets the per-ns query stats, but it's an expensive index lookup that we've already done for the full pipe-concatenated set of namespaces (i.e. to serve the query). IIUC this is effectively a second full scan.

Our options might be pretty limited here-- we can:
1. In the `SingleClusterPlanner`, split `LogicalPlan`s with pipe-concatenated namespaces into one plan per namespace. Materialize each, then concatenate/aggregate. **Benefit**: filodb clusters can still be sent namespace-regex; total count of remote plans remains low. **Cost**: risky planner rework.
2. In the `ShardKeyRegexPlanner`, set the max count of pipe-concatenated namespaces to 1. **Benefit**: stupid-simple configuration change. **Cost**: count of remote plans will increase; query fanout increases.

I'm leaning towards option 2 (at least as a first attempt). The benefit we've seen from reduced fanout has been negligible, so I don't expect query latencies to suddenly increase. Also, we can scale the max count of pipe-concatenated namespaces gradually back to 1 and abort if query latencies spike.

What do you think @amolnayak311?

PR #2064, opened 2025-09-12
feat(gateway): Supporting remote write protocol for prometheus kafka consumer for data completeness

MichaelBragilevsky:
**Pull Request checklist**

**Current behavior** :
Currently TestTimeseriesPrometheusConsumer uses pushgateway to produce data series to Prometheus. Pushgateway stores only last sample, but we want to store all historic data.
GatewayServer accepts only one metric type.

**New behavior** :
TestTimeseriesPrometheusConsumer  changed to use remote write protocol with protobuf to send historic data to Prometheus.
Added support for namespace and workspace to be propagated to GatewayServer.
Refactored GatewayServer to accept different metric types in one run.

### DIFF
--- a/gateway/src/main/scala/filodb/timeseries/TestTimeseriesPrometheusConsumer.scala
+++ b/gateway/src/main/scala/filodb/timeseries/TestTimeseriesPrometheusConsumer.scala
@@ -20,33 +20,15 @@ import monix.kafka.config.AutoOffsetReset
 import monix.reactive.Observable
 import org.apache.kafka.clients.consumer.ConsumerConfig
 import org.apache.kafka.common.serialization.{Deserializer, LongDeserializer}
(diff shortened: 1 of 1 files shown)


project/Dependencies.scala
```diff
@@ -148,6 +148,8 @@ object Dependencies {
     logbackDep,
     "io.monix"   %% "monix-kafka-1x" % monixKafkaVersion,
     "org.rogach" %% "scallop"        % "3.1.1",
+    "com.google.protobuf" % "protobuf-java" % "3.21.7",
```
sandeep6189: I think we have proto and snappy dependencies defined for other projects in Dependencies.scala . Should we make sure that we are using the same versions across ( to avoid runtime overrides ) ?

image

PR #2066, opened 2025-09-16
full merge integration to main for 0.9.31 release

sherali42:


sandeep6189 (approved):

version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "0.9.30.9"
+version in ThisBuild := "0.9.31"
```
alextheimer: Does anything rely on the patch version? I couldn't find a version in recent history that didn't have the patch version.

sherali42: done, added minor version just in case 👍

alextheimer (approved):

yu-shipit (approved):

PR #2079, opened 2025-10-17
feat(core): Process metrics including page faults

vishramachandran:
**Pull Request checklist**

Process metrics including page faults and context switch count with Otel.

### DIFF
--- a/core/src/main/scala/filodb.core/metrics/FilodbMetrics.scala
+++ b/core/src/main/scala/filodb.core/metrics/FilodbMetrics.scala
@@ -4,6 +4,7 @@ import java.nio.file.{Files, Paths}
 import java.time.Duration
 import java.util.concurrent.{ConcurrentHashMap, TimeUnit}
 
+import scala.collection.mutable.ListBuffer
 import scala.concurrent.duration.DurationInt
 
 import com.typesafe.config.Config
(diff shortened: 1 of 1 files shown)


core/src/main/scala/filodb.core/metrics/FilodbMetrics.scala
```diff
@@ -228,9 +233,42 @@ private class FilodbMetrics(filodbMetricsConfig: Config) extends StrictLogging {
...
+    implicit val sch: Scheduler = monix.execution.Scheduler.global
+    val strMem = "process_memory_bytes"
+    val memRss = FilodbMetrics.bytesGauge(strMem, Map("type" -> "rss"))
+    val memVms = FilodbMetrics.bytesGauge(strMem, Map("type" -> "vms"))
+    val strCpu = "process_cpu_time_total_seconds"
+    val cpuUserTime = FilodbMetrics.timeGauge(strCpu, TimeUnit.MILLISECONDS, Map("type" -> "user"))
+    val cpuSystemTime = FilodbMetrics.timeGauge(strCpu, TimeUnit.MILLISECONDS, Map("type" -> "system"))
+    val majorPageFaults = FilodbMetrics.gauge("process_major_page_faults_total")
+    val minorPageFaults = FilodbMetrics.gauge("process_minor_page_faults_total")
+    val contextSwitches = FilodbMetrics.gauge("process_context_switches_total")
```
alextheimer: Should these be counters?

alextheimer (approved):
LGTM -- PTAL at the one comment in case the metric types should be changed.

core/src/main/scala/filodb.core/metrics/FilodbMetrics.scala
vishramachandran: The are indeed OS level counters, but since they are read from an external source they cannot be incremented with the read value as such. Gauge is the best for now.

vishramachandran: I will change the metric name to add "cumul" to hint the reader

vishramachandran: I think I found a way to keep it a counter. Will make the change.

PR #2084, opened 2025-10-24
Exec plan sample limits: per-leaf + total cap

bhavanap1131-cmd:
Separate metadata query leaf limit from execPlanSamples limit:

This change introduces a distinct configuration for controlling the per-leaf result size in metadata (/series) queries.
Previously, the execPlanSamples parameter was applied as both the global and per-leaf limit, which could lead to inconsistent enforcement when intermediate plan limits are enabled.

With this update:

Added a new config key metadata-leaf-limit to define the maximum number of samples returned per leaf.
execPlanSamples continues to control the overall query limit for regular (non-metadata) queries.

### DIFF
--- a/coordinator/src/main/scala/filodb/coordinator/ProtoConverters.scala
+++ b/coordinator/src/main/scala/filodb/coordinator/ProtoConverters.scala
@@ -350,7 +350,7 @@ object ProtoConverters {
     cd.getInputColIdsList.asScala.toSeq.map(i => i.intValue())
   }
 
-  // TimeDownsampler
+  //  TimeDownsampler
   implicit class TimeDownsamplerToProtoConverter(td: filodb.core.downsample.TimeDownsampler) {
     def toProto: GrpcMultiPartitionQueryService.TimeDownsampler = {
       val builder = GrpcMultiPartitionQueryService.TimeDownsampler.newBuilder()


core/src/main/scala/filodb.core/query/QueryContext.scala
```diff
@@ -26,6 +26,7 @@ case object UnavailablePromQlQueryParams extends TsdbQueryParams
 
 case class PerQueryLimits(
         execPlanSamples: Int = 1000000,       // Limit on ExecPlan results in samples, default is 100K
+        execPlanLeafSamples: Int = 1000000,   // Limit on ExecPlanLeaf results in samples, default is 100K
```
alextheimer: Ack'd the other comments are wrong, but could you correct this one?

query/src/main/scala/filodb/query/exec/MetadataRemoteExec.scala
```diff
@@ -167,5 +167,5 @@ case class MetadataRemoteExec(queryEndpoint: String,
       if (response.partial.isDefined) response.partial.get else false, response.message)
   }
 
-  override def checkSamplesLimit(numResultSamples: Int, queryWarnings: QueryWarnings): Unit = {}
+  // override def checkSamplesLimit(numResultSamples: Int, queryWarnings: QueryWarnings): Unit = {}
```
alextheimer: Feel free to delete the line completely.

alextheimer (commented):
Could you update the relevant `ProtoConverters`, too? https://github.com/filodb/FiloDB/blob/9e937374ad0edd038295293abc54fcd2621e5545/query/src/main/scala/filodb/query/ProtoConverters.scala#L356

bhavanap1131-cmd:
test this please

query/src/main/scala/filodb/query/ProtoConverters.scala
alextheimer, line 375: Does this need an update, too?

PR #2130, opened 2026-02-03
 fix(queryplanner): fix metadata query routing to multiple partitions

bhavanap1131-cmd:
**Pull Request checklist**                                                                                 
                                                                                                         
  - The commits messages follow the contribution CONTRIBUTING.md
  - Tests for the changes have been added for bug fixes                                 
  - Docs have been added and updated for bug fixes                                 
                                                                                                         
  ---                                                                                                    
  **Current behavior:**                                                                                      
                                                                                                         
  Metadata queries (LabelValues, SeriesKeysByFilters, LabelNames, etc.) in MultiPartitionPlanner were
  unconditionally routed through getMetadataPartitions (legacy fallback path), regardless of whether the 
  shard-key filters were sufficient for direct partition routing. SeriesKeysByFilters plans were also not
   handled correctly since their filters are stored directly on the plan node rather than on RawSeries   
  leaves.         

  **New behavior:**

  Introduced resolveMetadataPartitions which routes directly via getPartitions when all shard-key filters
   are present and Equals, falling back to getMetadataPartitions otherwise. SeriesKeysByFilters plans are
   now handled explicitly. Test coverage added in MultiPartitionPlannerSpec for all routing conditions   
  across both plan types. Broken and misplaced tests removed from PlannerHierarchySpec.

  **Other information:**                                                                                     
   
  - Removed three tests from PlannerHierarchySpec that invoked MultiPartitionPlanner directly without    
  processMultiPartition = true -- the partition provider was never called. Behaviour is covered in
  MultiPartitionPlannerSpec.


coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
```diff
@@ -1061,10 +1061,16 @@ class MultiPartitionPlanner(val partitionLocationProvider: PartitionLocationProv
...
     // is a metadata operation and shares common components with other metadata endpoints.
+    val timeRange = TimeRange(queryParams.startSecs * 1000L, queryParams.endSecs * 1000L)
+    val shardKeyFilterGroups = LogicalPlan.getNonMetricShardKeyFilters(lp, dataset.options.shardKeyColumns)
     val partitions = lp match {
-      case lc: LabelCardinality       => getPartitions(lc, qContext.origQueryParams.asInstanceOf[PromQlQueryParams])
-      case _                          => getMetadataPartitions(lp.filters,
-        TimeRange(queryParams.startSecs * 1000, queryParams.endSecs * 1000))
+      case lc: LabelCardinality => getPartitions(lc, queryParams)
+      case _ => shardKeyFilterGroups.collectFirst {
+        case group
```
alextheimer: We should default to the usual `getMetadataPartitions` if any nonempty filter group doesn't have an `Equals` filter for all non-metric shard-key labels.

Example:
```
# Suppose:
nonMetricShardKeyColumns=[a, b]

# OK -- should find partitions with new logic.
[
    [Filter(a = "123"), Filter(b = "456")],
    [Filter(c = "123"), Filter(a = "789"), Filter(b = "012")]
]

# Default to getMetadataPartitions -- "b" not filtered by equality in the first group.
[
    [Filter(a = "123"), Filter(b =~ "456")],
    [Filter(c = "123"), Filter(a = "789"), Filter(b = "012")]
]

# Default to getMetadataPartitions -- second group is missing "a" filter.
[
    [Filter(a = "123"), Filter(b = "456")],
    [Filter(c = "123"), Filter(b = "012")]
]
```

alextheimer: Please add some test cases, as well

amolnayak311 (changes requested):
Please follow the connections given in https://github.com/filodb/FiloDB/blob/develop/CONTRIBUTING.md and update the PR title, and provide the summary of the changes

coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
```diff
@@ -1103,6 +1098,38 @@ class MultiPartitionPlanner(val partitionLocationProvider: PartitionLocationProv
...
+      })
+
+    val shouldFallback = shardKeyFilterGroups.exists(g => g.nonEmpty && !hasAllNonMetricEquals(g))
+
+    lp match {
+      case lc: LabelCardinality =>
+        getPartitions(lc, queryParams)
+
+      case _ =>
+        val fallback = getMetadataPartitions(lp.filters, timeRange)
```
alextheimer: This might incur a remote request -- call it only if `shouldFallback`

coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
```diff
@@ -1103,6 +1098,38 @@ class MultiPartitionPlanner(val partitionLocationProvider: PartitionLocationProv
...
+        if (shouldFallback) fallback
+        else shardKeyFilterGroups.collectFirst {
```
alextheimer: All nonempty filters should be accounted for. What you're looking for is something like:
```
shardKeyFilterGroups
    .filter(nonempty)
    .map(group -> routingMap)
    .flatMap(routingMap -> partitionLocationProvider.getPartitions(routingMap, timeRange))
    .distinct  // Probably we don't need anything more sophisticated than this.
```

coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
```diff
@@ -1103,6 +1098,43 @@ class MultiPartitionPlanner(val partitionLocationProvider: PartitionLocationProv
...
+      case lc: LabelCardinality =>
+        getPartitions(lc, queryParams)
+
+      case _ =>
+        if (shouldFallback) {
+          getMetadataPartitions(lp.filters, timeRange)
+        } else {
+          shardKeyFilterGroups
+            .filter(_.nonEmpty)
+            .filter(hasAllNonMetricEquals)
```
alextheimer: Correct me if I'm wrong: I think this check can be removed, since we'd only enter this `else` if `shouldFallback` is `false`, and `shouldFallback` is only `true` iff either (a) no filter groups exist, or (b) some group doesn't have all non-metric shard-key equality filters.

So, by the time we hit this filter, we've already guaranteed that all groups have all required `ColumnFilter`s.

coordinator/src/test/scala/filodb.coordinator/queryplanner/ShardKeyRegexPlannerSpec.scala
```diff
@@ -1179,4 +1179,700 @@ class ShardKeyRegexPlannerSpec extends AnyFunSpec with Matchers with ScalaFuture
       validatePlan(execPlan.children.head, expected)
     }
   }
+
```
alextheimer: Toss these new tests into `PlannerHierarchySpec` (since they're not testing either the `MultiPartitionPlanner` or `ShardKeyRegexPlanner` exclusively).

coordinator/src/test/scala/filodb.coordinator/queryplanner/ShardKeyRegexPlannerSpec.scala
```diff
@@ -1179,4 +1179,700 @@ class ShardKeyRegexPlannerSpec extends AnyFunSpec with Matchers with ScalaFuture
...
+        Seq(
+          ColumnFilter("_ws_", Equals("demo")),
+          ColumnFilter("_ns_", Equals("App-2"))
+        )
+      )
+    }
+
+    val skrp = new ShardKeyRegexPlanner(dataset, testMpp, testShardKeyMatcherFn, simplePartitionLocationProvider, queryConfig)
+
+    val lp = Parser.queryToLogicalPlan("""test{_ws_ = "demo", _ns_ =~ "App.*"}""", 1000, 1000)
```
alextheimer: `queryToLogicalPlan` will give a non-metadata plan. Use this method, instead: `metadataQueryToLogicalPlan`

coordinator/src/test/scala/filodb.coordinator/queryplanner/ShardKeyRegexPlannerSpec.scala
```diff
@@ -1179,4 +1179,700 @@ class ShardKeyRegexPlannerSpec extends AnyFunSpec with Matchers with ScalaFuture
...
+    execPlan.children.size shouldEqual 2
+
+    // Verify both partitions are represented
+    val childFilters = execPlan.children.flatMap(_.children.flatMap {
+      case mspe: MultiSchemaPartitionsExec => mspe.filters
+      case _ => Seq.empty
+    })
+
+    childFilters.count(f => f.column == "_ns_" && f.filter.valuesStrings.contains("App-1")) should be > 0
+    childFilters.count(f => f.column == "_ns_" && f.filter.valuesStrings.contains("App-2")) should be > 0
```
alextheimer: As written, these don't validate the correct set of partitions were included in the plan. You can use the `ExecPlan` method `printTree` to see how to access the partitions.

If it's easier, you can use a method similar to `should route raw data export to downSample cluster` in `PlannerHierarchySpec` to easily and visually validate that the plan is correct.

coordinator/src/test/scala/filodb.coordinator/queryplanner/ShardKeyRegexPlannerSpec.scala
```diff
@@ -1179,4 +1179,700 @@ class ShardKeyRegexPlannerSpec extends AnyFunSpec with Matchers with ScalaFuture
...
+    val testMppPartitionLocationProvider = new PartitionLocationProvider {
+      override def getPartitions(routingKey: Map[String, String], timeRange: TimeRange): List[PartitionAssignment] = {
+        if (routingKey.equals(Map("_ns_" -> "App-1", "_ws_" -> "demo")))
+          List(PartitionAssignment("remote", "remote-url", TimeRange(timeRange.startMs, timeRange.endMs)))
+        else
+          List(PartitionAssignment("local", "local-url", TimeRange(timeRange.startMs, timeRange.endMs)))
+      }
+
+      override def getMetadataPartitions(nonMetricShardKeyFilters: Seq[ColumnFilter], timeRange: TimeRange): List[PartitionAssignment] =
+        testPartitions(timeRange)
```
alextheimer: Make this return a partition that doesn't match any provided by `getPartitions`, then you can copy-paste this and cover the 5 main test cases:
1. All shard-key `Equals` filters included. `getPartitions` returns one partition.
2. All shard-key `Equals` filters included. `getPartitions` returns duplicate partitions.
3. All shard-key `Equals` filters included. `getPartitions` returns >1 unique partitions.
4. One shard-key filter missing; expect getMetadataPartitions output.
5. One shard-key filter is not `Equals`; expect getMetadataPartitions output.

coordinator/src/test/scala/filodb.coordinator/queryplanner/ShardKeyRegexPlannerSpec.scala
```diff
@@ -1179,4 +1179,700 @@ class ShardKeyRegexPlannerSpec extends AnyFunSpec with Matchers with ScalaFuture
...
+    val childFilters = execPlan.children.flatMap(_.children.flatMap {
+      case mspe: MultiSchemaPartitionsExec => mspe.filters
+      case _ => Seq.empty
+    })
+
+    childFilters.count(f => f.column == "_ns_" && f.filter.valuesStrings.contains("App-1")) should be > 0
+    childFilters.count(f => f.column == "_ns_" && f.filter.valuesStrings.contains("App-2")) should be > 0
+  }
+
+  it("should handle multipartition aggregate queries with map-reduce correctly") {
```
alextheimer: Most of the tests below here don't test metadata functionality; these can be removed.
There is a comment above that describes how you can build the tests:
> you can copy-paste this and cover the 5 main test cases: [...]

coordinator/src/test/scala/filodb.coordinator/queryplanner/PlannerHierarchySpec.scala
```diff
@@ -4118,4 +4118,118 @@ class PlannerHierarchySpec extends AnyFunSpec with Matchers with PlanValidationS
...
+      val lp = Parser.labelValuesQueryToLogicalPlan(
+        Seq("instance"), Some("""_ws_="demo", _ns_="App-1""""), labelValuesTime
+      )
+
+      val execPlan = mppForMetadata.materialize(lp, QueryContext(origQueryParams = labelValuesParams))
+
+      // Should route to single partition from getPartitions (wrapped in concat exec)
+      execPlan.isInstanceOf[LabelValuesDistConcatExec] shouldEqual true
+      execPlan.asInstanceOf[LabelValuesDistConcatExec].children.size shouldEqual 2
+      execPlan.asInstanceOf[LabelValuesDistConcatExec].children.head.isInstanceOf[MetadataRemoteExec] shouldEqual false
```
alextheimer: Could you add a check that this hits only the expected partition? Without any a-priori understanding of how these plans are built, it is tough to understand whether-or-not the current check is meaningful.

coordinator/src/test/scala/filodb.coordinator/queryplanner/PlannerHierarchySpec.scala
```diff
@@ -4118,4 +4118,118 @@ class PlannerHierarchySpec extends AnyFunSpec with Matchers with PlanValidationS
...
+    }
+
+    it("should use getPartitions when all shard-key Equals filters are present and return duplicate partitions") {
```
alextheimer: Minor title fix: the expectation is that it deduplicates the partitions.

coordinator/src/test/scala/filodb.coordinator/queryplanner/PlannerHierarchySpec.scala
```diff
@@ -4118,4 +4118,118 @@ class PlannerHierarchySpec extends AnyFunSpec with Matchers with PlanValidationS
...
+                                          timeRange: TimeRange): List[PartitionAssignment] =
+          metadataPartitionProvider.getMetadataPartitions(nonMetricShardKeyFilters, timeRange)
+      }
+
+      val mppDuplicate = new MultiPartitionPlanner(
+        duplicatePartitionProvider, singlePartitionPlanner, "localPartition", dataset, queryConfig
+      )
+
+      val lp = Parser.labelValuesQueryToLogicalPlan(
+        Seq("instance"), Some("""_ws_="demo", _ns_="App-1""""), labelValuesTime
```
alextheimer: Use a `ShardKeyRegexPlanner` for this test. We should confirm that a `_ns_=~"foo|bar"` query results in a plan that hits only one partition even if the same `PartitionAssignment` is returned for each namespace (`foo` and `bar`).

coordinator/src/test/scala/filodb.coordinator/queryplanner/PlannerHierarchySpec.scala
```diff
@@ -4118,4 +4118,118 @@ class PlannerHierarchySpec extends AnyFunSpec with Matchers with PlanValidationS
...
+    it("should fallback to getMetadataPartitions when one shard-key filter is not Equals") {
+      val lp = Parser.labelValuesQueryToLogicalPlan(
+        Seq("instance"), Some("""_ws_=~"demo.*", _ns_="App-1""""), labelValuesTime
+      )
+
+      val execPlan = mppForMetadata.materialize(lp, QueryContext(origQueryParams = labelValuesParams))
+
+      // Should route to metadata partitions when _ws_ uses regex instead of Equals
+      execPlan.isInstanceOf[LabelValuesDistConcatExec] shouldEqual true
+      execPlan.asInstanceOf[LabelValuesDistConcatExec].children.size should be > 1
```
alextheimer: Same as mentioned in another comment: all of these should check the partition names directly. This is correct as-is, but it might hide unexpected behavior if anyone ever updates these tests.

coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
```diff
@@ -310,11 +310,8 @@ class MultiPartitionPlanner(val partitionLocationProvider: PartitionLocationProv
           val valueSet = acc.getOrElseUpdate(colFilter.column, new mutable.HashSet[String])
           colFilter.filter match {
             case eq: Equals => valueSet.add(eq.value.toString)
-            case re: EqualsRegex if QueryUtils.containsPipeOnlyRegex(re.value.toString) =>
-              val values = QueryUtils.splitAtUnescapedPipes(re.value.toString)
-              values.foreach(valueSet.add)
             case _ => throw new IllegalArgumentException(
-              s"""shard keys must be filtered by equality or "|"-only regex. filter=$colFilter""")
+              s"""shard keys must be filtered by equality. filter=$colFilter""")
```
alextheimer: Rollback these diffs; only `resolveMetadataPartitions` should be updated.

coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
```diff
@@ -1103,6 +1095,38 @@ class MultiPartitionPlanner(val partitionLocationProvider: PartitionLocationProv
...
+      LogicalPlan.getNonMetricShardKeyFilters(lp, dataset.options.shardKeyColumns)
+    val nonMetricCols = dataset.options.nonMetricShardColumns
+
+    def buildRoutingMap(group: Seq[ColumnFilter]): Map[String, String] =
+      nonMetricCols.flatMap(col =>
+        group.collectFirst {
+          case ColumnFilter(c, Equals(v: String)) if c == col => c -> v
+          case ColumnFilter(c, filodb.core.query.Filter.EqualsRegex(v: String)) if c == col => c -> v
+        }
+      ).toMap
```
alextheimer, line 1432: You can remove the `Equals[Regex]` constraints:
```suggestion
    def buildRoutingMap(group: Seq[ColumnFilter]): Map[String, String] =
      nonMetricCols.flatMap(shardKeyCol =>
        group.collectFirst {
          case ColumnFilter(c, filter) if c == shardKeyCol => c -> v
        }
      ).toMap
```

coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
```diff
@@ -1104,18 +1101,15 @@ class MultiPartitionPlanner(val partitionLocationProvider: PartitionLocationProv
...
-        group.collectFirst { case ColumnFilter(c, Equals(v: String)) if c == col => c -> v }.get
+      nonMetricCols.flatMap(col =>
+        group.collectFirst {
+          case ColumnFilter(c, Equals(v: String)) if c == col => c -> v
+          case ColumnFilter(c, filodb.core.query.Filter.EqualsRegex(v: String)) if c == col => c -> v
+        }
       ).toMap
 
-    val shouldFallback = shardKeyFilterGroups.isEmpty || shardKeyFilterGroups.exists(g => !hasAllNonMetricEquals(g))
+    val shouldFallback = queryConfig.routingConfig.forceMetadataRegexFallback || shardKeyFilterGroups.isEmpty
```
alextheimer: What do you think about this? `queryConfig.routingConfig.useLegacyMetadataRouting`.

coordinator/src/test/scala/filodb.coordinator/queryplanner/MultiPartitionPlannerSpec.scala
```diff
@@ -2041,11 +2041,24 @@ class MultiPartitionPlannerSpec extends AnyFunSpec with Matchers with PlanValida
       override def getMetadataPartitions(nonMetricShardKeyFilters: Seq[ColumnFilter], timeRange: TimeRange): List[PartitionAssignment] = ???
     }
     val mpp = new MultiPartitionPlanner(dummyPartitionLocationProvider, localPlanner, "local", dataset, queryConfig)
-    val lp = Parser.queryRangeToLogicalPlan("""foo{job=~"abc|def"} + foo{job="ghi"}""", TimeStepParams(1000, 100, 2000))
+    val lp = Parser.queryRangeToLogicalPlan("""foo{job="abc"} + foo{job="def"} + foo{job="ghi"}""", TimeStepParams(1000, 100, 2000))
```
alextheimer: Accidental diff?

coordinator/src/test/scala/filodb.coordinator/queryplanner/MultiPartitionPlannerSpec.scala
```diff
@@ -2041,11 +2041,24 @@ class MultiPartitionPlannerSpec extends AnyFunSpec with Matchers with PlanValida
...
     mpp.getRoutingKeys(lp) shouldEqual expected
   }
 
+  it ("should throw IllegalArgumentException for regex filters on shard keys") {
+    val dummyPartitionLocationProvider = new PartitionLocationProvider {
+      override def getPartitions(routingKey: Map[String, String], timeRange: TimeRange): List[PartitionAssignment] = ???
+      override def getMetadataPartitions(nonMetricShardKeyFilters: Seq[ColumnFilter], timeRange: TimeRange): List[PartitionAssignment] = ???
+    }
+    val mpp = new MultiPartitionPlanner(dummyPartitionLocationProvider, localPlanner, "local", dataset, queryConfig)
+    val lp = Parser.queryRangeToLogicalPlan("""foo{job=~"abc|def"}""", TimeStepParams(1000, 100, 2000))
```
alextheimer: Use metadata parser methods only; this PR does not change non-metadata routing logic.

coordinator/src/test/scala/filodb.coordinator/queryplanner/PlannerHierarchySpec.scala
```diff
@@ -3116,10 +3116,7 @@ class PlannerHierarchySpec extends AnyFunSpec with Matchers with PlanValidationS
         Nil
       }  // i.e. filters for a scalar
     }
-
-    // Case 7: top k with regex, the resolved regex should all be two remote partition, one using PromQLRemoteExec and other PromQLGrpcRemoteExec, should fail
     // TODO
-    // Case 8: top k with regex, the resolved regex should all be two remote partition, both use PromQLGrpcRemoteExec, should be supported
```
alextheimer, line 3125: Accidental diff?

coordinator/src/test/scala/filodb.coordinator/queryplanner/PlannerHierarchySpec.scala
```diff
@@ -4120,6 +4117,7 @@ class PlannerHierarchySpec extends AnyFunSpec with Matchers with PlanValidationS
   }
 
   describe("Metadata Query Routing - MultiPartitionPlanner") {
+    // Shared instances to prevent re-creation on each test (causes CI timeout)
```
alextheimer: PRBs timeout if these are created per-test?

coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
alextheimer: Looks like these diffs are still present

coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
```diff
@@ -1103,6 +1096,37 @@ class MultiPartitionPlanner(val partitionLocationProvider: PartitionLocationProv
...
+  private def resolveMetadataPartitions(lp: MetadataQueryPlan, queryParams: PromQlQueryParams) = {
+    val timeRange = TimeRange(queryParams.startSecs * 1000L, queryParams.endSecs * 1000L)
+    val shardKeyFilterGroups =
+      LogicalPlan.getNonMetricShardKeyFilters(lp, dataset.options.shardKeyColumns)
+    val nonMetricCols = dataset.options.nonMetricShardColumns
+
+    def buildRoutingMap(group: Seq[ColumnFilter]): Map[String, String] =
+      nonMetricCols.flatMap(shardKeyCol =>
+        group.collectFirst {
+          case ColumnFilter(c, filter) if c == shardKeyCol => c -> filter.value.toString
```
alextheimer: I was wrong about removing the regex case; we can't convert regex filters to `Map[String, String]` for the `PartitionLocationProvider`.

Typically in the `MultiPartitionPlanner`, we only need to handle either pipe-concatenated `=~` or plain `=` filters; the `ShardKeyRegexPlanner` resolves everything else into one of those two filters. But I see it is different for metadata queries: we pass all filters as-is down to this logic. I think that might be why we originally decided to just route these queries to all of the workspaces's partitions, because in that case we wouldn't need to figure out exactly which of those partitions are relevant to the query.

We need a `shardKeyMapper` to map a set of `ColumnFilter`s to sets of shard-keys. For example:
```
shardkey1=foo
shardkey2=~bar.*

-> [
  {shardKey1=foo, shardKey2=bar1},
  {shardKey1=foo, shardKey2=bar2},
  {shardKey1=foo, shardKey2=bar3}
]
```

But the `MultiPartitionPlanner` does not have a `shardKeyMapper` because the original intent was that the `ShardKeyRegexPlanner` would more-or-less pass only `=` filters down to the `MultiPartitionPlanner`.

We have two options:
1. Give a `shardKeyMapper` to `MultiPartitionPlanner`, and use it here.
2. Use the `shardKeyMapper` at the `ShardKeyRegexPlanner` to find all relevant shard-key sets, create a `LogicalPlan` for each, then pass all individually down to the `MultiPartitionPlanner` for materialization.

The former is probably simpler, but it doesn't jive with the original intent of the `MultiPartitionPlanner`. @amolnayak311, what do you think? I think we can stick with (1) for now, and we can iterate toward (2) if it's absolutely needed. The complexity of (2) is very likely not worth the effort.

amolnayak311: @alextheimer 2 is cleaner and now the MPP is even more cluttered and now the purpose of SKRP is even more questionable.  I would prefer 2 but if the complexity if high we cab stick with 1, but we are very likely not to revisit this again and this would stay forever. :)

alextheimer: Maybe it is better to take a third option for now, then we can handle this correctly when we have time: we can just require that all shard-key filters are `=`. We'll fallback to the legacy routing logic if any other filters arrive here.

coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
alextheimer: Remove these diffs @bhavanap1131-cmd

coordinator/src/main/scala/filodb.coordinator/queryplanner/MultiPartitionPlanner.scala
```diff
@@ -1103,6 +1096,48 @@ class MultiPartitionPlanner(val partitionLocationProvider: PartitionLocationProv
...
+        }
+      ).toMap
+
+    val shouldFallback = queryConfig.routingConfig.useLegacyMetadataRouting || !areEqualFilters ||
+      shardKeyFilterGroups.isEmpty
+
+    lp match {
+      case lc: LabelCardinality =>
```
alextheimer, line 1441: Could you add a note here that mentions this preserves the legacy routing behavior?

coordinator/src/test/scala/filodb.coordinator/queryplanner/MultiPartitionPlannerSpec.scala
alextheimer: @bhavanap1131-cmd Should this diff be here?

coordinator/src/test/scala/filodb.coordinator/queryplanner/MultiPartitionPlannerSpec.scala
alextheimer: @bhavanap1131-cmd use metadata parser methods only

coordinator/src/test/scala/filodb.coordinator/queryplanner/PlannerHierarchySpec.scala
```diff
@@ -4118,4 +4115,244 @@ class PlannerHierarchySpec extends AnyFunSpec with Matchers with PlanValidationS
...
+  }
+
+  it ("should give correct routing keys for Equals filters in multi-partition queries") {
+    val partitionProvider = new PartitionLocationProvider {
+      override def getPartitions(routingKey: Map[String, String], timeRange: TimeRange): List[PartitionAssignment] = ???
+      override def getMetadataPartitions(nonMetricShardKeyFilters: Seq[ColumnFilter],
+                                         timeRange: TimeRange): List[PartitionAssignment] = ???
+    }
+    val mpp = new MultiPartitionPlanner(partitionProvider, singlePartitionPlanner, "local", dataset, queryConfig)
+    val lp = Parser.queryRangeToLogicalPlan(
```
alextheimer: Some of these still use `queryRangeToLogicalPlan`; only parse metadata plans, else the new logic will not be tested.

coordinator/src/test/scala/filodb.coordinator/queryplanner/PlannerHierarchySpec.scala
```diff
@@ -4118,4 +4115,244 @@ class PlannerHierarchySpec extends AnyFunSpec with Matchers with PlanValidationS
...
+      """foo{_ws_="demo",_ns_="ns1"} + foo{_ws_="demo",_ns_="ns2"}""",
```
alextheimer: A metadata plan would not produce this PromQL; stick with simple [instant selectors](https://prometheus.io/docs/prometheus/latest/querying/basics/#instant-vector-selectors).

core/src/main/scala/filodb.core/query/QueryConfig.scala
```diff
@@ -92,8 +95,9 @@ case class RoutingConfig(
                           enableApproximatelyEqualCheckInStitch: Boolean = true,
                           periodOfUncertaintyMs: Long                    = (5 minutes).toMillis,
                           tenantsWithDisabledRemoteStitch: Set[String]   = Set.empty,
-                          stitchDisabledTenantColumn: String             = ""
-
+                          stitchDisabledTenantColumn: String             = "",
+                          forceMetadataRegexFallback: Boolean            = false,
```
alextheimer: Can this config be removed?

coordinator/src/test/scala/filodb.coordinator/queryplanner/PlannerHierarchySpec.scala
bhavanap1131-cmd: Done. Rewrote the test to use Parser.metadataQueryToLogicalPlan which produces a SeriesKeysByFilters plan, the actual MetadataQueryPlan type that flows through resolveMetadataPartitions.

coordinator/src/test/scala/filodb.coordinator/queryplanner/PlannerHierarchySpec.scala
bhavanap1131-cmd: Fixed. The test now uses a simple instant selector foo{_ns_=~"ns.*"},  metadataQueryToLogicalPlan only accepts InstantExpression so binary PromQL was never valid here.

core/src/main/scala/filodb.core/query/QueryConfig.scala
bhavanap1131-cmd: Yes, removed. It was only defined in RoutingConfig with a default of false and never referenced in any Scala, conf, JSON, or YAML files.

coordinator/src/test/scala/filodb.coordinator/queryplanner/PlannerHierarchySpec.scala
alextheimer: Could you re-add these comments?

coordinator/src/test/scala/filodb.coordinator/queryplanner/PlannerHierarchySpec.scala
```diff
@@ -4118,4 +4115,256 @@ class PlannerHierarchySpec extends AnyFunSpec with Matchers with PlanValidationS
...
+  }
+
+  it ("should throw IllegalArgumentException for regex filters on _ns_ in getRoutingKeys") {
+    val partitionProvider = new PartitionLocationProvider {
+      override def getPartitions(routingKey: Map[String, String], timeRange: TimeRange): List[PartitionAssignment] = ???
+      override def getMetadataPartitions(nonMetricShardKeyFilters: Seq[ColumnFilter],
+                                         timeRange: TimeRange): List[PartitionAssignment] = ???
+    }
+    val mpp = new MultiPartitionPlanner(partitionProvider, singlePartitionPlanner, "local", dataset, queryConfig)
+    val lp = Parser.queryRangeToLogicalPlan(
```
alextheimer: One last `queryRangeToLogicalPlan` should be changed to a metadata-producing method

coordinator/src/test/scala/filodb.coordinator/queryplanner/MultiPartitionPlannerSpec.scala
```diff
@@ -2680,5 +2692,92 @@ class MultiPartitionPlannerSpec extends AnyFunSpec with Matchers with PlanValida
     }
   }
 
+  describe("resolveMetadataPartitions fallback behavior") {
```
alextheimer, line 2729: Could you add one case where a shard-key filter is missing?

coordinator/src/test/scala/filodb.coordinator/queryplanner/ShardKeyRegexPlannerSpec.scala
```diff
@@ -1179,4 +1179,59 @@ class ShardKeyRegexPlannerSpec extends AnyFunSpec with Matchers with ScalaFuture
       validatePlan(execPlan.children.head, expected)
     }
   }
+
+  // Metadata routing tests for MultiPartitionPlanner
+  describe("Metadata Routing - MultiPartitionPlanner") {
```
alextheimer: What do you think about omitting these tests, since the functionality is already sufficiently tested in `MultiPartitionPlannerSpec`?

coordinator/src/test/scala/filodb.coordinator/queryplanner/ShardKeyRegexPlannerSpec.scala
```diff
@@ -1179,4 +1179,59 @@ class ShardKeyRegexPlannerSpec extends AnyFunSpec with Matchers with ScalaFuture
...
+        Seq("instance"), Some("""_ws_="demo""""), labelValuesTime
+      )
+
+      val execPlan = mppForMetadata.materialize(lp, QueryContext(origQueryParams = labelValuesParams))
+
+      execPlan.isInstanceOf[LabelValuesDistConcatExec] shouldEqual true
+      execPlan.children.size shouldEqual 32
+    }
+
+    it("should handle label values query without complete shard keys") {
```
alextheimer: This is identical to the above test.

coordinator/src/test/scala/filodb.coordinator/queryplanner/PlannerHierarchySpec.scala
```diff
@@ -4118,4 +4115,256 @@ class PlannerHierarchySpec extends AnyFunSpec with Matchers with PlanValidationS
...
+    val labelValuesTime = TimeStepParams(1000, 20, 5000)
+
+    def extractMetadataEndpoints(plan: ExecPlan): Seq[String] = plan match {
+      case m: MetadataRemoteExec => Seq(m.queryEndpoint)
+      case c: LabelValuesDistConcatExec =>
+        c.children.collect { case m: MetadataRemoteExec => m.queryEndpoint }
+      case _ => Seq.empty
+    }
+
+    it("should use getPartitions when all shard-key Equals filters are present and route to single partition") {
```
alextheimer: What do you think about moving this to `MultiPartitionPlannerSpec` (or removing it if it's a duplicate there)? This file is typically used for tests where the root-level planner (`ShardKeyRegexPlanner`) is invoked to build a plan, but this test invokes the `MultiPartitionPlanner` directly.

coordinator/src/test/scala/filodb.coordinator/queryplanner/PlannerHierarchySpec.scala
```diff
@@ -4118,4 +4115,256 @@ class PlannerHierarchySpec extends AnyFunSpec with Matchers with PlanValidationS
...
+      // Should route to local partitions when complete shard-key filters are present
+      execPlan.isInstanceOf[LabelValuesDistConcatExec] shouldEqual true
+      val concatExec = execPlan.asInstanceOf[LabelValuesDistConcatExec]
+      // Should have 2 children (shard 0 and shard 1 from SinglePartitionPlanner)
+      concatExec.children.size shouldEqual 2
+      // All children should be LabelValuesExec (not remote)
+      concatExec.children.foreach(child => child.isInstanceOf[LabelValuesExec] shouldEqual false)
+    }
+
+    it("should use getPartitions when all shard-key Equals filters are present and deduplicate partitions") {
```
alextheimer: What do you think about moving this to `MultiPartitionPlannerSpec` (or removing it if it's a duplicate there)? This file is typically used for tests where the root-level planner (`ShardKeyRegexPlanner`) is invoked to build a plan, but this test invokes the `MultiPartitionPlanner` directly.

coordinator/src/test/scala/filodb.coordinator/queryplanner/PlannerHierarchySpec.scala
```diff
@@ -4118,4 +4115,256 @@ class PlannerHierarchySpec extends AnyFunSpec with Matchers with PlanValidationS
...
+      // Should route to local partitions (duplicate partitions are deduplicated at partition provider level)
+      execPlan.isInstanceOf[LabelValuesDistConcatExec] shouldEqual true
+      val concatExec = execPlan.asInstanceOf[LabelValuesDistConcatExec]
+      // Even with duplicate partition assignment, routes to local partition
+      concatExec.children.size shouldEqual 2
+      // All children should be LabelValuesExec
+      concatExec.children.foreach(child => child.isInstanceOf[LabelValuesExec] shouldEqual false)
+    }
+
+    it("should use getPartitions when all shard-key Equals filters are present and route to multiple unique partitions") {
```
alextheimer: What do you think about moving this to `MultiPartitionPlannerSpec` (or removing it if it's a duplicate there)? This file is typically used for tests where the root-level planner (`ShardKeyRegexPlanner`) is invoked to build a plan, but this test invokes the `MultiPartitionPlanner` directly.

coordinator/src/test/scala/filodb.coordinator/queryplanner/PlannerHierarchySpec.scala
```diff
@@ -4118,4 +4115,256 @@ class PlannerHierarchySpec extends AnyFunSpec with Matchers with PlanValidationS
...
+      // Should create concat exec with multiple partitions - actual behavior is LabelValuesDistConcatExec
+      execPlan.isInstanceOf[LabelValuesDistConcatExec] shouldEqual true
+      val concatExec = execPlan.asInstanceOf[LabelValuesDistConcatExec]
+      // Should have children for each shard (2 shards from local partition)
+      concatExec.children.size should be > 0
+      // All children should be LabelValuesExec
+      concatExec.children.foreach(child => child.isInstanceOf[LabelValuesExec] shouldEqual false)
+    }
+
+    it("should fallback to getMetadataPartitions when one shard-key filter is missing") {
```
alextheimer: What do you think about moving this to `MultiPartitionPlannerSpec` (or removing it if it's a duplicate there)? This file is typically used for tests where the root-level planner (`ShardKeyRegexPlanner`) is invoked to build a plan, but this test invokes the `MultiPartitionPlanner` directly.

coordinator/src/test/scala/filodb.coordinator/queryplanner/PlannerHierarchySpec.scala
```diff
@@ -4118,4 +4115,256 @@ class PlannerHierarchySpec extends AnyFunSpec with Matchers with PlanValidationS
...
+      val execPlan = skrpRegex.materialize(lp, QueryContext(origQueryParams = labelValuesParams))
+
+      // When regex filter used, should fallback to metadata partitions
+      execPlan.isInstanceOf[LabelValuesDistConcatExec] shouldEqual true
+      val concatExec = execPlan.asInstanceOf[LabelValuesDistConcatExec]
+      concatExec.children.size should be >= 1
+    }
+  }
+
+  it ("should fall back to legacy metadata routing when non-equals filter is present") {
```
alextheimer: What do you think about moving this to `MultiPartitionPlannerSpec` (or removing it if it's a duplicate there)? This file is typically used for tests where the root-level planner (`ShardKeyRegexPlanner`) is invoked to build a plan, but this test invokes the `MultiPartitionPlanner` directly.

coordinator/src/test/scala/filodb.coordinator/queryplanner/PlannerHierarchySpec.scala
```diff
@@ -4118,4 +4115,256 @@ class PlannerHierarchySpec extends AnyFunSpec with Matchers with PlanValidationS
...
+      """foo{_ns_=~"ns.*"}""",
+      TimeStepParams(startSeconds, step, endSeconds))
+    val metadataQueryParams = PromQlQueryParams("notUsedQuery", startSeconds, step, endSeconds)
+    mpp.materialize(lp, QueryContext(origQueryParams = metadataQueryParams,
+      plannerParams = PlannerParams(processMultiPartition = true)))
+    legacyPathUsed shouldEqual true
+    directPathUsed shouldEqual false
+  }
+
+  it ("should throw IllegalArgumentException for regex filters on _ns_ in getRoutingKeys") {
```
alextheimer: What do you think about moving this to `MultiPartitionPlannerSpec` (or removing it if it's a duplicate there)? This file is typically used for tests where the root-level planner (`ShardKeyRegexPlanner`) is invoked to build a plan, but this test invokes the `MultiPartitionPlanner` directly.

coordinator/src/test/scala/filodb.coordinator/queryplanner/PlannerHierarchySpec.scala
```diff
@@ -4118,4 +4115,256 @@ class PlannerHierarchySpec extends AnyFunSpec with Matchers with PlanValidationS
...
+      TimeStepParams(startSeconds, step, endSeconds), Antlr)
+    // getRoutingKeys should throw an exception for regex filters as only Equals filters are supported
+    // Pipe-only regexes (e.g. "ns1|ns2") are handled specially, but wildcard regexes should throw.
+    // Regex filters on shard keys are handled by ShardKeyRegexPlanner which wraps  MultiPartitionPlanner
+    an[IllegalArgumentException] should be thrownBy {
+      mpp.getRoutingKeys(lp)
+    }
+  }
+
+  it ("should handle non-equals filters gracefully in getPartitions delegation") {
```
alextheimer: What do you think about moving this to `MultiPartitionPlannerSpec` (or removing it if it's a duplicate there)? This file is typically used for tests where the root-level planner (`ShardKeyRegexPlanner`) is invoked to build a plan, but this test invokes the `MultiPartitionPlanner` directly.

coordinator/src/test/scala/filodb.coordinator/queryplanner/PlannerHierarchySpec.scala
```diff
@@ -4118,4 +4115,256 @@ class PlannerHierarchySpec extends AnyFunSpec with Matchers with PlanValidationS
...
+      val execPlan = mppForMetadata.materialize(lp, QueryContext(origQueryParams = labelValuesParams))
+
+      // When filter is missing, should fallback to metadata partitions
+      execPlan.isInstanceOf[LabelValuesDistConcatExec] shouldEqual true
+      val concatExec = execPlan.asInstanceOf[LabelValuesDistConcatExec]
+      // Should route to local partitions when metadata query routes to local
+      concatExec.children.size should be >= 1
+    }
+
+    it("should fallback to getMetadataPartitions when one shard-key filter is not Equals") {
```
alextheimer: This is the only test in `PlannerHierarchySpec` that queries through a root `ShardKeyRegexPlanner`, and I think it's probably the only one you need in this file. Could you rename the test to this, instead, just to prevent any connotation with `!=`?:
```
"should fallback to getMetadataPartitions when one shard-key filter is regex"
```

coordinator/src/test/scala/filodb.coordinator/queryplanner/MultiPartitionPlannerSpec.scala
```diff
@@ -2778,6 +2778,68 @@ class MultiPartitionPlannerSpec extends AnyFunSpec with Matchers with PlanValida
...
+        TimeStepParams(startSeconds, step, endSeconds))
+      val metadataQueryParams = PromQlQueryParams("notUsedQuery", startSeconds, step, endSeconds)
+      val execPlan = engine.materialize(lp, QueryContext(origQueryParams = metadataQueryParams,
+        plannerParams = PlannerParams(processMultiPartition = true)))
+
+      execPlan.isInstanceOf[MetadataRemoteExec] shouldEqual true
+      execPlan.asInstanceOf[MetadataRemoteExec].queryEndpoint shouldEqual "direct-remote-url"
+    }
+
+    it("should handle non-equals filters gracefully in getPartitions delegation") {
```
alextheimer: What non-equals filter is used here?

coordinator/src/test/scala/filodb.coordinator/queryplanner/MultiPartitionPlannerSpec.scala
```diff
@@ -2778,6 +2778,68 @@ class MultiPartitionPlannerSpec extends AnyFunSpec with Matchers with PlanValida
...
+      val lp = Parser.metadataQueryToLogicalPlan(
+        """foo{_ws_="demo"}""",
+        TimeStepParams(startSeconds, step, endSeconds))
+      val metadataQueryParams = PromQlQueryParams("notUsedQuery", startSeconds, step, endSeconds)
+      val execPlan = engine.materialize(lp, QueryContext(origQueryParams = metadataQueryParams,
+        plannerParams = PlannerParams(processMultiPartition = true)))
+
+      execPlan.isInstanceOf[MetadataRemoteExec] shouldEqual true
+      execPlan.asInstanceOf[MetadataRemoteExec].queryEndpoint shouldEqual "direct-remote-url"
+    }
```
alextheimer: These two tests show the behavior is incorrect; `getPartitions` should only be invoked if:
- all shard-key filters are present.
- all shard-key filters are `Equals`.

Looks like we're missing one `shouldFallback` check in `resolveMetadataPartitions`.

coordinator/src/test/scala/filodb.coordinator/queryplanner/MultiPartitionPlannerSpec.scala
```diff
@@ -2593,6 +2593,18 @@ class MultiPartitionPlannerSpec extends AnyFunSpec with Matchers with PlanValida
...
+    val dummyPartitionLocationProvider = new PartitionLocationProvider {
+      override def getPartitions(routingKey: Map[String, String], timeRange: TimeRange): List[PartitionAssignment] = ???
+      override def getMetadataPartitions(nonMetricShardKeyFilters: Seq[ColumnFilter], timeRange: TimeRange): List[PartitionAssignment] = ???
+    }
+    val mpp = new MultiPartitionPlanner(dummyPartitionLocationProvider, localPlanner, "local", dataset, queryConfig)
+    val lp = Parser.metadataQueryToLogicalPlan("""foo{job=~"abc|def"}""", TimeStepParams(1000, 100, 2000))
+    noException should be thrownBy {
+      mpp.getRoutingKeys(lp)
+    }
+  }
```
alextheimer: This test's name and its assertion contradict each other.

coordinator/src/test/scala/filodb.coordinator/queryplanner/MultiPartitionPlannerSpec.scala
bhavanap1131-cmd: The test didn't actually use any non-equals filter, it called getPartitionsTrait 
 directly with a plain routing map, bypassing resolveMetadataPartitions entirely. The test was hollow
 and the description was misleading. Removed it.

coordinator/src/test/scala/filodb.coordinator/queryplanner/MultiPartitionPlannerSpec.scala
bhavanap1131-cmd: Fixed. Added the missing allShardKeysPresent check to shouldFallback in resolveMetadataPartitions:
 val shouldFallback = queryConfig.routingConfig.useLegacyMetadataRouting || !areEqualFilters ||
 shardKeyFilterGroups.isEmpty ||
  !shardKeyFilterGroups.forall(group => nonMetricCols.forall(col => group.exists(_.column == col)))
 Both tests now correctly assert that a missing shard-key filter falls back to getMetadataPartitions 
  (legacy-remote-url), not getPartitions.

coordinator/src/test/scala/filodb.coordinator/queryplanner/MultiPartitionPlannerSpec.scala
bhavanap1131-cmd: Fixed. The name was a leftover from when the test used queryRangeToLogicalPlan and expected an
 IllegalArgumentException. It was later changed to use metadataQueryToLogicalPlan with
   noException but the description was never updated. Renamed to "should not throw for
  regex filters on shard keys in metadata queries

coordinator/src/test/scala/filodb.coordinator/queryplanner/PlannerHierarchySpec.scala
bhavanap1131-cmd: Analyzed, found to have broken assertions (shouldEqual false contradicting the comments), missing  
  processMultiPartition = true, and never actually invoking the partition provider
  Confirmed as duplicates - the non-fallback behavior is already properly covered by "should use      
  getPartitions directly when all shard key filters are Equals" in MultiPartitionPlannerSpec
  Deleted from PlannerHierarchySpec

coordinator/src/test/scala/filodb.coordinator/queryplanner/PlannerHierarchySpec.scala
bhavanap1131-cmd: Analyzed, found to have broken assertions (shouldEqual false contradicting the comments), missing  
  processMultiPartition = true, and never actually invoking the partition provider
  Confirmed as duplicates - the non-fallback behavior is already properly covered by "should use      
  getPartitions directly when all shard key filters are Equals" in MultiPartitionPlannerSpec
  Deleted from PlannerHierarchySpec

coordinator/src/test/scala/filodb.coordinator/queryplanner/PlannerHierarchySpec.scala
bhavanap1131-cmd: Analyzed, found to have broken assertions (shouldEqual false contradicting the comments), missing  
  processMultiPartition = true, and never actually invoking the partition provider
  Confirmed as duplicates - the non-fallback behavior is already properly covered by "should use      
  getPartitions directly when all shard key filters are Equals" in MultiPartitionPlannerSpec
  Deleted from PlannerHierarchySpec

coordinator/src/test/scala/filodb.coordinator/queryplanner/PlannerHierarchySpec.scala
```diff
@@ -4118,4 +4118,72 @@ class PlannerHierarchySpec extends AnyFunSpec with Matchers with PlanValidationS
...
+      val lp = Parser.labelValuesQueryToLogicalPlan(
+        Seq("instance"), Some("""_ws_=~"demo.*", _ns_="App-1""""), labelValuesTime
+      )
+
+      val execPlan = skrpRegex.materialize(lp, QueryContext(origQueryParams = labelValuesParams))
+
+      // When regex filter used, should fallback to metadata partitions
+      execPlan.isInstanceOf[LabelValuesDistConcatExec] shouldEqual true
+      val concatExec = execPlan.asInstanceOf[LabelValuesDistConcatExec]
+      concatExec.children.size should be >= 1
```
alextheimer: This should be `> 1` (since `getPartitions` returns `1`).

alextheimer (commented):
Huge thanks for all the iteration @bhavanap1131-cmd 
This PR can be merged after this one last comment is addressed.

coordinator/src/test/scala/filodb.coordinator/queryplanner/PlannerHierarchySpec.scala
bhavanap1131-cmd: Yes, corrected it.

alextheimer (approved):

amolnayak311 (approved):

alextheimer (commented):
It looks like some tests are still missing. What about `should not throw for
regex filters on shard keys in metadata queries`? That is mentioned in a comment just prior to the force-push.

Couple other tests that look like they are missing (just a few I see in the PR comments):
- should use getPartitions directly when a shard-key filter is missing
- should fallback to getMetadataPartitions when one shard-key filter is regex
- should handle non-equals filters gracefully in getPartitions delegation

There might be others, too. I would recommend using `git reflog` to find the state of the repo immediately prior to [this commit](https://github.com/filodb/FiloDB/pull/2130/changes/d58ebc589256bdf5151a40205c06d601ecb6e87a). You can open a separate PR with that state, and I'll take a look. If that looks good, we can hard-reset the state of the original PR to that branch.

alextheimer (approved):
Thanks for the commit surgery @bhavanap1131-cmd 🙏

PR #2193, opened 2026-04-09
feat(core): TimeSeriesShard billable cardinalities

alextheimer:
**Pull Request checklist**

Currently, `TimeSeriesShard` increments/decrements cardinality counts by `1` for all series.

This does not accurately reflect the resource-utilization weight of heavier series (i.e. histograms).

**New behavior :**

This PR updates `TimeSeriesShard` to instead count "billable" cardinalities; each series' "physical" active cardinality is multiplied by a configurable factor intended to reflect its resource-utilization weight.

For example, with a configuration such as this:
```
filodb {
  schema-name-to-billable-cardinality-per-active-series {
    prom-histogram = 15
  }
}
```
-- billable cardinalities would increment/decrement by `15` for `prom-histogram`s.

**BREAKING CHANGES**

Breaking changes may include:

**Other information**:

### DIFF
--- a/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
+++ b/core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
@@ -312,12 +312,13 @@ class TimeSeriesShard(val ref: DatasetRef,
   private val partKeyUpdatesPublishingEnabled = filodbConfig.getBoolean("memstore.index-updates-publishing-enabled")
 
   /**
-    * Active cardinality counts are updated according to the multipliers defined here.
-    * For example: if one prom-histogram series maps to 25 "virtual" series, then cardinality
-    *   counters will increment/decrement by 25 for each series accounted/unaccounted for.
+   * Billable cardinality counts are updated according to the multipliers returned here.
+   * For example: if one prom-histogram series maps to 25 "billable" series per "physical"
+   *   active series, then cardinality counters will increment/decrement by 25 for
+   *   each series accounted/unaccounted for.
     */
-  private val schemaNameToVirtualActiveCardinalityPerSeries =
-    filodbConfig.getConfig("cardinality.schema-name-to-virtual-active-cardinality-per-series")
+  val schemaNameToBillableCardinalityPerActiveSeries =
+    filodbConfig.getConfig("cardinality.schema-name-to-billable-cardinality-per-active-series")
       .entrySet().asScala.map { entry =>
         entry.getKey -> entry.getValue.unwrapped.asInstanceOf[Int]
       }.toMap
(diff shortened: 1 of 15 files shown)


vishramachandran:
We may need to publish two metrics - actual time series and  billed time series. There is a need to see both.

core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityTracker.scala
```diff
@@ -76,10 +85,17 @@ class CardinalityTracker(ref: DatasetRef,
...
+    require(
+      activeDelta == 0 && billableDelta == 0
+        || activeDelta == 1 && billableDelta > 0
+        || activeDelta == -1 && billableDelta < 0,
+      s"Invalid values for activeDelta / billableDelta; must have the same sign: " +
+      s"activeDelta=$activeDelta; billableDelta=$billableDelta; shardKey=$shardKey"
+    )
 
     flushCount match {
       case Some(threshold) => modifyCountWithAggregation(shardKey, threshold, totalDelta)
```
sandeep6189, line 97: Quick Question: we don't need billableDelta for this function modifyCountWithAggregation ? ( This function uses a Map to aggregate in memory card changes before writing to RocksDB )

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
```diff
@@ -311,6 +311,18 @@ class TimeSeriesShard(val ref: DatasetRef,
   // Configuration to enable/disable real-time publishing of index updates for downstream consumers.
   private val partKeyUpdatesPublishingEnabled = filodbConfig.getBoolean("memstore.index-updates-publishing-enabled")
 
+  /**
+   * Billable cardinality counts are updated according to the multipliers returned here.
+   * For example: if one prom-histogram series maps to 25 "billable" series per "physical"
+   *   active series, then cardinality counters will increment/decrement by 25 for
+   *   each series accounted/unaccounted for.
+    */
+  val schemaNameToBillableCardinalityPerActiveSeries =
```
sandeep6189, line 320: this is why we need `schema` ?

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
```diff
@@ -485,6 +485,7 @@ final case object TsCardExec {
    */
   val RESULT_SCHEMA = ResultSchema(Seq(ColumnInfo("group", ColumnType.StringColumn),
                                        ColumnInfo("active", ColumnType.LongColumn),
+                                       ColumnInfo("billable", ColumnType.LongColumn),
```
sandeep6189: Should we move to last ? ( Downstream services might need to be updated here for TSCardinalities query )

core/src/main/scala/filodb.core/memstore/ratelimit/CardinalityTracker.scala
alextheimer: Nope, it is not needed. From the method docs:
> NOTE: We are only cardinality count for total TS in aggregated fashion. We will add support for active TS if needed

The method also only accepts a "total" delta; it does not accept any "active" delta argument.

core/src/main/scala/filodb.core/memstore/TimeSeriesShard.scala
alextheimer: Yep, that's correct 👍

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
alextheimer: Good point -- updated: https://github.com/filodb/FiloDB/pull/2193/changes/f384d32262981631472727e4370656fb79bdb600

sandeep6189 (approved):

PR #2194, opened 2026-04-11
fix(core): support metric names with unicode characters

alextheimer:
**Pull Request checklist**

Currently, remote requests fail for valid PromQL queries such as:
```
{__name__="foo bar baz",label1="value1"}`
```
This happens because, when the query `LogicalPlan` is converted back into PromQL prior to remote execution, metric names are prepended as i.e.:
```
foo bar baz{label1="value1"}`
```
This is not the intended behavior; the string "foo bar baz" should represent a single metric name.

**New behavior :**

This PR updates the `LogicalPlanParser`'s `filtersToQuery` method to instead conditionally translate a `LogicalPlan`'s metric name into a typical `__name__="foo"` filter. The name-to-explicit-filter translation happens when the metric name does not match a regex; the regex tests whether-or-not the name is valid to prefix selector curly-braces.

After this change, metric names with any arbitrary unicode characters will be supported by remote-execution queries. 

**BREAKING CHANGES**

Breaking changes may include:

**Other information**:


coordinator/src/main/scala/filodb.coordinator/queryplanner/LogicalPlanParser.scala
```diff
@@ -36,20 +36,25 @@ object LogicalPlanParser {
 
   private def filtersToQuery(filters: Seq[(String, String, String)], columns: Seq[String],
```
alextheimer, line 43: Alternatively, to reduce the size of the diff, we could check metric names against a regex matcher for valid "prepend" characters. E.g.:
```
foobar123    // Matches! Metric name is valid to prepend to `{}` filters.
foo bar baz  // Doesn't match; metric name must be included in explicit `__name__` filter.
```
Matching metric names would be prepended; non-matching names would be added as an explicit `__name__` filter.

I went this route (everything's a `__name__` filter) for consistency's sake, but that isn't a hard requirement.

yu-shipit (approved):

PR #2198, opened 2026-04-17
add ts_of_(last|min|max)_over_time range functions

sherali42:
**Pull Request checklist**

These three new PromQL range functions return the timestamp (in seconds) of a specific sample within a time window, 

- ts_of_last_over_time: timestamp of last non-NaN sample
- ts_of_min_over_time: timestamp of sample with minimum value
- ts_of_max_over_time: timestamp of sample with maximum value

Each function has both sliding window (iterating) and chunked implementations. The chunked variants extend ChunkedRangeFunction[TransientRow] directly to access both timestamp and value vectors.

Added test case to catch regression issues parsing `time() + sum(rate(foo{_ws_=\"demo\", _ns_=\"App-0\"}[1i]))`

Files modified:
- PlanEnums.scala: RangeFunctionId entries
- InternalRangeFunction.scala: internal enum + mapping
- RangeFunction.scala: 6 implementation classes + factory registration
- query_service.proto: protobuf enum values 31-33
- ProtoConverters.scala: proto serialization mappings
- AggrOverTimeFunctionsSpec.scala: unit tests
- ParserSpec.scala: parser validation tests


query/src/main/scala/filodb/query/exec/rangefn/RangeFunction.scala
```diff
@@ -588,6 +599,88 @@ class TimestampIteratingFunction extends RangeFunction[TransientRow] {
   }
 }
 
+/**
+ * Returns the timestamp (in seconds) of the last non-NaN sample in the window.
+ * Like LastSampleFunction but returns the timestamp instead of the value.
+ */
+class TsOfLastSampleFunction extends RangeFunction[TransientRow] {
```
vishramachandran, line 606: Instead of rewriting the entire range function, can we reuse LastSampleFunction that already exists to emit the ts instead of value ? It can be done with a class parameter that toggles between emitting value vs ts. 

This 
* reuses code and removes code duplication
* ensures that the same logic exists in both functions

vishramachandran: Similar comment for all other functions, including chunked functions as well. The reason I am particular about this is that I had submitted a fix for LastSample logic to match Prometheus ( https://github.com/filodb/FiloDB/pull/2181 ), and it is not clear if that has been incorporated in the new functions.  It is easier to validate and maintain if the logic is reused. 

Please look at this fix PR as well https://github.com/filodb/FiloDB/pull/2178

vishramachandran (changes requested):
@sherali42 I have an important suggestion. In addition, I had found that 

```
(
  time() - 
  ts_of_last_over_time(
    sum by (cluster) (
      your_metric
    )[24h:]
  )
) > 300
```

... such a query does not parse with our antlr grammar. We need to fix that as well.
I leave it to you to see if that will be a separate PR or included in this one. A unit test will be necessary.

vishramachandran:
This should be closed since @yu-shipit  merged a PR for the same feature

PR #2214, opened 2026-04-28
Add partial result detection to metadata queries when shard limit is exceeded (Metadata limits PR2)

bhavanap1131-cmd:
Summary:
Metadata queries (/series, /labels, /label/values, label cardinality) previously had no way to signal that results were truncated when a shard hit its enforced limit. This PR propagates QuerySession into the metadata query paths so shards can flag resultCouldBePartial = true when the result count equals the limit, allowing callers to send a partial results warning to the user.
  
New Changes:
 - Added querySession: QuerySession parameter to TimeSeriesStore.partKeysWithFilters and its implementations in TimeSeriesShard and DownsampledTimeSeriesShard
 - Both shard implementations now set querySession.resultCouldBePartial = true with a reason when result.length == limit across all fetch paths (with/without first-last sample times, series scan, label values scan)
 - LabelValuesExec (no-filter path) sets the partial flag when returned label count equals the limit
 - LabelCardinalityExec replaces the hardcoded 1000000 cap with queryContext.plannerParams.enforcedLimits.execPlanLeafSamples to be consistent with the per-query limits.

### DIFF
--- a/core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
+++ b/core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
@@ -162,8 +162,14 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
                           fetchFirstLastSampleTimes: Boolean,
                           endTime: Long,
                           startTime: Long,
-                          limit: Int): Iterator[Map[ZeroCopyUTF8String, ZeroCopyUTF8String]] = {
-    partKeyIndex.partKeyRecordsFromFilters(filter, startTime, endTime, limit).iterator.map { pk =>
+                          limit: Int,
+                          querySession: QuerySession): Iterator[Map[ZeroCopyUTF8String, ZeroCopyUTF8String]] = {
+    val result = partKeyIndex.partKeyRecordsFromFilters(filter, startTime, endTime, limit)
+    if (result.length == limit) {
+      querySession.resultCouldBePartial = true
+      querySession.partialResultsReason = Some("Result may be partial since some shards exceeded the query limit")
+    }
+    result.iterator.map { pk =>
       val partKey = PartKeyWithTimes(pk.partKey, UnsafeUtils.arayOffset, pk.startTime, pk.endTime)
       schemas.part.binSchema.toStringPairs(partKey.base, partKey.offset).map(pair => {
         pair._1.utf8 -> pair._2.utf8
@@ -558,6 +564,7 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
 
       val rows = new mutable.HashSet[Map[ZeroCopyUTF8String, ZeroCopyUTF8String]]()
       var partLoopIndex = 0
+      var limitReached = false
       val matched = partKeyIndex.foreachPartKeyMatchingFilter(filters, startTime, endTime,
         nextPart => {
           if (rows.size < limit) {
@@ -573,9 +580,16 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
 
             if (currVal.nonEmpty) rows.add(currVal)
             partLoopIndex += 1
-          } else throw new CollectionTerminatedException
+          } else {
+            limitReached = true
+            throw new CollectionTerminatedException
+          }
         }
       )
+      if (limitReached) {
+        querySession.resultCouldBePartial = true
+        querySession.partialResultsReason = Some("Result may be partial since some shards exceeded the query limit")
+      }
       querySession.queryStats.getTimeSeriesScannedCounter(statsGroup).addAndGet(matched)
       rows.toIterator
     }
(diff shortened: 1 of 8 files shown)


core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
```diff
@@ -162,8 +162,14 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
...
                           endTime: Long,
                           startTime: Long,
-                          limit: Int): Iterator[Map[ZeroCopyUTF8String, ZeroCopyUTF8String]] = {
-    partKeyIndex.partKeyRecordsFromFilters(filter, startTime, endTime, limit).iterator.map { pk =>
+                          limit: Int,
+                          querySession: QuerySession): Iterator[Map[ZeroCopyUTF8String, ZeroCopyUTF8String]] = {
+    val result = partKeyIndex.partKeyRecordsFromFilters(filter, startTime, endTime, limit)
+    if (result.length == limit) {
+      querySession.resultCouldBePartial = true
+      querySession.partialResultsReason = Some("Result may be partial since some shards exceeded the query limit")
```
alextheimer: I'm surprised we have a "could be partial" flag and not an "is partial" flag 🤔

Still, since in 99.999% of cases the limits would genuinely have been exceeded, it might make sense to return a stronger claim so the "may" isn't confusing/concerning:
> Some shards returned a result size greater than $limit.

And the message should cue the user to their next steps, so it might be worth adding something like:
> Some shards returned a result size greater than $limit; apply more filters or reduce the query time-range.

alextheimer: The same also applies to other messages included with this PR.

core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
```diff
@@ -162,8 +162,14 @@ class DownsampledTimeSeriesShard(rawDatasetRef: DatasetRef,
                           fetchFirstLastSampleTimes: Boolean,
                           endTime: Long,
                           startTime: Long,
-                          limit: Int): Iterator[Map[ZeroCopyUTF8String, ZeroCopyUTF8String]] = {
-    partKeyIndex.partKeyRecordsFromFilters(filter, startTime, endTime, limit).iterator.map { pk =>
+                          limit: Int,
+                          querySession: QuerySession): Iterator[Map[ZeroCopyUTF8String, ZeroCopyUTF8String]] = {
+    val result = partKeyIndex.partKeyRecordsFromFilters(filter, startTime, endTime, limit)
+    if (result.length == limit) {
```
alextheimer, line 168: Should these iterators also have a similar check?:
```
SingleLabelValuesResultIterator
LabelValueResultIterator
```

core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
bhavanap1131-cmd: Updated all partial result reason messages across the PR to use the below format:
  s"Some shards returned a result size greater than $limit; apply more filters or reduce the query time-range."
 Applied consistently to all locations in DownsampledTimeSeriesShard, TimeSeriesShard, and MetadataExecPlan.

core/src/main/scala/filodb.core/downsample/DownsampledTimeSeriesShard.scala
bhavanap1131-cmd: Added resultCouldBePartial checks to both iterators. Two cases addressed:
   - DownsampledTimeSeriesShard: Both iterators already used a limitReached flag via
  CollectionTerminatedException, added the resultCouldBePartial check in the existing if
  (limitReached) block.
  - TimeSeriesShard (in-memory path): Both iterators used a while loop that stopped
   at the limit with no partial flag set - added if (rows.size == limit) check after eachloop to set the flag. Tests added for both paths in MetadataExecSpec.

query/src/main/scala/filodb/query/exec/MetadataExecPlan.scala
```diff
@@ -439,13 +441,22 @@ final case class LabelCardinalityExec(queryContext: QueryContext,
...
         if (labelNames.nonEmpty) {
           val sketchMap = scala.collection.mutable.Map[String, CpcSketch]()
+          val leafLimit = queryContext.plannerParams.enforcedLimits.execPlanLeafSamples
           labelNames.foreach { case label =>
             // GOTCHA: This approach will not catch cardinality of labels which are disabled for faceting
             // since their value lengths are > 1000. We expect the gateway to reject (or shorten) that data early on.
+            var count = 0
             memstore.singleLabelValueWithFilters(dataset, shard, filters, label.toString,
               endMs, startMs, querySession,
               queryContext.plannerParams.enforcedLimits.execPlanLeafSamples).foreach { labelValue =>
```
alextheimer: Use the `leafLimit` variable here, too, so it's clear the same limit is passed to this method.

alextheimer (approved):
Just one non-blocking comment 👍

alextheimer (commented):
Could you restore this PR with `git reflog`, as well?

PR #2219, opened 2026-05-01
fix(index): Fixing name collision for label `tags` in tantivy index

srivik:
Description:                                                                                                                                                                                             
                                                                                                                                                                                                           
  ## Summary                                                                                                                                                                                               
  - Fixes a bug where filtering by a label whose name matches the MapColumn name                                                                                                                           
    (e.g., `tags="b1.metal"` when the partition schema MapColumn is also named "tags")                                                                                                                     
    returns no data in the Tantivy index                                                                                                                                                                   
  - The root cause is Tantivy's `find_field_with_default` resolving the field name to                                                                                                                      
    the JSON field itself with an empty prefix, producing an incorrect query term                                                                                                                          
  - Group-by (`by (tags)`) was unaffected because it reads from partition key binary data                                                                                                                  
    directly, bypassing the index                                                                                                                                                                          
                                                                                                                                                                                                           
  ## Changes                                                                                                                                                                                               
  - Added `resolve_field_with_default()` in `query_parser.rs` that detects when a field                                                                                                                    
    name collides with the JSON column name and treats it as a subfield path                                                                                                                               
  - Applied the fix to equals, regex, prefix, and term-in query paths                                                                                                                                      
  - Applied equivalent fix in `reader.rs` for label value lookups                                                                                                                                          
                                                                                                                                                                                                           
  ## Test plan                                                                                                                                                                                             
  - [x] Rust unit tests: equals, regex, term-in, prefix with map column name collision                                                                                                                     
  - [x] Scala integration test: end-to-end indexing + querying with `tags:map` schema                                                                                                                      
  - [ ] Verify on deployment with `part-key-index-type = tantivy` that `tags="value"`                                                                                                                      
        filters now return correct results

### DIFF
--- a/core/src/rust/filodb_core/src/query_parser.rs
+++ b/core/src/rust/filodb_core/src/query_parser.rs
@@ -176,6 +176,29 @@ fn value_with_prefix(prefix: &str, value: &str) -> String {
     )
 }
 
+/// Resolve a field name to a (Field, prefix) pair, handling JSON map columns.
+///
+/// When the field name matches the JSON (map) column itself, we treat it as a
+/// subfield lookup within that JSON column rather than a direct field query.
+/// This handles the case where a label has the same name as the map column
+/// (e.g., a label "tags" inside a map column also named "tags").
+fn resolve_field_with_default<'a>(
+    schema: &'a Schema,
+    default_field: Option<Field>,
+    field_name: &'a str,
+) -> Option<(Field, &'a str)> {
+    let (field, prefix) = schema.find_field_with_default(field_name, default_field)?;
+
+    // If the resolved field is the default JSON field and the prefix is empty,
+    // it means the field name matched the JSON column name directly. Treat the
+    // field name as a subfield path within the JSON column instead.
+    if prefix.is_empty() && default_field == Some(field) {
+        Some((field, field_name))
+    } else {
+        Some((field, prefix))
+    }
+}
+
 /// Build a query using a specified field and value, handling JSON
 /// fields
 ///
@@ -194,7 +217,7 @@ fn query_with_field_and_value<T>(
 where
     T: FnOnce(Field, &str) -> Result<Box<dyn Query>, TantivyError>,
 {
-    let Some((field, prefix)) = schema.find_field_with_default(field, default_field) else {
+    let Some((field, prefix)) = resolve_field_with_default(schema, default_field, field) else {
         // If it's an invalid field then map to an empty query, which will emulate Lucene's behavior
         return Ok(Box::new(EmptyQuery));
     };
@@ -251,7 +274,7 @@ fn parse_term_in_query<'a>(
     for _ in 0..term_count {
         let (input, text) = parse_string(next_input)?;
 
-        if let Some((field, prefix)) = schema.find_field_with_default(&column, default_field) {
+        if let Some((field, prefix)) = resolve_field_with_default(schema, default_field, &column) {
             terms.push(Term::from_field_text(
                 field,
                 &value_with_prefix(prefix, &text),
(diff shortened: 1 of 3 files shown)


sandeep6189:
please update the PR header (according to the format) and add PR description

core/src/rust/filodb_core/src/query_parser.rs
```diff
@@ -176,6 +176,29 @@ fn value_with_prefix(prefix: &str, value: &str) -> String {
     )
 }
 
+/// Resolve a field name to a (Field, prefix) pair, handling JSON map columns.
```
sandeep6189, line 179: nitpick: lets add the following flow chart as code comment ? ( for future references ) - the same fix is in reader.rs, so any common place would work.
```
                                                                                                                                                                            
  Indexed columns:                                                                                                                                                          
    ┌────────────┐  ┌────────────┐  ┌────────────┐                                                                                                                          
    │   _ws_     │  │   _ns_     │  │  __name__  │  ← real indexed columns                                                                                                  
    └────────────┘  └────────────┘  └────────────┘                                                                                                                          
                                                                                                                                                                            
  JSON map column (the "default field"):                                                                                                                                    
    ┌─────────────────────────────────────────────┐                                                                                                                         
    │   tags  (JSON column)                       │                                                                                                                         
    │                                             │                                                                                                                         
    │   { "instance": "i-001",                    │
    │     "job": "node",                          │                                                                                                                         
    │     "tags": "some-value"  ← LABEL named     │         
    │   }                          same as column │                                                                                                                         
    └─────────────────────────────────────────────┘                                                                                                                         
                                                                                                                                                                            
  Now query: tags="some-value"                                                                                                                                              
                                                                                                                                                                            
  Behavior WITHOUT this function                                                                                                                                                  
----------------------------------------   
  schema.find_field_with_default("tags", Some(json_field))                                                                                                                  
         │                                                  
         │ "tags" IS an indexed field name (the JSON column itself!)
         │ So it returns the JSON column directly with empty prefix                                                                                                         
         │                                                                                                                                                                  
         ▼                                                                                                                                                                  
  returns (json_field, "")         ← prefix = empty                                                                                                                         
         │                                                                                                                                                                  
         ▼
  value_with_prefix("", "some-value") → "some-value"                                                                                                                        
         │                                                  
         ▼
  TermQuery: search json_field for literal value "some-value"
         │                                                                                                                                                                  
         ▼
  ✗ WRONG — searches the entire JSON blob for literal "some-value"                                                                                                          
     misses documents where "tags" subfield = "some-value"                                                                                                                  
                                                                                                                                                                            
  Behavior WITH this function                                                                                                                                                    
----------------------------------------                                                                                                                                                                            
  schema.find_field_with_default("tags", Some(json_field))                                                                                                                  
         │                                                  
         │ Returns (json_field, "")
         ▼
                                                                                                                                                                            
  resolve_field_with_default checks:
     prefix.is_empty() && default_field == Some(field)                                                                                                                      
     ↑                    ↑                                                                                                                                                 
     "no prefix"          "and the field IS the default JSON field"                                                                                                         
     │                                                                                                                                                                      
     │ This means: the user's field name matches the JSON column name                                                                                                       
     │             → they actually want a subfield lookup                                                                                                                   
     │                                                                                                                                                                      
     ▼                                                                                                                                                                      
  returns (json_field, "tags")     ← prefix = the field name itself                                                                                                         
         │                                                  
         ▼
  value_with_prefix("tags", "some-value") → "tags\0ssome-value"                                                                                                             
         │
         ▼                                                                                                                                                                  
  TermQuery: search json_field for "tags\0ssome-value"      
         │                                                                                                                                                                  
         ▼
  ✓ CORRECT — finds documents where the "tags" subfield in JSON = "some-value"  
```

sandeep6189 (approved):
Thank you for taking this up @srivik

PR #2227, opened 2026-05-11
perf(coord): Zero copy of binary record into Arrow VSR during query execution

vishramachandran:
**Pull Request checklist**

This PR achieves two things:
* Eliminates the on-heap RecordBuilder → Arrow copy in populateRvContentsIntoVsrs. A new SingleRecordBuilder writes BinaryRecords directly into the Arrow VarBinaryVector's off-heap data buffer. This avoids heap allocation in the hot path and improves performance in the arrow flight rpc path
* Migrates flight metadata (header, footer, RV metadata) from Kryo to Protobuf, removing the RvMetadata/RespHeader/RespFooter Kryo-registered classes. This fixes tech debt of using kryo for cross-partition calls.

Refactor of RecordBuilder was done very carefully:
* extensive unit tests were added, refactor was done without any semantic edits to code, thus RecordBuilder(new) + RecordBuilderBase == RecordBuilder(old)
* old tests were rerun to ensure none failed

### DIFF
--- a/coordinator/src/main/scala/filodb/coordinator/flight/FlightPlanDispatcher.scala
+++ b/coordinator/src/main/scala/filodb/coordinator/flight/FlightPlanDispatcher.scala
@@ -114,6 +114,8 @@ case class FlightPlanDispatcher(location: Location,
               footerThrowable = if (footer.hasThrowable) Some(footer.getThrowable.fromProto) else None
               qLogger.debug(s"FlightPlanDispatcher received footer for queryPlanId=${plan.planId} with stats: " +
                 s"${footerStats.get}, throwable: $footerThrowable")
+            } else {
+              qLogger.warn(s"FlightPlanDispatcher received metadata with unknown type for queryPlanId=${plan.planId}")
             }
           }
           // Error on stream is thrown as exception and will be handled at onErrorHandle below
--- a/coordinator/src/main/scala/filodb/coordinator/flight/FlightPlanDispatcher.scala
+++ b/coordinator/src/main/scala/filodb/coordinator/flight/FlightPlanDispatcher.scala
@@ -14,9 +14,10 @@ import org.apache.arrow.vector.{VectorLoader, VectorSchemaRoot, VectorUnloader}
 import filodb.coordinator.QueryScheduler
 import filodb.core.QueryTimeoutException
 import filodb.core.memstore.FiloSchedulers
-import filodb.core.query.{QuerySession, QueryStats}
+import filodb.core.query.{QuerySession, QueryStats, ResultSchema}
 import filodb.core.store.ChunkSource
 import filodb.query.{QueryError, QueryResponse, QueryResult, StreamQueryResponse}
+import filodb.query.ProtoConverters._
 import filodb.query.exec.{ExecPlan, ExecPlanWithClientParams, PlanDispatcher}
 
 case class FlightPlanDispatcher(location: Location,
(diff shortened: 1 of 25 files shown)


kvpetrov (approved):
LGTM
there are seem to be transient errors with 429, and I don't really see any changes in the dependencies, so, I would double check before committing

coordinator/src/main/scala/filodb/coordinator/flight/FlightPlanDispatcher.scala
```diff
@@ -101,27 +103,29 @@ case class FlightPlanDispatcher(location: Location,
...
-              case footer: RespFooter =>
-                respFooter = Some(footer)
-                qLogger.debug(s"FlightPlanDispatcher received RespFooter for queryPlanId=${plan.planId} with stats: " +
-                  s"${footer.queryStats}, throwable: ${footer.throwable}")
+            val meta = FlightProtoSerDeser.deserializeMetadata(stream.getLatestMetadata)
+            if (meta.hasHeader) {
+              resultSchema = Some(meta.getHeader.getResultSchema.fromProto)
+              qLogger.debug(s"FlightPlanDispatcher received header for queryPlanId=${plan.planId} " +
+                s"with schema: ${resultSchema.get}")
+            } else if (meta.hasFooter) {
```
alextheimer, line 111: Should there be an `else` `ERROR` log just in case?

coordinator/src/main/scala/filodb/coordinator/flight/FlightProtoSerDeser.scala
```diff
@@ -0,0 +1,78 @@
...
+  def deserializeFromBytes(bytes: Array[Byte]): ProtoRangeVector.RvMetadata =
+    ProtoRangeVector.RvMetadata.parseFrom(bytes)
+
+  // Writes proto bytes directly into the VarBinaryVector data buffer without an intermediate byte[].
+  // Mirrors the manual offset-chain bookkeeping in addFromReader — see the comment there for why
+  // VarBinaryVector.set() is not used.
+  private def writeProto(msg: com.google.protobuf.MessageLite, state: VsrPopulationState,
+                         needNewVec: () => Unit): Unit = {
+    val size = msg.getSerializedSize
+    if (state.bytesRemaining < size || state.rowNum >= maxNumRows) needNewVec()
```
alextheimer, line 59: Is it ever possible that an RVK is so huge it exceeds the vector max length?

alextheimer (approved):
LGTM 👍 
Extra brownie points for the descriptive comments -- those are super helpful.

coordinator/src/main/scala/filodb/coordinator/flight/FlightPlanDispatcher.scala
vishramachandran: Can add - right now the proto has only two "oneof"s

coordinator/src/main/scala/filodb/coordinator/flight/FlightProtoSerDeser.scala
vishramachandran: It is very rare because the RecordContainer max lengths are same as VSR max lengths right now. The part keys would not have made it into the TSDB if it were longer because they would be subject to the same RecordContainer max lengths. But it can happen rarely if queries add more keys to the RVK. I'll add some require calls if it won't spoil performance.

PR #2232, opened 2026-05-13
perf(query): prevent allocations during QueryStats updates

alextheimer:
**Pull Request checklist**

It was found during perf testing that functions in the new samples-scanned accounting infrastructure allocate more heap objects than expected. This happens because some `TrieMap` methods (e.g. `isEmpty`) create a read-only snapshot every time they are invoked. The snapshots are small, but the memory pressure will add up on a hot path.

**New behavior :**

This PR refactors existing code to prevent unnecessary allocations. Primarily, the implementation of `QueryStats` is now changed to support frequent entry iteration without any `Iterator` allocations. The most-important fields are now:
```
  /**
   * Stores all stats entries.
   * Helpful in scenarios where e.g. entries need frequent iteration but
   *   Iterator allocations should be prevented.
   * Also useful for efficient size() calculation (as opposed to TrieMap.size(),
   *   which allocates an Object on the heap and has some computation overhead).
   */
  private val entries = ArrayBuffer[(Seq[String], Stat)]()
  private val keyToEntryIndex = new TrieMap[Seq[String], Integer]()
```

Since `QueryStats`' previous backing `TrieMap` was exposed outside the class, clients invoked its methods directly. Some of the ostensibly-overkill methods included with this PR have been added to `QueryStats` to bridge that gap; the trie no longer contains the same data, so all instances of direct trie usage have been replaced with new `QueryStats` methods.

These changes allow the new samples-scanned accounting infrastructure to operate without incurring massive memory overhead.

One additional `QueryStats`-unrelated lambda allocation in `ChunkSetInfo.scala` was found while running performance tests; this accounted for ~10% of all allocations under heavy load. The allocation is also removed as part of this PR.

### DIFF
--- a/core/src/main/scala/filodb.core/query/QueryContext.scala
+++ b/core/src/main/scala/filodb.core/query/QueryContext.scala
@@ -4,7 +4,7 @@ import java.util.UUID
 import java.util.concurrent.atomic.{AtomicInteger, AtomicLong}
 
 import scala.collection.concurrent.TrieMap
-import scala.collection.mutable.SortedSet
+import scala.collection.mutable.{ArrayBuffer, SortedSet}
 import scala.concurrent.duration._
 
 import com.typesafe.scalalogging.StrictLogging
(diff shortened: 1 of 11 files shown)


core/src/main/scala/filodb.core/query/QueryContext.scala
```diff
@@ -402,16 +402,233 @@ case class Stat() {
...
+  private val keyToEntryIndex = new TrieMap[Seq[String], Integer]()
+
+  /**
+   * Set true if Nil is ever added as a key.
+   * Useful in cases where this needs to be known efficiently
+   *   (e.g. while samples-scanned counters are updated).
+   **/
+  @volatile private var containsNilKey = false;
+
+  private val lock = new AnyRef
```
vishramachandran: Use a read-write lock or concurrent map.
Looks like writers lock in this implementation, but readers don't.

The ArrayBuffer could be growing or shrinking under the covers by writers - so a read lock is needed.

vishramachandran (commented):
Thanks for the ChunksetInfo allocation reduction bonus!!

Looks good to me.

The only important fix that is needed is the readers lock or a concurrent map for correctness.

core/src/main/scala/filodb.core/query/QueryContext.scala
alextheimer: Ah man, you are right. I thought we could get away without any read locks because all samples-scanned counter updates are processed serially for each individual `QueryStats`, but that is not true for leaves because each series does not have its own `QueryStats`.

I will add read/write locks and perf test again to confirm performance is not impacted.

alextheimer: Actually wait, even at the leaves we should be OK with the new changes: `QueryStats` entries are populated serially in `ChunkSource.rangeVectors`; any concurrent counter updates after that will not create any new `QueryStats` entires. I'll still add read/write locks because that's safer and might not impact performance too much, but I think we can fallback to the original write-locks-only implementation if needed.

alextheimer: I have pushed some changes: read/write locks are now used for all except those with names prefixed by `unsafe`. Locking methods had a not-insignificant presence in the profiles.

For each, I have justified in a comment why skipping locks is OK. The profiles look much better now that locks are removed from methods used by the new samples-scanned accounting infrastructure.

vishramachandran (approved):
Looks better now from a functional and concurrency perspective. Good to keep in mind that ReentrantReadWriteLock is not the most performant type of lock in JDK11 for doing this. They have improved in later JDKs - so I am okay with this since there is an impending JDK upgrade on our roadmap. If we see any slowness - we should come back and investigate if we can improve and optimize.

PR #2250, opened 2026-05-21
fix(query): move SamplesScannedConfig to PlannerParams

alextheimer:
**Pull Request checklist**

Currently, `SamplesScannedConfig` resides in `QueryConfig`. However, the intent is that custom `SamplesScannedConfig`s can be passed to `QueryActor`; that is currently not possible because `QueryConfig` is not accepted as an execution parameter. Each actor's `QueryConfig` is loaded once from a local config.

**New behavior :**

`SamplesScannedConfig` is now moved to `PlannerParams`; `QueryActor`s accept this as part of a `QueryContext`.


core/src/main/scala/filodb.core/query/QueryContext.scala
```diff
@@ -180,6 +180,85 @@ object DownCluster {
   }
 }
 
+/**
```
alextheimer, line 187: This entire class (and its corresponding object class) was copy-pasted from `QueryConfig.scala`. There were no changes.

grpc/src/main/protobuf/query_service.proto
```diff
@@ -50,27 +50,28 @@ enum FailoverMode {
...
+    optional bool processMultiPartition                 = 13;
+    optional bool allowPartialResults                   = 14;
+    optional bool histogramMap                          = 15;
+    optional bool useProtoExecPlans                     = 16;
+    optional bool reduceShardKeyRegexFanout             = 17;
+    optional uint32 maxShardKeyRegexFanoutBatchSize     = 18;
+    optional bool allowNestedAggregatePushdown          = 19;
+    repeated DownPartition downPartitions               = 20;
+    optional FailoverMode failoverMode                  = 21;
+    optional SamplesScannedConfig samplesScannedConfig  = 22;
```
alextheimer, line 74: This is the only tangible change; all other rows are included in the diff only because their field numbers were visually shifted to the right.

grpc/src/main/protobuf/query_service.proto
```diff
@@ -455,8 +456,7 @@ message QueryConfig {
   repeated string grpcPartitionsDenyList     = 16;
   optional string plannerSelector            = 17;
   map<string,int32> recordContainerOverrides = 18;
-  SamplesScannedConfig samplesScannedConfig  = 19;
-  repeated string flightPartitionsDenyList   = 20;
+  repeated string flightPartitionsDenyList   = 19;
```
alextheimer, line 459: OK to change the number; this is not merged to `main`.

query/src/main/scala/filodb/query/ProtoConverters.scala
```diff
@@ -281,6 +286,146 @@ object ProtoConverters {
     }
   }
 
+  implicit class SamplesScannedConfigToProtoConverter(config: SamplesScannedConfig) {
```
alextheimer, line 288: All of this diff was copy-pasted from the other `ProtoConverters` file.

core/src/main/scala/filodb.core/query/QueryContext.scala
```diff
@@ -420,9 +567,23 @@ case class QueryStats() {
...
+
+  /**
+   * Invoked by reflection during deserialization.
+   */
+  private def readObject(in: java.io.ObjectInputStream): Unit = {
+    in.defaultReadObject()
+    lock = new ReentrantReadWriteLock()
+    readLock = lock.readLock()
+    writeLock = lock.writeLock()
+  }
```
vishramachandran: This is for Java serialization. Will this work with Kryo ? Almost certainly not.
Simpler may be to make it a getter and initialize one lazily when accessed with null value.

alextheimer: Good catch. I've added proper Kryo `read()` and `write()` methods to skip the extra `if (initialized)` on hot paths: https://github.com/filodb/FiloDB/pull/2250/changes#diff-6e2f1a29a962b325119951c0f0df9bb4c77b6cf9d634a3a673633ffac74d33c6R580

alextheimer: Also added ser/deser tests (credit to Claude for finding the `SerializationSpec` file -- I was unaware of any existing ser/deser tests).

I've confirmed the test fails if I break the Kryo interface methods in QueryStats; the new logic is definitely getting invoked.

core/src/main/scala/filodb.core/query/QueryContext.scala
```diff
@@ -422,9 +571,57 @@ case class QueryStats() {
...
-  private val lock = new ReentrantReadWriteLock()
-  private val readLock = lock.readLock()
-  private val writeLock = lock.writeLock()
+  // NOTE: all are 'var' to allow assignment from within read(...).
+  //   These fields will be left null if not explicitly set in read(...).
+  private var lock = new ReentrantReadWriteLock()
+  private var readLock = lock.readLock()
+  private var writeLock = lock.writeLock()
+
+  override def write(kryo: Kryo, output: Output): Unit = {
```
vishramachandran: This may be simpler https://github.com/filodb/FiloDB/pull/2279

alextheimer: These changes are now moved to the incoming JDK21 branch.

core/src/main/scala/filodb.core/query/QueryContext.scala
alextheimer: These changes are now merged into the incoming JDK21 branch.

core/src/main/scala/filodb.core/query/QueryContext.scala
```diff
@@ -422,9 +569,23 @@ case class QueryStats() {
...
+  //   readObject is preferred to avoid the lazy overhead given that
+  //   these variables are accessed on hot paths.
+  @transient private var lock = new ReentrantReadWriteLock()
+  @transient private var readLock = lock.readLock()
+  @transient private var writeLock = lock.writeLock()
+
+  /**
+   * Invoked by reflection during deserialization.
+   */
+  private def readObject(in: java.io.ObjectInputStream): Unit = {
```
siri-varma: What is the purpose of this ?

siri-varma (approved):
LGTM

core/src/main/scala/filodb.core/query/QueryContext.scala
alextheimer: Oof good catch-- I reverted my commits that included similar work, but it looks like there were even earlier commits that included more related changes. I will revert these, too.

PR #2254, opened 2026-05-22
fix(coordinator): force verbose=true on TsCardinalities HA failover

jatakiajanvi12:
## Summary

The HAPlanner's TsCardinalities cross-cluster failover hop was not forcing `verbose=true` on the buddy request. When the responder gates `_type` emission on `verbose`, the response omits the field and the local decoder fails:

```
errorType: query_execution_failed
error: "Attempt to decode value on failed cursor: DownField(_type),DownArray,DownField(data)"
```

This mirrors the pattern already in `MultiPartitionPlanner.scala:1430`, where verbose is forced to true on the remote query context for V2 cardinality cross-partition hops so the response always carries `_type`.

## Test plan

- [x] Added `HighAvailabilityPlannerSpec` test asserting `verbose=true` is set on both the `PromQlQueryParams` and the URL params of the resulting `MetadataRemoteExec`
- [x] Updated two existing TsCardinalities HA tests to expect `verbose=true` in the printed plan tree
- [x] `coordinator/Test/compile` succeeds locally
- [ ] CI runs the full test suites (local sbt forking blocked by sandbox)

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/queryplanner/HighAvailabilityPlanner.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryplanner/HighAvailabilityPlanner.scala
@@ -140,8 +140,12 @@ class HighAvailabilityPlanner(dsRef: DatasetRef,
           // TsCardinalities cannot be converted to PromQL — handle before convertToQuery
           rootLogicalPlan match {
             case lp: TsCardinalities =>
-              val newQueryContext = qContext.copy(plannerParams = qContext.plannerParams.
-                copy(processFailure = false, processMultiPartition = false))
+              // Force verbose=true so the buddy cluster's V2 cardinality response includes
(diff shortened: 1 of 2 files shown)


query/src/main/scala/filodb/query/PromCirceSupport.scala
```diff
@@ -58,12 +58,15 @@ object PromCirceSupport {
           } yield LabelCardinalitySampl(metric, card)
         } else if (c.downField("group").focus.nonEmpty) {
           // This is for TsCardinalities response
+          // `_type` is optional: some HTTP responders (e.g. Apple's public V2 cardinality
```
sandeep6189: remove the apple reference ( this is leaking internal apple information in public repo )

sandeep6189 (approved):

PR #2269, opened 2026-05-29
Add dispatch latency metric to FiloDB

siri-varma:
Problem

query_materialize_time_ns does not exist for grpc. It is only emitted on the http side of the query server api.

We need a similar metric on the grpc time to understand the time it takes to materialize data.

  Change

  Emit one metric at the shared convergence point both paths hit — QueryPlanner.dispatchExecPlan / dispatchStreamingExecPlan — tagged by entry transport:

  query_dispatch_latency_seconds{ source="http"|"grpc", dataset=... }

  - gRPC server passes source="grpc"; HTTP/CLI default to source="http".
  - querySource is a method argument, not a serialized QueryContext field → no Kryo/proto version-skew with data nodes.

  Why this is the right place

  Both the HTTP executor and the gRPC server call dispatchExecPlan after their own parse/materialize, so a metric here measures the same window for both — true apples-to-apples. It can't reuse the existing
  query_execute_plan_time_ns name because that's a query-service (MetricsPublisher, _ns) metric, and FiloDB code can only use FilodbMetrics (Kamon → _seconds); FiloDB also can't depend on query-service.

  Benefit

  One PromQL query compares the transports directly:
  histogram_quantile(0.99, sum by (le, source) (rate(query_dispatch_latency_seconds{...}[$interval])))

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/queryplanner/QueryPlanner.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/queryplanner/QueryPlanner.scala
@@ -1,5 +1,7 @@
 package filodb.coordinator.queryplanner
 
+import java.util.concurrent.TimeUnit
+
 import scala.concurrent.duration.FiniteDuration
 
 import kamon.Kamon
@@ -7,6 +9,7 @@ import monix.eval.Task
 import monix.execution.Scheduler
 import monix.reactive.Observable
 
+import filodb.core.metrics.FilodbMetrics
 import filodb.core.query.{QueryContext, QuerySession}
 import filodb.query.{LogicalPlan, QueryResponse, StreamQueryResponse}
 import filodb.query.exec.{ClientParams, ExecPlan, ExecPlanWithClientParams, UnsupportedChunkSource}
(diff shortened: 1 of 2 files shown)


coordinator/src/main/scala/filodb.coordinator/queryplanner/QueryPlanner.scala
```diff
@@ -41,13 +64,27 @@ trait QueryPlanner {
       execPlan.dispatcher.dispatch(ExecPlanWithClientParams(execPlan,
         ClientParams(execPlan.queryContext.plannerParams.queryTimeoutMillis),
         querySession), UnsupportedChunkSource())
-    }
+    }.guarantee(Task.eval {
```
alextheimer, line 67: This new metric would measure the full request latency of each dispatch. That includes ser/deser, network, and evaluation. Probably GRPC request latencies would be skewed super high just because GRPC is only used near the root of a query plan.

I think what you're actually looking for is a metric that measures only network latencies for failed-over queries, but it would probably take quite a bit of work to setup something like that.

PR #2290, opened 2026-06-24
feat(multiple): Preparing for JDK21 by making compatible changes in JDK11 itself

vishramachandran:
**Pull Request checklist**

Prepare for JDK21 by making changes in code without upgrading to JDK21 yet.
Also using virtual threads with reflection based on Java runtime. This allows us to compile code with Java 11 compatibility thus not forcing all library dependents to upgrade to JDK21 in lock step.

I am essentially bringing code changes from develop-jdk21 branch into here without build upgrades to JDK21.

### DIFF
--- a/coordinator/src/main/scala/filodb.coordinator/client/Serializer.scala
+++ b/coordinator/src/main/scala/filodb.coordinator/client/Serializer.scala
@@ -1,6 +1,7 @@
 package filodb.coordinator.client
 
 import java.net.URI
+import java.util.concurrent.locks.ReentrantReadWriteLock
 
 import com.esotericsoftware.kryo.{Kryo, Serializer => KryoSerializer}
 import com.esotericsoftware.kryo.io._
@@ -27,6 +28,7 @@ object KryoInit {
     kryo.addDefaultSerializer(classOf[ZeroCopyUTF8String], classOf[ZeroCopyUTF8StringSerializer])
     kryo.register(classOf[Schema], new SchemaSerializer)
     kryo.register(classOf[PartitionSchema], new PartSchemaSerializer)
+    kryo.register(classOf[ReentrantReadWriteLock], new ReentrantReadWriteLockSerializer)
 
     initOtherFiloClasses(kryo)
     initQueryEngine2Classes(kryo)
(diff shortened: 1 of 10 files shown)


core/src/main/scala/filodb.core/query/QueryContext.scala
```diff
@@ -570,8 +570,8 @@ case class QueryStats() {
   @volatile private var containsNilKey = false;
 
   private val lock = new ReentrantReadWriteLock()
```
kvpetrov, line 572: maybe instead:
@transient
private lazy val lock = new ReentrantReadWriteLock()

@transient
private lazy val readLock = lock.readLock()
...

coordinator/src/main/scala/filodb.coordinator/client/Serializer.scala
```diff
@@ -185,3 +187,14 @@ class PartSchemaSerializer extends KryoSerializer[PartitionSchema] {
     require(schema == schemas.part)
   }
 }
+
+class ReentrantReadWriteLockSerializer extends KryoSerializer[ReentrantReadWriteLock] {
```
kvpetrov, line 191: maybe we don't need this class?
@transient
private lazy val lock = new ReentrantReadWriteLock()

@transient
private lazy val readLock = lock.readLock()
...

kvpetrov (approved):
minor comment to use transient/lazy instead of faker serializer/deserializer, easier to read the code

core/src/main/scala/filodb.core/query/QueryContext.scala
vishramachandran: At the receiving end, we need it to not be null. A lock needs to get initialized. That's why it cannot just be annotated with transient.

coordinator/src/main/scala/filodb.coordinator/client/Serializer.scala
vishramachandran: same as above

core/src/main/scala/filodb.core/query/QueryContext.scala
kvpetrov: lazy is supposed to do that

PR #2297, opened 2026-06-26
fix(coord): Query warnings and partial result details need to be in flight response

vishramachandran:
**Pull Request checklist**

Flight RPC response did not propagate warnings and partial result flags up the call hierarchy. This PR enables that.

### DIFF
--- a/coordinator/src/main/scala/filodb/coordinator/flight/FlightPlanDispatcher.scala
+++ b/coordinator/src/main/scala/filodb/coordinator/flight/FlightPlanDispatcher.scala
@@ -14,7 +14,7 @@ import org.apache.arrow.vector.{VectorLoader, VectorSchemaRoot, VectorUnloader}
 import filodb.coordinator.QueryScheduler
 import filodb.core.QueryTimeoutException
 import filodb.core.memstore.FiloSchedulers
-import filodb.core.query.{QuerySession, QueryStats, ResultSchema}
+import filodb.core.query.{QuerySession, QueryStats, QueryWarnings, ResultSchema}
 import filodb.core.store.ChunkSource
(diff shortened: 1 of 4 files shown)


coordinator/src/main/scala/filodb/coordinator/flight/FlightPlanDispatcher.scala
```diff
@@ -127,7 +137,7 @@ case class FlightPlanDispatcher(location: Location,
         } else {
           val srvs = ArrowSerializedRangeVectorOps.convertVsrsIntoArrowSrvs(vsrs.toSeq, resultSchema.get)
           QueryResult(plan.queryContext.queryId, resultSchema.get, srvs,
-            footerStats.getOrElse(QueryStats()))
+            footerStats.getOrElse(QueryStats()), footerWarnings, footerMayBePartial, footerPartialResultReason)
```
alextheimer, line 140: Confirming: it is correct to use these empty `QueryStats` here?

alextheimer (approved):
Nice catch -- just one non-blocking comment ✅

coordinator/src/main/scala/filodb/coordinator/flight/FlightPlanDispatcher.scala
vishramachandran: If client did not get a footer (technically illegal state) we use empty QueryStats. Were you thinking of something else ?

alextheimer: Ah I misinterpreted this, then-- I thought this was handling even the happy path, too. In that case feel free to merge ✅

PR #2301, opened 2026-07-01
bug(coord): Arrow Flight RPC — query-limit safety, VSR memory fix, and hot-path optimizations

vishramachandran:
**Pull Request checklist**

Summary

  Three related concerns in the Arrow Flight result-streaming path are addressed together:

  1. Bug fix — Arrow memory leaks when a query limit exception fires mid-drain
  2. Bug fix — stale timestamp computed during null-row reconstruction in ArrowSerializedRangeVector
  3. Performance — per-row overhead in the serialisation hot path

  A suite of new regression tests is added that reproduce each query-limit scenario and assert zero Arrow memory is leaked.

  ---
  Changes

  FlightQueryResultStreaming — correct VSR ownership on exception

  The previous finishedVsrs.foreach { … } drain loop had a subtle ownership hazard: if checkResultBytes threw mid-iteration, the throwing VSR was neither in finishedVsrs
  (already removed by foreach) nor in freeVsrs (not yet returned), and the bracket's cleanup did not close it — a leak.

  This PR replaces the loop with drainFinishedVsrs(), which:
  - removes each VSR from finishedVsrs before sending it so the bracket always has a consistent view: unsent VSRs stay in finishedVsrs, sent VSRs go to freeVsrs, and the one
  that threw is closed inline in the catch block before re-throwing.
  - The bracket release is updated to close all four VSR categories: flightVsr, remaining finishedVsrs, all freeVsrs, and currentVsr as a safety net.

  ArrowSerializedRangeVectorOps — free-pool capacity and hot-path hoisting

  - MpscArrayQueue capacity raised from a hard-coded 5 to maxVsrs = 50; if the pool is ever full, haltAndCatchFire fires immediately instead of silently dropping a VSR.
  - canRemoveEmptyRows(…) and recordSchema.columns(1).colType were being evaluated on every row in the inner serialisation loop. Both are now hoisted to constants computed
  once before the loop.
  - Extracted updateVsrLengthInState() to eliminate three identical setValueCount / setRowCount call sites.

  RangeVector — timestamp derivation bug

  ArrowSerializedRangeVector's row iterator was reconstructing null-row timestamps by incrementing a mutable curTimestamp counter. When rows were skipped or re-used across
  VSR boundaries the counter could drift. The fix derives the timestamp directly: startMs + rowsRead * step. The mutable field is removed entirely.

  FiloDBSinglePartitionFlightProducerSpec — memory-safety regression tests

  Three new tests in FiloDBSinglePartitionFlightProducerSpec exercise the query-limit paths:

```
  ┌────────────────────────────────┬─────────────────────────────────────┬────────────────┐
  │              Test              │           Limit triggered           │  VSR position  │
  ├────────────────────────────────┼─────────────────────────────────────┼────────────────┤
  │ checkResultBytes on first VSR  │ execPlanResultBytes = 1 byte        │ VSR1 mid-drain │
  ├────────────────────────────────┼─────────────────────────────────────┼────────────────┤
  │ checkSamplesLimit              │ execPlanSamples = 1                 │ Before drain   │
  ├────────────────────────────────┼─────────────────────────────────────┼────────────────┤
  │ checkResultBytes on second VSR │ execPlanResultBytes = maxVecLen * 2 │ VSR2 mid-drain │
  └────────────────────────────────┴─────────────────────────────────────┴────────────────┘

```  

  Each test uses an isolated child allocator, asserts the response is a QueryError wrapping QueryLimitException, and asserts that testAllocator.getAllocatedMemory ==
  allocatedMemBeforeQuery after limitedSession.close().

  ---
  Testing

  sbt "coordinator/testOnly filodb.coordinator.flight.FiloDBSinglePartitionFlightProducerSpec"
  sbt "coordinator/testOnly filodb.coordinator.flight.ArrowSerializedRangeVectorSpec"
  sbt "coordinator/testOnly filodb.coordinator.ProtoConvertersSpec"

### DIFF
--- a/coordinator/src/main/scala/filodb/coordinator/flight/ArrowSerializedRangeVectorOps.scala
+++ b/coordinator/src/main/scala/filodb/coordinator/flight/ArrowSerializedRangeVectorOps.scala
@@ -34,20 +34,20 @@ object ArrowSerializedRangeVectorOps {
   // FIXME this should not be hard-coded
   val maxVecLen = 1048576 // 1 MB
   val maxNumRows = maxVecLen / 15
+  val maxVsrs = 50
 
   def emptyVectorSchemaRoot(allocator: BufferAllocator): VectorSchemaRoot = {
     VectorSchemaRoot.create(arrowSrvSchema, allocator)
   }
 
   case class VsrPopulationState(var flightVsr: VectorSchemaRoot = null,
-                                var currentVsr: VectorSchemaRoot = null,
-                                var currentIsRvkVec: BitVector = null,
-                                var currentRvkBrVec: VarBinaryVector = null,
-                                var rowNum: Int = -1,
-                                var bytesRemaining: Int = maxVecLen,
-                                // FIXME hard coded 5
-                                freeVsrs: MpscArrayQueue[VectorSchemaRoot] = new MpscArrayQueue[VectorSchemaRoot](5),
-                                finishedVsrs: ArrayBuffer[VectorSchemaRoot] = ArrayBuffer.empty)   {
+                        var currentVsr: VectorSchemaRoot = null,
+                        var currentIsRvkVec: BitVector = null,
+                        var currentRvkBrVec: VarBinaryVector = null,
+                        var rowNum: Int = -1,
+                        var bytesRemaining: Int = maxVecLen,
+                        freeVsrs: MpscArrayQueue[VectorSchemaRoot] = new MpscArrayQueue[VectorSchemaRoot](maxVsrs),
+                        finishedVsrs: ArrayBuffer[VectorSchemaRoot] = ArrayBuffer.empty)   {
 
     def finishAndGetCurrentVsr(): VectorSchemaRoot = {
       if (currentVsr != null) {
@@ -187,13 +187,15 @@ object ArrowSerializedRangeVectorOps {
         val startNs = Utils.currentThreadCpuTimeNanos
         try {
           ChunkMap.validateNoSharedLocks(execPlan)
+          val canRemoveEmptyRows = SerializedRangeVector.canRemoveEmptyRows(rv.outputRange, recordSchema)
+          lazy val col1Type = recordSchema.columns(1).colType
           Using.resource(rv.rows()) { rows =>
             while (rows.hasNext) {
               val nextRow = rows.next()
               // Don't encode empty-histogram / NaN data over the wire
-              if (!SerializedRangeVector.canRemoveEmptyRows(rv.outputRange, recordSchema) ||
-                recordSchema.columns(1).colType == DoubleColumn && !java.lang.Double.isNaN(nextRow.getDouble(1)) ||
-                recordSchema.columns(1).colType == HistogramColumn && !nextRow.getHistogram(1).isEmpty) {
+              if (!canRemoveEmptyRows ||
+                col1Type == DoubleColumn && !java.lang.Double.isNaN(nextRow.getDouble(1)) ||
+                col1Type == HistogramColumn && !nextRow.getHistogram(1).isEmpty) {
                 addFromReader(nextRow)
               } else {
                 addNullRow()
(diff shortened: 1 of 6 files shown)


coordinator/src/main/scala/filodb/coordinator/flight/ArrowSerializedRangeVectorOps.scala
```diff
@@ -34,20 +34,32 @@ object ArrowSerializedRangeVectorOps {
   // FIXME this should not be hard-coded
   val maxVecLen = 1048576 // 1 MB
   val maxNumRows = maxVecLen / 15
+  val maxVsrs = 50
```
alextheimer: What's the likelihood we might need to change one of these values? Or the likelihood each workload "class" might need these configured differently?

coordinator/src/main/scala/filodb/coordinator/flight/ArrowSerializedRangeVectorOps.scala
```diff
@@ -211,6 +223,26 @@ object ArrowSerializedRangeVectorOps {
...
+  /**
+   * Invokes `f` with each set-bit row index (ascending) in a bit-packed Arrow BitVector's data
+   * buffer at `bufAddr`, without testing cleared bits individually. Reads the buffer one 64-bit
+   * word at a time so runs of unset bits (the common case: many data rows between sparse RVK
+   * marker rows) are skipped in O(1) rather than tested one bit at a time. Arrow bit buffers are
+   * always allocated padded to an 8-byte boundary, so reading a full trailing word past `rowCount`
+   * is always in-bounds; the `rowIndex < rowCount` guard just discards any padding bits.
+   */
+  private def cforSetBitPositions(bufAddr: Long, rowCount: Int)(f: Int => Unit): Unit = {
+    val numWords = (rowCount + 63) >> 6
```
alextheimer: Could you replace these bit-shifts with division/multiplication for readability?

coordinator/src/main/scala/filodb/coordinator/flight/FlightQueryResultStreaming.scala
```diff
@@ -209,11 +210,50 @@ trait FlightQueryResultStreaming extends StrictLogging {
...
+                      execPlan.checkResultBytes(bytes, queryConfig, res.warnings)
+                      loader.load(rb)
+                    }
+                    logger.debug(s"Putting $label vsr into flight response for " +
+                      s"queryPlanId=${execPlan.planId} rowCount=${state.flightVsr.getRowCount} " +
+                      s"vectorSize0=${state.flightVsr.getVector(0).getValueCount} " +
+                      s"vectorSize1=${state.flightVsr.getVector(1).getValueCount}")
+                    listener.putNext()
+                    val offered = state.freeVsrs.offer(vsr) // return to free pool for reuse
+                    if (!offered) Shutdown.haltAndCatchFire(
```
alextheimer, line 245: When might this happen? Is this something that could be ignored until someone is available to update a config? Asking because: it might be better to increment a metric in that case.

alextheimer (approved):
LGTM -- just a couple non-blocking comments/questions.

coordinator/src/main/scala/filodb/coordinator/flight/ArrowSerializedRangeVectorOps.scala
vishramachandran: If we tune this to make sure it corresponds to our per-query limits, we should be fine and not need to tune. My thinking was that we make it configurable the first time we find the need to change. 50 is sufficient for now.

coordinator/src/main/scala/filodb/coordinator/flight/ArrowSerializedRangeVectorOps.scala
vishramachandran: Bit operations are fast. Let me try to add some comments.

coordinator/src/main/scala/filodb/coordinator/flight/FlightQueryResultStreaming.scala
vishramachandran: More than 50 VSRs present in the queue, and no room left in the pool queue. Unlikely. If it happens, we need to increase size of queue (50 above).

vishramachandran: Shutdown class already has a metric

PR #2321, opened 2026-07-28
Restores all tests removed by force-push on multiple_partitions branch

bhavanap1131-cmd:
This is a clean branch created from develop to address the missing tests flagged in PR #2130. It applies the same resolveMetadataPartitions routing logic using git diff from the original branch, with all tests fully restored.

  Changes:
  - MultiPartitionPlanner: adds resolveMetadataPartitions which routes metadata queries (LabelValues, LabelNames, SeriesKeysByFilters) through getPartitions when all shard-key filters are
  Equals, falling back to getMetadataPartitions for regex, non-equals, or missing shard-key filters
  - ShardKeyRegexPlanner / SinglePartitionPlanner: adds flightEnabled default parameter
  
  Tests restored and added:
  - should not throw for regex filters on shard keys in metadata queries
  - should fall back to getMetadataPartitions when a shard key filter is non-Equals (regex)
  - should fall back to getMetadataPartitions when a shard key filter is NotEquals
  - should use getPartitions directly when all shard key filters are Equals (LabelValues, LabelNames, SeriesKeysByFilters variants)
  - should fall back to getMetadataPartitions when a shard-key filter is missing (LabelValues and SeriesKeysByFilters variants)
  - should fallback to getMetadataPartitions when one shard-key filter is regex (PlannerHierarchySpec)
  - should handle non-equals filters gracefully in getPartitions delegation
  
  Note: should fall back to getMetadataPartitions when useLegacyMetadataRouting is true was not restored — the useLegacyMetadataRouting config flag was removed from QueryConfig in a subsequent commit and the test no longer compiles.


alextheimer (commented):
This needs to be restored to the state that @amolnayak311 had approved on April 14th on this PR: https://github.com/filodb/FiloDB/pull/2130

The commit state should be restored, as well; if commits `c1, c2, ... cn` were present when the original PR was approved, then commits `c1, c2, ... cn` should be present in this PR.

The timeline should line up: the most-recent commit should be dated just prior to Amol's approval.

bhavanap1131-cmd:
test this please

